PR #17, opened 2019-11-14
Fix the encrypted token migration issue encountered on HEAD

josephschorr:
This change ensures there is better messaging around the encrypted token migration, including a new phase to use for new installations, and fixes an issue encountered when running database migrations for new installations


sfxworks:
Thanks to this, the docker file referencing `FROM centos:7` now works! 
https://i.snipboard.io/0MCoDQ.jpg

Before there were startup errors that would now allow me to go beyond the config user/pass phase.

Tested with
`docker run --privileged=true --net=host -p 8443:8443 -e ENCRYPTED_ROBOT_TOKEN_MIGRATION_PHASE=new-installation quantomworks/quay config asdf`

PR #21, opened 2019-11-15
OpenSearch

alecmerdler:
### Description

Adds [OpenSearch](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md) capabilities to Quay. This allows searching [directly using the Chrome OmniBar](https://stackoverflow.com/questions/7630144/how-to-add-google-chrome-omnibox-search-support-for-your-site).

### Changes

- Add `/opensearch.xml` route handler and template (for specific hostname)
- Add `<link rel="search">` tag to `base.html`

### Demo

**Chrome OnmiBar search (using `ngrok` because `localhost` not allowed):**
quay-opensearch

Addresses https://issues.jboss.org/browse/PROJQUAY-14

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -131,6 +131,8 @@ b._i.push([a,e,d])};b.__SV=1.2}})(document,window.mixpanel||[]);
 mixpanel.init("{{ mixpanel_key }}", { track_pageview : false, debug: {{ is_debug }} });</script><!-- end Mixpanel -->
 {% endif %}
 
+  <link rel="search" type="application/opensearchdescription+xml" title="{{ config_set['REGISTRY_TITLE'] }}" href="/opensearch.xml" />
+
   </head>
   <body ng-class="pageClass + ' ' + (user.anonymous ? 'anon' : 'signedin')" class="co-img-bg-network">
     <div id="co-l-footer-wrapper">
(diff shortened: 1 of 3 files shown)


thomasmckay:
Does this work only against quay.io or will it run against a on-premise install?

alecmerdler:
This will work for any running Quay instance.

templates/base.html
```diff
@@ -131,6 +131,8 @@
 mixpanel.init("{{ mixpanel_key }}", { track_pageview : false, debug: {{ is_debug }} });</script><!-- end Mixpanel -->
 {% endif %}
 
+  <link rel="search" type="application/opensearchdescription+xml" title="Quay" href="/opensearch.xml" />
```
josephschorr: The title needs to come from the config: `REGISTRY_TITLE` or `REGISTRY_TITLE_SHORT`. Both should already be in the template's context

templates/opensearch.xml
```diff
@@ -0,0 +1,10 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/" xmlns:moz="http://www.mozilla.org/2006/browser/search/">
+<ShortName>Quay</ShortName>
```
josephschorr: Ditto here, and for the description

templates/base.html
alecmerdler: Done

templates/opensearch.xml
alecmerdler: Done

thomasmckay:
Thanks for this! Please open and reference a jira at https://issues.jboss.org/projects/PROJQUAY/issues

alecmerdler:
@thomasmckay Added JIRA.

thomasmckay:
@josephschorr Can you review this please?

PR #23, opened 2019-11-16
PROJQUAY-85 - Travis CI: Lint for Python syntax errors

cclauss:
### Description of Changes

#### Changes:

* ..
* ..

#### Issue: https://issues.redhat.com/browse/PROJQUAY-85

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@
 language: python
 python: 2.7
 
-sudo: required
-
 services:
   - docker
 
@@ -24,6 +22,7 @@ cache:
     - $HOME/docker
 
 stages:
+  - lint
   - build
   - test
   - clean
@@ -31,7 +30,18 @@ stages:
 # We should label the steps if Travis ever supports it:
 # https://github.com/travis-ci/travis-ci/issues/5898
 jobs:
+  allow_failures:
+    - python: 3.8
   include:
+    - stage: lint
+      script:
+        - pip install flake8
+        - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+    - stage: lint
+      python: 3.8
+      script:
+        - pip install flake8
+        - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
     - stage: build
       name: Build
       script: scripts/ci build
--- a/config_app/config_application.py
+++ b/config_app/config_application.py
@@ -1,4 +1,6 @@
+import logging
 from config_app.c_app import app as application
+from util.log import logfile_path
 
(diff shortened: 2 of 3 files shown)


thomasmckay:
This is great, thanks! We're going to do a mass change to PEP-8 python formatting so we'll wait to test and merge.

If you're willing, creating an issue to track this at https://issues.jboss.org/projects/PROJQUAY/issues would be great. (If you do not wish to do this, we'll create one on your behalf. Frequent contributors are expected to do this step, however.)

There are three commits in this PR, please squash them into one unless there is a reason to keep unique.

Welcome!

cclauss:
Given that these tests pass is there any reason to wait?

.travis.yml
```diff
@@ -24,14 +22,26 @@ cache:
...
 # We should label the steps if Travis ever supports it:
 # https://github.com/travis-ci/travis-ci/issues/5898
 jobs:
+  allow_failures:
+    - python: 3.8
   include:
+    - stage: lint
+      script:
+        - pip install flake8
+        - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
```
thomasmckay: Mind changing this to a config file?

What was the reason to confine the select to these four types?

cclauss: https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.

cclauss:
Closing in favor of #124.

PR #25, opened 2019-11-19
PROJQUAY-12 - remove mirror/rules API

thomasmckay:
Updated API to remove mirror/rules endpoint. Rule is now handled within body of config only.

https://issues.jboss.org/browse/PROJQUAY-12

### DIFF
--- a/endpoints/api/mirror.py
+++ b/endpoints/api/mirror.py
@@ -11,6 +11,7 @@ import features
 
 from auth.auth_context import get_authenticated_user
 from data import model
+from data.database import RepoMirrorRuleType
 from endpoints.api import (RepositoryParamResource, nickname, path_param, require_repo_admin,
                            resource, validate_json_request, define_json_response, show_if,
                            format_date)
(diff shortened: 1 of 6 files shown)


endpoints/api/mirror.py
```diff
@@ -231,7 +232,7 @@ def get(self, namespace_name, repository_name):
       'sync_retries_remaining': mirror.sync_retries_remaining,
       'sync_status': mirror.sync_status.name,
       'root_rule': {
-        'rule_type': 'TAG_GLOB_CSV',
+        'rule_type': 1,
```
kurtismullins: I believe that we should continue using `TAG_GLOB_CSV` or some human-friendly String.
The integer `1` is simply an internal enumeration and probably shouldn't be exposed through the API.

thomasmckay: @kurtismullins other tables seem to use a "kind" that maps a string to an int. should we future proof the rules and change "rule_type" to "rule_kind" in the API? we wouldn't change the db columns but could if we wanted in the future

https://docs.quay.io/api/swagger/#!/repository/listRepos
repo_kind

thomasmckay: Replying to myself... yes, if we're going to use strings let's stick with the `kind` pattern.

thomasmckay: @kurtis updated

thomasmckay:
@kurtismullins - working on the tests. please review based on the code. thanks!

PR #27, opened 2019-11-20
Reformat Python code with psf/black

cclauss:
### Description of Changes

As discussed at https://github.com/quay/quay/pull/23#issuecomment-555777993 a "mass PEP-8 Python formatting" is needed so this PR is the result of executing __black --target-version py27 .__ using the Python Software Foundation's __black__ code formatting tool.  https://github.com/psf/black

__black__ is NOT configurable to meet _local_ standards.  This is done on purpose so that _all_ Python code starts to look the same across _all_ projects.  You can have any color you want as long as it is __black__.

#### Changes:

* ..
* ..

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!


thomasmckay:
@jzelinskie Thoughts?

cclauss:
Black --safe (the default) builds an Abstract Syntax Tree before and then reformats the code and then builds a second AST after to do a diff on to ensure the new code is 100% syntax compatible with the original.  I doubt the current code can build a Python 3.7 AST.  The formatting with the current settings will be similar but not identical.  The trick is to surrender control to the tool and rerun it as your code progresses.

jzelinskie:
I definitely prefer this to anything else. Previously we tried using yapf, but because that uses the clang-format algorithm, which is only on a line-by-line basis, it didn't guarantee that the same expression written two different ways would always be exactly transformed into the same result.

This PR needs another commit that adds a linting step to the travis.yml so that PRs that do not conform to black fail.

thomasmckay:
Some devs are out of town and have not had a chance to ponder the full implications of this. Some questions for you @cclauss (I'm treating you as the expert :)...

1. Can you share the exact `black` command you ran? I'd like to run it myself to see what the various options do. Likely we'll bump the line length at least in the final version.

2. Importantly, would you recommend waiting until we make the move to python-3.7? I'm curious if the diff between `black -t 2.7` and `black -t 3.7` is large and if that will be disruptive.

Thanks!

cclauss:
@thomasmckay Please reread the original commit message because the answer to 1. is there:
* PR is the result of executing __black --target-version py27 .__

> I'd like to run it myself to see what the various options do.

There are not a lot of options (on purpose) as discussed above:
__black__ is NOT configurable to meet _local_ standards.  This is done on purpose so that _all_ Python code starts to look the same across _all_ projects.  You can have any color you want as long as it is __black__.

Finally, what I was trying to get across above on diff between `black -t 2.7` and `black -t 3.7` is format early and format as frequently as needed.

Once __black .__ is successful then we should add __black --check .__ to our continuous integration.

cclauss:
Closing because the formatting executes very quickly but these PRs touch so many files that conflicts also appear quickly.

thomasmckay:
Good point on running it with python version bumps and on PRs.

Really appreciate you initiating this topic. Stay tuned!

PR #28, opened 2019-11-20
PROJQUAY-17 - Update build.sh to support buildah or podman

sghosh151:
use buildah or podman instead of docker if available

### Description of Changes

* search for an use buildah or podman for container build
* motivation for the change (use podman - non root for build)
* contrast with previous behavior - perfrs buildah, then podman, then docker for container build

#### Changes:

* ..
* ..

#### Issue: https://issues.jboss.org/browse/PROJQUAY-17

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!


thomasmckay:
I've created a jira for this https://issues.jboss.org/browse/PROJQUAY-17

I added some acceptance criteria so docs and QE have something to work off of. I'd prefer to see a way for this to be explicitly controlled (via env var is my suggestion). Can you fill in details there?

Thanks!

thomasmckay:
@sghosh151 Were going to continue work on this?

sghosh151:
Yes, will get a patch later this week.  Just catching up at work after
KubeCon.

hazcod:
Ping @sghosh151

vwbusguy:
Is there anything I can do to help with this?  It's a feature we'd like to see since there's no way to do multi-staged docker builds using RHEL supplied packages, as a result.

thomasmckay:
Personally I have all three tools installed. Perhaps an arg to the script or an environment variable to override the default? I'm happy to have the default be _buildah_ if that is the concensus.

Thoughts @vwbusguy @sghosh151 @hazcod ?

vwbusguy:
@thomasmckay - Personally, I'm happy to have buildah as the default as well.  This is essentially the RHEL8 expected behavior.  I can also see the use case for someone wanting to specify a tool.  Part of the reason this PR exists is because it is currently hard-coded to only support Docker.  Might I suggest for now releasing this one as-is, if it works, and then adding support for an environment variable selector as a potential follow-up iteration?

PR #38, opened 2019-11-25
added scripts to build off both rhel7 and centos7

tparikh:
As we setup CI/CD pipeline for Quay, we may need to switch the source image for quay container images. We should achieve this code change for Jenkins job in the app-interface repository and not the upstream quay repository.

https://issues.jboss.org/browse/PROJQUAY-24


tparikh:
[test]

PR #42, opened 2019-11-25
Fix clearing of names on access tokens

josephschorr:
This fixes the logic on the token_name lookup and only applies it to non-temporary tokens


josephschorr:
@thomasmckay This was a noticed bug fix while verifying something else; no JIRA was filed or is necessary, as it was discovered before it was run.

josephschorr:
As part of the wider change of https://jira.coreos.com/browse/QUAY-1776

PR #47, opened 2019-12-02
PROJQUAY-34 - create and chown /certificates

thomasmckay:
The `certs_create.sh` expects _/certificates_ to be writable. Dockerfiles updated.

https://issues.jboss.org/browse/PROJQUAY-34


tparikh:
/lgtm

I have verified this patch with CentOS7 dockerfile.

PR #51, opened 2019-12-04
Break the encrypted token backfill into its own phase

josephschorr:
This will ensure that if we have an issue, we don't also have to contend with a schema change


josephschorr:
https://jira.coreos.com/browse/QUAY-2204

PR #22, opened 2019-11-15
PROJQUAY-13 - Rework SLIs and migrate to Prometheus PushGateway

jzelinskie:
Previously Quay had been using a [simple aggregator binary](https://github.com/coreos/prometheus-aggregator) written before Prometheus PushGateway existed. This PR replaces this process with the v1 release of Prometheus PushGateway and also applies numerous idiomatic improvements to the metrics being tracked by Quay as they've been reviewed by members of the Prometheus team themselves.

Notably, this PR also adds a metric for tracking greenlet content switching execution timing.

Thanks for feedback to @brancz @liliC @bwplotka @paulfantom @s-urbaniak @metalmatze

⚠️ This change affects how metrics are exposed by the Quay container and could potentially require manual changes to deployments

PR #59, opened 2019-12-05
dockerfile: add pushgateway to OSBS images

jzelinskie:
This change adds the missing installation commands to the OSBS Dockerfiles that were missed in #22 .


jzelinskie:
@thomasmckay Do I need to do anything specific to get these binaries into OSBS?

PR #65, opened 2019-12-10
PROJQUAY-39 modified quay liveness and readiness probes

tparikh:
The Liveness and Readiness probes were not working because the connection to port 8443 was being rejected due to SSL cert issues. As the cert is setup for external domain name, requests to `localhost` were being rejected with `400` error code.

The modified check uses `curl` with `-k` (insecure) flag to check if the application is ready and later check if it is still live.

https://issues.redhat.com/browse/PROJQUAY-39

### DIFF
--- a/deploy/openshift/quay-app.yaml
+++ b/deploy/openshift/quay-app.yaml
@@ -249,9 +249,3 @@ parameters:
   - name: IMAGE_PULL_POLICY
     value: "IfNotPresent"
     displayName: image pull policy
-  # - name:
-  #   value: ""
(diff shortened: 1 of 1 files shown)


tparikh:
@thomasmckay @kurtismullins PTAL

ibazulic:
If TLS is terminated outside of the instance, probes will fail again because the container is not listening on 8443. Also, when do the probes start hitting the instance? Quay needs 30 to 45 seconda to start and nginx will give out a 502 until all workers are up and running.

PR #89, opened 2019-12-13
deploy: add prom service for OpenShift

jzelinskie:


aditya-konarde:
We actually realized this needs to be in the quay-app OpenShift manifest. 

/hold

aditya-konarde:
This is good to go now

PR #95, opened 2019-12-16
openshift: use one template for all of quay

jzelinskie:


tparikh:
@jzelinskie can you open JIRA describing the changes and link JIRA to this PR.

PR #109, opened 2019-12-18
Add support for custom webhook bodies

josephschorr:
This will allow users to configure the body content of outgoing webhooks for notifications

Fixes https://jira.coreos.com/browse/QUAY-1322

### DIFF
--- a/static/directives/create-external-notification.html
+++ b/static/directives/create-external-notification.html
@@ -1,8 +1,5 @@
 <div class="create-external-notification-element">
   <form id="createForm" name="createForm" ng-submit="createNotification()">
-    <!-- Creating spinner -->
-    <div class="cor-loader" ng-show="status == 'creating'"></div>
-
     <!-- Event -->
     <table class="help-table">
       <tr>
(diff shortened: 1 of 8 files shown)


static/directives/create-external-notification.html
```diff
@@ -130,21 +127,18 @@
                          allowed-entities="['user', 'team', 'org']"
                          ng-switch-when="entity"></div>
 
-                    <div ng-if="getHelpUrl(field, currentConfig)"
-                         style="margin-top: 10px;  margin-bottom: 10px">
-                      See: <a href="{{ getHelpUrl(field, currentConfig) }}" ng-safenewtab>{{ getHelpUrl(field, currentConfig) }}</a>
-                    </div>
+                    <!-- Template -->
```
alecmerdler: Can you add some screenshots of the new UI?

static/js/directives/ui/create-external-notification.js
```diff
@@ -103,6 +103,9 @@ angular.module('quay').directive('createExternalNotification', function () {
         ApiService.createRepoNotification(data, params).then(function(resp) {
           $scope.status = '';
           $scope.notificationCreated({'notification': resp});
+        }, function(resp) {
+          $scope.status = 'error';
+          $scope.errorMessage = ApiService.getErrorMessage(resp, 'Could not create notification');
```
alecmerdler: So we are only performing server-side validation of the input?

util/jsontemplate.py
```diff
@@ -0,0 +1,81 @@
...
+from jsonpath_rw.lexer import JsonPathLexerError
+
+INLINE_PATH_PATTERN = r"\$\{([^}]*)\}"
+
+
+class JSONTemplateParseException(Exception):
+    """ Exception raised if a JSON template could not be parsed. """
+
+
+class JSONTemplate(object):
```
alecmerdler: Advantages of this versus just using [Jinja.Template](https://jinja.palletsprojects.com/en/2.10.x/api/#jinja2.Template)?

josephschorr: Simplicity and safety: This is guaranteed to produce JSON and unlike Jinja, we have no concerns about weird imports or execution of arbitrary code that could exhaust memory, or infinitely loop, etc.

static/js/directives/ui/create-external-notification.js
josephschorr: For the template? Yes. Doing so on the client would be a bit more work.

dongboyan77:
Verified
Screenshot from 2019-12-26 10-21-58

josephschorr:
@dongboyan77 Can you post a verification screenshot with a custom body as well? (unless the above is that and I missed it)

dongboyan77:
Screenshot from 2019-12-25 17-38-32
@josephschorr

josephschorr:
@dongboyan77 Does that mean it failed? Looks like it says the webhook was disabled

dongboyan77:
@josephschorr It succeed, it shows disabled in screenshot because I make the network down

SISheogorath:
Uhm, isn't the docs for this feature still missing? At least I couldn't find any, and the implemented `help_url` (https://docs.quay.io/guides/webhook-template.html) Points to a dead page.

PR #110, opened 2019-12-18
metrics: push/pull bytes metrics named consistent

jzelinskie:
These were named inconsistently. /shrug

### DIFF
--- a/data/registry_model/blobuploader.py
+++ b/data/registry_model/blobuploader.py
@@ -26,7 +26,7 @@ chunk_upload_duration = Histogram(
     labelnames=["region"],
 )
 pushed_bytes_total = Counter(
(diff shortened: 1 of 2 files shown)


data/registry_model/blobuploader.py
```diff
@@ -26,7 +26,7 @@
     labelnames=["region"],
 )
 pushed_bytes_total = Counter(
-    "quay_registry_pushed_bytes_total", "number of bytes pushed to the registry"
+    "quay_registry_image_pushed_estimated_bytes_total", "number of bytes pushed to the registry"
```
josephschorr: This isn't estimated. This is actual (since we can track every byte uploaded)

jzelinskie: That's a good point -- the push isn't an estimate.

PR #132, opened 2020-01-05
Fix the assertion in lookup of logs by namespace

josephschorr:
Adds additional testing and assertions as well

Fixes https://issues.redhat.com/browse/QUAY-2239

### DIFF
--- a/data/logs_model/table_logs_model.py
+++ b/data/logs_model/table_logs_model.py
@@ -9,7 +9,7 @@ from dateutil.relativedelta import relativedelta
 
 from data import model
 from data.model import config
-from data.database import LogEntry, LogEntry2, LogEntry3, UseThenDisconnect
+from data.database import LogEntry, LogEntry2, LogEntry3, BaseModel, UseThenDisconnect
 from data.logs_model.interface import (
     ActionLogsDataInterface,
     LogsIterationTimeout,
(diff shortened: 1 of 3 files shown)


data/logs_model/table_logs_model.py
```diff
@@ -279,15 +282,19 @@ def load_logs():
...
                 logs_query = model.log.get_logs_query(
-                    namespace=namespace_id,
+                    namespace_id=namespace_id,
                     repository=repository_id,
                     start_time=current_start_datetime,
                     end_time=current_end_datetime,
                 )
                 logs = list(logs_query)
                 for log in logs:
+                    assert isinstance(log, BaseModel)
```
kleesc: Why is this check needed?

data/logs_model/table_logs_model.py
```diff
@@ -279,15 +282,19 @@ def load_logs():
                 )
 
                 logs_query = model.log.get_logs_query(
-                    namespace=namespace_id,
+                    namespace_id=namespace_id,
```
kleesc: Would it be more consistent to just use the actual namespace here instead of adding `namespace_id` as parameter to `_logs_query`. Just because every other methods in this class uses the namespace's name. Also if `_logs_query` is able to take the `namespace_id`, should it also be able to accept a `repository_id`, `performer_id`, ...

data/logs_model/table_logs_model.py
josephschorr: Just to be absolutely sure we're getting row instances out of the query.

data/logs_model/table_logs_model.py
josephschorr: We'd have to do another lookup, which is slower and less safe.

PR #138, opened 2020-01-08
PROJQUAY-123 - enable repo for gpgme-devel

thomasmckay:
https://issues.redhat.com/browse/PROJQUAY-123

Updated Dockerfile.centos8 to install gpgme-devel. Also uncommented remaining aspects to build container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ RUN INSTALL_PKGS="\
         gcc-c++ git \
         openldap-devel \
         python3-devel \
+        python3-gpg \
         dnsmasq \
         memcached \
         openssl \
--- a/requirements.txt
+++ b/requirements.txt
@@ -122,7 +122,7 @@ pycparser==2.19
 pycryptodome==3.9.0
 pycryptodomex==3.9.0
 PyGithub==1.43.8
-# gpgme-devel not available on centos8 pygpgme==0.3
+gpg==1.10.0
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -122,7 +122,7 @@ pycparser==2.19
 pycryptodome==3.9.0
 pycryptodomex==3.9.0
 PyGithub==1.43.8
-# gpgme-devel not available on centos8 pygpgme==0.3
+pygpgme==0.3
```
Conan-Kudo: pygpgme does not work with current versions of gpgme. Use `python3-gpg` instead from EL8 repos. They should be visible as Python package `gpg` with pip, but only installable via RPM.

Dockerfile
```diff
@@ -26,12 +26,13 @@ RUN INSTALL_PKGS="\
...
         openldap-devel \
         python3-devel \
+        python3-gpg \
         dnsmasq \
         memcached \
         openssl \
         skopeo \
         " && \
-    yum -y --setopt=tsflags=nodocs --setopt=skip_missing_names_on_install=False install $INSTALL_PKGS && \
+    yum -y --enablerepo=PowerTools --setopt=tsflags=nodocs --setopt=skip_missing_names_on_install=False install $INSTALL_PKGS && \
```
Conan-Kudo: You should not need `PowerTools` anymore...

PR #142, opened 2020-01-09
Add a timeout to queuefile to ensure it doesn't block forever

josephschorr:
https://issues.redhat.com/browse/PROJQUAY-132

### DIFF
--- a/util/registry/queuefile.py
+++ b/util/registry/queuefile.py
@@ -1,9 +1,12 @@
+from multiprocessing.queues import Empty, Queue
+
+
 class QueueFile(object):
     """ Class which implements a file-like interface and reads QueueResult's from a blocking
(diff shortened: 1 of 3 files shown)


util/registry/queuefile.py
```diff
@@ -1,9 +1,13 @@
...
+
+
 class QueueFile(object):
     """ Class which implements a file-like interface and reads QueueResult's from a blocking
       multiprocessing queue.
   """
 
-    def __init__(self, queue, name=None):
+    def __init__(self, queue, name=None, timeout=None):
+        assert isinstance(queue, Queue)
```
kleesc: This is breaking unit tests since it's using a FakeQueue.

josephschorr: Removed

PR #151, opened 2020-01-13
Fix display of entities in the notifications table

josephschorr:


josephschorr:
Screen Shot 2020-01-13 at 4 36 00 PM

PR #160, opened 2020-01-15
Gevent monkey patch fix

kleesc:
### Description of Changes

https://issues.redhat.com/browse/PROJQUAY-169
* Monkey patch before importing other dependencies when starting a gunicorn

#### Changes:

* ..
* ..

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/registry.py
+++ b/registry.py
@@ -1,8 +1,3 @@
-# NOTE: Must be before we import or call anything that may be synchronous.
-from gevent import monkey
-
-monkey.patch_all()
-
 import endpoints.decorated  # Note: We need to import this module to make sure the decorators are registered.
 import features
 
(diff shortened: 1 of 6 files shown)


registry.py
```diff
@@ -1,8 +1,3 @@
-# NOTE: Must be before we import or call anything that may be synchronous.
-from gevent import monkey
-
-monkey.patch_all()
```
kurtismullins: @kleesc Assuming I understand correctly -- If this is removed, and Quay is ran without gunicorn (e.g. `python -m application.py`) then the monkey patching won't occur. Is this expected behaviour? Is there any downfall to keeping these monkey-patch lines in addition to adding them before the `gunicorn` workers?

kleesc: It's still being monkey patched in application.py.

kurtismullins: @kleesc Ahh, so you're basically identifying the entry-points and monkey patching them there? Smart!

PR #170, opened 2020-01-16
Remove Duplicate Code for Secscan Client

alecmerdler:
### Description

Found duplicate code block in `app.py`.


app.py
```diff
@@ -315,20 +315,6 @@ def _request_end(resp):
     instance_keys=instance_keys,
 )
 
-url_scheme_and_hostname = URLSchemeAndHostname(
```
alecmerdler: Duplicated from above.

PR #172, opened 2020-01-17
Disable Flake8 in CI on Python 2.7 branch

kurtismullins:
### Description of Changes

- It appears that Flake8 is raising issues with the existing Python 2.7 code.
  While it is possible that these could genuinely use some cleanup, it
  appears some errors come from the expectation of Python 3.x syntax. It  has
  been temporarily commented out until we can fix it.
- Related to: https://github.com/quay/quay/pull/162

Example CI Error:
```
35    E999 SyntaxError: invalid syntax
27    F821 undefined name 'basestring'
62
The command "scripts/ci lint" exited with 1.
```

#### Changes:

* Comment out execution of `flake8` in CI until all offending files are fixed or Flake8 is configured to verify against the correct Python version

#### Issue:

- This is a resolution to the CI failures caused in #162

**TESTING**

- The CI lint stage should no longer fail on `flake8` and `flake8` is not executed during CI at this time.

**BREAKING CHANGE**

n/a

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/scripts/ci
+++ b/scripts/ci
@@ -197,7 +197,7 @@ postgres() {
 case "$1" in
     lint)
         run_black
-        run_flake8
+        # run_flake8  TODO: Run this after offending files have been fixed. Ensure appropriate Python version is used.
         ;;
 
     build)


cclauss:
flake8 on Python 2 passes with #128

kurtismullins:
> flake8 on Python 2 passes with #128

Thanks for mentioning that @cclauss! I think the primary difference is that #128 targets the `quay/python3` branch while #162 was targeting `quay/master`. This is just a temporary work-around until we can wrap up the python3 work and get it merged into master.

One proposal I have is that once I rebase `python3` against `master` again, we can re-enable flake8 in the `python3` branch. What do you think about that approach?

cclauss:
After doing many Python 3 ports, I can say that the separate __python3__ branch almost never works.  A single branch that can run on both Python 2 and Python 3 is the best practice.

cclauss:
The errors above are because flake8 was running on Python 3, not on Python 2.

PR #181, opened 2020-01-20
Remove the active migration for encrypted tokens now that it is complete

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-185

### DIFF
--- a/Dockerfile.cirun
+++ b/Dockerfile.cirun
@@ -1,7 +1,6 @@
 FROM quay-ci-base
 RUN mkdir -p conf/stack
 RUN rm -rf test/data/test.db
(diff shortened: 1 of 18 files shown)


josephschorr:
@thomasmckay Done

josephschorr:
@thomasmckay Yes, as data/migrationutil.py is the framework for defining migrations. This PR removes the currently active migration, but it leaves the framework in place for future data migrations as needed. Note that the code won't be called right now since `ActiveMigration` is `None` as of this PR.

kurtismullins:
@josephschorr Just a heads-up, it looks like there are test failures.

josephschorr:
@kurtismullins updated

PR #183, opened 2020-01-20
Bytes unicode conversion

kleesc:
### Description of Changes

* Makes the conversions from str to bytes when needed.

#### Changes:

* ..
* ..

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/auth/test/test_basic.py
+++ b/auth/test/test_basic.py
@@ -65,8 +65,8 @@ def _token(username, password):
         ),
         (
             _token("usér", "passwôrd"),
-            ValidateResult(AuthKind.basic, error_message="Invalid Username or Password")
-        )
+            ValidateResult(AuthKind.basic, error_message="Invalid Username or Password"),
+        ),
     ],
 )
 def test_validate_basic_auth_token(token, expected_result, app):
@@ -116,14 +116,14 @@ def test_valid_app_specific_token(app):
 
 def test_invalid_unicode(app):
     token = b"\xebOH"
-    header = "basic " + b64encode(token).decode('ascii')
+    header = "basic " + b64encode(token).decode("ascii")
     result = validate_basic_auth(header)
     assert result == ValidateResult(AuthKind.basic, missing=True)
 
 
 def test_invalid_unicode_2(app):
-    token = "“4JPCOLIVMAY32Q3XGVPHC4CBF8SKII5FWNYMASOFDIVSXTC5I5NBU”".encode('utf-8')
-    header = "basic " + b64encode(b"devtable+somerobot:%s" % token).decode('ascii')
+    token = "“4JPCOLIVMAY32Q3XGVPHC4CBF8SKII5FWNYMASOFDIVSXTC5I5NBU”".encode("utf-8")
+    header = "basic " + b64encode(b"devtable+somerobot:%s" % token).decode("ascii")
     result = validate_basic_auth(header)
     assert result == ValidateResult(
         AuthKind.basic,
--- a/auth/test/test_basic.py
+++ b/auth/test/test_basic.py
@@ -19,8 +19,8 @@ from test.fixtures import *
 def _token(username, password):
     assert isinstance(username, str)
     assert isinstance(password, str)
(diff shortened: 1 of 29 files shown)


auth/test/test_basic.py
```diff
@@ -19,7 +19,8 @@
 def _token(username, password):
     assert isinstance(username, str)
     assert isinstance(password, str)
-    return "basic " + b64encode("%s:%s" % (username, password))
+    token_bytes = b"%s:%s" % (username.encode('ascii'), password.encode('ascii'))
```
josephschorr: What happens if the username or password is in unicode? While we don't support unicode usernames, we should make sure we don't raise an exception. Can you add a test for that (if we don't have one)?

avatars/avatars.py
```diff
@@ -101,7 +101,7 @@ def get_data(self, name, email_or_id, kind="user"):
         # Note: email_or_id may be None if gotten from external auth when email is disabled,
         # so use the username in that case.
         username_email_or_id = email_or_id or name
```
josephschorr: Ditto. Test needed for a unicode input here

digest/digest_tools.py
```diff
@@ -71,7 +71,7 @@ def sha256_digest_from_generator(content_generator):
   """
     digest = hashlib.sha256()
```
josephschorr: Please add a test for creating a digest with a unicode input string

endpoints/api/tag.py
```diff
@@ -32,10 +32,10 @@ def _tag_dict(tag):
         "reversion": tag.reversion,
     }
 
-    if tag.lifetime_start_ts > 0:
+    if tag.lifetime_start_ts and tag.lifetime_start_ts > 0:
```
josephschorr: Huh! `None > 0` worked in Python 2? Crazy

endpoints/v2/v2auth.py
```diff
@@ -133,7 +133,7 @@ def generate_registry_jwt(auth_result):
     token = generate_bearer_token(
         audience_param, subject, context, access, TOKEN_VALIDITY_LIFETIME_S, instance_keys
     )
-    return jsonify({"token": token})
+    return jsonify({"token": token.decode('ascii')})
```
josephschorr: Are we 100% sure JWT tokens are ASCII compatible? We should check the spec and link to the relative section

kleesc: The tokens should be a base64 encoded string joined by a `.` AFAIK.

alecmerdler: The first line of the [abstract in the RFC](https://tools.ietf.org/html/rfc7519): 
> JSON Web Token (JWT) is a compact, **URL-safe** means...
Yeah it's just base 64 character set + periods.

kurtismullins:
@kleesc Can you run the `black` command on your branch? I see this lint error in CI:
```
9 files would be reformatted, 827 files would be left unchanged.
```
The command that is used by CI is available in the `scripts/ci` file.

data/fields.py
```diff
@@ -72,12 +72,12 @@ class Base64BinaryField(TextField):
     def db_value(self, value):
         if value is None:
             return None
-        return base64.b64encode(value)
+        return base64.b64encode(value).decode('ascii')
```
alecmerdler: I ran into this in my conversion work as well. This means every time we want to use this field, we need to make sure we are expecting actual `bytes` when using this field. Does that make sense? Or should we just keep everything as strings?

To avoid headaches, this guy suggested using the ["Unicode Sandwich"](https://nedbatchelder.com/text/unipain.html#h_pain_relief) (decode incoming binary immediately -> Python code manipulates decoded strings -> encode outgoing strings it

kurtismullins: One potential pitfall is that not all unicode can be accurately converted to ASCII characters. We should verify the behavior on a case-by-case basis. My hunch is that it's fine but it'd be worth verifying the data being base64 encoded.

josephschorr: If the data is passed through `b64encode`, it should be fine, as that call will fail if it cannot encode the data

PR #184, opened 2020-01-20
PROJQUAY-181: updated nginx to redirect http to https for quay.io

tparikh:
[ci skip]

JIRA: https://issues.redhat.com/browse/PROJQUAY-181

http://quay.io should be redirected to https://quay.io

Set `FEATURE_PROXY_PROTOCOL` to `True` to enable proxy protocol support for HTTP traffic.

### DIFF
--- a/conf/init/nginx_conf_create.py
+++ b/conf/init/nginx_conf_create.py
@@ -123,6 +123,17 @@ def generate_rate_limiting_config(config):
         static_dir=STATIC_DIR,
     )
 
+def generate_hosted_http_base_config(config):
+    """
+  Generates hosted http base config from the app config
+  """
+    config = config or {}
+    feature_proxy_protocol = config.get("FEATURE_PROXY_PROTOCOL", False)
+
+    write_config(
+        os.path.join(QUAYCONF_DIR, "nginx/hosted-http-base.conf"),
+        feature_proxy_protocol=feature_proxy_protocol,
+    )
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


conf/init/nginx_conf_create.py
```diff
@@ -123,6 +123,16 @@ def generate_rate_limiting_config(config):
         static_dir=STATIC_DIR,
     )
 
+def generate_hosted_http_base_config(config):
+    """
+  Generates hosted http base config from the app config
+  """
+    config = config or {}
+    nginx_http_listen_directive_params = config.get("NGINX_HTTP_LISTEN_DIRECTIVE_PARAMS", "default_server")
```
josephschorr: Tejas - We should just, instead, add a FEATURE_PROXY_PROTOCOL and if set, add that to the nginx side.

conf/init/nginx_conf_create.py
```diff
@@ -123,6 +123,23 @@ def generate_rate_limiting_config(config):
         static_dir=STATIC_DIR,
     )
 
+def generate_hosted_http_base_config(config):
+    """
+  Generates hosted http base config from the app config
+  """
+    config = config or {}
+
+    feature_proxy_protocol = config.get("FEATURE_PROXY_PROTOCOL", False)
```
josephschorr: I'd send in the proxy protocol value and use an `{% if ` in the template

tparikh: PR updated

PR #195, opened 2020-01-22
Python 3 for `initdb`

alecmerdler:
### Description

Makes the necessary changes for `initdb.py` to work with Python 3.

### Testing

```
$ python initdb.py
```

Addresses https://issues.redhat.com/browse/PROJQUAY-164
Commits taken from https://github.com/quay/quay/pull/183


alecmerdler:
I'm not sure if it's a problem that `test/data/test.db` was changed as a result. Should it be the same under Python 3?

kurtismullins:
@alecmerdler Can you run `black` on this branch? It's currently failing in the "lint" stage.

util/security/secret.py
```diff
@@ -25,5 +25,10 @@ def convert_secret_key(config_secret_key):
         secret_key = str(bytearray(list(map(ord, config_secret_key))))
 
     # Otherwise, use the bytes directly.
-    assert len(secret_key)
-    return "".join(itertools.islice(itertools.cycle(secret_key), 32))
+    assert len(secret_key) > 0
+    print(secret_key)
```
kurtismullins: Not sure if this was intentional but we should probably pull this `print()` statement out. I assume it was there for debugging purposes.

util/security/secret.py
```diff
@@ -25,5 +25,10 @@ def convert_secret_key(config_secret_key):
...
+
+    # FIXME(alecmerdler): Testing
+    secret_key = b'\xa3l\x9d}%\xa90\xf4\xa5\x86=/\x8d\xc4\n\x83'
```
kurtismullins: I assume this was also left in for debugging purposes.

kleesc:
You'll probably need to remove the enclosing `str()` in L13 and L25 in `util/security/secret.py`

util/security/secret.py
```diff
@@ -22,8 +22,9 @@ def convert_secret_key(config_secret_key):
             pass
 
     if secret_key is None:
-        secret_key = str(bytearray(list(map(ord, config_secret_key))))
+        secret_key = bytearray(list(map(ord, config_secret_key)))
```
alecmerdler: This is causing differences in `test/data/test.db` when running in Python 2 versus Python 3. Investigating.

alecmerdler:
Closed by https://github.com/quay/quay/pull/183

PR #208, opened 2020-01-30
Better manifest errors

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-219

### DIFF
--- a/data/model/oci/retriever.py
+++ b/data/model/oci/retriever.py
@@ -1,7 +1,12 @@
+import time
+
 from image.docker.interfaces import ContentRetriever
 from data.database import Manifest
 from data.model.oci.blob import get_repository_blob_by_digest
 from data.model.storage import get_layer_path
+
+RETRY_COUNT = 2
+RETRY_DELAY = 0.25  # seconds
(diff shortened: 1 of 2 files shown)


data/model/oci/retriever.py
```diff
@@ -37,4 +42,14 @@ def get_blob_bytes_with_digest(self, digest):
...
+        # NOTE: Some storage engines are eventually consistent, and so we add a small
+        # retry here for retrieving the blobs from storage, as they may just have been
+        # written as part of the push process.
+        for retry in range(0, RETRY_COUNT):
+            try:
+                return self.storage.get_content(blob.locations, get_layer_path(blob))
+            except IOError:
+                time.sleep(RETRY_DELAY)
+
+        return None
```
kleesc: Shouldn't it reraise whatever the last failed result of `get_content` is instead?

josephschorr: No. This method should return `None` if the file is not found.

josephschorr:
@alecmerdler @kleesc Made another small change

PR #210, opened 2020-01-31
Fix missing Dockerfile in CI

kurtismullins:
In the Python3 work, we have introduced multiple changes to CI. Specifically, we have moved all test-dependencies from the base Dockerfile(s) into Dockerfile.cirun.

There is now the following error:
```
Unable to find image 'quay-ci:943072a-687-ci-run' locally

docker: Error response from daemon: pull access denied for quay-ci, repository does not exist or may require 'docker login'.
```

**Jira:**
https://issues.redhat.com/browse/PROJQUAY-218

### DIFF
--- a/scripts/ci
+++ b/scripts/ci
@@ -8,6 +8,7 @@ IMAGE="quay-ci"
 CACHE_DIR="${HOME}/docker"
 SHORT_SHA="${TRAVIS_COMMIT:0:7}"
 IMAGE_TAG="${SHORT_SHA}-${TRAVIS_BUILD_NUMBER}"
+BASE_IMAGE_TAR="${CACHE_DIR}/${IMAGE}-${IMAGE_TAG}-base.tar.gz"
 IMAGE_TAR="${CACHE_DIR}/${IMAGE}-${IMAGE_TAG}.tar.gz"
 
 MYSQL_IMAGE="mysql:5.7"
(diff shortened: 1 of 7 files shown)


kurtismullins:
This does not fix _every_ test as can be seen in this PR's Travis CI build. However, it's a good start and I want to cleanup how CI is performed overall.

scripts/ci
```diff
@@ -68,10 +75,11 @@ clean_cache() {
...
 
 fail_clean() {
-    if [[ $TRAVIS_TEST_RESULT -ne 0 ]]; then
-        echo "Job failed. Cleaning cache..."
-        clean_cache
-    fi
+    #if [[ $TRAVIS_TEST_RESULT -ne 0 ]]; then
+    #    echo "Job failed. Cleaning cache..."
+    #    clean_cache
+    #fi
```
thomasmckay: Just remove this?

scripts/ci
```diff
@@ -204,7 +212,7 @@ postgres() {
 case "$1" in
     lint)
         run_black
-        # run_flake8  TODO: Run this after offending files have been fixed. Ensure appropriate Python version is used.
+        #run_flake8  # TODO: Enable this once we're ready to tackle Flake8 revealed errors
```
thomasmckay: I thought we were going to enable this on python3 branch?

kurtismullins: Not until we're ready to fix the errors it reveals (non-trivial)

scripts/ci
kurtismullins: I'll clean it up in a future revision :) I have another approach I want to take on CI in general that is much quicker and simpler.

PR #215, opened 2020-02-03
Pre-flight the cleanup of all big tables referencing a purged repo

josephschorr:
This should prevent the final deletion transaction from timing out

Fixes https://issues.redhat.com/browse/PROJQUAY-239

### DIFF
--- a/data/model/gc.py
+++ b/data/model/gc.py
@@ -1,5 +1,7 @@
 import logging
 
+from peewee import fn
+
 from data.model import config, db_transaction, storage, _basequery, tag as pre_oci_tag
 from data.model.oci import tag as oci_tag
 from data.database import Repository, db_for_update
@@ -15,6 +17,15 @@ from data.database import (
     Label,
     TagManifestLabel,
     RepositoryState,
+    RepositoryBuild,
+    RepositoryBuildTrigger,
+    RepositoryActionCount,
+    Star,
+    AccessToken,
+    RepositoryNotification,
+    BlobUpload,
+    RepoMirrorConfig,
+    RepositoryPermission,
 )
 from data.database import RepositoryTag, TagManifest, Image, DerivedStorageForImage
 from data.database import TagManifestToManifest, TagToRepositoryTag, TagManifestLabelMap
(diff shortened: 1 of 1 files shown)


data/model/gc.py
```diff
@@ -95,6 +118,28 @@ def purge_repository(repo, force=False):
...
+        min_id = model.select(fn.Min(model.id)).where(model.repository == repo).scalar()
+        if min_id is None:
+            return
+
+        max_id = (
+            model.select(fn.Max(model.id))
+            .where(model.repository == repo, model.id <= min_id + chunk_size)
+            .scalar()
+        )
+        if min_id is None or max_id is None or min_id > max_id:
```
kurtismullins: (nitpick)
```
if min_id is None
```
This was checked back on Line 127 but doesn't hurt anything to be here.

kurtismullins:
```
if [...] min_id > max_id
```
Should we assert this is not true, as a sanity check? It shouldn't ever happen regardless because of your query but I do suppose things can change in the database after the `min_id` query was executed.

data/model/gc.py
```diff
@@ -95,6 +118,28 @@ def purge_repository(repo, force=False):
...
+    assert repo.state == RepositoryState.MARKED_FOR_DELETION or force
+
+    while True:
+        min_id = model.select(fn.Min(model.id)).where(model.repository == repo).scalar()
+        if min_id is None:
+            return
+
+        max_id = (
+            model.select(fn.Max(model.id))
+            .where(model.repository == repo, model.id <= min_id + chunk_size)
```
kurtismullins: (nitpick) It might be easier to read, and ensure the correct logic, by grouping these in parenthesis:
```
min_id + chunk_size
```
e.g.
```
.where(model.repository == repo, model.id <= (min_id + chunk_size))
```

data/model/gc.py
josephschorr: Since this is just an optimization, if it fails, I'd rather just leave early than assert, since that would break the worker.

data/model/gc.py
josephschorr: sure. I can change that

PR #216, opened 2020-02-03
ManifestSecurityStatus Database Model

alecmerdler:
### Description

Lays the groundwork for the integration of Clair v4. Adds a new database model and associated functions for manipulating and querying.

### Changes

- Introduces `ManifestSecurityStatus` database model
- Introduces `V4SecurityScanner` implementation designed to use new DB model and Clair v4 client

### Future Work

- Clair v4 API client
- Add new v4-enabled endpoints to `/endpoints/api/secscan.py` and deprecate old ones
- Update UI to consume new API responses

Addresses https://issues.redhat.com/browse/PROJQUAY-81


data/database.py
```diff
@@ -1840,6 +1840,47 @@ class RepoMirrorConfig(BaseModel):
...
+    COMPLETED = 4
+
+
+@unique
+class IndexerVersion(IntEnum):
+    """
+    Possible versions of security indexers.
+    """
+
+    V2 = 0
```
alecmerdler: Why would we have `V2` if Clair v2 implementation does not consume `ManifestSecurityStatus` data model?

josephschorr: Because we're going to be changing it to consume the MSS. That's how we can get off of the Image table without fully migrating to Clair V4

data/database.py
```diff
@@ -1840,6 +1840,47 @@ class RepoMirrorConfig(BaseModel):
...
+    V4 = 1
+
+
+class ManifestSecurityStatus(BaseModel):
+    """
+    Represents the security scan status for a particular container image manifest. 
+
+    Intended to replace the `security_indexed` and `security_indexed_engine` fields
```
josephschorr: Remove the Intended command

data/database.py
```diff
@@ -1840,6 +1840,47 @@ class RepoMirrorConfig(BaseModel):
...
+class IndexerVersion(IntEnum):
+    """
+    Possible versions of security indexers.
+    """
+
+    V2 = 0
+    V4 = 1
+
+
+class ManifestSecurityStatus(BaseModel):
```
josephschorr: You should also add this to the transitive_deletion lists under User and Repository, to prevent recursive queries on delete

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,125 @@
...
+  *opaque* to the security worker, and the security worker should *not* pull any data out or modify
+  the token in any way.
+  """
+
+
+class V4SecurityScanner(SecurityScannerInterface):
+    """
+    Implementation of the security scanner interface for Clair V4 API-compatible implementations.
+
+    In addition to consuming a new security scanner API, this class also uses the `ManifestSecurityStatus`
```
josephschorr: Remove this part; it'll be obvious

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,125 @@
...
+
+        validator = SecurityConfigValidator(
+            app.config.get("FEATURE_SECURITY_SCANNER", False),
+            app.config.get("SECURITY_SCANNER_ENDPOINT"),
+        )
+        if not validator.valid():
+            logger.warning("Failed to validate security scanner V2 configuration")
+            return
+        
+        url_scheme_and_hostname = URLSchemeAndHostname(
```
josephschorr: Unused

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,125 @@
...
+            app.config["PREFERRED_URL_SCHEME"], app.config["SERVER_HOSTNAME"]
+        )
+      
+
+    def load_security_information(self, manifest_or_legacy_image, include_vulnerabilities=False):
+        if not isinstance(manifest_or_legacy_image, Manifest):
+            # TODO(alecmerdler): Do we support legacy images?
```
josephschorr: Nope. Have this return `None` in that case

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,125 @@
...
+        except database.ManifestSecurityStatus.DoesNotExist:
+            return SecurityInformationLookupResult.with_status(
+                ScanLookupStatus.NOT_YET_INDEXED
+            )
+
+        if status.index_status == database.IndexStatus.FAILED:
+            return SecurityInformationLookupResult.with_status(ScanLookupStatus.FAILED_TO_INDEX)
+        if status.index_status == database.IndexStatus.UNSUPPORTED:
+            return SecurityInformationLookupResult.with_status(ScanLookupStatus.UNSUPPORTED_FOR_INDEXING)
+        if status.index_status in [database.IndexStatus.TIMED_OUT, database.IndexStatus.WAITING, database.IndexStatus.IN]:
```
josephschorr: `IndexStatus.IN`?

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,125 @@
...
+        # TODO(alecmerdler): Fetch data from Clair API
+        return SecurityInformationLookupResult.with_status(ScanLookupStatus.NOT_YET_INDEXED)
+
+
+    def perform_indexing(self, start_token=None):
+        min_id = None
+        if start_token is not None:
+            min_id = start_token.min_id
+        else:
+            min_id = self.app.config.get("SECURITY_SCANNER_INDEXING_MIN_ID")
```
josephschorr: We won't be using an INDEXING_MIN anymore. Instead, this class should (for now), take a whitelist of namespaces and only allow manifests in those namespaces to be indexed.

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,125 @@
...
+
+        return ScanToken(max_id + 1)
+
+
+    def register_model_cleanup_callbacks(self, data_model_config):
+        raise NotImplementedError
+
+
+    @property
+    def legacy_api_handler(self):
```
josephschorr: Have this property raise an unsupported exception

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,125 @@
...
+
+        for candidate, abt, num_remaining in iterator:
+            # TODO(alecmerdler): Call out to Clair v4 to `/index` the manifest
+            pass
+
+        return ScanToken(max_id + 1)
+
+
+    def register_model_cleanup_callbacks(self, data_model_config):
+        raise NotImplementedError
```
josephschorr: Have this `pass`

data/secscan_model/secscan_v4_model.py
alecmerdler: Is that a method on `IntEnum`?

josephschorr: No idea; you wrote it... not sure what it means :)

alecmerdler: Oh whoops

config.py
```diff
@@ -472,6 +472,9 @@ def create_transaction(db):
     # The version of the API to use for the security scanner.
     SECURITY_SCANNER_API_VERSION = "v1"
 
+    # Namespace whitelist for security scanner.
+    SECURITY_SCANNER_INDEXING_WHITELIST = []
```
josephschorr: Name this for V4, as it'll only be used for that until we're ready for a full rollout.

data/database.py
```diff
@@ -1840,6 +1840,47 @@ class RepoMirrorConfig(BaseModel):
...
+
+
+@unique
+class IndexerVersion(IntEnum):
+    """
+    Possible versions of security indexers.
+    """
+
+    V2 = 0
+    V4 = 1
```
josephschorr: Sets these to `2` and `4` to make it easier to see in the DB

data/database.py
```diff
@@ -1840,6 +1840,47 @@ class RepoMirrorConfig(BaseModel):
...
+
+class ManifestSecurityStatus(BaseModel):
+    """
+    Represents the security scan status for a particular container image manifest. 
+
+    Intended to replace the `security_indexed` and `security_indexed_engine` fields 
+    on the `Image` model.
+    """
+
+    manifest = ForeignKeyField(Manifest)
```
josephschorr: Add the `repository` here, to make it easier for us to GC entries in this table

data/database.py
```diff
@@ -1840,6 +1840,47 @@ class RepoMirrorConfig(BaseModel):
...
+    index_status = ClientEnumField(IndexStatus)
+    error_json = JSONField(default={})
+    last_indexed = DateTimeField(default=datetime.utcnow)
+    indexer_hash = CharField(max_length=128)
```
josephschorr: Add `index=True` for quick lookup of things to reindex

data/database.py
```diff
@@ -1840,6 +1840,47 @@ class RepoMirrorConfig(BaseModel):
...
+    Represents the security scan status for a particular container image manifest. 
+
+    Intended to replace the `security_indexed` and `security_indexed_engine` fields 
+    on the `Image` model.
+    """
+
+    manifest = ForeignKeyField(Manifest)
+    index_status = ClientEnumField(IndexStatus)
+    error_json = JSONField(default={})
+    last_indexed = DateTimeField(default=datetime.utcnow)
```
josephschorr: Add index=True for quick lookup of things to potentially reindex if we support that based on date/time

data/database.py
```diff
@@ -1840,6 +1840,47 @@ class RepoMirrorConfig(BaseModel):
...
+    indexer_hash = CharField(max_length=128)
+    indexer_version = ClientEnumField(IndexerVersion)
+    metadata_json = JSONField(default={})
```
josephschorr: This model needs a few compound indexes, like something like:

- ('repository, 'index_status') - For quick lookup of things to index
- ('repository', 'indexer_hash') - Ditto for the whitelisted reindexing

alecmerdler: But a `Repository` can have many `Manifests`, so it won't be a unique index, correct?

josephschorr: Correct. It should not be unique.

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,136 @@
...
+                              .scalar())
+
+        min_id = id_for(fn.Min(database.Manifest.id))
+        max_id = id_for(fn.Max(database.Manifest.id))
+        
+        if max_id is None or min_id is None or min_id > max_id:
+            return None
+
+        # FIXME(alecmerdler): Filter out any `Manifests` that are still being uploaded
+        def not_indexed_query():
```
alecmerdler: @josephschorr What do you think of these three queries? Are they accurate?

josephschorr: Import the tables from database: `from data.database import Manifest` and these need the namespace filter added.

alecmerdler: Done

config.py
```diff
@@ -472,6 +472,9 @@ def create_transaction(db):
     # The version of the API to use for the security scanner.
     SECURITY_SCANNER_API_VERSION = "v1"
 
+    # Namespace whitelist for security scanner.
+    SECURITY_SCANNER_V4_INDEXING_WHITELIST = []
```
josephschorr: drop indexing and make it `namespace_whitelist`

data/database.py
```diff
@@ -1901,6 +1901,53 @@ class RepoMirrorConfig(BaseModel):
     root_rule = ForeignKeyField(RepoMirrorRule)
 
 
+@unique
+class IndexStatus(IntEnum):
+    """
+    Possible statuses of manifest security scan progress.
+    """
+
```
josephschorr: Remove extra newline

data/database.py
```diff
@@ -1901,6 +1901,53 @@ class RepoMirrorConfig(BaseModel):
...
+    FAILED = -1
+    UNSUPPORTED = 0
+    TIMED_OUT = 1
+    WAITING = 2
```
josephschorr: If we're going to stick with the smaller terms, can we put comments explaining what they mean? What does `WAITING` mean here? Waiting to start? To complete?

data/database.py
```diff
@@ -1901,6 +1901,53 @@ class RepoMirrorConfig(BaseModel):
...
+    IN_PROGRESS = 3
+    COMPLETED = 4
+
+
+@unique
+class IndexerVersion(IntEnum):
+    """
+    Possible versions of security indexers.
+    """
+
```
josephschorr: Ditto remove newline

data/database.py
```diff
@@ -1901,6 +1901,53 @@ class RepoMirrorConfig(BaseModel):
...
+    error_json = JSONField(default={})
+    last_indexed = DateTimeField(default=datetime.utcnow, index=True)
+    indexer_hash = CharField(max_length=128, index=True)
+    indexer_version = ClientEnumField(IndexerVersion)
+    metadata_json = JSONField(default={})
+
+    class Meta:
+        database = db
+        read_only_config = read_only_config
+        indexes = ((("repository", "index_status"), False), (("repository", "indexer_hash"), False))
```
josephschorr: one index per line

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,152 @@
...
+
+        status = None
+        try:
+            status = ManifestSecurityStatus.get(manifest=manifest_or_legacy_image._db_id)
+        except ManifestSecurityStatus.DoesNotExist:
+            return SecurityInformationLookupResult.with_status(ScanLookupStatus.NOT_YET_INDEXED)
+
+        if status.index_status == IndexStatus.FAILED:
+            return SecurityInformationLookupResult.with_status(ScanLookupStatus.FAILED_TO_INDEX)
+        if status.index_status == IndexStatus.UNSUPPORTED:
```
josephschorr: newline before this line

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,152 @@
...
+            return SecurityInformationLookupResult.with_status(
+                ScanLookupStatus.UNSUPPORTED_FOR_INDEXING
+            )
+        if status.index_status in [
```
josephschorr: newline before this line.

In general: add a blank line before any branch statement

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def perform_indexing(self, start_token=None):
+        whitelisted_namespaces = self.app.config.get("SECURITY_SCANNER_V4_INDEXING_WHITELIST", [])
+
+        def eligible_manifests(base_query):
+            return (
+                base_query.join(Repository)
+                .join(User)
+                .where(
+                    User.username << whitelisted_namespaces if len(whitelisted_namespaces) else True
```
josephschorr: If the list is empty, just return the query unmodified

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,152 @@
...
+                )
+            )
+
+        def id_for(id_fn):
+            return eligible_manifests(Manifest.select(id_fn, Repository, User)).scalar()
+
+        min_id = id_for(fn.Min(Manifest.id))
+        max_id = id_for(fn.Max(Manifest.id))
```
josephschorr: This `max` is going to be very, very, very slow.

alecmerdler: Same with the `min`, I assume then. Should we remove the namespace filtering logic here (since it is already in the `batch_query` argument to `yield_random_entries()`) and just use:
```python
min_id = Manifest.select(fn.Min(Manifest.id)).scalar()
max_id = Manifest.select(fn.Max(Manifest.id)).scalar()
```
Would this work?

josephschorr: `min` might be a bit faster, but yeah, we're going to filter it anyway.

data/database.py
alecmerdler: `black` doesn't let me

josephschorr: Seriously? booooo

data/database.py
alecmerdler: `black` doesn't let me

data/secscan_model/secscan_v4_model.py
alecmerdler: Done

data/database.py
```diff
@@ -1901,6 +1903,51 @@ class RepoMirrorConfig(BaseModel):
...
 
 
+@unique
+class IndexStatus(IntEnum):
+    """
+    Possible statuses of manifest security scan progress.
+    """
+
+    MANIFEST_UNSUPPORTED = -2
+    FAILED = -1
```
josephschorr: This is missing states like "Not yet indexed", which we need in case we want to start creating these rows optimistically

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,157 @@
...
+
+        def index_error_query():
+            return (
+                eligible_manifests(Manifest.select())
+                .switch(Manifest)
+                .join(ManifestSecurityStatus)
+                .where(ManifestSecurityStatus.index_status == IndexStatus.FAILED)
+            )
+
+        def needs_reindexing_query():
```
josephschorr: Take the hash in to the method and define it as a constant at the top for now

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,157 @@
...
+            indexer_hash = ""
+            return (
+                eligible_manifests(Manifest.select())
+                .switch(Manifest)
+                .join(ManifestSecurityStatus)
+                .where(ManifestSecurityStatus.indexer_hash != indexer_hash)
+            )
+
+        # 4^log10(total) gives us a scalable batch size into the billions.
+        batch_size = int(4 ** log10(max(10, max_id - min_id)))
```
josephschorr: We should likely move this into a method, that takes in a query modifier, and then just yield from all three or somesuch

data/database.py
alecmerdler: We can easily add that in the future. Isn't it sufficient to simply check if the row exists as confirmation that it hasn't been indexed?

josephschorr: Yeah, I guess. Let's leave it for now then.

data/secscan_model/secscan_v4_model.py
alecmerdler: Switched to use [`itertools.chain()`](https://docs.python.org/3/library/itertools.html#itertools.chain) instead.

data/secscan_model/secscan_v2_model.py
```diff
@@ -59,6 +59,8 @@ def __init__(self, app, instance_keys, storage):
         self.app = app
         self._legacy_secscan_api = None
 
+        assert app.config
```
josephschorr: Why this assert?

alecmerdler: Leftover from debugging, removed.

data/migrations/versions/c9c6d3ba8ae2_add_manifestsecuritystatus_table.py
```diff
@@ -0,0 +1,76 @@
...
+        ],
+    )
+    # ### end population of test data ### #
+
+
+def downgrade(tables, tester, progress_reporter):
+    op = ProgressWrapper(original_op, progress_reporter)
+
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index('manifestsecuritystatus_repository_id_indexer_hash', table_name='manifestsecuritystatus')
```
josephschorr: You can remove the drop_index calls, since the drop_table will do that too

alecmerdler: Thanks

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,163 @@
...
+    def __init__(self, app, instance_keys, storage):
+        self.app = app
+
+        validator = SecurityConfigValidator(
+            app.config.get("FEATURE_SECURITY_SCANNER", False),
+            app.config.get("SECURITY_SCANNER_ENDPOINT"),
+        )
+
+        if not validator.valid():
+            logger.warning("Failed to validate security scanner V2 configuration")
```
josephschorr: Fix comment

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    Implementation of the security scanner interface for Clair V4 API-compatible implementations.
+    """
+
+    def __init__(self, app, instance_keys, storage):
+        self.app = app
+
+        validator = SecurityConfigValidator(
+            app.config.get("FEATURE_SECURITY_SCANNER", False),
+            app.config.get("SECURITY_SCANNER_ENDPOINT"),
```
josephschorr: This endpoint field is going to have to be different for V4, and we're going to need a different validator now that i think of it.

So: Please create a new validator in this package, and have it take in a V4 endpoint, and (for now), have it always return False for `valid`

alecmerdler: Done

data/secscan_model/secscan_v4_model.py
alecmerdler: Done

config.py
```diff
@@ -451,9 +451,12 @@ def create_transaction(db):
     FEATURE_SECURITY_SCANNER = False
     FEATURE_SECURITY_NOTIFICATIONS = False
 
-    # The endpoint for the security scanner.
+    # The endpoint for the (deprecated) V2 security scanner.
     SECURITY_SCANNER_ENDPOINT = "http://192.168.99.101:6060"
 
+    # The endpoint for the V4 security scanner.
+    SECURITY_SCANNER_ENDPOINT_V4 = "http://192.168.99.101:6060"
```
josephschorr: Define this to empty for now

data/secscan_model/secscan_v4_model.py
```diff
@@ -0,0 +1,163 @@
...
+
+class V4SecurityScanner(SecurityScannerInterface):
+    """
+    Implementation of the security scanner interface for Clair V4 API-compatible implementations.
+    """
+
+    def __init__(self, app, instance_keys, storage):
+        self.app = app
+
+        validator = V2SecurityConfigValidator(
```
josephschorr: This should be V4 validator

data/migrations/versions/c9c6d3ba8ae2_add_manifestsecuritystatus_table.py
```diff
@@ -0,0 +1,118 @@
...
+        ["manifest_id"],
+        unique=False,
+    )
+    op.create_index(
+        "manifestsecuritystatus_repository_id",
+        "manifestsecuritystatus",
+        ["repository_id"],
+        unique=False,
+    )
+    op.create_index(
```
josephschorr: You'll need to regenerate this for the changed indexes

alecmerdler: Done

josephschorr:
@alecmerdler Make sure you coordinate with whoever is on call, so that they know once this is merged there is an upcoming database migration

PR #221, opened 2020-02-06
The incremented counters are for v2 as a whole

josephschorr:
Currently, we mark them as v2_1, which isn't fully correct, as we serve both v2_1 and v2_2 through the same code paths


jzelinskie:
Can we do something just slightly smarter here by checking `request.accept_headers` to determine if the client intended to do v2_1 or v2_2? It's a shame to throw that data out entirely.

josephschorr:
We can on the manifest side, if you like, but not on the blob related ones.

josephschorr:
@jzelinskie Updated

PR #222, opened 2020-02-06
Fix base64 encode/decode error in config app.

kurtismullins:
### Description of Changes

This PR fixes one failing test in the configuration app's unit tests.

#### Changes:

* Explicitly read certificate files as bytes (previous behavior) and then convert them to unicode/strings after base64 encoding them (achieves previous behavior) in the configuration app.

#### Issue:
- Fixing unit test failures in the Python3 branch as we progress in our transition.

**TESTING**

- Configuration App tests should no longer fail

**BREAKING CHANGE**

n/a

---

## Reviewer Checklist

- [ ] It works!


kurtismullins:
*Note:* The tests still fail as expected but a base64-related error in the configuration app's tests is no longer occurring.

PR #226, opened 2020-02-13
Handle integrity errors being raised in v2 auth

josephschorr:
Instead of raising an unhandled exception, we now check for the repository during the create process. If it was created since we last checked, we simply use the reference. Otherwise, we return an error to the user.

Fixes https://issues.redhat.com/projects/PROJQUAY/issues/PROJQUAY-289

### DIFF
--- a/data/model/repository.py
+++ b/data/model/repository.py
@@ -79,13 +79,32 @@ def create_repository(
     yesterday = datetime.now() - timedelta(days=1)
 
     with db_transaction():
-        repo = Repository.create(
-            name=name,
-            visibility=Repository.visibility.get_id(visibility),
-            namespace_user=namespace_user,
-            kind=Repository.kind.get_id(repo_kind),
-            description=description,
-        )
+        # Check if the repository exists to avoid an IntegrityError if possible.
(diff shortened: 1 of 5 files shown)


data/model/repository.py
```diff
@@ -79,13 +79,32 @@ def create_repository(
...
+                visibility=Repository.visibility.get_id(visibility),
+                namespace_user=namespace_user,
+                kind=Repository.kind.get_id(repo_kind),
+                description=description,
+            )
+        except IntegrityError as ie:
+            # NOTE: This is a just-in-case fallback.
+            try:
+                Repository.get(namespace_user=namespace_user, name=name)
+                return None
```
kleesc: Isn't this already enforced by the database? If there's an integrity error when trying to create, then that repo shouldn't exists. And if for some reason it does, should it return None or that repo?

Otherwise LGTM

josephschorr: Well, the integrity error could be due to another reason, hence the extra check and the extra log. I just want to make sure we log for real unknown errors, and not otherwise.

PR #238, opened 2020-02-26
Fix performance of the global prom stats worker queries

josephschorr:
- Change the counts to use the MySQL information schema for estimates
- Change the lookup of active users to skip the unnecessary heavy filter


data/estimate.py
```diff
@@ -0,0 +1,11 @@
+def mysql_estimate_row_count(model_cls, db):
+    """ Uses the information_schema to retrieve the row count for a table. """
+    query = "SELECT table_name, table_rows FROM information_schema.tables WHERE table_schema = DATABASE() AND table_name = %s"
+    cursor = db.execute_sql(query, (model_cls._meta.table_name,))
+    res = cursor.fetchone()
```
jakedt: will this raise an error if there isn't a result? are you ok with that?

data/estimate.py
```diff
@@ -0,0 +1,11 @@
+def mysql_estimate_row_count(model_cls, db):
+    """ Uses the information_schema to retrieve the row count for a table. """
+    query = "SELECT table_name, table_rows FROM information_schema.tables WHERE table_schema = DATABASE() AND table_name = %s"
```
jakedt: you don't need to select out the table name here if you're using an `=` where clause

data/estimate.py
josephschorr: It will and I am, as it shouldn't happen in a working database

data/estimate.py
josephschorr: Good point.

PR #246, opened 2020-03-04
Optimize repository lookup queries to meet the expected maximums

josephschorr:
We were accidentally looking up more data that strictly allowed

Adds some additional assertions and testing as well

Fixes https://issues.redhat.com/browse/PROJQUAY-439

### DIFF
--- a/data/model/oci/tag.py
+++ b/data/model/oci/tag.py
@@ -191,6 +191,7 @@ def get_most_recent_tag_lifetime_start(repository_ids):
     specified repository or None if none.
     """
     assert len(repository_ids) > 0 and None not in repository_ids
+    assert len(repository_ids) <= 100
 
     query = (
         Tag.select(Tag.repository, fn.Max(Tag.lifetime_start_ms))
--- a/data/model/repository.py
+++ b/data/model/repository.py
@@ -386,8 +386,10 @@ def get_filtered_matching_repositories(
         if filter_user is None:
             return []
 
+        # NOTE: We add the offset to the limit here to ensure we have enough results
+        # for the take's we conduct below.
         iterator = _filter_repositories_visible_to_user(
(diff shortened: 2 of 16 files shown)


josephschorr:
@dongboyan77 Mind doing a QE pass?

josephschorr:
@dongboyan77 Actually, looks like I need to make more changes for this. So hold off.

data/model/oci/tag.py
```diff
@@ -191,6 +191,7 @@ def get_most_recent_tag_lifetime_start(repository_ids):
     specified repository or None if none.
     """
     assert len(repository_ids) > 0 and None not in repository_ids
+    assert len(repository_ids) <= 100
```
alecmerdler: Interesting approach. It's impossible to know this requirement just from the function caller's perspective of the function signature. Is that a normal pattern?

dongboyan77:
@josephschorr OK, I will verify it when you finished

data/model/oci/tag.py
josephschorr: It is documented in the interface that calls this, but yeah, in Python there isn't really a good way to know.

josephschorr:
@dongboyan77 Please verify when you have a chance.

dongboyan77:
@josephschorr OK

dongboyan77:
@josephschorr hi, QE complete verification:
1. Make sure paginated search works for them
Actual results: By default config, show 10 repos per page
2. Make sure autocompletion works for them
Screenshot from 2020-03-30 18-50-20

dongboyan77:
Found one issue: there is no Grid View in repo page
Screenshot from 2020-03-30 18-53-49

josephschorr:
@dongboyan77 This change removes the grid view entirely, as it cannot be paginated easily

dongboyan77:
@josephschorr OK, got it

PR #251, opened 2020-03-05
PROJQUAY-70 - default wildcard mirror tags

thomasmckay:
https://issues.redhat.com/browse/PROJQUAY-70

When no mirror tag pattern is specified (empty string), default to wildcard (*). Additional clean up of mirror tag list to strip whitespace and remove duplicates.

### DIFF
--- a/data/model/repo_mirror.py
+++ b/data/model/repo_mirror.py
@@ -527,6 +527,23 @@ def validate_rule(rule_type, rule_value):
         )
 
 
+def clean_rule(rule_type, rule_value):
+    if rule_type != RepoMirrorRuleType.TAG_GLOB_CSV:
+        return rule_value
+
+    cleaned_rule_value = []
+    for value in rule_value:
+        value = value.strip()
+        if value == "":
+            cleaned_rule_value.append("*")
+        else:
+            cleaned_rule_value.append(value)
+
+    cleaned_rule_value = list(dict.fromkeys(cleaned_rule_value))  # remove duplicates
+
+    return cleaned_rule_value
+
+
 def create_rule(
     repository,
     rule_value,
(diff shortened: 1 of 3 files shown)


data/model/repo_mirror.py
```diff
@@ -527,6 +527,23 @@ def validate_rule(rule_type, rule_value):
         )
 
 
+def clean_rule(rule_type, rule_value):
```
kurtismullins: I'm not opposed to doing this but we should also do validation and input sanitation on the frontend (the endpoint). It would be nice to enforce the input we receive from API Requests and give the appropriate response instead of silently mutating their data.

data/model/test/test_repo_mirroring.py
```diff
@@ -245,3 +248,10 @@ def test_release_mirror(initialized_db):
     mirror = release_mirror(mirror, RepoMirrorStatus.FAIL)
     assert mirror.sync_retries_remaining == 3
     assert mirror.sync_start_date > original_sync_start_date
+
+
+@pytest.mark.parametrize(
+    "rule_value, cleaned_rule_value", [(["", " 1.1 ", "", "1.1"], ["*", "1.1"])]
```
kurtismullins: Can you also add the following test case?
```
"1.1, latest, "
```
(Note the trailing comma and extra space)

It *shouldn't* happen with JSON but a quick search reveals that *some* browsers will allow it. I am curious if, as an edge-case, we will accidentally inject `*` into a list of other tags.

data/model/repo_mirror.py
alecmerdler: Agreed, we should always avoid "magical" behavior whenever possible. Moving validation to the API layer and adding helpful tips in the frontend keeps business logic out of our data layer.

data/model/test/test_repo_mirroring.py
thomasmckay: that's not how it works: a list is expected. if that comes in as a string that's what will be saved.

PR #252, opened 2020-03-06
Add ability to filter user lookups in LDAP user system

josephschorr:
This allows customers to specify additional restrictions on the users
that are allowed to be used

Fixes https://issues.redhat.com/browse/PROJQUAY-226

### DIFF
--- a/config_app/js/core-config-setup/config-setup-tool.html
+++ b/config_app/js/core-config-setup/config-setup-tool.html
@@ -971,6 +971,34 @@
               </td>
             </tr>
             <tr>
+              <td>Additional User Filter Expression:</td>
+              <td>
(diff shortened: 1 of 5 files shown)


vwbusguy:
I was able to test this today with a tag provided by @thomasmckay .  It worked as expected.  I was able to test this with some student workers who hadn't signed in before.  The one in the security group was able to get in, the one who wasn't reported "user not found".

You might note in the documentation that the memberOf needs to include the full DN, including the base_dn.

josephschorr:
@vwbusguy Thanks for the test! I've added additional notes. PTAL and let me know what you think.

vwbusguy:
@josephschorr - Thanks! Looks good to me.

PR #254, opened 2020-03-07
Fix Well-Known Endpoint to Use Correct Hostname

alecmerdler:
### Description

Ensure that the keys use the actual installation hostname and not `quay.io`.

Addresses https://issues.redhat.com/browse/PROJQUAY-441

### DIFF
--- a/endpoints/wellknown.py
+++ b/endpoints/wellknown.py
@@ -1,7 +1,7 @@
 import json
 import logging
 
-from app import get_app_url
+from app import app, get_app_url
 from auth.decorators import require_session_login
 from auth.auth_context import get_authenticated_user
 from flask import Blueprint, make_response, redirect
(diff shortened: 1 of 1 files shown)


endpoints/wellknown.py
```diff
@@ -20,12 +20,14 @@ def app_capabilities():
     manifest_security = "%s/api/v1/repository/{namespace}/{reponame}/manifest/{digest}/security"
     manifest_security_tmpl = manifest_security % get_app_url()
 
+    app_name = ".".join(app.config["SERVER_HOSTNAME"].split(".")[::-1])
```
thomasmckay: Strip off port (eg. quay.example.com:8443)

josephschorr:
@alecmerdler @thomasmckay This change is incorrect. The *keys* of the dictionary are to represent that this is a defined API for "quay.io" the *product*, not the website. Today, we'd probably want to change it to `io.projectquay`, since we're now open source, but it should NOT reflect the server hostname.

alecmerdler:
@josephschorr The reason I made this fix is because I discovered that the Container Security Operator [builds the keys based on the hostname](https://github.com/quay/container-security-operator/blob/master/secscan/wellknown.go#L34), meaning that it actually doesn't work for any registry other than Quay.io right now! Should the CSO be fixed instead?

josephschorr:
@alecmerdler How did we not notice that before?

Yes, the CSO should be fixed, not Quay

alecmerdler:
Closing in favor of https://github.com/quay/container-security-operator/pull/30.

PR #255, opened 2020-03-09
Add config tool support for configuring ephemeral builds

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-333

### DIFF
--- a/buildman/manager/executor.py
+++ b/buildman/manager/executor.py
@@ -8,6 +8,7 @@ import threading
 import time
 import uuid
 
+from abc import ABCMeta, abstractmethod
 from functools import partial, wraps
 
 import boto.ec2
@@ -82,6 +83,10 @@ class BuilderExecutor(object):
 
         default_websocket_scheme = "wss" if app.config["PREFERRED_URL_SCHEME"] == "https" else "ws"
         self.websocket_scheme = executor_config.get("WEBSOCKET_SCHEME", default_websocket_scheme)
+
+    @abstractmethod
+    def validate(self, executor_config):
+        """ Validates the given config, returning a tuple of (okay, err_msg). """
 
     @property
     def name(self):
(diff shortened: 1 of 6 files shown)


buildman/manager/executor.py
```diff
@@ -390,17 +423,30 @@ def __init__(self, *args, **kwargs):
             "BUILDER_VM_CONTAINER_IMAGE", "quay.io/quay/quay-builder-qemu-coreos:stable"
         )
 
+    def validate(self, executor_config):
+        try:
+            create_job = self._build_request(executor_config, "POST", self._jobs_path(), json={})
+            if int(create_job.status_code / 100) != 400:
```
kleesc: Should this be `int(status_code / 100) == 4`?

util/config/validators/validate_ephemeral_builder.py
```diff
@@ -0,0 +1,48 @@
...
+        for executor_config in executors:
+            executor_kind = executor_config.get("EXECUTOR")
+            if executor_kind is None:
+                raise ConfigValidationException("Missing executor kind")
+
+            if executor_kind == "kubernetes":
+                okay, err_msg = KubernetesExecutor.validate(executor_config)
+                if not okay:
+                    raise ConfigValidationException(err_msg)
+            elif kubernetes == "ec2":
```
kleesc: Did you mean `executor_kind`?

util/config/validators/validate_ephemeral_builder.py
```diff
@@ -0,0 +1,48 @@
...
+            return
+
+        if not config.get("BUILD_MANAGER"):
+            return
+
+        build_manager_config = config["BUILD_MANAGER"]
+        if build_manager_config[0] != "ephemeral":
+            return
+
+        if len(build_manager_config) != 2:
```
kleesc: Move this above to before checking for ephemeral.

buildman/manager/executor.py
josephschorr: No, because then it would succeed on auth errors, which we're explicitly trying to catch.

util/config/validators/validate_ephemeral_builder.py
josephschorr: Yes, good catch.

util/config/validators/validate_ephemeral_builder.py
josephschorr: Why? We check if there are any elements above the ephemeral check.

kleesc: The above will raise an IndexError instead of a ConfigValidationException.

buildman/manager/executor.py
kleesc: Wouldn't the current check return a single digit though? 
e.g `401 // 100` is `4`, not `400`

josephschorr: Oh... good catch. Copy-paste issue.

util/config/validators/validate_ephemeral_builder.py
josephschorr: `        if not config.get("BUILD_MANAGER"):` will catch it, but I'll change the order to make it nicer

buildman/manager/executor.py
```diff
@@ -390,17 +423,30 @@ def __init__(self, *args, **kwargs):
             "BUILDER_VM_CONTAINER_IMAGE", "quay.io/quay/quay-builder-qemu-coreos:stable"
         )
 
+    def validate(self, executor_config):
+        try:
+            create_job = self._build_request(executor_config, "POST", self._jobs_path(), json={})
+            if create_job.status_code != 400:
```
kleesc: This should check for `==` instead of `!=`. Also, still need that integer division to catch __all__ 4XX:
`if create_job.status_code // 100 == 4:`

josephschorr: I think you might be misinterpreting the check here. The check is to see if we get a 400; if we do, then we know we have permission to start a Job, and it failed because we *purposefully* gave it an empty body. If we get a *non-400* (like a 401 or 403), we fail, because then we know we don't have permission.

kleesc: Ahh. Gotcha.

kurtismullins:
@josephschorr and @kleesc 
Is this Pull Request still relevant now that the buildman rewrite is underway and the config_app module has been removed/deprecated in favor of quay/config-tool?

flavianmissi:
This will likely not go in as is, as it's pretty old, so I'm closing it.

PR #258, opened 2020-03-10
Ensure that temporary tags cannot be written for repositories marked for deletion

josephschorr:
The purge code assumes that once it has deleted tags, no new ones will
be added, so we need to enforce that state

### DIFF
--- a/data/model/gc.py
+++ b/data/model/gc.py
@@ -77,6 +77,10 @@ def purge_repository(repo, force=False):
     """
     assert repo.state == RepositoryState.MARKED_FOR_DELETION or force
 
+    # Update the repo state to ensure nothing else is written to it.
+    repo.state = RepositoryState.MARKED_FOR_DELETION
+    repo.save()
+
     # Delete the repository of all Appr-referenced entries.
     # Note that new-model Tag's must be deleted in *two* passes, as they can reference parent tags,
     # and MySQL is... particular... about such relationships when deleting.
--- a/data/model/tag.py
+++ b/data/model/tag.py
@@ -19,6 +19,7 @@ from data.model import (
 from data.database import (
     RepositoryTag,
     Repository,
+    RepositoryState,
     Image,
     ImageStorage,
     Namespace,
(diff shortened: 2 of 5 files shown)


data/model/gc.py
```diff
@@ -77,6 +77,10 @@ def purge_repository(repo, force=False):
     """
     assert repo.state == RepositoryState.MARKED_FOR_DELETION or force
 
+    # Update the repo state to ensure nothing else is written to it.
+    repo.state = RepositoryState.MARKED_FOR_DELETION
```
kurtismullins: :100:

data/model/tag.py
```diff
@@ -396,20 +397,27 @@ def create_temporary_hidden_tag(repo, image_obj, expiration_s):
...
-        lifetime_start_ts=now_ts,
-        lifetime_end_ts=expire_ts,
-        hidden=True,
-    )
-    return tag_name
+
+    # Ensure the repository is not marked for deletion.
+    with db_transaction():
+        current = Repository.get(id=repo)
+        if current.state == RepositoryState.MARKED_FOR_DELETION:
```
kurtismullins: Do you think this will heavily impact the database load?

data/model/test/test_tag.py
```diff
@@ -352,13 +353,22 @@ def test_get_most_recent_tag(initialized_db):
     # Create a hidden tag that is the most recent.
     repo = model.repository.get_repository("devtable", "simple")
     image = model.tag.get_tag_image("devtable", "simple", "latest")
-    model.tag.create_temporary_hidden_tag(repo, image, 10000000)
+    model.tag.create_temporary_hidden_tag(repo, image, 10000000) is not None
```
kurtismullins: I haven't ran this but should this be an assert statement?

data/model/tag.py
josephschorr: I don't believe so, as we already create this under a transaction elsewhere and it should only be per uploaded blob. That being said, it will have a small impact, but its the only safe way.

data/model/test/test_tag.py
josephschorr: yes! good catch!

kurtismullins:
(nitpick)

From your commit message:
>The purge code assumes that ones it has deleted tags

```
s/ones/once
```

PR #261, opened 2020-03-10
PROJQUAY-454 - version from env or git

thomasmckay:
https://issues.redhat.com/browse/PROJQUAY-454

    Version is determined in following order:
    1. First line VERSION file
    2. environment variable QUAY_RELEASE
    3. GIT_HEAD file
    4. git rev-parse HEAD

Added a quay-entrypoint command of "version" that will print the first non-blank line from VERSION.

VERSION file is created during container build. The first line will be the ARG QUAY_VERSION and the second will be `git describe --tags`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,8 +7,6 @@ screenshots
 tools
 test/data/registry
 venv
-.git
-!.git/HEAD
 .gitignore
 .github
 Bobfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,31 @@
+# Builder
+
+FROM centos:7 as builder
+
+ENV QUAYDIR /quay-registry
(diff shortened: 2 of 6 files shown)


.dockerignore
```diff
@@ -7,8 +7,6 @@ screenshots
 tools
 test/data/registry
 venv
-.git
-!.git/HEAD
```
josephschorr: This will result in the `.git` directory being in the image, which is a no-go.

thomasmckay: @josephschorr no, i remove in the Dockerfile (i'll update Dockerfile.rhel7, etc. when we're agreed on this PR)

thomasmckay:
won't work with quay.io builders

PR #263, opened 2020-03-11
Mark any server hostnames with `quay.io` as reserved

josephschorr:
Usage of `quay.io` in a server hostname breaks the superuser API
(on purpose) to ensure that API is never available in production.

This change ensures this is messaged properly.

Fixes https://issues.redhat.com/browse/PROJQUAY-80

### DIFF
--- a/config_app/js/core-config-setup/config-setup-tool.html
+++ b/config_app/js/core-config-setup/config-setup-tool.html
@@ -56,6 +56,12 @@
             <tr>
               <td>Server Hostname:</td>
               <td>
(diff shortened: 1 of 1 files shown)


josephschorr:
Screen Shot 2020-03-11 at 2 21 31 PM

config_app/js/core-config-setup/config-setup-tool.html
```diff
@@ -56,6 +56,12 @@
             <tr>
               <td>Server Hostname:</td>
               <td>
+                <div class="co-alert co-alert-danger" ng-if="config.SERVER_HOSTNAME.indexOf('quay.io') >=0 "
```
alecmerdler: Can we add this as validation to `schema.py` as well?

PR #273, opened 2020-03-18
Remove the use of a transaction for User and Repository delete

josephschorr:
While a transaction is obviously safer, with the number of tables
and rows referencing these tables now, a transaction is potentially
locking up a significant chunk of the database. Since we're already
performing cleanup before calling the delete, including disabling
new data being written for the User or Repository, deletion without
a transaction should (usually) be sufficient; if it isn't, an
IntegrityError will be raised, and the workers can retry continuing
the GC operation

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -145,11 +145,12 @@ def delete_instance_filtered(instance, model_class, delete_nullable, skip_transi
 
     Callers *must* ensure that any models listed in the skip_transitive_deletes must be capable
     of being directly deleted when the instance is deleted (with automatic sorting handling
-    dependency order).
+    dependency order) - for example, the Manifest and ManifestBlob tables for Repository will
+    always refer to the *same* repository when Manifest references ManifestBlob, so we can safely
+    skip transitive deletion for the Manifest table.
 
-    For example, the RepositoryTag and Image tables for Repository will always refer to the
-    *same* repository when RepositoryTag references Image, so we can safely skip
-    transitive deletion for the RepositoryTag table.
+    Callers *must* catch IntegrityError's raised, as this method will *not* delete the instance
+    under a transaction, to avoid locking the database.
     """
     # We need to sort the ops so that models get cleaned in order of their dependencies
     ops = reversed(list(instance.dependencies(delete_nullable)))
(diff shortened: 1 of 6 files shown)


jzelinskie:
I think removing this transaction is fine, but I'm curious as to what's causing it to take so long. Spanning across many tables might be the cause -- but I'd like to be 100% sure of that because this is supposed to be the bastion of ensuring that the delete has properly occurred.

josephschorr:
@jzelinskie Yeah, I've seen evidence of the transaction timing out, which likely means it is taking a long time. Granted, this was also before we started pre-deleting everything, but I still suspect it is taking an unusually long time given the number of tables involved. For safety, I'd prefer to remove the transaction regardless.

kleesc:
Shouldn't most tables be skipped because of `skip_transitive_delete`?

josephschorr:
@kleesc No. `skip_transitive_delete` means that if you have a table like `RepositoryTag` that references `Repository` and `Image`, that it won't try to delete all references `Image` rows *through* `RepositoryTag`, since peewee doesn't know that the `repository` field on both those tables are the same.

Peewee will still generate (correctly) queries of the form `delete * from {Table} where repository={the repo}` for all tables with a foreign key constraint.

jzelinskie:
I spent a few minutes seeing if we could potentially keep the transaction and rather change its isolation level. [According to the documentation](https://dev.mysql.com/doc/refman/8.0/en/innodb-transaction-isolation-levels.html#isolevel_read-committed), UPDATES and DELETES should only be doing row-level locking. This would imply that if this was slow, it would be because of some kind of contention on the _rows_ and not the _tables_.

I think we need more evidence as to what the problem is before we go forward with applying a possible solution.

josephschorr:
@jzelinskie Sure, but I'd wager it is *also* locking the indexes, and we have *quite* a few of those are high shared-access tables

jzelinskie:
That's for sure -- an interesting experiment would be to determine whether or not a DELETE with 0 rows selected for deletion still locks indexes on the table. Following this idea, we could have some preemptive queries that determine what tables we can skip in this final transaction.

josephschorr:
@jzelinskie I mean, we could just check them all to see if there are any matching rows, and then issue the transaction for the remainder... but that still provides incomplete protection, which makes it not worth it IMO.

jzelinskie:
The only issue I see with this code now is that if someone kills the process or the machine forcefully shuts down during one of these deletes, only a partial deletion will have occurred. This isn't the end of the world, but the final delete wouldn't have occurred yet, so it'd be visible that the delete never 100% finished.

josephschorr:
@jzelinskie Yeah, although I think that's okay, because then the repo delete will just be picked up again by the worker and continue

PR #275, opened 2020-03-18
Remove the shared registry model impl now that we have a single registry model

josephschorr:


jzelinskie:
I think removing this transaction is fine, but I'm curious as to what's causing it to take so long. Spanning across many tables might be the cause -- but I'd like to be 100% sure of that because this is supposed to be the bastion of ensuring that the delete has properly occurred.

PR #277, opened 2020-03-23
Fix status response codes to conform to the OCI spec

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-509


josephschorr:
Verified by running the full clients test suite as well

PR #278, opened 2020-03-23
Fix support for empty manifests in V2_2

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-510

### DIFF
--- a/test/registry/protocol_v2.py
+++ b/test/registry/protocol_v2.py
@@ -359,7 +359,7 @@ class V2Protocol(RegistryProtocol):
             "history": [history_for_image(image) for image in images],
         }
 
-        if images[-1].config:
+        if images and images[-1].config:
             config["config"] = images[-1].config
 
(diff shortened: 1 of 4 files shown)


test/registry/protocol_v2.py
```diff
@@ -359,7 +359,7 @@ def history_for_image(image):
             "history": [history_for_image(image) for image in images],
         }
 
-        if images[-1].config:
+        if images and images[-1].config:
```
alecmerdler: Can you explain what this is doing?

josephschorr: Sure. It is grabbing the config of the top layer in the history for use as the overall config of the manifest. V1 had config for each image (since each layer is its own image), while V2 only has config at the most recent layer.

alecmerdler: So there is no config for empty manifests?

josephschorr: There is, this is just part of the test and is necessary for us to override the default empty one

PR #279, opened 2020-03-23
Implement experimental support for OCI manifests

josephschorr:
Specification: https://github.com/opencontainers/image-spec

Epic: https://issues.redhat.com/browse/PROJQUAY-446
Fixes https://issues.redhat.com/browse/PROJQUAY-476
Fixes https://issues.redhat.com/browse/PROJQUAY-477


image/oci/test/test_oci_manifest.py
```diff
@@ -0,0 +1,168 @@
...
+    }
+  ],
+  "annotations": {
+    "com.example.key1": "value1",
+    "com.example.key2": "value2"
+  }
+}"""
+
+
+SAMPLE_REMOTE_MANIFEST = """{
```
kleesc, line 41: Is this one different from the one above?

image/oci/test/test_oci_manifest.py
```diff
@@ -0,0 +1,168 @@
...
+    assert not manifest.has_remote_layer
+    assert manifest.has_legacy_image
+    assert manifest.annotations == {u"com.example.key1": u"value1", u"com.example.key2": u"value2"}
+
+
+def test_parse_basic_remote_manifest():
+    manifest = OCIManifest(Bytes.for_string_or_unicode(SAMPLE_REMOTE_MANIFEST))
+    assert not manifest.is_manifest_list
+    assert (
+        manifest.digest == "sha256:dd18ed87a00474aff683cee7160771e043f1f0eadd780232715bc0678a984a5e"
```
kleesc, line 104: How is the sha computed in this test different from the one above?

image/oci/test/test_oci_manifest.py
```diff
@@ -0,0 +1,168 @@
...
+        },
+        "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7",
+        7023,
+    )
+
+    manifest = OCIManifest(Bytes.for_string_or_unicode(SAMPLE_MANIFEST))
+    assert manifest.get_manifest_labels(retriever) == {
+        u"com.example.key1": u"value1",
+        u"com.example.key2": u"value2",
+        "foo": "bar",
```
kleesc, line 153: Why are the first two unicode while the this one bytes?

kleesc: Can we remove the `u` prefixes then. It kinda suggests that the first two labels are parsed differently than the last.

kleesc:
Do we have/need registry tests for the new content types?

image/oci/test/test_oci_manifest.py
josephschorr: Sure

josephschorr:
> Do we have/need registry tests for the new content types?

@kleesc Yes, in the registry test suite. Look at the fixtures to see where I added those.

PR #284, opened 2020-03-25
Make sure we don't send zero value metrics to Prometheus in verbs

josephschorr:
The `inc` function rejects these with an exception

Fixes https://issues.redhat.com/browse/PROJQUAY-456


josephschorr:
@jzelinskie It shouldn't be, but there appears to be a bug somewhere in ACI generation that can lead to empty images. Since ACI is deprecated, I'd rather just patch this and wait for us to remove ACI.

PR #285, opened 2020-03-26
Stop relying on GitHub's admin permissions for a repository

josephschorr:
It appears GitHub has narrowed the permissions returned, so while we
might still have the ability to admin the *repo*, the permissions
list is returning `false`, so we disallow people to create triggers,
which is wrong.

Fixes https://issues.redhat.com/browse/PROJQUAY-523


josephschorr:
@kleesc It used too... that's what's so odd.

kleesc:
@josephschorr If it's a org repo, do they need to be an owner instead of member?
https://github.com/quay/quay/pull/285/files#diff-3d4fb223ed7ca8dc11006526fa4564cdL332

josephschorr:
@kleesc Its possible they changed the meaning, but previously when we requested `repo` scope, it meant we got admin on the repos. Anyway, I'm waiting for a response from GitHub and this is a workaround either way.

PR #289, opened 2020-03-27
PROJQUAY-524 - relax reading CHANGELOG.md

thomasmckay:
https://issues.redhat.com/browse/PROJQUAY-524

### DIFF
--- a/_init.py
+++ b/_init.py
@@ -31,7 +31,7 @@ config_provider = get_config_provider(
 def _get_version_number_changelog():
     try:
         with open(os.path.join(ROOT_DIR, "CHANGELOG.md")) as f:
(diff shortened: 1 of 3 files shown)


_init.py
```diff
@@ -31,7 +31,7 @@
 def _get_version_number_changelog():
     try:
         with open(os.path.join(ROOT_DIR, "CHANGELOG.md")) as f:
-            return re.search(r"(v[0-9]+\.[0-9]+\.[0-9]+)", f.readline()).group(0)
+            return f.readline()[4:-1]
```
kleesc: Probably better to just add to whatever the previous regex was not covering

kleesc: @thomasmckay I'm fine with this. Just need to make sure the name is on the first line.

PR #293, opened 2020-03-27
Change verbs to use a DerivedStorageForManifest table instead of DerivedStorageForImage

josephschorr:
This allows us to deprecate the DerivedStorageForImage table.

Fixes https://issues.redhat.com/browse/PROJQUAY-519


data/migrations/dba_operator/e8e27f36f87d-databasemigration.yaml
```diff
@@ -0,0 +1,62 @@
+
+---
+apiVersion: dbaoperator.app-sre.redhat.com/v1alpha1
```
alecmerdler, line 3: I thought we were going to start putting migrations in their own, separate PRs to decouple from application logic in case we need to roll back.

josephschorr: We can certainly do so; I created this PR before we decided to do so

josephschorr:
Closing since it has a data migration; I'll merge this change into the overall image deprecation PR

PR #298, opened 2020-03-31
Deprecation Decorator for API Resources

alecmerdler:
### Description

Introduces `@deprecated` decorator for API resources, which adds the [`Deprecation: true` header](https://tools.ietf.org/id/draft-dalal-deprecation-header-01.html#RFC7234) to the response. Applies this decorator to `getRepoImageSecurity`.

### Testing

```shell
$ TZ=UTC TEST=true PYTHONPATH="." py.test --verbose -x ./endpoints/api/test/test_deprecated_route.py
```

Addresses https://issues.redhat.com/browse/PROJQUAY-282
Addresses https://issues.redhat.com/browse/PROJQUAY-537

### DIFF
--- a/endpoints/wellknown.py
+++ b/endpoints/wellknown.py
@@ -24,7 +24,10 @@ def app_capabilities():
         "appName": "io.quay",
         "capabilities": {
             "io.quay.view-image": {"url-template": view_image_tmpl,},
-            "io.quay.image-security": {"rest-api-template": image_security_tmpl,},
+            "io.quay.image-security": {
+                "rest-api-template": image_security_tmpl,
+                "deprecated": True,
+            },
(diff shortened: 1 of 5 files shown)


endpoints/wellknown.py
```diff
@@ -24,7 +24,10 @@ def app_capabilities():
         "appName": "io.quay",
         "capabilities": {
             "io.quay.view-image": {"url-template": view_image_tmpl,},
-            "io.quay.image-security": {"rest-api-template": image_security_tmpl,},
+            "io.quay.image-security": {
+                "rest-api-template": image_security_tmpl,
+                "deprecated": True,
```
alecmerdler, line 29: @josephschorr Does it make sense to add this here?

josephschorr: Yup!

PR #304, opened 2020-04-01
Migrate from Travis+Makefiles to GitHub Workflows+Tox

jzelinskie:
### Description of Changes

Many quality of life changes around local development and continuous integration.

#### Changes:

* introduces GitHub workflows for CI
* removes travis and CI-specific makefiles and dockerfiles
* support registry tests via tox-docker
* update dockerfiles to use requirements-dev for test utilities
* add pylint to dev requirements
* fix existing tox configuration -- this should let us drop makefiles for ci/testing in a follow up
* init pyproject.toml -- PEP518 declares this a common configuration file for many tools
* add tox and remove black from dev requirements
* pylint: drop 2 space ident configuration
* canonicalize requirements.txt -- should improve diffing these files
* drop unused docutils dependency
* move all mock packages into requirements-dev
* merge requirements-dev and requirements-tests -- there were conflicting versions across these, hopefully this reduces maintenance burden
* remove yapf configuration -- quay uses black now

## Reviewer Checklist

- [ ] It works!


BillDett:
This looks like a fairly substantial change to Quay development workflow- we need to get together and review/understand as a team as it has a (hopefully) positive impact on day to day work.  Were we tracking this work anywhere in PROJQUAY?

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+    - uses: actions/checkout@v2
+    - name: Set up Python 2.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 2.7
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get install libgpgme-dev libldap2-dev libsasl2-dev
```
josephschorr, line 60: Is it possible to extract these deps into a file and then maybe use that both here and in the Dockerfile?

Dockerfile
```diff
@@ -43,7 +43,7 @@ COPY . .
 RUN scl enable python27 "\
     pip install --upgrade setuptools==44 pip && \
     pip install -r requirements.txt --no-cache && \
-    pip install -r requirements-tests.txt --no-cache && \
+    pip install -r requirements-dev.txt --no-cache && \
```
josephschorr, line 46: Why are we installing the dev deps now?

.github/workflows/CI.yaml
jzelinskie: Unfortunately, these packages have different names on Debian / RHEL ecosystems.

util/canonicaljson.py
```diff
@@ -15,5 +15,5 @@ def canonicalize(json_obj):
         sorted_obj = sorted({key: canonicalize(val) for key, val in json_obj.items()}.items())
         return collections.OrderedDict(sorted_obj)
     elif isinstance(json_obj, (list, tuple)):
-        return [canonicalize(val) for val in json_obj]
+        return sorted([canonicalize(val) for val in json_obj])
```
josephschorr, line 18: This is incorrect: It would alter the order of items in the list, and result in producing different JSON data

jzelinskie: Ah, this is a good point. It still fixes the bug, though! Maybe I could add a helper function in this module that recursively sorts lists if you don't care about the order? There really isn't an equivalent for sets in JSON.

josephschorr: Sure. I'd have it take in a sort key as well, so you can at least define the order for sorting. Then, make sure the sorting is off for any registry usage of this method

data/secscan_model/test/test_secscan_v4_model.py
```diff
@@ -295,15 +296,16 @@ def test_features_for():
...
+            Layer(
+                "sha256:b05ac1eeec8635442fa5d3e55d6ef4ad287b9c66055a552c2fd309c334563b0a",
+                "",
+                "",
+                4,
+                features_for(vuln_report),
+            )
+        ).to_dict()
+
+        assert json.dumps(canonicalize(features_for_sec_info)) == json.dumps(
```
alecmerdler, line 309: Thanks for fixing this!

util/canonicaljson.py
jzelinskie: While I expect this to rarely be used, I added it as a kwarg.

PR #306, opened 2020-04-02
PROJQUAY-551 - verify ldap user filter

thomasmckay:
https://issues.redhat.com/browse/PROJQUAY-551

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -147,11 +147,23 @@ class LDAPUsers(FederatedUsers):
         referral_dn = referral_uri[len("ldap:///") :]
         return referral_dn
 
+    def _clean_user_filter(self, query):
+        if not query:
+            return None
+
+        if query[0] == "(" and query[-1] == ")":
+            return query
+
+        if query[0] == "(" or query[-1] == ")":
+            raise ldap.LDAPError("Filter must be fully enclosed in parenthesis")
+
+        return "(%s)" % query
+
     def _add_user_filter(self, query):
         if not self._ldap_user_filter:
             return query
 
-        return u"(&({0}){1})".format(self._ldap_user_filter, query)
+        return u"(&{0}{1})".format(self._clean_user_filter(self._ldap_user_filter), query)
 
     def _ldap_user_search_with_rdn(self, conn, username_or_email, user_search_dn, suffix=""):
         query = u"(|({0}={2}{3})({1}={2}{3}))".format(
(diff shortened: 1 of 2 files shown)


data/users/externalldap.py
```diff
@@ -147,11 +147,21 @@ def _get_ldap_referral_dn(self, referral_exception):
         referral_dn = referral_uri[len("ldap:///") :]
         return referral_dn
 
+    def _clean_user_filter(self, query):
+        if not query:
+            return None
+
+        if self._ldap_user_filter[0] == "(":
```
josephschorr, line 154: Check for the ")" on the other end too and if not present, fail with `None`

data/users/externalldap.py
```diff
@@ -262,27 +272,33 @@ def at_least_one_user_exists(self):
 
         has_pagination = not self._force_no_pagination
         with self._ldap.get_connection() as conn:
+            search_flt = self._clean_user_filter(self._ldap_user_filter)
```
josephschorr, line 275: Can you add a test for this method being filtered in the LDAP test suite? https://github.com/quay/quay/blob/master/test/test_ldap.py#L632

data/users/externalldap.py
thomasmckay: This will be picked up by the verification on the ldap side. This is mostly just to catch the cases where user puts in something just as a straight string like "someOtherField=someOtherValue". Anything more complicated than that we'll leave to the validation.

data/users/externalldap.py
thomasmckay: added unit test

data/users/externalldap.py
```diff
@@ -147,11 +147,20 @@ def _get_ldap_referral_dn(self, referral_exception):
         referral_dn = referral_uri[len("ldap:///") :]
         return referral_dn
 
+    def _clean_user_filter(self, query):
+        if not query:
+            return None
+
+        if query[0] == "(":
+            return query
```
josephschorr, line 155: Please add `assert query[-1] == ')'`  here

test/test_ldap.py
```diff
@@ -644,6 +644,42 @@ def test_ldap_user_filtering_valid_users(self):
...
+                admin_dn,
+                admin_passwd,
+                user_rdn,
+                uid_attr,
+                email_attr,
+                ldap_user_filter="abc=def",
+            )
+            mock_clean_user_filter = Mock()
+            ldap._clean_user_filter = mock_clean_user_filter
+            ldap.at_least_one_user_exists()
```
josephschorr, line 680: assert that it does exist

data/users/externalldap.py
thomasmckay: it is now wrapped in parens if the string does not start _and_ end with one

PR #321, opened 2020-04-07
[WIP] Openshift Integrated OAuth Authentication

mosen:
### Description of Changes

This change adds features and configuration for integrating with the OpenShift 3.x/4.x integrated OAuth service.

When running Project Quay inside an OpenShift cluster, the administrator should be presented with the option to configure Integrated OAuth, such that logging into the OpenShift console confers access to Project Quay.

The integration will include OAuth token scopes so that the authentication principal and full display name are carried over to Project Quay. Additionally, group memberships from OpenShift should be carried over if possible, so that access may be granted to OpenShift groups from Project Quay in a future change.

The functionality will not impact Project Quay permissions at this time. The user is authenticated to Quay but there will be no further provisioning. This is regarded as beyond the scope of this change.

The end user motivation for the change: To provide a seamless experience when deploying Project Quay into an OpenShift v3/v4 cluster so that no further configuration of Authentication is required, unless there is a requirement to separate Project Quay from the operation of the cluster.

The current behaviour: When deploying Project Quay into an OpenShift cluster, none of the available authentication methods support integration with the OpenShift RBAC/console identity. It is up to the user to establish an LDAP or OIDC connection via an external IdP which matches their OpenShift identity.

#### Changes:

* Add Configuration Feature flag and Configuration Variable, similar to Google, Gitlab etc OIDC configuration parameters.
* Communicate with the OpenShift cluster API to determine the OAuth endpoints (token/userinfo).
* Accepts OpenShift OAuth service authentication flow for authenticating OpenShift users.
* (TODO) Provide configuration UI via quay config-app.
* (TODO) Validated Unit and Integration Testing

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!


mosen:
Apologies, seem to have commit history from python3 branch due to rebase.

alecmerdler:
@mosen Thanks a ton for this contribution! We'd love to help review it, would you mind fixing up the rebase?

mosen:
Rebasing

mosen:
I'm going to branch off python3 again and cherry pick changes back in. Rebase is giving me a few headaches.

mosen:
Ready to review, no config-app UI yet.

thomasmckay:
https://issues.redhat.com/browse/PROJQUAY-673

util/config/validators/validate_access.py
```diff
@@ -15,11 +15,12 @@ def validate(cls, validator_context):
             # Make sure we have at least one OIDC enabled.
             github_login = config.get("FEATURE_GITHUB_LOGIN", False)
             google_login = config.get("FEATURE_GOOGLE_LOGIN", False)
+            openshift_login = config.get("FEATURE_OPENSHIFT_LOGIN", False)
 
             login_manager = OAuthLoginManager(config, client=client)
             custom_oidc = [s for s in login_manager.services if isinstance(s, OIDCLoginService)]
 
-            if not github_login and not google_login and not custom_oidc:
+            if not github_login and not google_login and not openshift_login and not custom_oidc:
```
alecmerdler, line 23: Now that this list is growing, we could use 
```python
if not any([github_login, google_login, openshift_login]) and not custom_oidc:
```

oauth/services/openshift.py
```diff
@@ -0,0 +1,187 @@
...
+        except OAuthExchangeCodeException as oce:
+            raise OAuthLoginException(str(oce))
+
+        # Retrieve the user's information with the token.
+        try:
+            user_info = self.get_user_info(http_client, token)
+        except OAuthGetUserInfoException as oge:
+            raise OAuthLoginException(str(oge))
+
+        # if user_info.get("id", None) is None:
```
alecmerdler, line 161: Is this intended to be commented out?

oauth/services/openshift.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def _load_oauth_config_via_discovery(self, is_debugging):
+        """
+        Attempts to load the OAuth config via the OAuth discovery mechanism using OpenShift Default Service DNS.
+
+        If is_debugging is True, non-secure connections are alllowed. Raises an
+        DiscoveryFailureException on failure.
+        """
+        oauth_server = self.config.get("OPENSHIFT_SERVER", DEFAULT_OAUTH_HOST)
+        # if not oidc_server.startswith("https://") and not is_debugging:
```
alecmerdler, line 111: Remove commented-out code.

mosen:
Thanks @alecmerdler for your valuable time in peer review. Let me address all these issues and come back.

util/config/validators/validate_access.py
mosen: yep agreed, doesnt read very well for a giant list of and/or conditions

oauth/services/openshift.py
mosen: Let me see whether I can actually fetch profile information from the user-info endpoint. Otherwise automatically created Quay users just get the login carried over and nothing else.

mosen:
Sorry to change tack @alecmerdler do you have any feedback on the following proposal:
What I've done here is treated OpenShift as a garden variety OIDC provider. What might be even better is to do some automated discovery of client ID and client secret if they co-exist in the namespace where Quay is deployed?

I'm envisioning this user experience:
- User installs quay-operator, creates QuayEcosystem.
- QuayEcosystem optionally registers OAuth resource for quay as a client to openshift-oauth (optional QuayEcosystem property)
- QuayEcosystem optionally places OAuth secret into target namespace, default quay-enterprise.
- When Quay config is done, There is now the option to automatically discover the configuration. Integrated authentication can be one click.
- Advanced (hidden) section is still available which allows manual input of client id and client secret. This is to cover off scenarios where Quay relies on openshift authentication but does not reside in same cluster.

mosen:
Oh, theres an even more magical scenario, but i think it makes config debugging harder:
- detect the openshift.default.svc API endpoint and assume that default auth is openshift therefore user lands on already-configured Quay.
I kinda don't like having implicit magic like this though

mosen:
Sorry will have to rebase/cherry pick again because the upstream force pushed rebase python3 branch caused me some headaches with rewritten history branching off py3.

oauth/services/openshift.py
mosen: Removed

oauth/services/openshift.py
mosen: This one still pending user-info endpoint tests

util/config/validators/validate_access.py
mosen: Fixed not pushed

PR #325, opened 2020-04-07
Fix our handling of chunked uploads to meet the OCI spec

josephschorr:
We were using the wrong header name and format


josephschorr:
https://issues.redhat.com/browse/PROJQUAY-574

PR #344, opened 2020-04-21
Clair v4 `IndexState` Endpoint

alecmerdler:
Fixes https://issues.redhat.com/browse/PROJQUAY-610

### DIFF
--- a/util/secscan/v4/api.py
+++ b/util/secscan/v4/api.py
@@ -79,7 +79,7 @@ class SecurityScannerAPIInterface(object):
 Action = namedtuple("Action", ["name", "payload"])
 
 actions = {
-    "State": lambda: Action("State", ("GET", "state", None)),
+    "IndexState": lambda: Action("IndexState", ("GET", "index_state", None)),
     "Index": lambda manifest: Action("Index", ("POST", "index_report", manifest)),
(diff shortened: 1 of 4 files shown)


alecmerdler:
@thomasmckay This will need to be cherry-picked into `3.3-release`.

ldelossa:
Should the Interface impl have it's method changed as well? 

https://github.com/quay/quay/pull/340/files#diff-de04d2aed57baa3c456f91913d059b04R47

alecmerdler:
@ldelossa That's just the name of the method. Unless you are planning on adding `matcher_state`, then it would maybe make sense. The methods can't be 1:1 with the routes, for example `POST /index_report --> index()` and `GET /index_report --> `index_report()`.

util/secscan/v4/api.py
```diff
@@ -204,5 +204,6 @@ def is_valid_response(action, resp={}):
         try:
             validate(resp, schema, resolver=resolver)
             return True
-        except Exception:
+        except Exception as e:
+            logger.exception(e)
```
josephschorr, line 208: `logger.exception` will already add the exception contents; please put a descriptive message on this

alecmerdler: Fixed

PR #345, opened 2020-04-21
Add additional validation to webhook URLs to catch common mistakes

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-614

### DIFF
--- a/config.py
+++ b/config.py
@@ -721,3 +721,7 @@ class DefaultConfig(ImmutableConfig):
     # Feature Flag: Whether to allow Helm OCI content types.
     # See: https://helm.sh/docs/topics/registries/
     FEATURE_EXPERIMENTAL_HELM_OCI_SUPPORT = False
+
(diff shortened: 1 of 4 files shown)


config.py
```diff
@@ -721,3 +721,7 @@ def create_transaction(db):
     # Feature Flag: Whether to allow Helm OCI content types.
     # See: https://helm.sh/docs/topics/registries/
     FEATURE_EXPERIMENTAL_HELM_OCI_SUPPORT = False
+
+    # The set of hostnames disallowed from webhooks, beyond localhost (which will
+    # not work due to running inside a container).
+    WEBHOOK_HOSTNAME_BLACKLIST = []
```
alecmerdler, line 727: What is the justification for this new config field?

josephschorr: In case on-prem customers want to blacklist additional URL endpoints. For example, they may have a host that is inaccessible to the registry, so this would allow them to configure that hostname here, and give immediate feedback to users setting up webhooks

PR #346, opened 2020-04-22
Fix search filter for LDAP user lookup when no user filter is specified

josephschorr:
https://issues.redhat.com/browse/PROJQUAY-617

The LDAP class defined the filter using a default Python parameter, so
we need to mimic it


josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-617

PR #347, opened 2020-04-22
Fix wrong reference name in ClairSecurityScannerAPI

kleesc:
### Description of Changes

* Fix a few references using the wrong name

#### Issue: https://issues.redhat.com/browse/PROJQUAY-620

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/data/secscan_model/test/test_secscan_interface.py
+++ b/data/secscan_model/test/test_secscan_interface.py
@@ -47,7 +47,7 @@ def test_load_security_information(repository, v4_whitelist, initialized_db):
     "next_token, expected_next_token",
     [
         (None, SplitScanToken("v4", V4ScanToken(56))),
-        (SplitScanToken("v4", V4ScanToken(1)), SplitScanToken("v4", None)),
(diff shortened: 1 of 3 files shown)


data/secscan_model/test/test_secscan_interface.py
```diff
@@ -47,7 +47,7 @@ def test_load_security_information(repository, v4_whitelist, initialized_db):
     "next_token, expected_next_token",
     [
         (None, SplitScanToken("v4", V4ScanToken(56))),
-        (SplitScanToken("v4", V4ScanToken(1)), SplitScanToken("v4", None)),
```
alecmerdler, line 50: We still need to test this case.

kleesc: Is it a valid case though? Looking at what `perform_indexing` does, it looks like the it will always return a token if there's a `Manifest` with a higher `id` than the one parameterized for that test.

We can see from the previous one that there is one with `id=56`...

kleesc: And fallback to `v2` if there is not.

PR #356, opened 2020-04-28
GC query improvements

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-658

### DIFF
--- a/data/model/oci/tag.py
+++ b/data/model/oci/tag.py
@@ -12,6 +12,7 @@ from data.database import (
     ImageStorage,
     MediaType,
     RepositoryTag,
+    RepositoryState,
     TagManifest,
     TagManifestToManifest,
     get_epoch_timestamp_ms,
@@ -652,6 +653,8 @@ def find_repository_with_garbage(limit_to_gc_policy_s):
                 ~(Tag.lifetime_end_ms >> None),
                 (Tag.lifetime_end_ms <= expiration_timestamp),
                 (Namespace.removed_tag_expiration_s == limit_to_gc_policy_s),
+                (Namespace.enabled == True),
+                (Repository.state != RepositoryState.MARKED_FOR_DELETION),
             )
             .limit(GC_CANDIDATE_COUNT)
             .distinct()
--- a/data/model/storage.py
+++ b/data/model/storage.py
@@ -69,43 +69,25 @@ def add_storage_placement(storage, location_name):
         pass
 
 
-def _orphaned_storage_query(candidate_ids):
+def _is_storage_orphaned(candidate_id):
     """
-    Returns the subset of the candidate ImageStorage IDs representing storages that are no longer
-    referenced by images.
+    Returns the whether the given candidate storage ID is orphaned. Must be executed
+    under a transaction.
     """
-    # Issue a union query to find all storages that are still referenced by a candidate storage. This
-    # is much faster than the group_by and having call we used to use here.
-    nonorphaned_queries = []
-    for counter, candidate_id in enumerate(candidate_ids):
-        query_alias = "q{0}".format(counter)
+    with ensure_under_transaction():
+        try:
+            ManifestBlob.get(blob=candidate_id)
+            return False
+        except ManifestBlob.DoesNotExist:
+            pass
 
-        # TODO: remove the join with Image once fully on the OCI data model.
-        storage_subq = (
-            ImageStorage.select(ImageStorage.id)
-            .join(Image)
-            .where(ImageStorage.id == candidate_id)
-            .limit(1)
-            .alias(query_alias)
-        )
+        try:
+            Image.get(storage=candidate_id)
+            return False
+        except Image.DoesNotExist:
+            pass
 
-        nonorphaned_queries.append(ImageStorage.select(SQL("*")).from_(storage_subq))
-
-        manifest_storage_subq = (
-            ImageStorage.select(ImageStorage.id)
-            .join(ManifestBlob)
-            .where(ImageStorage.id == candidate_id)
-            .limit(1)
-            .alias(query_alias)
-        )
-
-        nonorphaned_queries.append(ImageStorage.select(SQL("*")).from_(manifest_storage_subq))
-
-    # Build the set of storages that are missing. These storages are orphaned.
-    nonorphaned_storage_ids = {
-        storage.id for storage in _basequery.reduce_as_tree(nonorphaned_queries)
-    }
-    return list(candidate_ids - nonorphaned_storage_ids)
+        return True
 
 
 def garbage_collect_storage(storage_id_whitelist):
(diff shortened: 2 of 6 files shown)


workers/namespacegcworker.py
```diff
@@ -18,6 +19,13 @@ class NamespaceGCWorker(QueueWorker):
     """
 
     def process_queue_item(self, job_details):
+        try:
+            with GlobalLock("LARGE_GARBAGE_COLLECTION", lock_ttl=NAMESPACE_GC_TIMEOUT + 60):
```
kleesc, line 23: Does this need to be using the same lock as the repository gc worker?
This worker is performing a delete  on the DeletedNamespace table, does it need to be performed exclusively from the repo gc?

workers/repositorygcworker.py
```diff
@@ -18,6 +19,13 @@ class RepositoryGCWorker(QueueWorker):
     """
 
     def process_queue_item(self, job_details):
+        try:
+            with GlobalLock("LARGE_GARBAGE_COLLECTION", lock_ttl=REPOSITORY_GC_TIMEOUT + 60):
```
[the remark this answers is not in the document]
josephschorr: For *full* deletion, yes. Not for incremental GC.

workers/namespacegcworker.py
josephschorr: Yes, since they could both GC the same repository, at the same time.

workers/gc/gcworker.py
```diff
@@ -39,13 +49,31 @@ def _garbage_collection_repos(self):
...
+                        repository = Repository.get(id=repo_ref.id)
+                    except Repository.DoesNotExist:
+                        return
+
+                    # Make sure the repository is not marked for deletion or the parent namespace
+                    # is not disabled. If either is the case, then the repository will be GCed
+                    # as part of another process.
+                    if (repository.state == RepositoryState.MARKED_FOR_DELETION) or (
+                        not repository.namespace_user.enabled
+                    ):
```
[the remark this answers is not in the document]
josephschorr: `black` decided it

data/model/storage.py
```diff
@@ -69,43 +69,25 @@ def add_storage_placement(storage, location_name):
...
+            return False
+        except ManifestBlob.DoesNotExist:
+            pass
 
-        nonorphaned_queries.append(ImageStorage.select(SQL("*")).from_(manifest_storage_subq))
+        try:
+            Image.get(storage=candidate_id)
+            return False
+        except Image.DoesNotExist:
+            pass
```
jzelinskie, line 88: I wonder if unioning these two queries and using `COUNT()` to see if the number of results are greater than zero is less overhead for the database (e.g. is it cheaper than opening another connection).

workers/gc/gcworker.py
```diff
@@ -39,13 +49,31 @@ def _garbage_collection_repos(self):
             assert features.GARBAGE_COLLECTION
 
             try:
-                repository = Repository.get(id=repo_ref.id)
-            except Repository.DoesNotExist:
-                return
+                with GlobalLock(
+                    "REPO_GARBAGE_COLLECTION_%s" % repo_ref.id, lock_ttl=REPOSITORY_GC_TIMEOUT + 60,
+                ) if not skip_lock_for_testing else empty_context():
```
jzelinskie, line 54: I didn't know you could put conditionals in `with`, neat.

workers/gc/gcworker.py
```diff
@@ -39,13 +49,31 @@ def _garbage_collection_repos(self):
...
+                    try:
+                        repository = Repository.get(id=repo_ref.id)
```
jzelinskie, line 56: Moving this `get by id` and the check for `marked_for_deletion` to outside the global lock will reduce the duration which a process is holding the lock (and thus contention). Whenever locking always move things outside the critical path.

data/model/oci/tag.py
```diff
@@ -652,6 +653,8 @@ def find_repository_with_garbage(limit_to_gc_policy_s):
                 ~(Tag.lifetime_end_ms >> None),
                 (Tag.lifetime_end_ms <= expiration_timestamp),
                 (Namespace.removed_tag_expiration_s == limit_to_gc_policy_s),
+                (Namespace.enabled == True),
```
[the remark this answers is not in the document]
josephschorr: No; the namespace deletion code takes care of that.

workers/namespacegcworker.py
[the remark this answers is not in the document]
josephschorr: Yeah, its just safety padding. I can move it into a constant if you like.

data/model/storage.py
josephschorr: Well, it doesn't open another connection, but it does run another query.

We could LEFT OUTER JOIN and then just retrieve the first result (if any), I suppose, but then we're holding two tables at once instead of a simple index lookup.

workers/gc/gcworker.py
josephschorr: Yeah, I discovered it recently.

workers/gc/gcworker.py
josephschorr: I actually put it into the lock to ensure that no one else is coming along and modifying before we check. This is the verification check (the lookup query also checks the state).

data/model/storage.py
```diff
@@ -185,54 +167,47 @@ def placements_to_filtered_paths_set(placements_list):
...
+                ).execute()
 
-        signatures_removed = (
-            ImageStorageSignature.delete()
-            .where(ImageStorageSignature.storage << orphaned_storage_ids)
-            .execute()
-        )
-        logger.debug("Removed %s image storage signatures", signatures_removed)
+            # Remove all orphaned storages
+            TorrentInfo.delete().where(TorrentInfo.storage == storage_id_to_check).execute()
```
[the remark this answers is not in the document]
josephschorr: Eventually, yes but for now we need to remove the rows, otherwise the deletion will fail due to a foreign key constraint

josephschorr:
@jzelinskie Let me know if you want me to make that one change

data/model/storage.py
jzelinskie: In the grand scheme of things, doing so would be premature optimization. If you were running many, many instances of Quay, you'd benefit from having GC reduce the number of queries it makes, but that's always going to be a drop in the bucket compared to the queries made by the API.

workers/namespacegcworker.py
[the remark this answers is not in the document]
jzelinskie: Please, do.

workers/gc/gcworker.py
jzelinskie: It sounds like this check is unnecessary if the gc call also checks this value and is under the lock. The only value I see from checking the repository here is to avoid taking the lock before you find out you didn't need to gc.

josephschorr: Yeah, I figure better to be careful, but it is extra code. Should I remove it?

jzelinskie: Yeah -- this is just wasting time under the lock. My vote is to totally remove it and only add it back before the lock is taken if we notice lock contention.

PR #362, opened 2020-04-29
Ensure Security Scanning Works Without Clair V4

alecmerdler:
### Description

The logic in the secscan model prevented indexing or retrieving vulnerability information for manifests when Quay was configured without Clair v4. Changes the logic to skip model splitting and adds tests.

### Changes

- Adds no-op implementations for `V2SecurityScanner` and `V4SecurityScanner` which are used if either are not configured (properly)
- Adds tests to ensure no-op implementations work 

Fixes https://issues.redhat.com/browse/PROJQUAY-662
Fixes https://issues.redhat.com/browse/PROJQUAY-640

### DIFF
--- a/data/secscan_model/__init__.py
+++ b/data/secscan_model/__init__.py
@@ -2,9 +2,9 @@ import os
 import logging
 from collections import namedtuple
 
-from data.secscan_model.secscan_v2_model import V2SecurityScanner
-from data.secscan_model.secscan_v4_model import V4SecurityScanner
-from data.secscan_model.interface import SecurityScannerInterface
+from data.secscan_model.secscan_v2_model import V2SecurityScanner, NoopV2SecurityScanner
+from data.secscan_model.secscan_v4_model import V4SecurityScanner, NoopV4SecurityScanner
+from data.secscan_model.interface import SecurityScannerInterface, InvalidConfigurationException
 from data.database import Manifest
 from data.registry_model.datatypes import Manifest as ManifestDataType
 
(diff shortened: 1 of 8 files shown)


data/secscan_model/__init__.py
```diff
@@ -24,18 +24,25 @@ def configure(self, app, instance_keys, storage):
...
+        if self._v4_model is not None:
+            logger.info("v4 whitelist: `%s`", self._v4_namespace_whitelist)
         logger.info("===============================")
 
         return self
 
     def perform_indexing(self, next_token=None):
         if next_token is None:
+            if self._v4_model is None:
+                return SplitScanToken("v4", None)
```
alecmerdler, line 38: Returning this skips us [down to `self._model.perform_indexing(None)`](https://github.com/quay/quay/pull/362/files#diff-a0d14413673a3951f56735fcda8fea49R49) on the next call.

data/secscan_model/__init__.py
```diff
@@ -18,22 +18,28 @@
 class SecurityScannerModelProxy(SecurityScannerInterface):
     def configure(self, app, instance_keys, storage):
         # TODO(alecmerdler): Just use `V4SecurityScanner` once Clair V2 is removed.
-        self._model = V2SecurityScanner(app, instance_keys, storage)
-        self._v4_model = V4SecurityScanner(app, instance_keys, storage)
+        try:
+            self._model = V2SecurityScanner(app, instance_keys, storage)
+        except Exception:
```
josephschorr, line 23: I'd prefer if we caught something more specific than just general `Exception` here

data/secscan_model/secscan_v4_model.py
```diff
@@ -62,12 +81,13 @@ def __init__(self, app, instance_keys, storage):
 
         validator = V4SecurityConfigValidator(
```
josephschorr, line 82: Can we add a check above these that sees if the SECURITY_SCANNER_V4_ENDPOINT is `None`, and if so, not write any warning log?

alecmerdler:
@thomasmckay This will need to be cherry-picked into Quay 3.3.

PR #371, opened 2020-05-04
Deprecate Image rows and move to in-memory synthesized legacy images

josephschorr:
This PR makes a number of changes so that we can stop relying on the Image table:

1. Changes derived storage to be stored per manifest, rather than per image
2. Moves uploaded blob tracking to its own dedicated table instead of (mis)using Image and a temporary RepositoryTag. RepositoryTag is now marked as deprecated and inserts to it will be blocked.
3. Changes the Manifest table to add additional metadata columns such as total compressed size and config media type
4. Switches all registry model code (except legacy V2 security scanning) to use in-memory synthesized legacy images, rather than loading rows from the table

Note that Image rows are still being written, and we will continue to have to do so until backfill completes.


data/registry_model/registry_oci_model.py
```diff
@@ -506,75 +451,15 @@ def change_repository_tag_expiration(self, tag, expiration_date):
...
-        if isinstance(manifest_or_legacy_image, Manifest):
-            image = oci.shared.get_legacy_image_for_manifest(manifest_or_legacy_image._db_id)
-            if image is None:
-                return SecurityScanStatus.UNSUPPORTED
-        else:
-            try:
-                image = database.Image.get(id=manifest_or_legacy_image._db_id)
-            except database.Image.DoesNotExist:
-                return None
+        # TODO: change from using the Image row once we've moved all security info into MSS.
```
alecmerdler, line 458: Per our discussion, we'd actually want to remove this function completely and only access this information using the `secscan_model`.

data/registry_model/registry_oci_model.py
```diff
@@ -308,11 +271,19 @@ def list_repository_tag_history(
             repository_ref._db_id, page, size, specific_tag_name, active_tags_only, since_time_ms
         )
 
-        # TODO: do we need legacy images here?
-        legacy_images_map = oci.tag.get_legacy_images_for_tags(tags)
+        # TODO: Remove this once the layers compressed sizes have been fully backfilled.
+        tags_missing_sizes = [tag for tag in tags if tag.manifest.layers_compressed_size is None]
+        legacy_images_map = {}
+        if tags_missing_sizes:
+            legacy_images_map = oci.tag.get_legacy_images_for_tags(tags_missing_sizes)
```
alecmerdler, line 278: This doesn't work:
```
gunicorn-web stdout | 2020-05-12 21:34:22,891 [414] [ERROR] [gunicorn.error] Error handling request /api/v1/repository/big-boi/bad-image/tag/?limit=100&page=1&onlyActiveTags=true
Traceback (most recent call last):
  ...
  File "/quay-registry/endpoints/api/tag.py", line 93, in get
    active_tags_only=active_tags_only,
  File "/quay-registry/data/registry_model/registry_oci_model.py", line 278, in list_repository_tag_history
    legacy_images_map = oci.tag.get_legacy_images_for_tags(tags_missing_sizes)
AttributeError: 'module' object has no attribute 'get_legacy_images_for_tags'
```

josephschorr: Fixed

flavianmissi:
Stale PR - closing. Please rebase and reopen.

PR #373, opened 2020-05-05
Fix Re-Indexing for V4 Security Scanner

alecmerdler:
### Description

We were incorrectly comparing the full JSON object to the actual string value, causing infinite re-indexing. Also, re-indexes a manifest if retrieving its security information from the scanner API fails.

### Changes 

- Fix `index_state` comparison
- Delete existing `ManifestSecurityStatus` if fetching info from security scanner service fails
- Adds `SECURITY_SCANNER_V4_REINDEX_THRESHOLD` config field

Fixes https://issues.redhat.com/browse/PROJQUAY-674


data/secscan_model/secscan_v4_model.py
```diff
@@ -128,6 +128,8 @@ def load_security_information(self, manifest_or_legacy_image, include_vulnerabil
         try:
             report = self._secscan_api.vulnerability_report(manifest_or_legacy_image.digest)
         except APIRequestFailure as arf:
+            status.delete_instance()
```
josephschorr, line 131: You'll need to check if we're in read only mode here, and if so, not call this delete (or just catch and ignore the readonlymodeexception)

data/secscan_model/secscan_v4_model.py
```diff
@@ -195,7 +197,7 @@ def needs_reindexing_query(indexer_hash):
             yield_random_entries(not_indexed_query, Manifest.id, batch_size, max_id, min_id,),
             yield_random_entries(index_error_query, Manifest.id, batch_size, max_id, min_id,),
             yield_random_entries(
-                lambda: needs_reindexing_query(indexer_state),
+                lambda: needs_reindexing_query(indexer_state.get("state", "")),
```
josephschorr, line 200: I'd prefer if the `.get` occurred at the API level

data/secscan_model/secscan_v4_model.py
alecmerdler: Do we need to add that down where we create the `ManifestSecurityStatus` as well?

data/secscan_model/secscan_v4_model.py
alecmerdler: All the rest of our methods return the full JSON. If we decide this endpoint needs more fields, then the secscan model will be able to use those.

data/secscan_model/secscan_v4_model.py
josephschorr: Nah, since the workers won't be run when in read only mode

data/secscan_model/secscan_v4_model.py
josephschorr: Okay.

config.py
```diff
@@ -478,6 +478,9 @@ def create_transaction(db):
     # Namespace whitelist for security scanner.
     SECURITY_SCANNER_V4_NAMESPACE_WHITELIST = []
 
+    # Minimum number of seconds before re-indexing a manifest the with security scanner.
```
josephschorr, line 481: `the with`?

data/secscan_model/secscan_v4_model.py
```diff
@@ -163,6 +170,10 @@ def eligible_manifests(base_query):
         if max_id is None or min_id is None or min_id > max_id:
             return None
 
+        can_reindex = (datetime.utcnow() - ManifestSecurityStatus.last_indexed) > timedelta(
```
josephschorr, line 173: This won't work efficiently on MySQL because it'll have to compute for each row, rather than using the index; instead, do this:

```
can_reindex_threshold = datetime.utcnow() - timedelta(....)
can_reindex = (ManifestSecurityStatus.last_indexed <= can_reindex_threshold)
```

data/secscan_model/secscan_v4_model.py
```diff
@@ -163,6 +170,10 @@ def eligible_manifests(base_query):
         if max_id is None or min_id is None or min_id > max_id:
             return None
 
+        reindex_threshold = lambda: datetime.utcnow() - timedelta(
```
josephschorr, line 173: Let's add a test for this: write the MSS row, change the indexer state, and make sure it isn't picked up again since we're within the reindexing exclusion window

alecmerdler: I already did [here](https://github.com/quay/quay/pull/373/files#diff-d1556d15ee5435b7ee2ca49b3a797803R279).

PR #376, opened 2020-05-07
LDAP: Defined found_results before try block so it's always assigned

tux-o-matic:
### Description of Changes
Up until now, the "if not found_results" line could throw an UnboundLocalError because the variable was assigned inside a try block which could fail with the variable never assigned.

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -442,6 +442,7 @@ class LDAPUsers(FederatedUsers):
                     break
 
                 while True:
+                    found_results = 0
                     try:
                         if has_pagination:
                             _, rdata, _, serverctrls = conn.result3(msgid)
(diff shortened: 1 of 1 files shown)


tux-o-matic:
@josephschorr , it might be a bit more complex to simulate the error which triggers that bug in unit testing where LDAP is mocked.
In the current code, the bug happens if *LDAPObject.result()* or *LDAPObject.result3()* throws an exception which isn't handled to *break* the while loop.
My fix at just makes sure that the variable is assigned before calling LDAP.

josephschorr:
@tux-o-matic you could `patch` the LDAP call on the mock ldap and force it to return an exception in the test. Might be a good test to have regardless, and I think we might do so already elsewhere (I may be wrong)

tux-o-matic:
@josephschorr, from what I can see there is only one use of *patch* in the LDAP test and that's just to use mocked data with the LDAP module.
Overwriting part of that just to be able to use *side_effect* is beyond me.
PS: I can't seem to find instructions to setup a local development environment for Quay other than building a container. No obvious way to run the unit test.

josephschorr:
@tux-o-matic You can run `tox`, but for now, I'll move and merge this.

PR #389, opened 2020-05-20
Update the k8 templates to work with EKS

bb03:
This still isn't functional, so I could definitely use some help with it.
As of right now I am able to spin up the config app, but configuration blows up with at least one "Contact support message". I'm pretty sure that with some minor configuration changes I could get it up and running smoothly.
Also, I would like to see more inline documentation around whats expected for certain secrets and the format of certain files. I also have lots of questions around permissions and setting the right contexts to run the database as. Getting this to work really shouldn't be as painful as it was for me so far.

### Description of Changes

#### Changes:

* ..
* ..

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/deploy/k8s/postgres-clair-deployment.yaml
+++ b/deploy/k8s/postgres-clair-deployment.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1 
 kind: Deployment
 metadata:
   labels:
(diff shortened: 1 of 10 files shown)


deploy/k8s/postgres-clair-deployment.yaml
```diff
@@ -16,18 +16,21 @@ spec:
...
           value: clair
-        - name: POSTGRESQL_DATABASE
+        - name: POSTGRES_DATABASE
           # Set the name of the Clair postgres database
           value: clair
-        - name: POSTGRESQL_PASSWORD
+        - name: POSTGRES_PASSWORD
           # Set the password for the Clair postgress user
           value: test123
-        image: registry.access.redhat.com/rhscl/postgresql-10-rhel7:1-35
```
thomasmckay, line 30: I'd prefer to continue using the Red Hat postgresql, what is the reason for changing?

bb03: Short answer: It didn't work.  I don't remember why off the top of my head, but I found that using the more generic postgres image didn't error out.  When it's run in your environment, do you see any issues with it, or does it work fine?

bb03: @thomasmckay I put the image definition back.  Let me know if you have any other comments.

bb03:
I got things to work for me.  I probably won't be making any more changes for the time being.  Thanks!

PR #393, opened 2020-05-26
Fixed repo list showing 'null' for state in api

jonathankingfc:
### Description of Changes

- Added Repository.state to query to return in api call to list repos
- Added check for state in test_get_repo
- Added api test for listRepos endpoint
## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/data/model/repository.py
+++ b/data/model/repository.py
@@ -265,6 +265,7 @@ def get_visible_repositories(
             Namespace.username,
             Repository.visibility,
             Repository.kind,
+            Repository.state,
         )
         .switch(Repository)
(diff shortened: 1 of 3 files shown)


jonathankingfc:
This PR resolves this issue: https://issues.redhat.com/browse/PROJQUAY-590

data/model/repository.py
```diff
@@ -265,6 +265,7 @@ def get_visible_repositories(
             Namespace.username,
             Repository.visibility,
             Repository.kind,
+            Repository.state,
```
alecmerdler, line 268: @jonathankingfc Can you add/update a unit test that confirms this change?

PR #405, opened 2020-06-02
Make sure to immediately delete federated login links on user delete

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-760

### DIFF
--- a/data/model/user.py
+++ b/data/model/user.py
@@ -1293,6 +1293,9 @@ def _delete_user_linked_data(user):
     # falling and only occurs if a superuser is being deleted.
     ServiceKeyApproval.update(approver=None).where(ServiceKeyApproval.approver == user).execute()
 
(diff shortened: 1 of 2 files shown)


data/model/user.py
```diff
@@ -1293,6 +1293,9 @@ def _delete_user_linked_data(user):
     # falling and only occurs if a superuser is being deleted.
     ServiceKeyApproval.update(approver=None).where(ServiceKeyApproval.approver == user).execute()
 
+    # Delete any federated user links.
+    FederatedLogin.delete().where(FederatedLogin.user == user).execute()
```
kleesc, line 1297: I'm assuming this whole function is too large to wrap it in a transaction along with the user delete?

josephschorr: Yes

PR #410, opened 2020-06-03
[PROJQUAY-480]  Add Prometheus Gauge for number of Gunicorn workers

kurtismullins:
### Description of Changes

- A new Prometheus Gauge is added to keep tabs on the count of the quantity of gunicorn workers for each gunicorn parent process.
- Each time the quantity is changed (e.g. if it scales up), then the quantity is updated automatically.

NOTE: This may _not_ include defunct/zombie/orphaned processes. This relies on gunicorn's built-in hook to produce this value. As it relies upon the gunicorn hook, it is also not providing that value to the health-check.

#### Changes:

- New metric for `quay_gunicorn_workers` is now available

#### Issue:

- https://issues.redhat.com/browse/PROJQUAY-480

**TESTING** ->

- Start Quay
- Verify that `quay_gunicorn_workers` metrics are available

**BREAKING CHANGE** 

n/a

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/conf/gunicorn_local.py
+++ b/conf/gunicorn_local.py
@@ -6,6 +6,8 @@ sys.path.append(os.path.join(os.path.dirname(__file__), "../"))
 import logging
 
 from Crypto import Random
+from prometheus_client import Gauge
+
 from util.log import logfile_path
 from util.workers import get_worker_count, get_worker_connections_count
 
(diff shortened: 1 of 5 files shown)


conf/gunicorn_local.py
```diff
@@ -31,3 +40,14 @@ def when_ready(server):
     logger.debug(
         "Starting local gunicorn with %s workers and %s worker class", workers, worker_class
     )
+
+
+def nworkers_changed(server, new_value, old_value):
```
alecmerdler, line 45: This is defined as a lifecycle hook in Gunicorn?

kurtismullins:
I am closing this Pull Request. I believe it would be better to use other means, such as a side-car prometheus exporter, to expose this data.

PR #411, opened 2020-06-04
quay-app template - update apiVersions and remove default readOnly

maorfr:
cc @josephschorr @tparikh

### DIFF
--- a/deploy/openshift/quay-app.yaml
+++ b/deploy/openshift/quay-app.yaml
@@ -4,7 +4,7 @@ kind: Template
 metadata:
   name: quay
 objects:
-- apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


deploy/openshift/quay-app.yaml
```diff
@@ -213,7 +213,6 @@ objects:
           - containerPort: 8443
           volumeMounts:
           - name: configvolume
-            readOnly: false
```
tparikh, line 216: Why is this changed needed?

maorfr: openshift removes this field, causing qontract-reconcile to think that it should apply again.
no harm in keeping it, just makes sense to me to not have fields with default values that are removed.

tparikh:
LGTM

PR #412, opened 2020-06-04
quay-app template - update apiVersions and add ServiceAccount

maorfr:
cc @josephschorr @tparikh 

replaces #409

### DIFF
--- a/deploy/openshift/quay-app.yaml
+++ b/deploy/openshift/quay-app.yaml
@@ -4,7 +4,7 @@ kind: Template
 metadata:
   name: quay
 objects:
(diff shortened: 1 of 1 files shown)


deploy/openshift/quay-app.yaml
```diff
@@ -206,7 +212,6 @@ objects:
           - containerPort: 8443
           volumeMounts:
           - name: configvolume
-            readOnly: false
```
tparikh, line 209: Why is this change needed?

tparikh:
LGTM

PR #419, opened 2020-06-04
Repo mirroring column

jonathankingfc:
### Description of Changes

Added UI change to show repo state when repo mirroring is enabled. 

#### Changes:
- Repo view on front end now has extra column to show repo state when FEATURE_REPO_MIRROR = true. 
repo_mirror_on

- Column is not present when FEATURE_REPO_MIRROR=false
repo_mirror_off

## Reviewer Checklist
@alecmerdler 
- [ ] It works!


jonathankingfc:
@alecmerdler

alecmerdler:
@jonathankingfc Can you add screenshots to the pull request description?

static/directives/repo-list-table.html
```diff
@@ -62,6 +68,11 @@
          </span>
          <span class="empty" ng-if="::!repository.last_modified">(Empty Repository)</span>
        </td>
+       <td class="state" ng-show="repoMirroringEnabled">
```
alecmerdler, line 71: We use [these CSS classes for the background color](https://sourcegraph.com/github.com/quay/quay/-/blob/static/js/directives/ui/repo-state/repo-state.component.css#L62:6) in other places in the UI.

static/css/directives/ui/repo-list-table.css
```diff
@@ -19,6 +19,11 @@
   font-size: 13px;
 }
 
+.repo-list-table .state {
```
alecmerdler, line 22: Use `repo-state` in case there is a name conflict with another CSS class. Just `state` is pretty common.

crozzy:
LGTM but it appears like there are a couple of unaddressed comments from @alecmerdler, have these been resolved offline?

harishsurf (approved):
lgtm

PR #425, opened 2020-06-05
OpenShift deploy template updates for log shipper

tparikh:
This PR removes load balancer service that is no longer used. It also sets log stream name to pod name and makes ticker for syslog configurable.


jfchevrette:
LGTM

PR #429, opened 2020-06-05
[PROJQUAY-775] [PROJQUAY-759] Expose gunicorn and zombie processes

kurtismullins:
### Description of Changes

- Queries local system (or container) for quantity of gunicorn and zombie processes
- Reports quantities to Prometheus
- Health check exposes whether ~any~ too many zombie processes exist

**TODO**
- ~Unit Tests~
- ~Verify the data in Prometheus is usable as expected when running multiple Quay pods~
- ~Ensure the worker is started by `supervisord` similar to all other workers~

#### Changes:

- New health-check value: `defunct_processes`
- New prometheus metrics: `quay_gunicorn_process_count` and `quay_zombie_process_count`
- New configuration variable: `MAX_DEFUNCT_PROCESS_COUNT`
- New configuration variable: `LOCAL_PROMETHEUS_STATS_FREQUENCY`

#### Issue:

- https://issues.redhat.com/browse/PROJQUAY-759
- https://issues.redhat.com/browse/PROJQUAY-775

**TESTING**

- Check prometheus data for zombie process and gunicorn process counts
- Verify the health-check appropriately displays True or False depending on the number of zombie processes on the same machine/namespace.
- Verify that the configuration variables are used.

**BREAKING CHANGE**

n/a

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/health/processes.py
+++ b/health/processes.py
@@ -1,0 +1,20 @@
+import logging
+
+import psutil
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_gunicorn_processes():
+    """
+    Return all gunicorn processes running within the current system or namespace.
+    """
+    return [p for p in psutil.process_iter() if p.name() == "gunicorn"]
+
+
+def get_all_zombies():
+    """
+    Return all defunct (zombie) processes within the current system or namespace.
+    """
+    return [p for p in psutil.process_iter() if p.status() == "zombie"]
--- a/health/services.py
+++ b/health/services.py
@@ -6,6 +6,7 @@ import psutil
 
 from app import build_logs, storage, authentication, instance_keys
 from health.models_pre_oci import pre_oci_model as model
+from health.processes import get_all_zombies
 
 logger = logging.getLogger(__name__)
 
@@ -175,6 +176,20 @@ def _check_disk_space(for_warning):
     return _check_disk_space
 
 
+def _check_zombie_process_count(app):
+    """
+    Returns False if too many zombie processes are running.
+    """
+    processes = get_all_zombies()
+    count = len(processes)
+    msg = "Found %s zombie processes." % count
+
+    max_zombies = app.config["MAX_DEFUNCT_PROCESS_COUNT"]
+    is_healthy = count <= max_zombies  # Only report False if too many zombies exist
+
+    return (is_healthy, msg)
+
+
 _INSTANCE_SERVICES = {
     "registry_gunicorn": _check_gunicorn("v1/_internal_ping"),
     "web_gunicorn": _check_gunicorn("_internal_ping"),
@@ -182,6 +197,7 @@ _INSTANCE_SERVICES = {
     "service_key": _check_service_key,
     "disk_space": _check_disk_space(for_warning=False),
     "jwtproxy": _check_jwt_proxy,
+    "defunct_processes": _check_zombie_process_count,
 }
 
 _GLOBAL_SERVICES = {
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -113,7 +113,7 @@ peewee==3.11.2
 pillow==6.1.0
 ply==3.11
 prometheus-client==0.7.1
-psutil==5.6.6
+psutil==5.6.3
```
kurtismullins, line 116: This is not needed. I'll revert this change before moving the PR out of `draft`

workers/localpromstats.py
```diff
@@ -0,0 +1,64 @@
...
+    def _report_stats(self):
+        logger.debug("Reporting local statistics to prometheus.")
+        gunicorn_process_gauge.set(get_gunicorn_process_count())
+        zombie_process_gauge.set(get_zombie_process_count())
+
+
+def main():
+    logging.config.fileConfig(logfile_path(debug=False), disable_existing_loggers=False)
+
+    if not app.config.get("PROMETHEUS_PUSHGATEWAY_URL"):
```
josephschorr, line 54: Add a feature flag as well to turn off the worker if we decide to do so

health/processes.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    Queries for all processes on the current system and handles any exceptions that may occur.
+
+    NOTE: On systems running more than Quay, this may include processes unrelated to Quay.
+    """
+    try:
+        return [p for p in psutil.process_iter()]
+    except Exception as e:
+        logger.exception("Unable to list all processes")
+    return []  # Don't allow this to break the application
```
jzelinskie, line 19: This whole file is a lot of boiler plate for 3 extremely simple list comprehensions.

I'm not sure why you wouldn't want to still raise this exception. Not being able to list processes on the system is a pretty bad failure and I'd rather that be explicit than have a function lie to me and tell me that there are 0 processes running or that there are 0 zombies when they're actually could be. I'd catch the exception in the worker and _not_ push a value as a result, rather than pushing a value of 0 to the gauge that's just plain wrong.

health/services.py
```diff
@@ -175,13 +176,28 @@ def _check_disk_space(app):
...
 
+def _check_zombie_process_count(app):
+    """
+    Returns True if too many zombie processes are running.
+    """
+    processes = get_all_zombies()
+    count = len(processes)
+    msg = "Found %s zombie processes." % count
+
+    max_zombies = app.config.get("MAX_DEFUNCT_PROCESS_COUNT", 0)
```
jzelinskie, line 187: This is absolutely for another day, but this stands out to me now:
since there is supposed to be a default in the app config for every value, I think we should be using `[]` to index into the `app.config` so that we get a crash in the scenario when the programmer has forgotten to add their sane default in the config. Having a default value in case of a lack of default value seems a little crazy.

jzelinskie:
Putting this here for now:  I think node_exporter supports finding zombies via https://github.com/prometheus/node_exporter/blob/594f417bdf6f49e145f1937af99e849377412354/collector/processes_linux.go#L62

kurtismullins:
@josephschorr @jzelinskie 

I am going to close this Pull Request unless you believe there's value in keeping it open. Just let me know either way. Thanks!

kurtismullins:
Closing this pull request as I don't believe we are going to implement this. IMHO I believe that in k8s/openshift, exporting data using a side-car container is the way to go. In non-k8s deployments, we can determine other tools which might be useful and offer them as suggestions in the documentation when/if the need arises.

PR #431, opened 2020-06-09
Redis connection improvements

kleesc:
### Description of Changes

* `health_check_interval` will attempt a roundtrip ping before running a command if the connection has been idle for more that the set value. If the ping fails, the redis client should close and reestablish the socket connection.
* Wrap calls to build phase updates (Redis calls) with try blocks. If there is an exception, calls `_build_finished` with its result set to `ERROR`. This is to allow the buildmanager to handle Redis connection errors more gracefully. 

#### Issue: [PROJQUAY-752](https://issues.redhat.com/browse/PROJQUAY-752)

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/buildman/component/buildcomponent.py
+++ b/buildman/component/buildcomponent.py
@@ -110,7 +110,14 @@ class BuildComponent(BaseComponent):
         )
 
         self._current_job = build_job
-        self._build_status = StatusHandler(self.build_logs, build_job.repo_build.uuid)
+        build_id = build_job.repo_build.uuid
+        try:
+            self._build_status = StatusHandler(self.build_logs, build_id)
+        except Exception as bse:
+            logger.exception("Failed to set phase for build ID: %s - %s", build_id, bse)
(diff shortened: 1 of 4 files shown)


buildman/component/buildcomponent.py
```diff
@@ -318,47 +318,66 @@ def _on_log_message(self, phase, json_data):
...
     @trollius.coroutine
     def _determine_cache_tag(
         self, command_comments, base_image_name, base_image_tag, base_image_id
     ):
-        with self._build_status as status_dict:
-            status_dict["total_commands"] = len(command_comments) + 1
+        try:
+            with self._build_status as status_dict:
+                status_dict["total_commands"] = len(command_comments) + 1
+        except Exception as bse:
```
josephschorr, line 378: Can we `logger.exception` here as well?

PR #433, opened 2020-06-09
Fix handling of client connections in pymemcache cache

josephschorr:
pymemcache is apparently not thread safe, so our reuse of the client
was causing the occasional hang on read. This change has us open a new
connection per request, and then close it once complete. We also make
sure to close the DB connection before making the memcache connection,
in case the memcache connection takes a bit of time

We should investigate switching to the PooledClient library for reusing
connections (safely) once we verify this change works

### Description of Changes

#### Changes:

* ..
* ..

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!


data/cache/impl.py
```diff
@@ -34,6 +37,20 @@ def retrieve(self, cache_key, loader, should_cache=is_not_none):
         pass
 
 
+class DisconnectWrapper(DataModelCache):
+    """ Wrapper around another data model cache that disconnects from the database before
```
alecmerdler, line 41: Fix docstring to match our convention:
```python
"""
Some docstring.
"""
```

josephschorr: Done

data/cache/impl.py
```diff
@@ -93,6 +111,16 @@ def retrieve(self, cache_key, loader, should_cache=is_not_none):
 _JSON_TYPE = 2
 
 
+@contextmanager
+def memcache_client(client_builder):
+    try:
+        client = client_builder()
+        yield client
+    finally:
```
alecmerdler, line 119: Doesn't this have to `yield` at every codepath? If an exception is raised before `yield client`, then you'll get: `TypeError: 'NoneType' object is not an iterator`.

alecmerdler: https://amir.rachum.com/blog/2017/03/03/generator-cleanup/

data/cache/test/test_cache.py
```diff
@@ -28,6 +34,9 @@ def test_caching(cache_type):
 
 
 def test_memcache():
+    global DATA
```
alecmerdler, line 37: What is this madness

josephschorr: Allows us to reassign the top level variable; its just for texting

data/cache/impl.py
josephschorr: I'll add one in the finally

alecmerdler:
```python 
@contextmanager
def memcached_client(client_builder):
    client = None
    try:
        client = client_builder()
    finally:
        yield client
        if client is not None:
            client.cloes()
```

data/cache/test/test_cache.py
alecmerdler: Hmm, not a fan but it's okay if just for testing...

PR #437, opened 2020-06-11
Change APPR to lazy load the iofile and tar file

josephschorr:


endpoints/appr/cnr_backend.py
```diff
@@ -12,7 +17,97 @@
...
+        self._lazy_load_file()
+        return self._io_file
+
+    def _lazy_load_file(self):
+        if self._io_file is not None:
+            return
+
+        self._io_file = io.BytesIO(self.blob)
+
+    def _lazy_load_tar(self):
```
jakedt, line 55: Do you need any locking around these methods?

endpoints/appr/cnr_backend.py
```diff
@@ -12,7 +17,97 @@
 from util.request import get_request_ip
 
 
+class LazyPackage(object):
```
jakedt, line 20: Can you add a comment saying what this is and why it will be helpful?

endpoints/appr/cnr_backend.py
josephschorr: I don't believe so, as this is only called per thread

endpoints/appr/cnr_backend.py
josephschorr: Sure

josephschorr:
@jakedt Yes, I tested it where I did GET calls only and ensured that the lazy loaded items were not hit

jakedt:
When are they loaded?

josephschorr:
> When are they loaded?

As far as I can tell, they aren't used at all, except for `.io_file`, which is used in our blob *writing* code here: https://github.com/quay/quay/blob/master/endpoints/appr/models_cnr.py#L327

endpoints/appr/cnr_backend.py
josephschorr: Its literally instantiated per request... what are your concerns around locking? (especially give the likelyhood this isn't even really used)

jakedt: Right now we're instantiating one per request, but we've also retroactively added fully hydrated object caches to hot routes in the past. This is a textbook example of when you need locking, or at least an atomic compare and swap. You've got shared state and multiple methods on the object can update that state. Moreover, something that is masquerading as a simple property read is actually mutating the object state. Why not just add it?

josephschorr: Mostly a concern about additional overhead, and a belief that the "at worst" scenario is that we simply reload the data, since the attributes being filled in will have the same result even if we call the loading code twice

jzelinskie: Here's my general concern:

A list of heap allocated objects of this type are stashed in a per-gunicorn-worker cache. Two greenlets get the same reference returned from the cache concurrently. Now depending on where/when the greenlets context swap, you have a scenario where they could both be mutating memory in the object in a way that corrupts the state.

endpoints/appr/cnr_backend.py
jzelinskie: Why not make this a doc string?

josephschorr: Because I don't want this to be treated as a "defined" class for use otherwise. Its merely a workaround

PR #439, opened 2020-06-15
[PROJQUAY-795] Monkey patch top of remaining gunicorn entrypoints

kurtismullins:
### Description of Changes

Full coverage of monkey patching for gunicorn.

#### Changes:

Includes monkey patching at the top of remaining gunicorn entrypoint files. This is consistent with the other `gunicorn*.py` files.

#### Issue:

- https://issues.redhat.com/browse/PROJQUAY-795

**TESTING**

- Run the container. Ensure nothing related to verbs is broken.

**BREAKING CHANGE**

n/a

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/conf/gunicorn_verbs.py
+++ b/conf/gunicorn_verbs.py
@@ -1,3 +1,8 @@
+# NOTE: Must be before we import or call anything that may be synchronous.
+from gevent import monkey
+
+monkey.patch_all()
+
 import sys
 import os
 
--- a/conf/gunicorn_local.py
+++ b/conf/gunicorn_local.py
@@ -1,3 +1,8 @@
+# NOTE: Must be before we import or call anything that may be synchronous.
+from gevent import monkey
+
+monkey.patch_all()
+
 import sys
 import os
 


conf/gunicorn_verbs.py
```diff
@@ -1,3 +1,8 @@
+# NOTE: Must be before we import or call anything that may be synchronous.
+from gevent import monkey
```
josephschorr, line 2: We don't use gevent in verbs, so this isn't needed

kurtismullins: @josephschorr In that case, would you be opposed to me removing the `gunicorn_verbs.py` file and pointing supervisord to start verbs as a standard python module?

Current:
```
command=nice -n 10 gunicorn -c %(ENV_QUAYCONF)s/gunicorn_verbs.py verbs:application
```

Proposed (untested)
```
python -m verbs:application
```

josephschorr: No. It still uses *gunicorn*, just not *gevent*. It uses sync workers.

kurtismullins: Thanks! Closing this accordingly.

kurtismullins:
The verbs module uses the synchronous worker and therefore does not require monkey patching. Closing this Pull Request accordingly.

PR #440, opened 2020-06-15
keystonev2: populate user.name into UserInformation

zhouhaibing089:
The behavior between keystonev2 and keystonev3 is different today.
In keystonev3 implementation, the username comes from [user.name][1],
but keystonev2 just uses the incoming `username_or_email`.

[1]: https://github.com/quay/quay/blob/f4179e5e719ebc8e795064b8703da8a29d877b3b/data/users/keystone.py#L316

### DIFF
--- a/data/users/keystone.py
+++ b/data/users/keystone.py
@@ -137,7 +137,7 @@ class KeystoneV2Users(FederatedUsers):
             return (None, "Missing email field for user %s" % user_id)
 
         email = user.email if hasattr(user, "email") else None
-        return (UserInformation(username=username_or_email, email=email, id=user_id), None)
+        return (UserInformation(username=user.name, email=email, id=user_id), None)
 
     def query_users(self, query, limit=20):
         return (None, self.federated_service, "Unsupported in Keystone V2")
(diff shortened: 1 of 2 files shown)


data/users/keystone.py
```diff
@@ -137,7 +137,7 @@ def verify_credentials(self, username_or_email, password):
             return (None, "Missing email field for user %s" % user_id)
 
         email = user.email if hasattr(user, "email") else None
-        return (UserInformation(username=username_or_email, email=email, id=user_id), None)
+        return (UserInformation(username=user.name, email=email, id=user_id), None)
```
josephschorr, line 140: And `name` is guaranteed to be present?

zhouhaibing089: Yes, see https://sergslipushenko.github.io/html/api-ref-identity-admin-v2.html

zhouhaibing089: Specifically below:

> A `user` object, which shows the `username`, `roles_links`, `id`, `roles`, and `name`.

josephschorr: Shouldn't this be `username` then?

zhouhaibing089: Honestly, I didn't investigate harder to figure out what's the difference between `username` and `name`. This is just being consistent with the field keystonev3 uses(though appears `username` isn't  a field anymore).

josephschorr: Can you run a test and verify?

zhouhaibing089: Yeah, everything looks good!

PR #442, opened 2020-06-15
Deprecate Image rows and move to in-memory synthesized legacy images [Python 3]

josephschorr:
**NOTE:** This change includes a database migration which adds columns to the `Manifest` table. This change will therefore lock that table while the operation occurs, so care should be taken before deployment (likely done in a readonly state)

This PR makes a number of changes so that we can stop relying on the Image table:

- Moves uploaded blob tracking to its own dedicated table instead of (mis)using Image and a temporary RepositoryTag. RepositoryTag is now marked as deprecated and inserts to it will be blocked.
- Changes the Manifest table to add additional metadata columns such as total compressed size and config media type
- Switches all registry model code (except legacy V2 security scanning) to use in-memory synthesized legacy images, rather than loading rows from the table
- Removes support for indexing in Clair V2 and instead only allows for reading existing Clair V2 information; All new manifests are indexed with Clair V4 exclusively
- Removes all APIs and support code for Verbs: squashing, ACI conversion and torrenting is all removed
- Removes all code around derived storage, since it was only used for verbs
- Removes all code around GPG signing, since it was only used for ACI conversion
- Stops writing Image rows (except in testing)
- A number of other small query improvements


BillDett:
Any dependencies we can drop (e.g. gpg) as a result of this refactoring?

josephschorr:
@BillDett Yes, this drops the GPG dependency

data/model/gc.py
```diff
@@ -376,6 +418,16 @@ def _purge_pre_oci_tag(tag, context, allow_non_expired=False):
         reloaded_tag.delete_instance()
 
 
+def _purge_uploaded_blob(uploaded_blob, context, allow_non_expired=False):
```
kurtismullins, line 421: I know this is a "private" function and the name is relatively explanatory, but can we add a docstring describing its expected behavior? e.g.

```
Deletes a blob from every persistent storage
```

data/model/oauth.py
```diff
@@ -352,8 +352,13 @@ def lookup_access_token_by_uuid(token_uuid):
 
 def lookup_access_token_for_user(user_obj, token_uuid):
     try:
-        return OAuthAccessToken.get(
-            OAuthAccessToken.authorized_user == user_obj, OAuthAccessToken.uuid == token_uuid
+        return (
```
kurtismullins, line 355: We should grab this query and verify that the query plan is using indexes. I assume it's called often.

data/model/oci/manifest.py
```diff
@@ -392,6 +321,121 @@ def _create_manifest(
...
 
 
+def _build_blob_map(
+    repository_id,
+    manifest_interface_instance,
+    retriever,
+    storage,
+    raise_on_error=False,
+    require_empty_layer=True,
+):
```
kurtismullins, line 331: Would you mind to include a docstring for this function? It has a lot of arguments and is relatively complex

data/secscan_model/secscan_v2_model.py
```diff
@@ -164,80 +143,13 @@ def load_security_information(self, manifest_or_legacy_image, include_vulnerabil
...
-                    logger.debug("Another worker pre-empted us for layer: %s", candidate.id)
-                    abt.set()
-                except APIRequestFailure:
-                    logger.exception("Security scanner service unavailable")
-                    return
-
-                unscanned_images.set(num_remaining)
-
-        return next_token
+        raise NotImplementedError("Unsupported for this security scanner version")
```
kurtismullins, line 155: :+1:

data/secscan_model/secscan_v4_model.py
```diff
@@ -178,16 +170,14 @@ def eligible_manifests(base_query):
         # TODO(alecmerdler): Filter out any `Manifests` that are still being uploaded
         def not_indexed_query():
             return (
-                eligible_manifests(Manifest.select())
-                .switch(Manifest)
+                Manifest.select()
                 .join(ManifestSecurityStatus, JOIN.LEFT_OUTER)
                 .where(ManifestSecurityStatus.id >> None)
```
kurtismullins, line 175: `ManifestSecurityStatus` objects can be created without an ID?

josephschorr: This is a left outer join; if the MSS row doesn't exist, "it"'s ID will be NULL

kurtismullins: Ahhh, that makes sense. Thanks for pointing that out.

data/model/gc.py
josephschorr: It deletes an UploadedBlob entry, actually

data/model/oauth.py
josephschorr: All ForeignKeyField's are indexed automatically in peewee, and the uuid is too

data/model/oci/manifest.py
josephschorr: Done

data/model/blob.py
```diff
@@ -192,7 +157,12 @@ def get_blob_upload_by_uuid(upload_uuid):
     Loads the upload with the given UUID, if any.
     """
     try:
-        return BlobUpload.select().where(BlobUpload.uuid == upload_uuid).get()
+        return (
+            BlobUpload.select(BlobUpload, ImageStorageLocation)
```
kleesc, line 161: Why the extra join?

data/database.py
```diff
@@ -1414,6 +1418,19 @@ class Meta:
         )
 
 
+class UploadedBlob(BaseModel):
```
kleesc, line 1421: Is this new table to avoid adding an extra new column to the BlobUpload table?

data/registry_model/registry_oci_model.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import sys
```
kleesc, line 2: Not used

config.py
```diff
@@ -739,3 +735,6 @@ def create_transaction(db):
 
     # Feature Flag: Whether the repository action count worker is enabled.
     FEATURE_REPOSITORY_ACTION_COUNTER = True
+
+    # TEMP FEATURE: Backfill the sizes of manifests.
+    FEATURE_MANIFEST_SIZE_BACKFILL = True
```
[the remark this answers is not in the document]
josephschorr: .... What the hell?

Its... gone....

I'll add it back in tomorrow

data/registry_model/shared.py
```diff
@@ -0,0 +1,17 @@
+import uuid
+
+from hashids import Hashids
+
+
+class SyntheticIDHandler(object):
+    def __init__(self, hash_salt=None):
+        self.hash_salt = hash_salt or str(uuid.uuid4())
+        self.hashids = Hashids(alphabet="0123456789abcdef", min_length=64, salt=self.hash_salt)
```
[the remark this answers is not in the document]
josephschorr: Yes, I think this is the best option because its both (1) tested and (2) handles all the restrictions we need

data/model/blob.py
josephschorr: Caller is using the location and, without the join, is making another SQL call

data/database.py
josephschorr: And because they have different lookup semantics; we could reuse BlobUpload but we'd have to verify the content checksum is in there

data/registry_model/shared.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def encode(self, manifest_id, layer_index=0):
+        encoded = self.hashids.encode(manifest_id, layer_index)
+        assert len(encoded) == 64
+        return encoded
+
+    def decode(self, synthetic_v1_id):
+        return self.hashids.decode(synthetic_v1_id)
```
kurtismullins, line 17:
```
self.hash_salt = hash_salt or str(uuid.uuid4())
```

Is there a test that verifies an `.encode()` followed by a `.decode()` with, and without, a value for `hash_salt`?

test/registry/registry_tests.py
```diff
@@ -872,10 +873,18 @@ def test_image_replication_empty_layers(
             credentials=credentials,
         )
 
+<<<<<<< HEAD
```
kurtismullins: Looks like a merge diff got left in here.

test/test_api_usage.py
```diff
@@ -3326,8 +3325,12 @@ def test_listdeletecreateandmovetag(self):
             params=dict(repository=ADMIN_ACCESS_USER + "/complex", tag="sometag"),
         )
 
+<<<<<<< HEAD
```
kurtismullins: Another merge diff.

conf/supervisord.conf.jnj
```diff
@@ -170,6 +170,14 @@ autostart = {{ config['teamsyncworker']['autostart'] }}
 stdout_events_enabled = true
 stderr_events_enabled = true
 
+[program:repomirrorworker]
```
kurtismullins: Might be related to the missing worker comment ... but why was this added? Copy+Paste job?

util/config/schema.py
```diff
@@ -17,7 +17,11 @@
     "FEATURE_REPOSITORY_GARBAGE_COLLECTION",
     "FEATURE_REPOSITORY_ACTION_COUNTER",
     "APP_REGISTRY_PACKAGE_LIST_CACHE_WHITELIST",
+<<<<<<< HEAD
```
kurtismullins: Diff left in

data/database.py
josephschorr: @kurtismullins have a better name? :) I couldn't come up with one, although I didn't spend a lot of time on it.

image/docker/schema2/list.py
```diff
@@ -249,6 +249,19 @@ def manifests(self, content_retriever):
             for m in manifests
         ]
 
+    @property
+    def amd64_linux_manifest_digest(self):
+        """ Returns the digest of the AMD64+Linux manifest in this list, if any, or None
+            if none.
+        """
+        for manifest_ref in self._parsed[DOCKER_SCHEMA2_MANIFESTLIST_MANIFESTS_KEY]:
+            platform = manifest_ref[DOCKER_SCHEMA2_MANIFESTLIST_PLATFORM_KEY]
```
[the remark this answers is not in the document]
josephschorr: Yes, since we validate via the schema defined above.

data/registry_model/shared.py
josephschorr: I trust hashids to be tested and we do have a test somewhere that sends in an invalid value

conf/supervisord.conf.jnj
josephschorr: Bad merge after rebase

data/registry_model/registry_oci_model.py
josephschorr: Removed

data/registry_model/shared.py
josephschorr: Added another explicit unit test just in case

test/registry/registry_tests.py
josephschorr: Yeah, these were removed in a followup change. Since all of this will be squashed on merge, I figured it was easier to leave it in then rebase

config.py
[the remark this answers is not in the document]
josephschorr: Found it in the history and brought it back

test/registry/registry_tests.py
kurtismullins: Sounds good! I'll close these ones

data/model/oauth.py
kurtismullins: Great! Closing accordingly.

data/database.py
kurtismullins: haha, great question! What would you consider the primary difference between BlobUpload and UploadedBlob? That might lead into some sort of way to differentiate the names.

config.py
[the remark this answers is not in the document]
kurtismullins: > TEMP FEATURE: Backfill the sizes of manifests.

Since this is mentioned as a temp feature, what is the condition to remove it? And, can you add that as a `TODO(after-something)` statement?

data/database.py
```diff
@@ -1414,6 +1418,19 @@ class Meta:
...
+    """
+    UploadedBlob tracks a recently uploaded blob and prevents it from being GCed
+    while within the expiration window.
+    """
+
+    id = BigAutoField()
+    repository = ForeignKeyField(Repository)
+    blob = ForeignKeyField(ImageStorage)
+    uploaded_at = DateTimeField(default=datetime.utcnow)
+    expires_at = DateTimeField(index=True)
```
kurtismullins, line 1431: Should `expires_at` have a default value? e.g. `now + 24 hours`

josephschorr: I explicitly made sure it didn't, since if it wasn't specified then row creation will fail on the client, and I want that to be the case :)

kurtismullins: Smart thinking! Resolving

data/database.py
kurtismullins: I can't come up with a better name either, so I am going to resolve this thread.

workers/manifestbackfillworker.py
```diff
@@ -0,0 +1,95 @@
...
+                if layers_compressed_size is None:
+                    layers_compressed_size = 0
+            except ManifestException as me:
+                logger.warning(
+                    "Got exception when trying to parse manifest %s: %s", manifest_row.id, me
+                )
+
+            assert layers_compressed_size is not None
+            updated = (
+                Manifest.update(layers_compressed_size=layers_compressed_size)
```
kleesc, line 70: `confg_media_type`need to be backfilled?

data/model/oauth.py
```diff
@@ -352,8 +352,13 @@ def lookup_access_token_by_uuid(token_uuid):
...
+            OAuthAccessToken.select(OAuthAccessToken, User)
```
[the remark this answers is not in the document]
josephschorr: Yes, this was done on purpose because some calls *do* use the User, and its better to have it present to prevent a subsequent lookup

data/secscan_model/secscan_v2_model.py
```diff
@@ -111,12 +100,8 @@ def __init__(self, app, instance_keys, storage):
...
         )
 
-        # NOTE: This import is in here because otherwise this class would depend upon app.
-        # Its not great, but as this is intended to be legacy until its removed, its okay.
-        from util.secscan.analyzer import LayerAnalyzer
-
-        self._target_version = app.config.get("SECURITY_SCANNER_ENGINE_VERSION_TARGET", 3)
-        self._analyzer = LayerAnalyzer(app.config, self._legacy_secscan_api)
+    def register_model_cleanup_callbacks(self, data_model_config):
+        pass
```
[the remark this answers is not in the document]
josephschorr: No, we'll need it for Clair V4 eventually

workers/manifestbackfillworker.py
josephschorr: Doesn't need to be, but we could try to infer it

josephschorr: Done, although for schema1 manifests and manifest lists, it remains None

config.py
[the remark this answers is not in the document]
josephschorr: It should be removed once it has been run on production and run on prem; so basically, we remove it in the next version

data/model/oci/manifest.py
```diff
@@ -392,6 +321,131 @@ def _create_manifest(
...
+    # than trying to load it on a per repository basis, and that is unnecessary anyway since
+    # this layer is predefined.
+    if EMPTY_LAYER_BLOB_DIGEST in digests:
+        digests.remove(EMPTY_LAYER_BLOB_DIGEST)
+        blob_map[EMPTY_LAYER_BLOB_DIGEST] = get_shared_blob(EMPTY_LAYER_BLOB_DIGEST)
+        if not blob_map[EMPTY_LAYER_BLOB_DIGEST]:
+            if raise_on_error:
+                raise CreateManifestException("Unable to retrieve specialized empty blob")
+
+            logger.warning("Could not find the special empty blob in storage")
```
kurtismullins, line 356: I am not sure under which conditions this would happen. Should the error output be a bit more actionable? Something along the lines of:

```
raise CreateManifestException("Unable to retrieve specialized empty blob. Has any data been pushed?")
```
I don't actually know what the correct action or steps to take would be. I'm just thinking that this warning is useful but I don't know why it's occurring or how to fix it. I imagine _if_ it happened, it'd be an important thing to solve as it's used quite often.

It might also help in the "self-sevice" direction.

data/model/oci/manifest.py
```diff
@@ -392,6 +321,131 @@ def _create_manifest(
...
+            return None
+
+    if digests:
+        query = lookup_repo_storages_by_content_checksum(repository_id, digests, with_uploads=True)
```
kurtismullins, line 360: (nitpick) Is this a query or a list of storage rows? Might be useful to change the variable name if it's not a query.

data/model/oci/manifest.py
```diff
@@ -392,6 +321,131 @@ def _create_manifest(
...
+                    manifest_interface_instance.digest,
+                    repository_id,
+                )
+
+                if raise_on_error:
+                    raise CreateManifestException("Unknown blob `%s`" % digest_str)
+
+                return None
+
+    # Special check: If the empty layer blob is needed for this manifest, add it to the
```
kurtismullins, line 376: Just for my understanding, in Docker v2 the empty layer is not needed?

data/model/oci/manifest.py
```diff
@@ -392,6 +321,131 @@ def _create_manifest(
...
+    # blob map. This is necessary because Docker decided to elide sending of this special
+    # empty layer in schema version 2, but we need to have it referenced for schema version 1.
+    if require_empty_layer and EMPTY_LAYER_BLOB_DIGEST not in blob_map:
+        try:
+            requires_empty_layer = manifest_interface_instance.get_requires_empty_layer_blob(
+                retriever
+            )
+        except ManifestException as ex:
+            if raise_on_error:
+                raise CreateManifestException(str(ex))
```
kurtismullins, line 386: Why not just re-raise the `ManifestException`?

util/repomirror/api.py
```diff
@@ -96,7 +96,7 @@ def repository_mirror(self, repository):
         Posts the given repository to the repo mirror for processing, blocking until complete.
 
         Returns the analysis version on success or raises an exception deriving from
-        AnalyzeLayerException on failure. Callers should handle all cases of AnalyzeLayerException.
+        RepoMirrorException on failure. Callers should handle all cases of RepoMirrorException.
```
kurtismullins, line 99: :100:

data/model/gc.py
```diff
@@ -194,7 +202,27 @@ def _purge_repository_contents(repo):
         if not found:
             break
 
-    # TODO: remove this once we're fully on the OCI data model.
+    # Purge any uploaded blobs that have expired.
+    while True:
+        found = False
+        for uploaded_blobs in _chunk_iterate_for_deletion(
+            UploadedBlob.select().where(UploadedBlob.repository == repo)
```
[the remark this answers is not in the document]
josephschorr: This is deleting *all* contents of a repository

data/model/oci/manifest.py
josephschorr: It shouldn't really ever happen, since we no longer actually use the empty layer in normal pushes. If it does, then its an internal error, and it should be reported

data/model/oci/manifest.py
josephschorr: Its a query that returns the storage rows

data/model/oci/manifest.py
josephschorr: Correct. Its only needed for Docker V2 Schema 1

data/model/oci/manifest.py
josephschorr: Because the interface can only raise `CreateManifestException`

PR #443, opened 2020-06-15
Get container namespace from fieldref

tparikh:
We no longer have to provide namespace in deploy config.


tparikh:
@rrati @josephschorr

rrati:
lgtm

PR #444, opened 2020-06-16
Redis Model Cache (PROJQUAY-788)

alecmerdler:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-788

**Changelog:** Add model cache implementation backed by Redis.

**Docs:** N/a

**Testing:** Add the following section to `config.yaml`:

```yaml
DATA_MODEL_CACHE_CONFIG:
  engine: redis
  host: <redis-hostname>
  port: 6379
```

**Details:** Adds an implementation of the model cache which uses Redis. The goal is when deploying a cluster of Quay app containers, they will all use a shared Redis instance for caching requests (as opposed to a per-container memcache), which will reduce load on the database for expensive queries. [Enhancements proposal](https://github.com/quay/enhancements/blob/main/enhancements/redis-cache.md).

*NOTE:* When using this model cache implementation, the Redis dependency changes from being low priority (effectively ephemeral) to being critical (and should be HA).
*NOTE:* The Quay container will still have `memcached` running even if not being used, unless we modify the container entrypoint to read the `config.yaml`.

~Blocked by https://github.com/quay/quay/pull/745~

### DIFF
--- a/data/cache/impl.py
+++ b/data/cache/impl.py
@@ -2,17 +2,17 @@ import logging
 import json
 import os
 
+from abc import ABCMeta, abstractmethod
 from contextlib import contextmanager
 from datetime import datetime
-
-from abc import ABCMeta, abstractmethod
+from pymemcache.client.base import PooledClient
+from prometheus_client import Counter
+from redis import StrictRedis, RedisError
 from six import add_metaclass
 
 from data.database import CloseForLongOperation
-
-from pymemcache.client.base import PooledClient
-
 from util.expiresdict import ExpiresDict
+from util.locking import GlobalLock, LockNotAcquiredException
 from util.timedeltastring import convert_to_timedelta
 from util.workers import get_worker_connections_count
 
@@ -20,8 +20,17 @@ from util.workers import get_worker_connections_count
 logger = logging.getLogger(__name__)
 
 
+cache_count = Counter(
+    "quay_model_cache", "number of attempts to retrieve from the model cache", labelnames=["type"]
+)
+
+
 def is_not_none(value):
     return value is not None
+
+
+def lock_key_for(cache_key):
+    return "LOCK_" + cache_key
 
 
 @add_metaclass(ABCMeta)
@@ -81,7 +90,11 @@ class InMemoryDataModelCache(DataModelCache):
         result = self.cache.get(cache_key.key, default_value=not_found)
         if result != not_found:
             logger.debug("Found result in cache for key %s: %s", cache_key.key, result)
+            cache_count.labels("hit").inc()
+
             return json.loads(result)
+        else:
+            cache_count.labels("miss").inc()
 
         logger.debug("Found no result in cache for key %s; calling loader", cache_key.key)
         result = loader()
@@ -153,7 +166,7 @@ class MemcachedModelCache(DataModelCache):
                 raise Exception("Unknown flags for value: {1}".format(flags))
 
             return PooledClient(
-                self.endpoint,
+                server=self.endpoint,
                 no_delay=True,
                 timeout=self.timeout,
                 connect_timeout=self.connect_timeout,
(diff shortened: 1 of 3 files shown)


alecmerdler:
Rebased with `master`.

data/cache/impl.py
```diff
@@ -213,3 +213,72 @@ def retrieve(self, cache_key, loader, should_cache=is_not_none):
...
+
+    def retrieve(self, cache_key, loader, should_cache=is_not_none):
+        # TODO(alecmerdler): We might want to have different behavior based on `cache_key` (using "sets" for `/tags/list`, single value for others...)
+        not_found = None
+        client = self._get_client()
+        if client is not None:
+            logger.debug("Checking cache for key %s", cache_key.key)
+            try:
+                cached_result = client.get(cache_key.key)
+                if cached_result != not_found:
```
kleesc, line 251: Not specific to this implementation, but we could instrument this to measure the hit rate.
e.g A `Counter` metrics for  cache hits divided by a `Counter` metrics for `retrieve` calls

kleesc (commented):
> When using this model cache implementation, the Redis dependency changes from being low priority (effectively ephemeral) to being critical (and should be HA).

It depends. Most deployment aren't large enough or don't generate enough traffic to cause a thundering herd issue, and clearing the cache wouldn't be an issue.

kleesc:
@alecmerdler The current ticket linked is the wrong one. It should be PROJQUAY-788

data/cache/impl.py
```diff
@@ -213,3 +213,72 @@ def retrieve(self, cache_key, loader, should_cache=is_not_none):
...
+    def _get_client(self):
+        if self.client is not None:
+            return self.client
+
+        args = {
+            "host": self.host,
+            "port": self.port,
+            "socket_connect_timeout": 1,
+            "socket_timeout": 2,
+            "single_connection_client": True,
```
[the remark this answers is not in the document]
kleesc: I _think_ StrictRedis already uses a connection pool under the hood.

data/cache/impl.py
```diff
@@ -20,6 +19,14 @@
 logger = logging.getLogger(__name__)
 
 
+cache_retrieves = Counter(
```
alecmerdler, line 22: @kleesc What do you think of these metrics?

kleesc:
@alecmerdler Like you previously mentioned, I think we should make it so that two requests can't write to the same key at once (grab a Lock).

alecmerdler:
@kleesc Ready for another review after being rebased with the `GlobalLock` fix.

data/cache/impl.py
```diff
@@ -213,3 +219,80 @@ def retrieve(self, cache_key, loader, should_cache=is_not_none):
...
+                        json.dumps(result),
+                        ex=int(expires.total_seconds()) if expires else None,
+                    )
+                    logger.debug(
+                        "Cached loaded result for key %s with expiration %s: %s",
+                        cache_key.key,
+                        result,
+                        cache_key.expiration,
+                    )
+                except LockNotAcquiredException:
```
kleesc, line 289: I'm not sure if this is going to catch it. I think the GlobalLock context manager needs to be inside the try block.

kleesc (changes requested):

data/cache/impl.py
```diff
@@ -213,3 +219,80 @@ def retrieve(self, cache_key, loader, should_cache=is_not_none):
...
+                    return json.loads(cached_result)
+            except RedisError as re:
+                logger.warning("Got exception when trying to retrieve key %s", cache_key.key)
+
+        logger.debug("Found no result in cache for key %s; calling loader", cache_key.key)
+        result = loader()
+        logger.debug("Got loaded result for key %s: %s", cache_key.key, result)
+        if self.client is not None and should_cache(result):
+            # NOTE: This assumes that the Redis defined in `DATA_MODEL_CACHE_CONFIG` is the same as `USER_EVENTS_REDIS`.
+            with GlobalLock(cache_key.key, lock_ttl=5):
```
kleesc, line 267: The key will need to be different from the actual key. Maybe add a `LOCK_` prefix or suffix.

kleesc (changes requested):

kleesc (dismissed):
LGTM

alecmerdler:
I'm seeing slow pushes (stuck in `preparing` state) when running this on my k8s cluster. Going to do a bit more testing to ensure it's unrelated to this code change.

kleesc (approved):
LGTM

PR #448, opened 2020-06-18
Raise Orchestrator errors on redis errors in orchestrator

kleesc:
### Description of Changes

* Re-raise the exceptions when using the redis orchestrator so that the build manager can handle them.
* Add missing `yield From` on some async calls
* Update the build phase on log failure

#### Changes:

* ..
* ..

#### Issue: https://issues.redhat.com/browse/PROJQUAY-752

**TESTING** ->

**BREAKING CHANGE** ->

---

## Reviewer Checklist

- [ ] It works!

### DIFF
--- a/buildman/component/buildcomponent.py
+++ b/buildman/component/buildcomponent.py
@@ -7,6 +7,7 @@ import trollius
 
 from autobahn.wamp.exception import ApplicationError
 from trollius import From, Return
+from redis import RedisError
 
 from buildman.server import BuildJobResult
 from buildman.component.basecomponent import BaseComponent
@@ -17,7 +18,7 @@ from buildman.jobutil.workererror import WorkerError
 
 from app import app
 from data.database import BUILD_PHASE, UseThenDisconnect
-from data.model import InvalidRepositoryBuildException
+from data.model import build as build_model, InvalidRepositoryBuildException
 from data.registry_model import registry_model
 from util import slash_join
 
@@ -114,8 +115,10 @@ class BuildComponent(BaseComponent):
         try:
             self._build_status = StatusHandler(self.build_logs, build_id)
         except Exception as bse:
-            logger.exception("Failed to set phase for build ID: %s - %s", build_id, bse)
-            self._build_status_failure(bse)
+            logger.exception(
+                "Failed to set phase for start build for build ID: %s - %s", build_id, bse
+            )
+            yield From(self._build_status_failure(bse))
             raise Return()
 
         self._image_info = {}
@@ -344,7 +347,6 @@ class BuildComponent(BaseComponent):
                         )
                         raise Return(False)
                 except InvalidRepositoryBuildException:
-                    build_id = self._current_job.repo_build.uuid
                     logger.warning("Build %s was not found; repo was probably deleted", build_id)
                     raise Return(False)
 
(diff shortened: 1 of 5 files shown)


buildman/orchestrator.py
```diff
@@ -784,10 +808,17 @@ def _publish(self, **kwargs):
...
-        yield From(self._client.delete(slash_join(key, REDIS_EXPIRING_SUFFIX)))
-        yield From(self._publish(event=KeyEvent.DELETE, key=key, value=value))
+        try:
+            value = yield From(self._client.get(key))
+            if value is None:
+                raise KeyError
+            yield From(self._client.delete(key))
+            yield From(self._client.delete(slash_join(key, REDIS_EXPIRING_SUFFIX)))
+            yield From(self._publish(event=KeyEvent.DELETE, key=key, value=value))
+        except RedisError as rce:
```
josephschorr, line 818: RedisError handler is repeated here

buildman/manager/ephemeral.py
```diff
@@ -566,6 +566,12 @@ def schedule(self, build_job):
                 realm,
                 build_uuid,
             )
+        except OrchestratorConnectionError:
+            logger.exception(
```
josephschorr, line 570: Does this not need to return or re-raise?

kleesc: The above one was not being re-raised or returned, so I'm just logging them, since these are just for job metrics.

kleesc:
@josephschorr Ping. `yield From` was missing from some async calls, so it was essentially doing nothing on exceptions... 🤦

buildman/orchestrator.py
```diff
@@ -784,10 +808,17 @@ def _publish(self, **kwargs):
...
         assert not self.is_canceller_only
 
-        value = yield From(self._client.get(key))
-        yield From(self._client.delete(key))
-        yield From(self._client.delete(slash_join(key, REDIS_EXPIRING_SUFFIX)))
-        yield From(self._publish(event=KeyEvent.DELETE, key=key, value=value))
+        try:
+            value = yield From(self._client.get(key))
+            if value is None:
+                raise KeyError
```
josephschorr, line 814: Why this change?

kleesc:
@jzelinskie ping

buildman/orchestrator.py
kleesc: From the interface:
>    Missing keys should return KeyError; otherwise, errors should raise an
    OrchestratorError.

buildman/component/buildcomponent.py
```diff
@@ -360,19 +362,26 @@ def _on_log_message(self, phase, json_data):
                 # If the json data contains an error, then something went wrong with a push or pull.
                 if "error" in log_data:
                     yield From(self._build_status.set_error(log_data["error"]))
+        except RedisError as re:
+            # Move forward with the build if it is a Redis error
+            logger.warning("Failed to set phase on log for build ID: %s - %s", build_id, bse)
+            build_model.update_phase_then_close(build_id, phase)
```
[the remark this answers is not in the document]
kleesc: @jzelinskie Sure, but it was already yielding control of the event loop to that same synchronous call whenever the worker logged a phase change in `buildstatus.py`.

Since that database call is coupled with the Redis call in `buildstatus.py`, if Redis is down, the build phase won't get updated correctly. 

In this case, we're making sure that we allow the build to move forward, or marked as failure if Redis is not available.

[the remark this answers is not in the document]
jzelinskie: Ideally, we'd wrap this with the same async wrapper we use for etcd, but I don't think it's worth the effort (and complexity) if we can get by with this until Python 3 or a new build system.

PR #452, opened 2020-06-22
Fix the RepositoryActionCount worker to properly select rows

josephschorr:
This selects only the rows missing, rather than all repositories

### DIFF
--- a/workers/repositoryactioncounter.py
+++ b/workers/repositoryactioncounter.py
@@ -24,8 +24,10 @@ class RepositoryActionCountWorker(Worker):
         self.add_operation(self._run_counting, POLL_PERIOD_SECONDS)
 
     def _run_counting(self):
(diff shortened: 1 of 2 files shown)


workers/repositoryactioncounter.py
```diff
@@ -24,8 +24,10 @@ def __init__(self):
         self.add_operation(self._run_counting, POLL_PERIOD_SECONDS)
 
     def _run_counting(self):
+        yesterday = date.today() - timedelta(days=1)
+
         def batch_query():
-            return database.Repository.select()
+            return model.repositoryactioncount.missing_counts_query(yesterday)
```
kleesc, line 30: I guess it's fine assume it won't get backed more than one day.

josephschorr: Correct

PR #455, opened 2020-06-24
PROJQUAY-740 mypy type annotations

mosen:
### Description of Changes

* This change adds mypy and Typing as dev dependencies and *SOME* coverage of type annotations in the python 2 compatible comment format.

#### Changes:

* mypy and typing added to requirements-dev.txt
* types annotated in almost all places that they cannot be inferred.
* efforts were taken not to impact any functional component.
* errors reported by mypy that could be corrected by a functional change have not been fixed, so you will see errors running mypy.
* makefile target added: `make types-test` but does not exist as a dependency of `test` target.
* unfixable non-inferred situations are ignored using annotation. including but not limited to: re-imports of imported dependencies, imports of pseudo-private functions and objects, monkey patching of imported modules, usage of attributes declared dynamically at runtime.

#### Issue: [PROJQUAY-740](https://issues.redhat.com/browse/PROJQUAY-740)

**TESTING** ->

Execute `mypy .` at the root of the application, you will receive outstanding errors for:
* Dynamically declared FEATURES_ attributes
* Re-declaration of functions and classes already declared in same file
* Mocking of functions and objects that do not meet the mocked objects signature.
* Duplicate declarations caused by `from x import *` on multiple lines.

Otherwise, the typing is consistent.

**BREAKING CHANGE** ->

I deliberately did not fix any of the above to avoid creating any functional difference in the application.
Additional imports have been added but I do not expect side effects.

---

## Reviewer Checklist

- [ ] It works!


mosen:
No worries ill run it through black

alecmerdler:
@mosen This is awesome, thanks! Type annotations are one of the things I've really been forward to adding to Quay now that we are on Python 3. After discussion, our team would prefer to forego the Python 2 compatible comment format in favor of the straight Python 3 syntax. We will handle any fixes we have to make when cherry-picking into our Python 2 branches. Do you mind updating your pull request to use the normal Python 3 syntax?

mosen:
Hi @alecmerdler  I'm glad you said that. the comment syntax really forces some ugly combinations of annotations that could be shorter. I will reformat this as the native type annotation style.

mosen:
The one part i could not work out, was how the global `features.<attribute>` was being accessed, since the features module did not define any variables or something like a magic method in its own scope, there must be some black magic going on that im finding difficult to annotate. I'll have to define features as unchecked in this case

mosen:
- Rewrote comment style as py3.6 native type annotations
- Made mypy more strict about returning information on things such as unreachable code paths, useless expressions etc. which is actually really good and has highlighted a lot of things.
- Format w/black as per request.

mosen:
I realise the CI is breaking, but i see unit testing faillures that have nothing to do with this change.

Makefile
```diff
@@ -75,6 +75,9 @@ full-db-test: ensure-test-db
 clients-test:
 	cd test/clients; python clients_test.py
 
+types-test:
+	mypy --py2 .
```
kurtismullins, line 79: @mosen 
Should this be `py3` now?

mosen: Sorry.. I'll fix this one and see if i can force py3 style annotations at the same time

mosen: the version flag is omitted and the `python_version` option inside mypy.ini denotes 3.6 as the currently supported version to check types for.

kurtismullins:
Hey @mosen, I overlooked your updated PR. My apologies.
I noticed the CI build had several failures but they were not obvious to their cause.
I just restarted the CI builds to see if they pass now.

Makefile
kurtismullins: Thanks! I'm not sure that it makes any significant difference, but I believe we are aiming for Python 3.8. We can always adjust later if needed as I believe Py3.8 is backwards compatible with Py3.6 syntax.

I'll go ahead and resolve this.

kurtismullins:
Hey @mosen, would you be able/willing to rebase this on `master`?

mosen:
Rebased and force pushed, thanks @kurtismullins

alecmerdler:
@mosen Thanks for the rebase, looks like some tests are failing but I'll check this out later. Excited to get this merged!

mosen:
@alecmerdler similar issue to the openshift oauth
```
    def test_ensure_schema_defines_all_fields():
        for key in vars(DefaultConfig):
            has_key = key in CONFIG_SCHEMA["properties"] or key in INTERNAL_ONLY_PROPERTIES
>           assert has_key, "Property `%s` is missing from config schema" % key
E           AssertionError: Property `__annotations__` is missing from config schema
E           assert False
```
mypy typings are interfering with this test specifically. I believe you can change the behaviour when running tests so this doesn't happen, I'll have a look

mosen:
revisiting this now as I have time

mosen:
rebased on upstream master

mosen:
tests now passing, please review @alecmerdler when you get a chance

flavianmissi:
> sure there might be a whole bunch of merge conflicts now :)

Yeah :/ we're really sorry it took us this long.
Just @ me when it's ready for another round and I'll get to it ASAP so we avoid leaving this PR in a similar situation again 🚀

mosen:
Rebased and force pushed just checking mypy errors / tests to validate that no regressions were introduced by the rebase

mosen:
type definitions corrected for master as of now. 
mypy reports 245 errors which i deliberately did not take any action on.
@flavianmissi have a look when you're ready

mosen:
wrt `buildman/manager/etcd_canceller.py` its probably a deleted file in upstream merged with my local branch so probably a mistake.

PR #463, opened 2020-06-29
Have the RAC worker fast-check whether it needs to do work

josephschorr:
The RAC worker will now check the count of entries for the day, versus
the number of repos. If all entries are found, the worker will go to
sleep.

### DIFF
--- a/workers/repositoryactioncounter.py
+++ b/workers/repositoryactioncounter.py
@@ -15,7 +15,7 @@ from workers.worker import Worker, with_exponential_backoff
 
 logger = logging.getLogger(__name__)
 
-POLL_PERIOD_SECONDS = 30 * 60  # 30 minutes
+POLL_PERIOD_SECONDS = 4 * 60 * 60  # 4 hours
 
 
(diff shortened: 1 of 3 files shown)


workers/repositoryactioncounter.py
```diff
@@ -34,8 +34,16 @@ def batch_query():
         if min_id is None or max_id is None:
             return
 
-        # 4^log10(total) gives us a scalable batch size into the billions.
-        batch_size = int(4 ** log10(max(10, max_id - min_id)))
+        # Check for the number RAC entries vs number of repos. If they are the same,
+        # nothing more to do.
+        repo_count = model.repository.get_repository_count()
+        rac_count = model.repositoryactioncount.found_entry_count(yesterday)
+        if rac_count >= repo_count:
```
kleesc, line 41: `==` or `>=`? Is there a case where the RAC should ever be higher than the repo count (repo deletions?)

josephschorr: Repository gets deleted and removed between one check and the other.

PR #466, opened 2020-07-01
Have appr blob redirect to storage when possible

josephschorr:


jzelinskie:
I've confirmed that operator-marketplace and operator-registry both leverage [openapi](https://github.com/go-openapi/runtime/tree/master/client) to interact with Quay, which does not appear to touch the default client settings for `CheckRedirect` which defines the policy for redirects in the standard Go client.

jzelinskie:
cc @ecordell

PR #468, opened 2020-07-02
DRAFT: Investigate adding security rule to repo mirroring

josephschorr:
DRAFT: Investigate adding security rule to repo mirroring

This new rule allows for skipping tags that have known security vulns
with severity that has not be explicitly allowed

### DIFF
--- a/workers/repomirrorworker/__init__.py
+++ b/workers/repomirrorworker/__init__.py
@@ -1,9 +1,13 @@
+import base64
 import os
 import re
 import traceback
 import fnmatch
 import logging.config
 
+import requests
+
+from contextlib import contextmanager
 from prometheus_client import Gauge
 
 import features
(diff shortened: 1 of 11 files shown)


workers/repomirrorworker/__init__.py
```diff
@@ -83,260 +80,217 @@ def process_mirrors(skopeo, token=None):
...
+    )
 
-    :param result: SkopeoResults object
-    :return: Message to display
-    """
+    had_failure = False
+    for tag in tags:
+        # Re-claim the mirror between each operation, to ensure no other worker
+        # preempts use
+        mirror = claim_mirror(mirror)
```
thomasmckay, line 217: After this, the caller of this method will have an obsolete reference to the mirror (sync_transaction_id will have changed). This method should return the mirror to the caller for continued use.

Perhaps return the mirror from this method and raise an Exception instead of returning had_failure?

workers/repomirrorworker/mirror_rules.py
```diff
@@ -0,0 +1,129 @@
...
+            continue
+
+        severities_found = set()
+        for feature in secscan_info.json().get("data", {}).get("Layer", {}).get("Features", []):
+            for vuln in feature.get("Vulnerabilities", []):
+                severities_found.add(vuln.get("Severity") or "Unknown")
+
+        logger.debug("Found severities for tag `%s`: %s", tag, severities_found)
+        if not severities_found:
+            yield tag
```
thomasmckay, line 108: Is this correct? If it is, I need some python learning. :)

PR #486, opened 2020-07-21
PROJQUAY-889 Fix cert generation

ibazulic:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-889

**Changelog:** 
Added the ability to create certificates based on hostname parsed as an env. variable.

**Docs:** 
When starting Quay config app, parse the additional `˙-e QUAY_CONFIG_HOSTNAME=some.hostname` to assign a CN to the created certificate. If the variable is not present, Quay config app will create default certificates with a bad CN that may not work in all browsers.

**Testing:** 
Run Quay config app with `-e QUAY_CONFIG_HOSTNAME=some.hostname`˙. 

**Details:** 
n/a


ibazulic:
For some reason, this pull pulled a bunch of additional PR requests. Which was not intended. Creating a new PR.

PR #487, opened 2020-07-21
PROJQUAY-889 Enable adding of specific CN to certs used by the config app

ibazulic:
**Issue**: https://issues.redhat.com/browse/PROJQUAY-889

**Changelog**:
Added the ability to create certificates based on hostname parsed as an env. variable.

**Docs**:
When starting Quay config app, parse the additional `-e QUAY_CONFIG_HOSTNAME=some.hostname` to assign a CN to the created certificate. If the variable is not present, Quay config app will create default certificates with a bad CN that may not work in all browsers.

**Testing**:
Run Quay config app with -e `QUAY_CONFIG_HOSTNAME=some.hostname`.

**Details**:
n/a

### DIFF
--- a/config_app/init/certs_create.sh
+++ b/config_app/init/certs_create.sh
@@ -19,7 +19,7 @@ else
             -subj "/C=US/ST=NY/L=NYC/O=Dis/CN=self-signed" \
             -keyout quay-config-key.pem  -out quay-config.pem
         else
-            openssl req -new -newkey rsa:4096 -days 3650 -nodes -x509 \
+        openssl req -new -newkey rsa:4096 -days 3650 -nodes -x509 \
             -subj "/C=US/ST=NY/L=NYC/O=Dis/CN=${QUAY_CONFIG_HOSTNAME}" \
             -keyout quay-config-key.pem  -out quay-config.pem
     fi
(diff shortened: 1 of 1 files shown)


ibazulic:
@kurtismullins can I get a review for this one as well? :-)

config_app/init/certs_create.sh
```diff
@@ -14,9 +14,15 @@ else
...
-        -keyout quay-config-key.pem  -out quay-config.pem
+    if [ -z "$QUAY_CONFIG_HOSTNAME" ];      then
+        openssl req -new -newkey rsa:4096 -days 3650 -nodes -x509 \
+            -subj "/C=US/ST=NY/L=NYC/O=Dis/CN=self-signed" \
+            -keyout quay-config-key.pem  -out quay-config.pem
+        else
+            openssl req -new -newkey rsa:4096 -days 3650 -nodes -x509 \
+            -subj "/C=US/ST=NY/L=NYC/O=Dis/CN=${QUAY_CONFIG_HOSTNAME}" \
+            -keyout quay-config-key.pem  -out quay-config.pem
+    fi
```
thomasmckay, line 25: indentation doesn't look correct

ibazulic: Corrected to be the same as before.

ibazulic: I force pushed changes so it should be okay now.

PR #492, opened 2020-07-24
Fix missing joins or lookups to reduce query count

josephschorr:
Peewee will invoke an automatic lookup if these are missing, which can
cause excess queries to be made to the database. By adding the missing
join's and model classes into the 'select', we remove these extra calls

**Issue:** https://issues.redhat.com/browse/PROJQUAY-900

**Changelog:** None, see the commit

**Docs:**  None

**Testing:** Any API touched should be tested for correctness:

- App Registry push/pull and listing
- Listing build triggers and starting a build via a triggered push
- Viewing repository notifications
- Creating and using an OAuth application
- Looking at the default permissions for an organization and adding a new default permission
- Inviting someone to a team and having them accept
- Deleting a repository and verify GC works
- Deleting a namespace and verify GC works
- Have a user confirm their email for a new account

**Details:** See above


data/database.py
```diff
@@ -1391,8 +1391,8 @@ class ExternalNotificationMethod(BaseModel):
 class RepositoryNotification(BaseModel):
     uuid = CharField(default=uuid_generator, index=True)
     repository = ForeignKeyField(Repository)
-    event = ForeignKeyField(ExternalNotificationEvent)
-    method = ForeignKeyField(ExternalNotificationMethod)
+    event = EnumField(ExternalNotificationEvent)
```
alecmerdler, line 1394: Doesn't this require a migration?

josephschorr: Nope. EnumField is just a wrapper around ForeignKeyField

PR #496, opened 2020-07-28
Fix clients test for Python 3

josephschorr:
Must now be run from the quay root directory with PYTHONPATH=.

**Issue:** None

**Changelog:** None

**Docs:** None

**Testing:** This is a test suite

**Details:** Fix the clients test suite for Python 3

### DIFF
--- a/test/clients/clients_test.py
+++ b/test/clients/clients_test.py
@@ -9,7 +9,7 @@ from threading import Thread
 
 from termcolor import colored
 
-from .client import DockerClient, PodmanClient, Command, FileCopy
+from test.clients.client import DockerClient, PodmanClient, Command, FileCopy
 
 
 def remove_control_characters(s):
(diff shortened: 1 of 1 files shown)


test/clients/clients_test.py
```diff
@@ -59,7 +51,7 @@ def _check_vagrant():
         for path in os.environ.get("PATH").split(":")
     )
     vagrant_plugins = subprocess.check_output([vagrant_command, "plugin", "list"])
-    return (vagrant, "vagrant-scp" in vagrant_plugins)
+    return (vagrant, "vagrant-scp" in str(vagrant_plugins))
```
kleesc, line 54: Should use `decode` instead of `str` here. It'll end up being a string repr of that `bytes` otherwise. e.g `"b'somebytes'"`

test/clients/clients_test.py
```diff
@@ -153,15 +145,15 @@ def _run_and_wait(command, error_allowed=False):
     failed = result != 0 and not error_allowed
     # vagrant scp doesn't report auth failure as non-0 exit
     failed = failed or (
-        len(command) > 1 and command[1] == "scp" and "authentication failures" in output
+        len(command) > 1 and command[1] == "scp" and "authentication failures" in str(output)
```
kleesc, line 148: Same here and below where `str` is used.

test/clients/clients_test.py
josephschorr: That explains a lot!

PR #497, opened 2020-07-28
Add an endpoint for tracking the secscan backfill

josephschorr:
Fixes https://issues.redhat.com/browse/PROJQUAY-912

**Issue:** https://issues.redhat.com/browse/PROJQUAY-912

**Changelog:** None

**Docs:** Provide

**Testing:** Start a pre-filled Quay with Clair V4 attached and watch `/secscan/_backfill_status`

**Details:** None

### DIFF
--- a/endpoints/secscan.py
+++ b/endpoints/secscan.py
@@ -1,6 +1,7 @@
 import logging
 
-from flask import make_response, Blueprint
+from flask import make_response, Blueprint, jsonify
+from data.database import ManifestSecurityStatus, Manifest
 from endpoints.decorators import anon_allowed
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


endpoints/secscan.py
```diff
@@ -11,3 +12,12 @@
...
 def internal_ping():
     return make_response("true", 200)
+
+
+@secscan.route("/_backfill_status")
+@anon_allowed
+def manifest_security_backfill_status():
+    manifest_count = Manifest.select().count()
+    mss_count = ManifestSecurityStatus.select().count()
+    return jsonify({"backfill_percent": mss_count / float(manifest_count),})
```
alecmerdler, line 22: I mean this is really just reporting what percent of manifests in the registry have been scanned. Could be used not only for tracking backfill, but in general as well.

josephschorr: That's the point, though: we'll ideally remove this once done

PR #500, opened 2020-07-30
[PROJQUAY-917] Fix CSRF token encoding

kurtismullins:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-917

**Changelog:** Fix CSRF token encoding in Python 3

**Docs:** n/a

**Testing:** Using any form in the UI, such as creating a User, should no longer fail with a CSRF token error.

**Details:** The CSRF token is being encoded in a way that causes the front-end to incorrectly escape certain sequences contained within the token. This causes CSRF validation to fail as the token created by the back-end and stored in the session does not match what is sent by the browser. This change fixes that issue.

### DIFF
--- a/endpoints/csrf.py
+++ b/endpoints/csrf.py
@@ -30,7 +30,7 @@ def generate_csrf_token(session_token_name=_QUAY_CSRF_TOKEN_NAME, force=False):
     Returns the generated token.
     """
     if session_token_name not in session or force:
-        session[session_token_name] = base64.b64encode(os.urandom(48)).decode("ascii")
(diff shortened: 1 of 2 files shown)


endpoints/csrf.py
```diff
@@ -30,7 +30,7 @@ def generate_csrf_token(session_token_name=_QUAY_CSRF_TOKEN_NAME, force=False):
     Returns the generated token.
     """
     if session_token_name not in session or force:
-        session[session_token_name] = base64.b64encode(os.urandom(48)).decode("ascii")
+        session[session_token_name] = base64.urlsafe_b64encode(os.urandom(48)).decode("UTF-8")
```
josephschorr, line 33: Make lowercase `utf-8`

PR #512, opened 2020-08-10
Mark invalid manifests in Secscan V4

josephschorr:
If we get an invalid manifest when indexing, we now mark the manifest
as invalid for security indexing

Fixes https://issues.redhat.com/browse/PROJQUAY-949

### DIFF
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -214,6 +214,28 @@ class V4SecurityScanner(SecurityScannerInterface):
         for candidate, abt, num_remaining in iterator:
             manifest = ManifestDataType.for_manifest(candidate, None)
             layers = registry_model.list_manifest_layers(manifest, self.storage, True)
+            if layers is None:
+                logger.warning(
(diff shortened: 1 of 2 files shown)


data/secscan_model/secscan_v4_model.py
```diff
@@ -214,6 +214,28 @@ def needs_reindexing_query(indexer_hash):
         for candidate, abt, num_remaining in iterator:
             manifest = ManifestDataType.for_manifest(candidate, None)
             layers = registry_model.list_manifest_layers(manifest, self.storage, True)
+            if layers is None:
```
alecmerdler, line 217: Should we add a test for this?

josephschorr: Yes, we should, but I wanted to get this fixed for QE. Should I do so as a followup or would you like to do so?

alecmerdler: We can do it as a followup.

josephschorr: Added it to this PR

PR #508, opened 2020-08-07
PROJQUAY-941 - Initialize the swift retry count correctly so that it is never None

kleesc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-941

**Changelog:** 

**Docs:** 

**Testing:** 

**Details:** 

------
(_This section may be deleted._)
**All fields are required.** If a field is not applicable (eg. no relevant CHANGELOG.md), specify "none" or "n/a".

Issue: This is the PROJQUAY jira reference. Pull-request title must start with issue name "PROJQUAY-1234 - ".

Changelog: One line description to be added to CHANGELOG.md during release builds. Typically starts with "Added:", "Fixed:", "Note:", etc.

Docs: Detailed description of changes necessary to docs.projectquay.io. Examples would be addition of config.yaml, indication of UI changes and screenshot impact, and changes in behavior of features.

Testing: Detailed description of how to test changes manually. This section combined with the _Docs_ section above must be sufficiently clear for full test cases to be performed.

Details: Other information meant for pull-request reviewers and developers.

PR #513, opened 2020-08-10
Fallback to reading 0 bytes if readable is not implemented.

kleesc:
Gunicorn will wrap the wsgi stream in its own wrapper class (see `class Body`).
This class is not compliant with the file-object api from Python (missing `readable`).

This issue never showed up before because of
https://github.com/quay/quay/blob/master/storage/swift.py#L418 always evaluating to `False` in Python2 which
was fixed by https://github.com/quay/quay/pull/508.

**Issue:** https://issues.redhat.com/browse/PROJQUAY-941

**Changelog:** 

**Docs:** 

**Testing:** 

**Details:**

### DIFF
--- a/util/registry/test/test_filelike.py
+++ b/util/registry/test/test_filelike.py
@@ -149,3 +149,69 @@ def test_slice_explictread():
     assert stream.read(2) == b"is"
     assert stream.read(5) == b" a"
     assert len(b"is a") == stream.tell()
+
+
+def test_non_filelike_obj_read_limit():
+    from gunicorn.http.body import Body
+
+    content = b"this will not really be a real fileobj"
+    fileobj = BytesIO(content)
+    body = Body(fileobj)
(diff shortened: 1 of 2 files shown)


util/registry/test/test_filelike.py
```diff
@@ -149,3 +149,12 @@ def test_slice_explictread():
...
     assert len(b"is a") == stream.tell()
+
+
+def test_non_filelike_obj():
+    from gunicorn.http.body import Body
+
+    fileobj = BytesIO(b"this will not really be a real fileobj")
+    body = Body(fileobj)
+    ls = LimitingStream(body)
+    assert ls.readable()
```
josephschorr, line 160: Execute a read as well, and ensure it finds the correct data and is limited

kleesc: Done

PR #518, opened 2020-08-12
Fix label application to manifests

josephschorr:
We now apply labels to manifests in batches as an optimization and we
now apply the *action* labels to tags whenever they are moved,
regardless of whether the manifest has just been created

Fixes https://issues.redhat.com/browse/PROJQUAY-279

### DIFF
--- a/data/model/oci/label.py
+++ b/data/model/oci/label.py
@@ -22,6 +22,28 @@ from util.validation import is_json
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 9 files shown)


ibazulic:
@HammerMeetNail we need this patch to fix the bug for the label creation. As discussed in the JIRA, we currently process labels on manifest creation, not on tag creation, so when multiple tags are created from the same manifest, only the first pushed tag will have expiry correctly set, others won't.

PR #533, opened 2020-08-25
Added config-tool as startup script

jonathankingfc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1008

**Changelog:** 
- Added config-tool script to startup
- Quay container will show errors and exit on startup if configuration is invalid. 

**Docs:** 
-N/A

**Testing:** 
- Being sent to QE

**Details:** 
- N/A
------

### DIFF
--- a/conf/init/01_validate_config.sh
+++ b/conf/init/01_validate_config.sh
@@ -1,0 +1,6 @@
+#! /bin/sh
+QUAYPATH=${QUAYPATH:-"."}
+QUAYCONF=${QUAYCONF:-"$QUAYPATH/conf"}
+
+echo "Validating Configuration"
+config-tool validate -c $QUAYCONF/stack/ --mode online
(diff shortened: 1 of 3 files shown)


conf/init/01_validate_config.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/sh
+QUAYPATH=${QUAYPATH:-"."}
+QUAYCONF=${QUAYCONF:-"$QUAYPATH/conf"}
+
+echo "Validating Configuration"
+config-tool validate -c $QUAYCONF/stack/ --mode online
```
thomasmckay, line 6: add env var to skip check

thomasmckay: For now, let's make this a non-failing check for testing purposes.

Perhaps also change the order of this to after the certs_create.sh and certs_install.sh are run.

PR #536, opened 2020-08-31
PROJQUAY-1021 chore: Version lock black to 19.10b0

arajkumar:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1021

**Details:** 

Github Actions step which checks python code style installs `black` without specifying a version which would always installs the latest one. This would lead to unpredictable CI build failures.

As a remedy, I proposed to version lock black to 19.10b0.

Other option is to modify the Quay code style using 20.8b1. See https://github.com/quay/quay/pull/537


arajkumar:
Fixed in #535

PR #542, opened 2020-09-02
[PROJQUAY-1035] fix: Convert gunicorn worker counts to int for comparisons

kurtismullins:
**Issue:**
https://issues.redhat.com/browse/PROJQUAY-1035

**Changelog:** 
This is a fix specific to the Python2 -> Python3 update. Nothing special needed in the changelog.

**Docs:** 
n/a - bug fix.

**Testing:** 
Deploy Quay to kubernetes using the following environment variables:
```
          - name: WORKER_COUNT_REGISTRY
            value: '8'
          - name: WORKER_COUNT_WEB
            value: '4'
          - name: WORKER_COUNT_SECSCAN
            value: '2'
```

**Details:** 
Fixes the following traceback:
```
Failed to read config file: /quay-registry/conf/gunicorn_web.py
Traceback (most recent call last):
File "/usr/local/lib/python3.6/site-packages/gunicorn/app/base.py", line 111, in get_config_from_filename
spec.loader.exec_module(mod)
File "<frozen importlib._bootstrap_external>", line 678, in exec_module
File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
File "/quay-registry/conf/gunicorn_web.py", line 21, in <module>
workers = get_worker_count("web", 2, minimum=2, maximum=32)
File "/quay-registry/util/workers.py", line 32, in get_worker_count
return max(override_value, minimum)
TypeError: '>' not supported between instances of 'int' and 'str'
```


util/workers.py
```diff
@@ -28,6 +28,10 @@ def get_worker_count(worker_kind_name, multiplier, minimum=None, maximum=None):
 
     # Check for an override via an environment variable.
     override_value = os.environ.get("WORKER_COUNT_" + worker_kind_name.upper())
+
+    if override_value:
+        override_value = int(override_value)
```
kleesc, line 33: nit: You're already checking override_value below. Also check for a `ValueError` when doing the `int` conversion.

kurtismullins: I could handle the potential `ValueError` but want it to fail-fast if Quay is deployed with a non-integer.

I'm not opposed to giving a better error message though. Thoughts @kleesc ?

util/workers.py
```diff
@@ -26,13 +28,16 @@ def get_worker_count(worker_kind_name, multiplier, minimum=None, maximum=None):
     minimum = minimum or multiplier
     maximum = maximum or (multiplier * multiplier)
 
-    # Check for an override via an environment variable.
+    # Check for a process-specific override via an environment variable.
     override_value = os.environ.get("WORKER_COUNT_" + worker_kind_name.upper())
-    if override_value is not None:
+    if override_value:
```
kleesc, line 33: This should still check against `None`

util/workers.py
```diff
@@ -26,13 +28,16 @@ def get_worker_count(worker_kind_name, multiplier, minimum=None, maximum=None):
...
+        override_value = int(override_value)
         return max(override_value, minimum)
 
+    # Check for a generic override via an environment variables
     override_value = os.environ.get("WORKER_COUNT")
-    if override_value is not None:
+    if override_value:
```
kleesc, line 39: Ditto.

kleesc (changes requested):

util/workers.py
kurtismullins: Why against None?

kleesc: To be consistent with how the other checks are done.
And in this case its generally better, since if for some reason the value of override_value was 0, it would still evaluate correctly.

kurtismullins: That is a good point -- if the environment variable is set, and somehow ends up with a value which evaluates to False, then we should handle that case accordingly.

Fixed.

kleesc (approved):

PR #551, opened 2020-09-10
PROJQUAY-1073

jonathankingfc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1073

**Changelog:** 

**Docs:** 

- With the new config-tool integration, there are no changes to how a user needs to bring up the Quay registry. There is, however, an updated process to validate a config bundle before Quay startup. On startup, Quay will print a table in which components either pass or fail validation. Validation errors will include messages on what caused the error. If a config bundle fails to be validated, Quay will exit and the startup process is cancelled. However, users may add an env variable (IGNORE_VALIDATION) that will bypass an exit on failed validation (the default behavior is to shut down Quay on startup if config bundle is invalid).

- The new config-tool includes a new backend engine for the config app. There are very few UX and UI changes to take note of. 

- For the config-app, the command is slightly different. Users should run the following command:

`sudo podman run --rm -it --name quay_config -p 7070:8080 -v $QUAY/quay-config:/conf/stack quay:<tag> config <password> [operator-endpoint]`

This will bring up the config editor with the provided username and password on port 7070. Note that users must mount in their Quay config directory to /conf/stack in order for the config tool to be able to read it. Also, if the config-tool is being used to configure the Quay Operator, a user can specify an endpoint at which the config-tool should post a reconfigured secret (Reach out to @alecmerdler for more details regarding the new Quay Operator).

- As for UI changes, the config app is now a single page web application. As noted earlier, users no longer have to upload a config.tar.gz because their config directory should be mounted into the container at startup. In addition to this, the config-app no longer asks a user whether or not they would like to create a bundle from scratch. If no confug-bundle is mounted, the user will be provided a page with default values filled in. At this point, a user must fill in the required fields. Once all required fields are filled, the user will be shown a "Validate Config" button. 

If validation is successful, they will be shown a success modal with the options to "Download config" and a "Commit to Operator" button (only if they are running the config app from within the Operator). If validation is failed, users will be shown a failure modal with the validation errors. 

- The config-app exposes an API for users to use. Once the container is running in config-app mode (the same command specified above), you can go to the endpoint `/swagger/index.html` to see swagger documentation regarding the API. The frontend uses this API for all of its functionality. 

TO USE TLS:
- TLS is optionally available if a user mounts a public private key pair. If no certs are provided, the config-tool will default to http. To use TLS, include the following flags in your `podman run` command:
```
        -v ./privateKey.key:${CONFIG_TOOL_PRIVATE_KEY} \
	-v ./publicKey.crt:${CONFIG_TOOL_PUBLIC_KEY} \
        -e CONFIG_TOOL_PUBLIC_KEY=${CONFIG_TOOL_PUBLIC_KEY} \
	-e CONFIG_TOOL_PRIVATE_KEY=${CONFIG_TOOL_PRIVATE_KEY} \
```
With these flags, we are mounting our keypair at a specific location which is then provided to the config-tool through the environment variables. 

**Testing:** 

**Details:** 
------

### DIFF
--- a/test/test_secscan.py
+++ b/test/test_secscan.py
@@ -92,7 +92,7 @@ class TestSecurityScanner(unittest.TestCase):
             # Retrieve the results.
             result = self.api.get_layer_data(manifest, include_vulnerabilities=True)
             self.assertIsNotNone(result)
-            self.assertEquals(result["Layer"]["Name"], security_scanner.layer_id(manifest))
+            self.assertEqual(result["Layer"]["Name"], security_scanner.layer_id(manifest))
 
 
 if __name__ == "__main__":
--- a/util/secscan/v4/api.py
+++ b/util/secscan/v4/api.py
@@ -81,13 +81,14 @@ class SecurityScannerAPIInterface(object):
 Action = namedtuple("Action", ["name", "payload"])
 
 actions = {
-    "IndexState": lambda: Action("IndexState", ("GET", "index_state", None)),
-    "Index": lambda manifest: Action("Index", ("POST", "index_report", manifest)),
+    "IndexState": lambda: Action("IndexState", ("GET", "/indexer/api/v1/index_state", None)),
+    "Index": lambda manifest: Action("Index", ("POST", "/indexer/api/v1/index_report", manifest)),
     "GetIndexReport": lambda manifest_hash: Action(
-        "GetIndexReport", ("GET", "index_report/" + manifest_hash, None)
+        "GetIndexReport", ("GET", "/indexer/api/v1/index_report/" + manifest_hash, None)
     ),
     "GetVulnerabilityReport": lambda manifest_hash: Action(
-        "GetVulnerabilityReport", ("GET", "vulnerability_report/" + manifest_hash, None,)
+        "GetVulnerabilityReport",
+        ("GET", "/matcher/api/v1/vulnerability_report/" + manifest_hash, None,),
     ),
 }
 
@@ -97,7 +98,7 @@ class ClairSecurityScannerAPI(SecurityScannerAPIInterface):
         self._client = client
         self._blob_url_retriever = blob_url_retriever
 
-        self.secscan_api_endpoint = urljoin(endpoint, "/api/v1/")
+        self.secscan_api_endpoint = endpoint
 
     def state(self):
         try:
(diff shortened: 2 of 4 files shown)


util/config/provider/baseprovider.py
```diff
@@ -41,14 +41,14 @@ def import_yaml(config_obj, config_file):
...
-    else:
-        logger.debug("Skipping config schema validation because setup is not complete")
+    # if config_obj.get("SETUP_COMPLETE", True):
+    #     try:
+    #         validate(config_obj, CONFIG_SCHEMA)
+    #     except ValidationError:
+    #         # Todo: Change this into a real error
+    #         logger.exception("Could not validate config schema")
+    # else:
+    #     logger.debug("Skipping config schema validation because setup is not complete")
```
thomasmckay, line 51: Will `SETUP_COMPLETE` in config.yaml still be used?

thomasmckay (changes requested):

thomasmckay (changes requested):
Ideally, remote the existing `config_app` dir completely. Placing new config-tool results (js and binary) under `/quay-registry/config-tool` would also be good to allow chmod, etc. to remain unchanged.

util/config/provider/baseprovider.py
thomasmckay: @jonathankingfc can you file jira for removing SETUP_COMPLETE uses?

thomasmckay (changes requested):

thomasmckay (approved):

thomasmckay:
LGTM
Let's get jira for SETUP_COMPLETE

quay-entrypoint.sh
```diff
@@ -34,41 +34,45 @@ cat << "EOF"
...
-
  Build, Store, and Distribute your Containers
-
-
 EOF
 
 # Custom environment variables for use in conf/supervisord.conf
 # The gunicorn-registry process DB_CONNECTION_POOLING must default to true
 export DB_CONNECTION_POOLING_REGISTRY=${DB_CONNECTION_POOLING:-"true"}
+export CONFIG_APP_PASSWORD=${CONFIG_APP_PASSWORD:-"\"\""}
```
thomasmckay, line 43: Running container with `-e CONFIG_APP_PASSWORD=password` complains about missing password

thomasmckay (changes requested):

thomasmckay:
issue: The config app UI is shown beneath the username/password login: This will leak information, right?

thomasmckay:
issue: storage doesn't appear configurable?
Quay-Config-Editor

thomasmckay:
issue: basic configuration section thinks its fields are required
issue: time machine does not have allowed expirations and defuault is blank

thomasmckay (approved):
LGTM

PR #554, opened 2020-09-15
sec: implement jwt signing to ClairV4

ldelossa:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-847

**Changelog:** Adds JWT support to ClairV4 client.

**Docs:** Two new config values are introduced. These values must be documented and will determine how Quay uses JWT to Clair.

**Testing:** Start Quay with new config values and confirm requests do not fail with http 401

**Details:**  PR Implements JWT signing toward ClairV4 directly in Quay. This eliminates the need for JWT Proxy.

### DIFF
--- a/util/config/schema.py
+++ b/util/config/schema.py
@@ -702,6 +702,11 @@ CONFIG_SCHEMA = {
             "description": "The number of seconds between indexing intervals in the security scanner. Defaults to 30.",
             "x-example": 30,
         },
+        "SECURITY_SCANNER_V4_PSK": {
+            "type": "string",
+            "description": "A base64 encoded string used to sign JWT(s) on Clair V4 requests. If 'None' jwt signing will not occur.",
+            "x-example": "PSK",
+        },
         # Repository mirroring
         "REPO_MIRROR_INTERVAL": {
             "type": "number",
--- a/util/secscan/v4/api.py
+++ b/util/secscan/v4/api.py
@@ -2,8 +2,11 @@ import logging
 import requests
 import json
 import os
+import jwt
+import base64
 
 from collections import namedtuple
+from datetime import datetime, timedelta
 from enum import Enum
 
 from abc import ABCMeta, abstractmethod
@@ -48,7 +51,7 @@ class SecurityScannerAPIInterface(object):
     @abstractmethod
     def state(self):
         """
-        The state endpoint returns a json structure indicating the indexer's internal configuration state. 
+        The state endpoint returns a json structure indicating the indexer's internal configuration state.
         A client may be interested in this as a signal that manifests may need to be re-indexed.
         """
         pass
@@ -56,7 +59,7 @@ class SecurityScannerAPIInterface(object):
     @abstractmethod
     def index(self, manifest, layers):
         """
-        By submitting a Manifest object to this endpoint Clair will fetch the layers, 
+        By submitting a Manifest object to this endpoint Clair will fetch the layers,
         scan each layer's contents, and provide an index of discovered packages, repository and distribution information.
         Returns a tuple of the `IndexReport` and the indexer state.
         """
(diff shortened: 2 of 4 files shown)


util/secscan/v4/api.py
```diff
@@ -196,6 +206,15 @@ def _perform(self, action):
 
         return resp
 
+    def _sign_jwt(self):
```
kurtismullins, line 209: In this function, It appears there's two primary paths:
```
if self.jwt_psk
```
and
```
else
```
For readability purposes, it might make more sense to wrap everything that should happen when `self.jwt_psk` evaluates to true/exists, and put the alternate/fallback into an `else` statement.

kurtismullins (changes requested):
Some minor requests. Most are nitpicks/style suggestions while the requirements.txt related changes should probably be fixed before going to master.

Looks good so far!

requirements-dev.txt
```diff
@@ -16,3 +16,4 @@ pytest-sugar
 tox
 tox-docker
 tqdm
+PyJWT==1.7.1
```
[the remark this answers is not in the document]
ldelossa: is there no value to pinning dependency versions so library changes do not break the code?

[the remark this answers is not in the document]
kurtismullins: I would pin it with a version in requirements.txt, exclude it from the requirements-dev.txt, and remove the version number from requirements-nover.txt.

(Using Github app on my phone, apologies for the lack of formatting or if I messed up any of the file names)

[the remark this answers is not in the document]
ldelossa: Cool thats done.

requirements.txt
```diff
@@ -186,3 +186,4 @@ xmltodict==0.12.0
 yapf==0.29.0
 zipp==2.1.0
 zope.interface==4.7.1
+PyJWT==1.7.1
```
kleesc, line 189: Duplicate. It;s already installed.

kleesc (dismissed):

util/secscan/v4/api.py
```diff
@@ -197,6 +224,33 @@ def _perform(self, action):
...
+        # early return on psk
+        if self.jwt_psk:
+            payload = {
+                "iss": self.instance_keys.service_name,
+                "exp": datetime.utcnow() + timedelta(minutes=5),
+            }
+            token = jwt.encode(payload, self.jwt_psk, algorithm="HS256").decode("utf-8")
+            return token
+
+        payload = {
```
alecmerdler, line 244: Remove this, we'll only use the PSK or nothing.

util/config/schema.py
```diff
@@ -702,6 +702,16 @@
             "description": "The number of seconds between indexing intervals in the security scanner. Defaults to 30.",
             "x-example": 30,
         },
+        "SECURITY_SCANNER_V4_SIGN_JWT": {
```
alecmerdler, line 705: Only add a single field: `SECURITY_SCANNER_V4_JWT_PSK` and make the decision to sign/verify requests with JWT based on the presence or absence of that field.

util/secscan/v4/api.py
```diff
@@ -94,9 +97,27 @@ def vulnerability_report(self, manifest_hash):
 
 
 class ClairSecurityScannerAPI(SecurityScannerAPIInterface):
-    def __init__(self, endpoint, client, blob_url_retriever):
+    """
+    if sign_jwt is true this class wil sign JWT requests to ClairV4
```
alecmerdler, line 101: Capitalize beginning of sentances.

alecmerdler (changes requested):
Let's remove the keyserver stuff and only sign JWT's with the pre-shared key, if it exists.

util/secscan/v4/api.py
```diff
@@ -197,6 +218,20 @@ def _perform(self, action):
...
 
+    def _sign_jwt(self):
+        """
+        Sign and return a jwt.
+
+        If self.jwt_psk is provided a pre-shared key will be used as the signing key.
+        """
+        payload = {
+            "iss": "quay",
+            "exp": datetime.utcnow() + timedelta(minutes=5),
```
kurtismullins, line 229: We should make this timedelta threshold configurable. Not necessary for this PR but probably good for inclusion as a TODO comment.

alecmerdler (approved):

kurtismullins (approved):
I have not actually tested this but from a visual overview, it LGTM.

PR #555, opened 2020-09-16
add: setup.py

kurtismullins:
**Work-In-Progress / Draft**

NOTE: The Jira actually asks for a `setup.cfg` instead of a `setup.py`. Feel free to create a new Pull Request and this one can be dismissed.

**Issue:** https://issues.redhat.com/browse/PROJQUAY-971

**Changelog:** add: setup.py

**Docs:** n/a

**Testing:** n/a

**Details:**
- Used for "container-first" images. Not intended for general use.
- This is only an example to get one started with fleshing out a setup.py. It breaks on `dumb-init`. See the comment in `requirements.txt` for context and a solution.
- It appears to run somewhat successfully but does not install all packages. Since it's aimed at the container-first builds, one will have to test it to see how well it works.


kurtismullins:
Closing as we do not need a `setup.py` file.

PR #561, opened 2020-09-29
PROJQUAY-1079 - do not rescan bogus manifests

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1079

**Changelog:** 

**Docs:** 

**Testing:** 
Confirm that when quay sends manifests to clair that it cannot parse, that the row in db is updated to -2 and not rescanned further.

**Details:** 
A helm repo was being pushed to clair with an empty layer array. Clair responds w/ a 400/BadRequest. Treat this now as an indication that clair _cannot_ scan this and should not retry. A further check was added to prevent empty layers from being sent to begin with.

### DIFF
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -24,7 +24,7 @@ from data.registry_model.datatypes import Manifest as ManifestDataType
 from data.registry_model import registry_model
 from util.migrate.allocator import yield_random_entries
 from util.secscan.validator import V4SecurityConfigValidator
-from util.secscan.v4.api import ClairSecurityScannerAPI, APIRequestFailure
+from util.secscan.v4.api import ClairSecurityScannerAPI, APIRequestFailure, InvalidContentSent
 from util.secscan import PRIORITY_LEVELS
 from util.secscan.blob import BlobURLRetriever
 from util.config import URLSchemeAndHostname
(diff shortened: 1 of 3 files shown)


data/secscan_model/secscan_v4_model.py
```diff
@@ -190,6 +190,7 @@ def needs_reindexing_query(indexer_hash):
                 Manifest.select()
                 .join(ManifestSecurityStatus)
                 .where(
+                    ManifestSecurityStatus.index_status != IndexStatus.MANIFEST_UNSUPPORTED,
```
thomasmckay, line 193: do not attempt to rescan unsupported manifests

data/secscan_model/secscan_v4_model.py
```diff
@@ -211,10 +212,25 @@ def needs_reindexing_query(indexer_hash):
...
+                    indexer_hash="none",
+                    indexer_version=IndexerVersion.V4,
+                    metadata_json={},
+                )
+
         for candidate, abt, num_remaining in iterator:
             manifest = ManifestDataType.for_manifest(candidate, None)
             layers = registry_model.list_manifest_layers(manifest, self.storage, True)
-            if layers is None:
+            if layers is None or layers == []:
```
thomasmckay, line 233: helm charts have an empty array of layers, which is not scannable

alecmerdler: Maybe `len(layers) == 0` might be more readable.

data/secscan_model/secscan_v4_model.py
```diff
@@ -211,10 +212,25 @@ def needs_reindexing_query(indexer_hash):
             ),
         )
 
+        def manifest_unsupported_status(manifest):
```
alecmerdler, line 215: Perhaps rename to `mark_manifest_unsupported` to better represent the effect calling this function has.

data/secscan_model/secscan_v4_model.py
alecmerdler: Can we add a unit test for this?

alecmerdler (changes requested):
Looks great, let's just add a unit test to prevent regressions.

alecmerdler (approved):

PR #562, opened 2020-09-30
[PROJQUAY-822] security: Hide sensitive LDAP log values

kurtismullins:
**Issue:** 

- https://issues.redhat.com/browse/PROJQUAY-822

**Changelog:**

- security: Modified an LDAP logging statement to exclude potentially sensitive data.

**Docs:**

- n/a

**Testing:**

Run Quay with LDAP integration and DEBUG=true. When logging in, one should see a log line similar to the following:
```
2020-10-05 23:56:01,460 [24247] [DEBUG] [data.users.externalldap] Found matching DNs: ['uid=euclid,dc=example,dc=com']
```

**Details:** 

Exclude potential passwords and other sensitive data from a given LDAP logging statement.

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -217,7 +217,9 @@ class LDAPUsers(FederatedUsers):
             if err_msg is not None:
                 return (None, err_msg)
 
-            logger.debug("Found matching pairs: %s", pairs)
+            dn_lst = [pair[0] for pair in pairs]
+            logger.debug("Found matching DNs: %s" % dn_lst)
+
             results = [LDAPUsers._LDAPResult(*pair) for pair in take(limit, pairs)]
 
             # Filter out pairs without DNs. Some LDAP impls will return such pairs.


data/users/externalldap.py
```diff
@@ -217,7 +217,21 @@ def _ldap_user_search(self, username_or_email, limit=20, suffix=""):
             if err_msg is not None:
                 return (None, err_msg)
 
-            logger.debug("Found matching pairs: %s", pairs)
+            # This may reveal sensitive information. Hide it unless Quay is started
+            # with an explicit flag used for debugging user-related issues.
+            if os.environ.get("USERS_DEBUG") == "1":
```
kleesc, line 222: Do we ever need to log the resulting attributes? Just debug log the `DN` (first item in the pair) which should contain enough to debug on the ldap side if needed.

kurtismullins: Good suggestion. I've updated the code accordingly.

kleesc (approved):

PR #563, opened 2020-10-01
[PROJQUAY-813] fix: Use database SSL certificate during migrations

kurtismullins:
**DRAFT**

**Issue:** 
- https://issues.redhat.com/browse/PROJQUAY-813

**Changelog:** 
- fix: Use database SSL certificate during migrations

**Docs:** 
- No new documentation needed. This just fixes expected behavior.

**Testing:** 
- Ensure that the Quay container, including migrations (alembic) run successfully when `config.yaml` specifies an SSL certificate to connect to the database.

**Details:** 
- Alembic uses its own connection mechanism for performing database schema migrations. It does not share the same connection mechanism defined within the standard Quay processes. It was excluding additional database connection arguments such as any provides SSL/TLS certificate.
- `threadlocals` is an old connection argument used by Peewee. As far as I can tell, it is no longer necessary to enable or disable this and has been excluded from the documentation and source code in the latest release of Peewee. It is *not* used by the Postgres driver `psycopg2` or the MySQL driver `pymysql` as far as I can see. Accordingly, it has been removed and any accidental inclusion of it within `config.yaml` from legacy installs will ignore it in Alembic. It's already ignored by Quay, itself, during the database connection initialization.

### DIFF
--- a/config.py
+++ b/config.py
@@ -195,7 +195,6 @@ class DefaultConfig(ImmutableConfig):
     # DB config
     DB_URI = "sqlite:///test/data/test.db"
     DB_CONNECTION_ARGS = {
-        "threadlocals": True,
         "autorollback": True,
     }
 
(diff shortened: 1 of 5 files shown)


kurtismullins:
Closing. This is superseded by #596

PR #572, opened 2020-10-12
PROJQUAY-1133 - turn off extra config workers

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1133

**Changelog:** none

**Docs:** none

**Testing:** 
Confirm only config-editor service is running

**Details:**


jonathankingfc (dismissed):
LGTM

kurtismullins (changes requested):
I assume the security scanning worker will still be required until users upgrade from ClairV2 to ClairV4. Thoughts?

Also, if the manifestbackfillworker is no longer needed, why not remove it?

PR #577, opened 2020-10-14
Renamed FEATURE_EXPERIMENTAL_HELM_OCI_SUPPORT feature flag

BillDett:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1142
Removed 'EXPERIMENTAL' from Helm V3 feature flag name.

**Changelog:** 

**Docs:** 
As referenced in 3.4 release notes.

**Testing:** 

**Details:**


alecmerdler (changes requested):
This means anyone who was using this config field before will need to update their `config.yaml` to use the new key to continue using this feature. Is that acceptable?

thomasmckay:
@jonathankingfc When config schema changes, is there anything that is required in config-tool?

BillDett:
@alecmerdler - I think it would be acceptable.  if the feature is no longer experimental then the flag name would be misleading.  Only tire-kickers should be using it while it's experimental so I think asking them to update config isn't unreasonable.

thomasmckay (approved):
LGTM

BillDett (commented):
@alecmerdler - any further concerns here over merging this?

alecmerdler (approved):

PR #582, opened 2020-10-20
Implement executor methods to get the count of running builders

kleesc:
Used by the manager to schedule builds based on the current running
count. Uses the specific executors' api to get the count of running
builders instead of Redis/Orchestrator.

This is due to issues encountered in the past where the manager would
have problems scheduling builds, and go into a weird state when
Redis was unavailable.

Remove wamp's REALM parameters from executor

Remove asyncio from executor

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1169

### DIFF
--- a/buildman/manager/executor.py
+++ b/buildman/manager/executor.py
@@ -1,4 +1,3 @@
-import asyncio
 import datetime
 import hashlib
 import io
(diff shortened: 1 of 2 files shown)


thomasmckay (approved):
LGTM

kurtismullins (approved):
From a visual review, it looks good to me.
Ship it!

PR #586, opened 2020-10-22
Set default QUAY_CONFIG_READ_ONLY_FIELD_GROUPS

jonathankingfc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1179

**Changelog:** 

- Set default QUAY_CONFIG_READ_ONLY_FIELD_GROUPS value so that supervisor template can compile when value is not set. 

**Docs:** 

**Testing:** 

**Details:** 

------


conf/supervisord.conf.jnj
```diff
@@ -257,10 +257,8 @@ environment=
   PYTHONPATH=%(ENV_QUAYDIR)s,
   CONFIG_APP_PASSWORD=%(ENV_CONFIG_APP_PASSWORD)s,
   ENV_OPERATOR_ENDPOINT=%(ENV_OPERATOR_ENDPOINT)s,
-  CONFIG_EDITOR_STATIC_CONTENT_PATH=%(ENV_QUAYDIR)s/config_app/static,
-  MY_POD_NAMESPACE=must_fix,
```
alecmerdler, line 261: Will these be passed to the config editor backend?

jonathankingfc: MY_POD_NAME and MY_POD_NAMESPACE can be removed since they are set by the operator and are not required by regular Quay instances. CONFIG_EDITOR_STATIC_CONTENT_PATH is set below. That line shows as removed since the comma was removed.

jonathankingfc: these changes are tested and working

alecmerdler (approved):

PR #587, opened 2020-10-23
PROJQUAY-702 Add the grpc definition

kleesc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-702

- Add the grpc proto definition and generated files
- Add the grpc dependencies to requirements.txt


requirements.txt
```diff
@@ -58,6 +58,7 @@ geoip2==3.0.0
 gevent==1.4.0
 gipc==1.0.1
 greenlet==0.4.15
+grpcio==1.30.0
```
thomasmckay, line 61: Please add these two packages into alphabetical list in requirements-osbs.txt. Running `pipdeptree` indicates that their dependencies are already present (six and setuptools).

thomasmckay (changes requested):

requirements.txt
kleesc: done

thomasmckay (approved):
LGTM

PR #591, opened 2020-10-28
Buildman rewrite

kleesc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-703

**Changelog:** 

Rewrites the buildmanager to use gRPC
    
Rewrite of the current buildmanager using gRPC.
This deprecates the enterprise type builder, as individual nodes will no longer keep track of build states/statuses because of WAMP.

Also removes trollius, which was required by the WAMP servers.
Instead, gRPC uses a threaded model to serve its requests.

Deprecates etcd as state trakcing for build states in favor of Redis only.

Defines a state interface to manage/transition build states, implemented by the
buildmanager.

**Docs:** 
Example config:
```
FEATURE_BUILD_SUPPORT: True
BUILD_MANAGER:
- ephemeral
- ALLOWED_WORKER_COUNT: 1
  ORCHESTRATOR_PREFIX: buildman/testing/
  ORCHESTRATOR:
    REDIS_HOST: quay-redis-host
    REDIS_SKIP_KEYSPACE_EVENT_SETUP: false # Setup pubsub events at runtime.

  EXECUTORS:
  - EXECUTOR: ec2
    EC2_REGION: us-east-1
    COREOS_AMI: ami-02545325b519192df # Fedora CoreOS. Quay uses an Ignition template to bootstrap the instance
    AWS_ACCESS_KEY: SOMEACCESSKEY
    AWS_SECRET_KEY: SOMESECRETKEY
    EC2_INSTANCE_TYPE: t2.small
    EC2_VPC_SUBNET_ID: SOMESUBNET
    EC2_SECURITY_GROUP_IDS:
    - SOMESECURITYGROUP
    EC2_KEY_NAME: SOMESSHKEY
    QUAY_USERNAME: SOMEUSERNAME_WITH_PUSH
    QUAY_PASSWORD: SOMEPASSWORD_WITH_PUSH
    WORKER_IMAGE: quay.io/quay/quay-builder
    WORKER_TAG: master
    VOLUME_SIZE: 52G # Worker instance spec
    BLOCK_DEVICE_SIZE: 58 # Worker instance spec
    SETUP_TIME: 180 # Time the executor wait for the build to start
```

**Testing:** Setup Quay using an `ephemeral` manager, orchestrator and executor. Try and start a build. 

**Details:**


thomasmckay (approved):
LGTM

thomasmckay (approved):
LGTM

thomasmckay (dismissed):
LGTM

buildman/manager/ephemeral.py
```diff
@@ -41,271 +58,114 @@
...
-                    realm_id,
-                    build_uuid,
-                    executor_name,
-                    execution_id,
-                )
-                await self.terminate_executor(executor_name, execution_id)
+        self._manager_config = {}
+        self._orchestrator = None
 
+    def initialize(self, manager_config):
```
kurtismullins, line 121: This might be a dumb question, but why not `initialize()` in `__init__()`? I don't see that method defined in the ABC.

buildman/manager/ephemeral.py
```diff
@@ -41,271 +58,114 @@
...
 
-        executor = executor_klass(executor_config, self.manager_hostname)
+        executor = executor_klass(executor_config, self._manager_hostname)
         if executor.name in self._executor_name_to_executor:
             raise Exception("Executor with name %s already registered" % executor.name)
 
         self._ordered_executors.append(executor)
         self._executor_name_to_executor[executor.name] = executor
 
+    def generate_build_token(self, token_type, build_id, job_id, expiration):
```
kurtismullins, line 151: This just seems like a very thin wrapper around `build_token()`. Why not call it directly? Same with `verify_build_token()`.

conf/supervisord.conf.jnj
```diff
@@ -44,9 +44,7 @@ stderr_events_enabled = true
 environment=
   PYTHONPATH=%(ENV_QUAYDIR)s
 command=python -m buildman.builder
-autostart = false
-; TODO (kleesc): Re-enable this eventually
-; autostart = {{ config['builder']['autostart'] }}
+autostart = {{ config['builder']['autostart'] }}
```
kurtismullins, line 47: :100:

buildman/manager/ephemeral.py
```diff
@@ -330,169 +186,412 @@ def _metric_prefix(self):
...
+            }
+
+        # If the build args have no buildpack, mark it as a failure before sending
+        # it to a builder instance.
+        if not build_args["build_package"] and not build_args["git"]:
+            logger.error(
+                "Failed to start job %s: insufficient build args - No package url or git",
+                job_id,
+            )
+            # TODO: callback job_completed for cleanup
```
kurtismullins, line 409: Should this be included in the current Pull Request -- or do you expect it to come later?

buildman/jobutil/buildjob.py
```diff
@@ -145,6 +145,26 @@ def _determine_cached_tag_by_tag(self):
 
             return None
 
+    def extract_dockerfile_args(self):
+        dockerfile_path = self.build_config.get("build_subdir", "")
+        context = self.build_config.get("context", "")
+        if not (dockerfile_path == "" or context == ""):
+            # This should not happen and can be removed when we centralize validating build_config
```
kurtismullins, line 152: Should a `TODO` be added here, or in a future Pull-Request, related to centralized validation of build_config?

kurtismullins (commented):
It's a lot on my end to digest as I have never walked through the build-manager code prior to this review. I will not block the merge as I see Tom has ack'd it but I do have some questions/comments from a high-level left in my review.

As far as I can tell, nothing is standing out to me as needing to be addressed. I believe it was discussed that automated test enhancements will come in a future commit. Let's test it out!

kurtismullins (dismissed):
LGTM

buildman/interface.py
```diff
@@ -0,0 +1,95 @@
...
+
+class BuildJobDoesNotExistsError(BuildJobError):
+    """Raised when trying to delete a job that does not exists."""
+
+
+class BuildJobExpiredError(BuildJobError):
+    """Raised when trying to update a job that has already expired."""
+
+
+class BuildJobResult(object):
```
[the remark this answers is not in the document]
kleesc: I just reused the previously defined class for job results.

buildman/interface.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+
+    ERROR = "error"
+    EXPIRED = "expired"
+    COMPLETE = "complete"
+    INCOMPLETE = "incomplete"
+    CANCELLED = "cancelled"
+
+
+RESULT_PHASES = {
```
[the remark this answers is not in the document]
kleesc: Ditto. This just maps the job results to the appropriate database defined build phases

buildman/jobutil/buildstatus.py
```diff
@@ -17,7 +17,6 @@ class StatusHandler(object):
     """
 
     def __init__(self, build_logs, repository_build_uuid):
-        self._current_phase = None
```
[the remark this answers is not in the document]
kleesc: This is not used anymore actually. I will remove it.

buildman/manager/ephemeral.py
```diff
@@ -41,271 +58,114 @@
...
+WORK_CHECK_TIMEOUT = 10
+SETUP_LEEWAY_SECONDS = 30
 
+# Schedule retry durations
 RETRY_IMMEDIATELY_SLEEP_DURATION = 0
 TOO_MANY_WORKERS_SLEEP_DURATION = 10
+CREATED_JOB_TIMEOUT_SLEEP_DURATION = 10
 
+CREATED_JOB_TIMEOUT = 15
+JOB_TIMEOUT_SECONDS = 300
```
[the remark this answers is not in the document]
kleesc: I think we should just set them to sane defaults. I don't think user should worry about how the different timeout values are.

buildman/manager/ephemeral.py
kleesc: This just saves us from having to explicitly pass the hostname and instance keys. But I can call it directly if you want.

buildman/manager/ephemeral.py
kleesc: Updated

buildman/manager/ephemeral.py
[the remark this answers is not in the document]
kurtismullins: Sounds good! If we find out that we need more dynamic configuration around these, we can always change it later.

buildman/manager/ephemeral.py
kurtismullins: It's not very important. Sometimes our code is a little hard to follow since we have a ton of abstraction layers. Just hoping to reduce that over time. I'll go ahead and resolve this one.

Thanks!

buildman/interface.py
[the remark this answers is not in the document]
kurtismullins: Ahh that makes sense; Iterative changes.

Perhaps one day we can take advantage of semantically obvious things like `enum.Enum` but it is far from a blocker. Thanks for the clarification!

kurtismullins (approved):
LGTM

PR #596, opened 2020-11-11
[PROJQUAY-813] fix: Schema migrations with MySQL and SSL

kurtismullins:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-813
**Changelog:** fix: Schema migrations with MySQL/MariaDB and SSL

**Docs:** 
- MySQL certificates now require the server's hostname or IP to be specified in both the SAN (Subject Alternative Names) and CN (Common Name) due to changes in Python >= 3.7.

**Testing:** 
- Ensure Quay can successfully run migrations when using a MySQL/MariaDB database with SSL enforced.

**Details:** 
- Alembic, the package responsible for performing schema migrations, does not share the same database connection as Quay. The config.yaml configuration portion to specify a CA bundle and other MySQL-specific SSL options are now passed to the connection used by Alembic.
- The schema migration code has been cleaned up a bit and currently unused/unsupported functionality has been removed from it to reduce complexity until they're needed or supported.
- No Postgres SSL/TLS support was explicitly added or modified in this change.
- The Prometheus-based metrics/reporting to share the status of migrations has been removed. In Quay v3.4.0, there will be no background migrations executed and it should be fairly obvious when it's finished. This behavior can be easily re-implemented in a clean and documented manner if it's needed.

### DIFF
--- a/data/migrations/env.py
+++ b/data/migrations/env.py
@@ -1,58 +1,43 @@
 import logging
 import os
 
-from logging.config import fileConfig
-from functools import partial
 from urllib.parse import unquote
 
 from alembic import context, op as alembic_op
 from alembic.script.revision import ResolutionError
 from alembic.util import CommandError
-from sqlalchemy import engine_from_config, pool
 from peewee import SqliteDatabase
+from sqlalchemy import create_engine
 
+from app import app
 from data.database import all_models, db, LEGACY_INDEX_MAP
 from data.migrations.tester import NoopTester, PopulateTestDataTester
 from data.model.sqlalchemybridge import gen_sqlalchemy_metadata
 from release import GIT_HEAD, REGION, SERVICE
 from util.morecollections import AttrDict
-from util.parsing import truthy_bool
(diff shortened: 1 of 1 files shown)


data/migrations/env.py
```diff
@@ -128,48 +97,20 @@ def run_migrations_online():
...
-
-    revision_to_migration = {}
-
-    def process_revision_directives(context, revision, directives):
-        script = directives[0]
-        migration = Migration()
-        revision_to_migration[(script.rev_id, revision)] = migration
-        migration.add_hints_from_ops(script.upgrade_ops)
-
+    engine = get_engine()
```
kleesc, line 104: Will this know where to look for migrations? We were previously getting it from the `ini` file (`data/migrations`).

kurtismullins: Yeah. It still looks in the right place to apply migrations and also uses the correct models (metadata) when generating new migrations.

thomasmckay (approved):
LGTM

PR #597, opened 2020-11-11
Removes the http scheme from the server address

kleesc:
-  The http scheme is specified as a dial option on the grpc Go client, not as part of the address.


kurtismullins (approved):
LGTM

alecmerdler (approved):

PR #599, opened 2020-11-17
PROJQUAY-1292 - do not scan manifest lists

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1292

**Changelog:** 
none

**Docs:** 
none

**Testing:** 
Push manifest list and see it is not scanned (no errors in securityworker logs)

**Details:** 
securityworker was erroring from assert, causing images not to be scanned. Error was result of a manifestlist making it to scanner. Both situations corrected.

### DIFF
--- a/data/registry_model/registry_oci_model.py
+++ b/data/registry_model/registry_oci_model.py
@@ -511,9 +511,15 @@ class OCIModel(RegistryDataInterface):
             logger.exception("Could not parse and validate manifest `%s`", manifest._db_id)
             return None
 
-        return self._list_manifest_layers(
-            manifest_obj.repository_id, parsed, storage, include_placements
-        )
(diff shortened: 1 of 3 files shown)


image/docker/schema2/test/test_list.py
```diff
@@ -19,6 +19,9 @@
 from image.docker.test.test_schema1 import MANIFEST_BYTES as v21_bytes
 from util.bytes import Bytes
 
+from app import storage
```
alecmerdler, line 22: Are these imports unused?

thomasmckay: corrected, thanks!

alecmerdler (approved):

PR #600, opened 2020-11-18
PROJQUAY-1258 - correct ldap python3 call

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1258

**Changelog:** 
none

**Docs:** 
none

**Testing:** 
Able to login w/ ldap

**Details:** 
Changes in python3 ldap

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -53,7 +53,6 @@ class LDAPConnection(object):
         allow_tls_fallback=False,
         timeout=None,
         network_timeout=None,
-        bytes_mode=None,
     ):
         self._ldap_uri = ldap_uri
         self._user_dn = user_dn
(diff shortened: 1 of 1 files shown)


data/users/externalldap.py
```diff
@@ -53,26 +53,19 @@ def __init__(
...
-        self._bytes_mode = bytes_mode or False
         self._conn = None
 
     def __enter__(self):
         trace_level = 2 if os.environ.get("USERS_DEBUG") == "1" else 0
 
-        self._conn = ldap.initialize(
-            self._ldap_uri, trace_level=trace_level, bytes_mode=self._bytes_mode
-        )
+        self._conn = ldap.initialize(self._ldap_uri, trace_level=trace_level)
```
kleesc, line 68: Does `bytes_mode` affect the cert issue? If not, it might be better to leave that option there, given that it will default to True otherwise. 
> If not given explicitly, python-ldap will default to bytes_mode=True, but if a unicode value is supplied to it, it will warn and use that value.

data/users/externalldap.py
```diff
@@ -83,6 +76,9 @@ def __enter__(self):
             logger.debug("TLS Fallback enabled in LDAP")
             self._conn.set_option(ldap.OPT_X_TLS_TRY, 1)
 
+        # Must come _after_ all other TLS options
+        self._conn.set_option(ldap.OPT_X_TLS_NEWCTX, ldap.OPT_OFF)
```
kleesc, line 80: What's the difference between `OPT_X_TLS_TRY` and `OPT_X_TLS_NEWCTX`, and what does `OPT_OFF` do?

kleesc (changes requested):

data/users/externalldap.py
kleesc: I guess it's not going to have an effect in Py3, given that all str are unicode.

thomasmckay: bytes_mode cannot be True in python3. Just legacy cruft.

data/users/externalldap.py
thomasmckay: Try is the "fallback to non-tls".

newctx explained https://github.com/python-ldap/python-ldap/issues/55

kleesc (approved):
LGTM

PR #602, opened 2020-11-20
Do not leave 0-byte multipart files in Cloud Storage

kurtismullins:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1123

**Changelog:** fix: Do not leave orphaned 0-byte multi-part uploads in Cloud Storage

**Docs:** n/a

**Testing:**
- Push images. Ensure there are no objects still sitting in an "incomplete" state after the image push is finalized successfully. There is a separate case where objects with > 0 bytes are left in the storage. This does not fix that issue.

**Details:**
- Prior to this Pull Request, zero-byte objects are created and left in a multi-part pending state. Due to a conditional statement in the code, those are never cleaned up. This Pull Request takes the approach of simply cleaning up those objects immediately. Eventually, the Cloud Storage logic should be refactored or fixed to never write these abandoned files in the first place.

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -396,6 +396,13 @@ class _CloudStorage(BaseStorageV2):
     def stream_upload_chunk(self, uuid, offset, length, in_fp, storage_metadata, content_type=None):
         self._initialize_cloud_conn()
 
+        # Do not create zero-byte objects.
+        # TODO: length=-1 appears to be a special case and used in tests. Why? Document it.
+        if length == 0:
+            logger.debug("Attempted to upload a zero-byte chunk to cloud storage. Ignoring.")
+            new_metadata = copy.deepcopy(storage_metadata)
+            return 0, new_metadata, None
+
         # We are going to upload each chunk to a separate key
         chunk_path = self._rel_upload_path(str(uuid4()))
         bytes_written, write_error = self._stream_write_internal(
@@ -407,6 +414,12 @@ class _CloudStorage(BaseStorageV2):
         # We are only going to track keys to which data was confirmed written
         if bytes_written > 0:
             new_metadata[_CHUNKS_KEY].append(_PartUploadMetadata(chunk_path, offset, bytes_written))
+
+        # Do not leave zero-byte chunks in Object Storage
+        # TODO: Determine a clean way to eliminate calling this function when it's not needed.
+        else:
+            logger.debug("Wrote zero bytes to Object Storage. Removing orphaned object.")
+            self.remove(chunk_path)
 
         return bytes_written, new_metadata, write_error
 
--- a/storage/test/test_cloud_storage.py
+++ b/storage/test/test_cloud_storage.py
@@ -1,6 +1,7 @@
 import os
 
 from io import BytesIO
+from mock import patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


thomasmckay:
When I tried reproducing this, I would start `podman push` then ctrl-c it part way through. This left non-zero sized files in the `/uploads` dir. When are those cleaned up?

kurtismullins:
@thomasmckay 

> When I tried reproducing this, I would start `podman push` then ctrl-c it part way through. This left non-zero sized files in the `/uploads` dir. When are those cleaned up?

That's a good question. I'm not sure off-hand.

This Pull Request addresess the issue where multi-part uploads are initiated but never marked as completed. In Noobaa/RHOCS, this will be shown as "incomplete" uploads. If Quay is creating those objects in the `/uploads` directory and leaving them as "in-progress" multi-part objects, then this should resolve that.

If the files you're talking about are finalized, then that requires a different mechanism to go through and "garbage collect" the uploads folder. I'm not sure if anything like that exists in Quay. I would consider that a separate issue, if not.

storage/cloud.py
```diff
@@ -396,6 +396,13 @@ def initiate_chunked_upload(self):
     def stream_upload_chunk(self, uuid, offset, length, in_fp, storage_metadata, content_type=None):
         self._initialize_cloud_conn()
 
+        # Do not create zero-byte objects.
+        # TODO: length=-1 appears to be a special case and used in tests. Why? Document it.
```
kleesc, line 400: Length `-1` means to write the whole `in_fp` stream, vs just `length` bytes if set.

storage/cloud.py
```diff
@@ -408,6 +415,12 @@ def stream_upload_chunk(self, uuid, offset, length, in_fp, storage_metadata, con
         if bytes_written > 0:
             new_metadata[_CHUNKS_KEY].append(_PartUploadMetadata(chunk_path, offset, bytes_written))
 
+        # Do not leave zero-byte chunks in Object Storage
+        # TODO: Determine a clean way to eliminate calling this function when it's not needed.
+        else:
```
kleesc, line 420: Woudn't the above check for zero-length prevent these zero bytes objects to be written in the first place. i.e Don't need to remove them anymore.

storage/test/test_cloud_storage.py
```diff
@@ -275,3 +276,19 @@ def test_rechunked(max_size, parts):
...
+    """
+    with patch.object(storage_engine, "remove") as mock_remove:
+
+        upload_id, metadata = storage_engine.initiate_chunked_upload()
+        data = b""
+        bytes_written, new_metadata, errors = storage_engine.stream_upload_chunk(
+            upload_id, 0, -1, BytesIO(data), metadata
+        )
+        assert len(data) == bytes_written
+        assert mock_remove.call_count == 1
```
kleesc, line 293: Likewise, the shouldn't need to remove anymore, since it's not being written?

kleesc:
@kurtismullins Is this just an issue with Nooba. What is causing the zero-bytes to be written in the first place?

storage/cloud.py
kurtismullins: Good to know. Thank you!

storage/cloud.py
kurtismullins: I don't believe so. In the case where it's explicitly defined as 0 bytes when this file is called, it is short-circuited.

In other cases, I believe that the -1 you explained earlier leads to the creation and writing of a 0 byte file. I'm not exactly sure if that's how they get created, but that initial short-circuit did not fix it for all cases during my personal testing.

kurtismullins:
> @kurtismullins Is this just an issue with Nooba. What is causing the zero-bytes to be written in the first place?

@kleesc IIRC, Ivan has witnessed this on AWS S3. I can also see it in a GCP Object Storage bucket I'm using for a long-running Quay deployment.

kleesc:
Resolved with #705

PR #603, opened 2020-11-23
PROJQUAY-1255 grpc nginx ssl

kleesc:

**Changelog:** 

**Docs:** 

**Testing:** 

**Details:**

### DIFF
--- a/buildman/builder.py
+++ b/buildman/builder.py
@@ -5,13 +5,19 @@ import socket
 
 import features
 
-from app import app, userfiles as user_files, build_logs, dockerfile_build_queue, instance_keys
+from app import (
+    app,
+    userfiles as user_files,
+    build_logs,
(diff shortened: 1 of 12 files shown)


kurtismullins (dismissed):
LGTM

buildman/builder.py
```diff
@@ -89,7 +87,7 @@ def run_build_manager():
         build_manager_config[1],
         instance_keys,
     )
-    server.run("0.0.0.0", controller_port, ssl=ssl_context)
```
kurtismullins, line 92: Why is this no longer required?

kleesc: Wasn't used. Probably from the old code. Instead, the ca is specified as a config option, or set to use the `ssl.cert` otherwise.

kurtismullins (approved):
LGTM

PR #605, opened 2020-11-27
PROJQUAY-1255 - Json + jinja syntax fix in the ignition cloud config template

kleesc:

**Changelog:** 
- Missing key / syntax fixes in ignition cloud config template introduced in https://github.com/quay/quay/pull/603.

**Docs:** 

**Testing:** 

**Details:**

### DIFF
--- a/buildman/manager/executor.py
+++ b/buildman/manager/executor.py
@@ -25,6 +25,7 @@ import release
 from _init import ROOT_DIR, OVERRIDE_CONFIG_DIRECTORY
 from app import app
 from buildman.container_cloud_config import CloudConfigContext
+from buildman.server import DEFAULT_GRPC_SERVER_PORT
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


buildman/manager/executor.py
```diff
@@ -170,7 +171,7 @@ def generate_cloud_config(
                 build_uuid=build_uuid,
                 quay_username=quay_username,
                 quay_password=quay_password,
-                manager_hostname=manager_hostname,
+                manager_hostname=manager_hostname.split(":", 1)[0] + ":" + str(DEFAULT_GRPC_SERVER_PORT),
```
kurtismullins, line 174: Can you define this logic outside of the function call arguments and either wrap it in a try/except statement or throw an assertion on it?

kurtismullins (changes requested):
@kleesc I have one request but otherwise it looks good.

buildman/manager/executor.py
kleesc: Moved it outside the function call. It doesn't need a try block since `split` just returns the original string if couldn't find the delimiter.

kurtismullins (approved):
LGTM

buildman/manager/executor.py
kurtismullins: Thank you!

PR #608, opened 2020-12-01
PROJQUAY-1128 - Allow basic auth on the secscan api endpoint when anonymous api resource is set.

kleesc:
Allow the auth flow to use basic auth on the secscan api resource when
FEATURE_ANONYMOUS_ACCESS (e.g CSO uses pull secrets to fetch secscan
resources)

This will still not allow anonymous acces to the secscan api resource (no auth)

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1128

**Changelog:** 

**Docs:** 

**Testing:** 
- set `FEATURE_ANONYMOUS_ACCESS` to `false`
- Try querying the secscan api with basic auth -> should succeed
- Try querying the secscan api with auth token -> should succeed
- Try querying the secscan api with no auth -> should fail with 401

- set `FEATURE_ANONYMOUS_ACCESS` to `True`
- Try querying the secscan api with basic auth -> should succeed
- Try querying the secscan api with auth token -> should succeed
- Try querying the secscan api with no auth -> should fail with 401

Examples: 
- Basic auth: `curl -k --user <user>:<password> https://quayhost.io/api/v1/repository/<namespace>/<repo>/manifest/<digest>/security`
- Bearer auth: `curl -k -H 'Authorization: Bearer <TOKEN> https://quayhost.io/api/v1/repository/<namespace>/<repo>/manifest/<digest>/security`

**Details:**

### DIFF
--- a/endpoints/api/secscan.py
+++ b/endpoints/api/secscan.py
@@ -24,6 +24,7 @@ from endpoints.api import (
     disallow_for_app_repositories,
     deprecated,
 )
+from endpoints.decorators import anon_allowed
 from endpoints.exception import NotFound, DownstreamIssue
 from endpoints.api.manifest import MANIFEST_DIGEST_ROUTE
 from util.parsing import truthy_bool
@@ -86,6 +87,7 @@ class RepositoryImageSecurity(RepositoryParamResource):
     """
 
     @process_basic_auth_no_pass
+    @anon_allowed
     @require_repo_read
     @nickname("getRepoImageSecurity")
     @deprecated()
@@ -119,6 +121,7 @@ class RepositoryManifestSecurity(RepositoryParamResource):
     """
 
     @process_basic_auth_no_pass
+    @anon_allowed
     @require_repo_read
     @nickname("getRepoManifestSecurity")
     @disallow_for_app_repositories
--- a/endpoints/api/test/test_secscan.py
+++ b/endpoints/api/test/test_secscan.py
@@ -1,6 +1,7 @@
 import base64
 
 import pytest
+from mock import patch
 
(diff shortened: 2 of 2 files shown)


kleesc:
@kurtismullins @thomasmckay ping

kurtismullins (approved):
LGTM

kurtismullins:
@kleesc Re-reading your testing instructions, it looks like that configuration value would have no affect on authentication in either case. 

It looks like in both cases (true/false) the three steps should always respond the same (success, success, fail).

Am I mis-reading that? Or is that expected?

kleesc:
@kurtismullins Correct. This is different from what the current behaviour is, where: 
- querying the secscan api with basic auth -> fails when `FEATURE_ANONYMOUS_ACCESS` is `false`

kurtismullins:
@kleesc 

> @kurtismullins Correct. This is different from what the current behaviour is, where:
> 
> * querying the secscan api with basic auth -> fails when `FEATURE_ANONYMOUS_ACCESS` is `false`

Ahh, that makes sense.

I was unable to produce a valid (basic) Authorization Header on my end. Oauth worked, however. Have you had any luck?

kurtismullins (dismissed):
@kleesc 

Can you provide the instructions you used to perform a Basic Authentication call to test these endpoints? Preferably with cURL or similar. 

It will be helpful for anyone verifying this Pull Request and also for QE when they later test it.

Thank you!

kleesc:
@kurtismullins Added `cURL` example in the PR description. You also need security scanning enabled to have these endpoints available.

thomasmckay (approved):
LGTM

PR #612, opened 2020-12-07
Add metrics to app instance's peewee connection pool

kleesc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1358

Add some metrics to Peewee:
- Connections in `PooledDatabase`s
- `connect`/`close` calls counts

Example: Screen Shot 2020-12-09 at 10 45 15 AM

**Changelog:** 

**Docs:** 

**Testing:**

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -18,7 +18,12 @@ import toposort
 from enum import IntEnum, Enum, unique
 from peewee import *
 from peewee import __exception_wrapper__, Function
-from playhouse.pool import PooledMySQLDatabase, PooledPostgresqlDatabase, PooledSqliteDatabase
+from playhouse.pool import (
+    PooledDatabase,
+    PooledMySQLDatabase,
+    PooledPostgresqlDatabase,
+    PooledSqliteDatabase,
+)
 
 from sqlalchemy.engine.url import make_url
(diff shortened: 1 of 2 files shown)


kleesc:
@kurtismullins ping

kurtismullins:
@kleesc Is the idea you've implemented here to update the connection count gauge on every method call of the connection pools?

kleesc:
@kurtismullins Yes

kurtismullins:
@kleesc If that's the case, then I believe the only observable connection count (by default) will be the registry. I don't believe any other connections, including those used by the workers, are pooled.

Do we want to include those as well?

kleesc:
@kurtismullins Can you explain why only the registry worker exposing metrics?
Each process has their own copy of `db` (separate pool). The metrics are labeled by `pid`, `process name` and `host`. And they should each publish the pool count the prom using these labels whenever the  `PooledDatabase` is accessed.

kurtismullins (approved):
LGTM. Let's follow standard procedures before pushing this to production just in case we encounter any unexpected side effects.

thomasmckay (approved):
LGTM

PR #619, opened 2020-12-10
PROJQUAY-1376 - Handle non 200 api response from executors

kleesc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1376

Allow the build manager to requeue (if allowed) if the requests to the executors' api return a non-200 response.
 
**Changelog:** 

**Docs:** 

**Testing:**

### DIFF
--- a/buildman/manager/ephemeral.py
+++ b/buildman/manager/ephemeral.py
@@ -583,13 +583,17 @@ class EphemeralBuilderManager(BuildStateInterface):
         logger.debug("Scheduling build %s", build_id)
 
         allowed_worker_count = self._manager_config.get("ALLOWED_WORKER_COUNT", 1)
-        if self._running_workers() >= allowed_worker_count:
-            logger.warning(
-                "Could not schedule build %s. Number of workers at capacity: %s.",
-                build_id,
-                self._running_workers(),
-            )
+        try:
+            if self._running_workers() >= allowed_worker_count:
+                logger.warning(
+                    "Could not schedule build %s. Number of workers at capacity: %s.",
+                    build_id,
+                    self._running_workers(),
+                )
             return False, TOO_MANY_WORKERS_SLEEP_DURATION
+        except Exception as exe:
+            logger.warning("Failed to get worker count from executors: %s", exe)
+            return False, EPHEMERAL_API_TIMEOUT
 
         job_id = self._job_key(build_id)
         try:
--- a/buildman/manager/executor.py
+++ b/buildman/manager/executor.py
@@ -429,6 +429,9 @@ class KubernetesExecutor(BuilderExecutor):
         tls_ca = self.executor_config.get("K8S_API_TLS_CA")
         service_account_token = self.executor_config.get("SERVICE_ACCOUNT_TOKEN")
 
+        if tls_ca:
+            request_options["verify"] = tls_ca
+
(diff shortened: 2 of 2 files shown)


buildman/manager/executor.py
```diff
@@ -402,6 +406,19 @@ def __init__(self, *args, **kwargs):
     def running_builders_count(self):
         q = {"labelSelector": "build,time,manager,quay-sha"}
         jobs_list = self._request("GET", self._jobs_path(), params=q)
+        if jobs_list.status_code != 200:
```
kurtismullins, line 409: (nitpick) It might more useful to future developers if this was renamed from `jobs_list` to `req` and then use the variable name `jobs_list` to reference the actual list of jobs.

buildman/manager/executor.py
```diff
@@ -215,10 +215,14 @@ def _get_conn(self):
 
     @property
     def running_builders_count(self):
-        ec2_conn = self._get_conn()
-        resp = ec2_conn.describe_instances(
-            Filters=[{"Name": "tag:Name", "Values": ["Quay Ephemeral Builder"]}]
-        )
+        try:
```
kurtismullins, line 218: Any reason you decided to put this try/except statement at this level, instead of within the `_get_conn()` method?

buildman/manager/ephemeral.py
```diff
@@ -583,13 +583,17 @@ def schedule(self, build_id):
         logger.debug("Scheduling build %s", build_id)
 
         allowed_worker_count = self._manager_config.get("ALLOWED_WORKER_COUNT", 1)
-        if self._running_workers() >= allowed_worker_count:
-            logger.warning(
-                "Could not schedule build %s. Number of workers at capacity: %s.",
-                build_id,
-                self._running_workers(),
-            )
+        try:
```
kurtismullins, line 586: (Same as below) Why did you put this try/except block at this level instead of within the `_running_workers()` method?

kurtismullins (dismissed):
@kleesc I left two questions and made a nitpick comment but overall, it looks good to me. Just ping me if you decide to make any changes and I will be happy to re-review.

buildman/manager/executor.py
kurtismullins: I did overlook that this also catches any error which might occur when hitting the API to fetch the instances. That make sense to be covered here.

kleesc: Doesn't really make sense to try/catch this in `_get_conn`, since we're trying to catch client errors on the api call. `_get_conn` would not raise any exception.

buildman/manager/ephemeral.py
kleesc: What would `_running_workers()` raise/return? The caller should be the one handling the exception.

kleesc: `_running_workers()` will just raise whatever the executor raises, and it's up to the caller to handle it.

kurtismullins: Ahh, I see. You ultimately want to return False and the exception here; not let the exception bubble up. That makes sense.

buildman/manager/executor.py
kurtismullins: :+1:

kurtismullins (approved):
LGTM

PR #631, opened 2020-12-18
PROJQUAY-1397 - Add a DEBUG flag to the executors to prevent cleanup

kleesc:
When set to true, DEBUG will prevent the build nodes from shutting
down after the quay-builder service is done or fails, and will prevent the
build manager from cleaning up the instances (terminating EC2
instances or deleting k8s jobs).

This will allow debugging builder node issues, and should not be set
in a production environment.

The lifetime service will still exist. i.e The instance will still
shutdown after ~2h (EC2 instances will terminate, k8s jobs will
complete)

Setting DEBUG will also affect ALLOWED_WORKER_COUNT, as the
unterminated instances/jobs will still count towards the total number
of running workers.

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1397
**Changelog:** 

**Docs:** 

**Testing:** 
- Set DEBUG to True in the executor(s)' config
- Start a build from Quay
- After a build completes or fails, the worker nodes should still be running in the executors.
- 
**Details:**

### DIFF
--- a/buildman/manager/executor.py
+++ b/buildman/manager/executor.py
@@ -62,6 +62,21 @@ def observe(metric, *labels):
     return decorator
 
 
+def persist_for_debugging(func):
+    """
+    Wrapper for stop_builder that prevents the workers from being cleaned up (for testing purposes only)
+    """
+
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


kurtismullins (approved):
LGTM

It might be helpful to add this to the documentation.

PR #635, opened 2021-01-04
Add a --shrink flag when resizing the qemu image.

kleesc:
Newer version of qemu will return a non-zero error when --shrink is not set when resizing an image.

**Changelog:** 
- Add `--shrink` to the qemu-img resize command

**Docs:** 

**Testing:** 

**Details:**


thomasmckay (approved):
LGTM

PR #642, opened 2021-01-07
PROJQUAY-1392 prevent over shrinking vm image

kleesc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1392

**Changelog:** 
- Prevent shrinking builder vm image, as that might truncate existing partitions. If the configured volume size is smaller than the default vdisk size, the image will not be resized.
 
**Docs:** 

**Testing:** 

**Details:**


thomasmckay (approved):
LGTM

PR #657, opened 2021-01-14
Fix Default Value for Build Manager Hostname

alecmerdler:
Fixes exception raised for referencing variable before assignment.


kleesc (approved):
LGTM

PR #660, opened 2021-01-20
Remove extra extra parameter when calling commit_sha

kleesc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1441

**Changelog:** 
- Remove extra param not needed in `commit_sha` call when a private key is used during a build

**Docs:** 

**Testing:** 

**Details:**


thomasmckay (approved):
LGTM

PR #667, opened 2021-01-25
Set default REPO_MIRROR_SERVER_HOSTNAME value to match SERVER_HOSTNAME

kleesc:
Set `REPO_MIRROR_SERVER_HOSTNAME` to a value instead of None in order to
match the defined json schema for the config.

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1499

**Changelog:** 
- Set default `REPO_MIRROR_SERVER_HOSTNAME` value to `SERVER_HOSTNAME` instaed of `None`

**Docs:** 

**Testing:** 

**Details:**

PR #673, opened 2021-02-10
PROJQUAY-1562 repo mirror config schema

kleesc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1562
Revert #667 . It wrongly assumed that `REPO_MIRROR_SERVER_HOSTNAME` shouldn't be null given the previous config schema. `REPO_MIRROR_SERVER_HOSTNAME` _can_ be null and should have the same value as `SERVER_HOSTNAME` if not explicitly set.

**Changelog:** 
- Reverts #667 
- Allow `REPO_MIRROR_SERVER_HOSTNAME` to be null in the jsonschema

**Docs:** 

**Testing:** 

**Details:**

### DIFF
--- a/config.py
+++ b/config.py
@@ -503,7 +503,7 @@ class DefaultConfig(ImmutableConfig):
     REPO_MIRROR_TLS_VERIFY = True
 
     # Replaces the SERVER_HOSTNAME as the destination for mirroring.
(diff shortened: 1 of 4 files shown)


thomasmckay (approved):
LGTM

PR #675, opened 2021-02-12
Fix Swift assertion for chunked upload

kleesc:
- Allow any binary stream (IOBase, including Werkzeug's LimitedStream) to Swift's type assertion
- Allow LimitingStream from util.registry.filelike to seek backward, since it is required by the Swift client in order to retry operations, if it is configured to do so

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1508

**Changelog:** 
- Update use of `_pyio` to `io` (`io` is implemented in C instead of pure Python) in Swift's implementation
- Fix Swift storage allowed stream types
- Fix Swift storage connection retries

**Docs:** 

**Testing:** 

**Details:**

### DIFF
--- a/util/registry/filelike.py
+++ b/util/registry/filelike.py
@@ -23,7 +23,12 @@ class BaseStreamFilelike(object):
     def tell(self):
         return self._cursor_position
 
-    def seek(self, index, whence=WHENCE_ABSOLUTE):
+    def seek(self, index, whence=WHENCE_ABSOLUTE, allow_backward=False):
+        if allow_backward and index < self._current_position:
+            if not self._fileobj.seekable():
+                raise IOError("Cannot seek backward on stream: fileobj is not seekable")
+            self._cursor_position = self._fileobj.seek(index, whence)
+
         num_bytes_to_ff = 0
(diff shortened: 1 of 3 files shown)


util/registry/filelike.py
```diff
@@ -23,7 +23,12 @@ def read(self, size=READ_UNTIL_END):
     def tell(self):
         return self._cursor_position
 
-    def seek(self, index, whence=WHENCE_ABSOLUTE):
+    def seek(self, index, whence=WHENCE_ABSOLUTE, allow_backward=False):
+        if allow_backward and index < self._current_position:
+            if not self._fileobj.seekable():
+                raise IOError("CAnnot seek backward on stream: fileobj is not seekable")
```
alecmerdler, line 29: Typo

kleesc: Fixed

alecmerdler (dismissed):

kleesc:
@alecmerdler Can you re-approve? I rebased on master after merging #676 that was blocking CI build step.

alecmerdler (approved):

PR #678, opened 2021-02-17
PROJQUAY-1284

jonathankingfc:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1284

**Changelog:** 
- Added `FEATURE_FIPS` flag to Quay config. When this feature is enabled, Quay will use sha256 hash instead of md5. This allows future FIPS users to utilize compliant hashes without breaking existing deployments (non-FIPS). 

**Docs:** 

**Testing:** 
- Tested TNG deployment on FIPS-enabled OCP. The issue is reproduced without `FEATURE_FIPS` enabled (as expected) but works correctly when `FEATURE_FIPS=true`.

**Details:** 

------

### DIFF
--- a/avatars/avatars.py
+++ b/avatars/avatars.py
@@ -1,6 +1,7 @@
 import hashlib
 import math
 import logging
+import features
 
 from requests.exceptions import RequestException
 
@@ -107,7 +108,16 @@ class BaseAvatar(object):
         # so use the username in that case.
         username_email_or_id = email_or_id or name
         username_email_or_id = Bytes.for_string_or_unicode(username_email_or_id).as_unicode()
-        hash_value = hashlib.md5(username_email_or_id.strip().lower().encode("utf-8")).hexdigest()
+
+        # If fips mode is enabled, we must use 256. In future Quay versions, this should be the default. Flag is set for backwards compatibility.
+        if features.FIPS:
+            hash_value = hashlib.sha256(
+                username_email_or_id.strip().lower().encode("utf-8")
+            ).hexdigest()
+        else:
+            hash_value = hashlib.md5(
+                username_email_or_id.strip().lower().encode("utf-8")
+            ).hexdigest()
 
         byte_count = int(math.ceil(math.log(len(colors), 16)))
         byte_data = hash_value[0:byte_count]
--- a/config.py
+++ b/config.py
@@ -379,6 +379,9 @@ class DefaultConfig(ImmutableConfig):
     # logging in via OIDC or a non-database internal auth provider.
     FEATURE_USERNAME_CONFIRMATION = True
 
+    # Feature Flag: If set to true, Quay will run using FIPS compliant hash functions.
+    FEATURE_FIPS = False
+
     # If a namespace is defined in the public namespace list, then it will appear on *all*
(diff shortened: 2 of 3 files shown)


avatars/avatars.py
```diff
@@ -30,10 +31,11 @@ class BaseAvatar(object):
     Base class for all avatar implementations.
     """
 
-    def __init__(self, preferred_url_scheme, colors, http_client):
+    def __init__(self, preferred_url_scheme, colors, http_client, feature_fips):
```
kleesc, line 34: Can you set `feature_fips` default to false?

config.py
```diff
@@ -379,6 +379,9 @@ def create_transaction(db):
     # logging in via OIDC or a non-database internal auth provider.
     FEATURE_USERNAME_CONFIRMATION = True
 
+    # Feature Flag: If set to true, Quay will run using FIPS compliant hash functions.
+    FEATURE_FIPS = False
```
kleesc, line 383: Is the plan to be eventually fully FIPS compliant? It looks like the crypto ciphers are only covered by FIPS 140.

kleesc (changes requested):
@jonathankingfc Can you also take a look in `storage/` for the different storage implementations. I know some of them uses md5 for content checksum, so I'm not sure if that's FIPS compliant or not.

avatars/avatars.py
kleesc: Actually, you don't really need this. Since your config flag starts with `FEATURE_`, you could just `import features` and check with:
```
import features
...

features.fips
```

jonathankingfc: Just updated PR to use features module instead of manually passing the value.

avatars/avatars.py
```diff
@@ -107,7 +108,16 @@ def get_data(self, name, email_or_id, kind="user"):
...
         username_email_or_id = Bytes.for_string_or_unicode(username_email_or_id).as_unicode()
-        hash_value = hashlib.md5(username_email_or_id.strip().lower().encode("utf-8")).hexdigest()
+
+        # If fips mode is enabled, we must use 256. In future Quay versions, this should be the default. Flag is set for backwards compatibility.
+        if features.FIPS:
+            hash_value = hashlib.sha256(
+                username_email_or_id.strip().lower().encode("utf-8")
+            ).hexdigest()
+        else:
+            hash_value = hashlib.md5(
```
kleesc, line 118: Can probably move this into GravatarAvatar specific implentation, since it's the only one that needs to use md5: https://en.gravatar.com/site/implement/hash/

jkeam: I'm a little worried about the size of that refactor, as `get_data` is used by a few publicly used methods (`get_data_for_external_user`, `get_data_for_org`, `get_data_for_team`, `get_data_for_user`).  

Perhaps we could accept this PR to solve the ticket related to FIPS and open a feature request to think about refactoring `GravatarAvatar` at a later time?

Also I'd love to help contribute anyway I can.  I'm particularly excited about getting this ticket merged in, so anything I can do to help move this along I'd be more than happy to help!

jkeam:
Just saw this closed, any particular reason why?

PR #679, opened 2021-02-19
Deprecate use of pycryptodome for AES

kleesc:
Since we're already using the `cryptography` package elsewhere, there is no need to have 2 different crypto packages as dependencies.
**NOTE**: `pycryptodome` is still needed for `pyjwkest`, but the latter (https://github.com/IdentityPython/pyjwkest) is not being maintained anymore, so we should eventually move it to some other JOSE library:
- https://github.com/mpdavis/python-jose or 
- https://github.com/latchset/jwcrypto

**Issue:** 
- https://issues.redhat.com/browse/PROJQUAY-1619
- https://issues.redhat.com/browse/PROJQUAY-224

**Changelog:** 
- Use `cryptography` instead of `pycryptodome` for AES

**Docs:** 

**Testing:** 
- Generate an encrypted password with a version of Quay using `cryptography` -> encrypted password should work using `docker login`
- Generate an encrypted password using a version of Quay still using `pycryptodome`.
  - Restart an instance of Quay with this PR included -> The encrypted password should still work.

**Details:**

### DIFF
--- a/data/users/__init__.py
+++ b/data/users/__init__.py
@@ -172,7 +172,7 @@ class UserAuthentication(object):
             return None
 
         try:
-            data = json.loads(message)
+            data = json.loads(message.decode("utf-8"))
         except ValueError:
             return None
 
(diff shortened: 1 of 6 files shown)


thomasmckay (changes requested):
whenever requirements.txt is changed, see if requirements-osbs.txt can be updated as well

kleesc:
@jonathankingfc @thomasmckay

thomasmckay (approved):
LGTM

PR #687, opened 2021-03-08
Use registry-name instead of registry-title as registry directive

syed:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1519
Pull-request title must start with "PROJQUAY-1519 - "

**Changelog:** 
* Use `registry-name` instead of `registry-title` as the class directive


thomasmckay (approved):
LGTM

PR #690, opened 2021-03-12
tags: apply tag expiry to created tags pointing to existing manifest

kleesc:
Apply a manifest's `quay.expires-after` label expiry value to new tags pointing
to that existing manifest. Before, that label would be only applied when the
manifest was created, and new tags targeting that manifest would not
have the manifest's corresponding expiry value set.

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1713

**Changelog:** 
- Fix issue where new tags targeting an existing manifest would not have manifest's `quay.expires-after` label applied

**Docs:** 

**Testing:** 
- Push an image with a valid `quay.expires-after` label/value -> the tag used should have the corresponding expiration
- Create another tag pointing to the same manifest -> the new tag should also have an expiration
  - NOTE: this should work for using both the registry api and tags api to create the tag
- Changing expiration on individual tags should not affect other tags -> new tags are always created with expiration set to `"tag creation time"` + `quay.expires-after`
- Moving a tag should apply the expiration label of the newly targeted manifest

### DIFF
--- a/data/model/oci/tag.py
+++ b/data/model/oci/tag.py
@@ -24,7 +24,6 @@ from data.database import (
     ExternalNotificationEvent,
     db_random_func,
 )
-from data.model.oci.shared import get_legacy_image_for_manifest
 from data.model import config
 from image.docker.schema1 import (
     DOCKER_SCHEMA1_CONTENT_TYPES,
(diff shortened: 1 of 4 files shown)


thomasmckay (approved):
LGTM

thomasmckay (approved):
LGTM

thomasmckay (approved):
LGTM

alecmerdler (approved):

PR #693, opened 2021-03-16
PROJQUAY-1409 - remove nodejs from final container

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1409

**Changelog:** 
none

**Docs:** 
none

**Testing:** 
none

**Details:** 
nodejs not needed in final container build stage

### DIFF
--- a/Dockerfile.osbs
+++ b/Dockerfile.osbs
@@ -1,4 +1,4 @@
-FROM registry.redhat.io/ubi8:8.2 AS build-npm
+FROM registry.redhat.io/ubi8:latest AS build-npm
 
 COPY $REMOTE_SOURCE $REMOTE_SOURCE_DIR
 WORKDIR $REMOTE_SOURCE_DIR/app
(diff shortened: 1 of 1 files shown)


Dockerfile.osbs
```diff
@@ -86,7 +86,6 @@ RUN INSTALL_PKGS="\
         memcached \
         openssl \
         skopeo \
-        nodejs \
```
alecmerdler, line 89: It looks like we still need `npm` (part of NodeJS) to complete the step where we download static assets for the frontend.

thomasmckay: I think it's alright to remove it from Dockerfile.osbs but I'll add it back to our regular Dockerfile and defer removing there until the restructuring PR goes in.

alecmerdler: As long as we don't run any `npm` commands later on in that container build (which it doesn't look like we do), then it should be fine.

alecmerdler (approved):

PR #705, opened 2021-03-24
storage: abort unfinished mpu when no bytes are written

kleesc:
Explicitly call abort on a mpu if no bytes were written.
Noobaa and Rados will not clean the artifacts, resulting in empty
files being stuck in an "uploading" state.

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1123

**Changelog:** 
- Explicitly abort MultipartUpload where no bytes were uploaded

**Docs:** 

**Testing:** 

**Details:**

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -380,8 +380,6 @@ class _CloudStorage(BaseStorageV2):
                     )
                     write_error = e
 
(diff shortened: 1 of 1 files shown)


thomasmckay (approved):
LGTM

PR #709, opened 2021-03-26
format: remove extra comma from s3 connect_kwargs

kleesc:
The way it was formatted (with the top-level parentheses), the extra
comma was causing the connection parameters to be passed as a 1-tuple
instead of a string.

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1787

**Changelog:** 
- Fix formating issue resulting in tuple instead of string

**Docs:** 

**Testing:** 

**Details:**


thomasmckay (approved):
LGTM

PR #712, opened 2021-04-01
gc: fix GlobalLock ttl unit and increase gc workers lock timeout

kleesc:
Correctly converts the given ttl from seconds to milliseconds when
passed to Redis (redlock uses 'px', not 'ex'). Also increase the lock
timeout of gc workers to 1 day.

Some iteration, for repos with large numbers of tags (1000s), will
take more than 15 minutes to complete. This change will prevent multiple
workers GCing the same repo, and one possibly preempting
another. GlobalLock's ttl will make the lock available again when
expired, but will not actually stop execution of the current GC
iteration until the GlobalLock context is done. Having a 1 day timeout
should be enough.

NOTE: The correct solution would have GlobalLock should either renew
the lock until the caller is done, or signal that it is no longer
valid to the caller.

**Issue:** https://issues.redhat.com/browse/PROJQUAY-1823

### DIFF
--- a/util/locking.py
+++ b/util/locking.py
@@ -14,6 +14,10 @@ class LockNotAcquiredException(Exception):
     """
 
 
+# TODO(kleesc): GlobalLock should either renew the lock until the caller is done,
+# or signal that it is no longer valid to the caller. Currently, GlobalLock will
+# just silently expire the redis key, making the lock available again while any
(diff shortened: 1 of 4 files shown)


alecmerdler (approved):

thomasmckay (approved):
LGTM

PR #713, opened 2021-04-06
PROJQUAY-1803 - specify S3 signature version

thomasmckay:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1803

**Changelog:** 

**Docs:** 

**Testing:** 
- test push/pull on s3 us-east-2, us-east-1
- test push/pull on other cloud storage

**Details:** 
us-east-2 does not support signature v2

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -380,6 +380,8 @@ class _CloudStorage(BaseStorageV2):
                     )
                     write_error = e
 
+                    multipart_uploads_completed.inc()
+
                     if cancel_on_error:
                         try:
                             mp.abort()
@@ -398,8 +400,6 @@ class _CloudStorage(BaseStorageV2):
                     "Parts": [{"ETag": p.e_tag, "PartNumber": p.part_number} for p in upload_parts],
                 },
             )
-        else:
-            mp.abort()
 
         return total_bytes_written, write_error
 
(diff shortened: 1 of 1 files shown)


kleesc:
@thomasmcka Boto3 should already be using sigv4 by default.

storage/cloud.py
```diff
@@ -711,16 +715,23 @@ def __init__(
...
+        client_kwargs = {
+            "config": Config(signature_version='s3v4')
+        }
+        resource_kwargs = {}
         if host or endpoint_url:
-            connect_kwargs["endpoint_url"] = endpoint_url or _build_endpoint_url(
+            client_kwargs["endpoint_url"] = endpoint_url or _build_endpoint_url(
+                host, port=port, is_secure=True
+            )
+            resource_kwargs["endpoint_url"] = endpoint_url or _build_endpoint_url(
```
kleesc, line 726: I don't think you don't need to specify two different params set. 
What's passed to the low-level client should also be passed to the Resource.
So I'd add the config to the existing _connect_params:
`connect_kwargs["config"] = Config(signature_version="s3v4"` instead.

kleesc (changes requested):

storage/cloud.py
thomasmckay: updated, thanks

PR #733, opened 2021-04-09
Quayio branch build fix

kleesc:
Changes so that the quayio branch is buildable:
- Cherrypick 25531104 (Moving some static s3 assets to repo)
- Pin pip to last python2 supported version
- Update front-end yarn build


thomasmckay (approved):
LGTM

PR #735, opened 2021-04-09
Add Timeout and Retry To OAuth Token Exchange (PROJQUAY-1335)

alecmerdler:
**Issue:** https://issues.redhat.com/browse/PROJQUAY-1335

**Changelog:** Add timeout to OAuth token exchange to avoid ECONNRESET from RH SSO.

**Docs:** N/a

**Testing:** `tox`

**Details:** Attempting to log into to quay.io using Red Hat SSO would occasionally result in an `HTTP 404`/`HTTP 502` response from the server. Exception logging revealed this to be the result of a `ECONNRESET` when sending a `POST` request to the SSO service for token exchange. As this is a problem in a service we do not control, our only option is to be more defensive on Quay's side. Adding a timeout here should alleviate this problem.

### DIFF
--- a/oauth/base.py
+++ b/oauth/base.py
@@ -1,5 +1,7 @@
 import copy
 import logging
+import requests
+import time
 import urllib.request, urllib.parse, urllib.error
 import urllib.parse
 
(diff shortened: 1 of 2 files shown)


alecmerdler:
This will need to be cherry-picked into the `quayio` branch.

kleesc (approved):
LGTM

kleesc (approved):
LGTM

kleesc:
@alecmerdler Is this ready to be merged?

alecmerdler:
@kleesc I wasn't able to reproduce the issue we were seeing on quay.io, so this may not be the final fix. But it's worth adding.

PR #737, opened 2021-04-09
queueworker: prevent stop event on WorkerSleepException (PROJQUAY-1857)

kleesc:
Prevents the queueworker from setting the event to stop the poll_queue
job when a WorkerSleepException is raised. On WorkerSleepException,
the worker should instead skip this iteration (go to sleep). e.g when
the NamespaceGCWorker can't acquire a lock because it is already taken
by some other worker.

Reverts the gcworkers job timeout from 24h to 3h. In case of a
deadlock between processes (for example, redeploying the app will not
clear the existing Redis keys), 24h is too long waiting for the locks to
expires so that the workers can resume work.

Add missing Counter increment for on row deletion on the Manifest table.

### DIFF
--- a/data/model/gc.py
+++ b/data/model/gc.py
@@ -586,6 +586,8 @@ def _garbage_collect_manifest(manifest_id, context):
     if deleted_manifest_legacy_image:
         gc_table_rows_deleted.labels(table="ManifestLegacyImage").inc(deleted_manifest_legacy_image)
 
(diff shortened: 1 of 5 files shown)


alecmerdler (approved):

kleesc:
@alecmerdler Will need reapproval

alecmerdler (approved):

PR #741, opened 2021-04-13
lock: reuse redis client when creating locks (PROJQUAY-1872)

kleesc:
By default, Redlock creates a new client per instance. Using the
provided factory allows Redlock to reuse a single connection per
instance and avoid running out of connections. e.g When a worker tries
to get a lock, it should not open new connections every time.

Increase sleep duration between queue polls on
WorkerSleepException. This will give more time before retrying after
failing to acquire a lock.

### DIFF
--- a/util/locking.py
+++ b/util/locking.py
@@ -1,7 +1,7 @@
 import logging
 
 from redis import RedisError
-from redlock import RedLock, RedLockError
+from redlock import RedLockFactory, RedLockError
 
 from app import app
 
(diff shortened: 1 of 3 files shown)


util/locking.py
```diff
@@ -26,13 +26,26 @@ class GlobalLock(object):
...
     """
 
+    lock_factory = None
+
+    def __new__(cls):
+        if cls.lock_factory is None:
+            cls.lock_factory = RedLockFactory(
+                connection_details=[
+                    {
+                        **dict(app.config["USER_EVENTS_REDIS"]),
```
alecmerdler, line 36: Is this only for the Redis instance being used for user events? Technically there can be separate Redis instances for user events and the build manager.

kleesc: Yes, `USER_EVENTS_REDIS` was the one used by default prior to this change. Do we want to make this configurable? Especially if we intend to use this elsewhere (#444)

alecmerdler: We can add this later.

alecmerdler (approved):

PR #746, opened 2021-04-14
PROJQUAY-1880: Adding index on uploading column in the imagestore table

syed:
This follows from the DB testing on the AWS dev cluster. We found adding this index increases the filtering of imagestore x manifestblob joins from 10% to 50%

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -1072,7 +1072,7 @@ class ImageStorage(BaseModel):
     uuid = CharField(default=uuid_generator, index=True, unique=True)
     image_size = BigIntegerField(null=True)
     uncompressed_size = BigIntegerField(null=True)
-    uploading = BooleanField(default=True, null=True)
+    uploading = BooleanField(default=True, index=True, null=True)
     cas_path = BooleanField(default=True)
     content_checksum = CharField(null=True, index=True)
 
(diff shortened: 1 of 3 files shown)


kleesc:
NOTE: Even though the `uploading` column has a low cardinality (bool), we believe that adding an index can still help with queries since one of the two possible values is selective enough (`True` vs `False`). i.e Only a small set of the entire column set is ever set to True.

josephschorr:
@kleesc Is `uploading` even used anymore? It was originally used because we created the `ImageStorage` and `Image` rows as part of the V1 upload process... but once we moved to a V2-centric model, everything is now blob based and the system only (as far as I recall) creates the `ImageStorage` rows once the upload has completed, always with `uploading=False`: https://github.com/quay/quay/blob/f774e4c6b6c674c822057a1d3dd49a3d5c36e6ca/data/model/blob.py#L83

You might be able to get away with dropping the column and checks entirely, if it is never set to `True` anywhere

kleesc:
@josephschorr You're right. AFAICT we don't explicitly query for blobs using this column anymore since #442. Looks like the only places where it could still be set were if it ever needed to upload the empty layer (shouldn't happen) and in tests. I guess the few blobs stuck in an uploading state are just waiting to get GC'd eventually.

Thanks!

kleesc:
Closing as this will be addressed by #764 #767 .

PR #753, opened 2021-04-19
quayio: returns value from delete_user calls

kleesc:
This was already fixed in `master` but not in the quayio branch, and is preventing the namespacegcworker from correctly reporting its metrics.


thomasmckay (approved):
LGTM

PR #757, opened 2021-04-20
modelcache: Increase the cache TTL for blob and tags (PROJQUAY-1878)

syed:
Increase the cache TTL for blob and tags. This follows from our
analyzis of production quay.io memcache stats. We see that the number of
refreshes are significanlty more than the cache hits. This suggests that
our default cache expiry is too low.


kleesc:
I don't mind increasing the blob cache expiration since it's content addressable, but I think we should leave the current tags list as is for now, as this will affect UX.

Also, if we're going to start changing these values, we should keep these as default and make them configurable instead.
These values may work for some, but might not with others, say with different eviction policies.

PR #764, opened 2021-04-21
chore: remove uploading filtering from imagestorage queries

kleesc:
The "uploading" column is an artifact from depending on writing to the
Image table (see BlobUpload table instead). As of 3.4, Quay no longer
writes to that table, and is only needed until quayio moves away from
Clair v2, after which work to remove "glue" code and fully deprecate
the Image table (amongst other tables) can start.

This is done as a separate commit from the actual migration so that it
can be cherrypicked.

### DIFF
--- a/data/model/blob.py
+++ b/data/model/blob.py
@@ -80,7 +80,6 @@ def store_blob_record_and_temp_link_in_repo(
         except ImageStorage.DoesNotExist:
             storage = ImageStorage.create(
                 content_checksum=blob_digest,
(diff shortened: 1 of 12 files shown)


thomasmckay:
@kleesc  how should this be tested?

kleesc:
@thomasmckay This change shouldn't change any functionality. So making sure the standard registry operations still work.

alecmerdler (approved):

PR #765, opened 2021-04-21
modelcache: Add config for cache ttl to keys (PROJQUAY-1878)

syed:
Adds a configuration option to modify the cache expiry timeout for ModelCache objects

### DIFF
--- a/config.py
+++ b/config.py
@@ -653,6 +653,12 @@ class DefaultConfig(ImmutableConfig):
     DATA_MODEL_CACHE_CONFIG = {
         "engine": "memcached",
         "endpoint": ("127.0.0.1", 18080),
+        "repository_blob_cache_ttl": "60s",
+        "catalog_page_cache_ttl": "60s",
+        "namespace_geo_restrictions_cache_ttl": "240s",
+        "active_repo_tags_cache_ttl": "120s",
+        "appr_applications_list_cache_ttl": "3600s",
+        "appr_show_package_cache_ttl": "3600s",
     }
 
     # Defines the number of successive failures of a build trigger's build before the trigger is
--- a/data/cache/cache_key.py
+++ b/data/cache/cache_key.py
@@ -9,48 +9,57 @@ class CacheKey(namedtuple("CacheKey", ["key", "expiration"])):
     pass
 
 
-def for_repository_blob(namespace_name, repo_name, digest, version):
+def for_repository_blob(namespace_name, repo_name, digest, version, cache_config):
     """
     Returns a cache key for a blob in a repository.
     """
-    return CacheKey("repo_blob__%s_%s_%s_%s" % (namespace_name, repo_name, digest, version), "60s")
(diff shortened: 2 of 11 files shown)


kleesc:
@syed Why not make this part of #757 ?

syed:
@kleesc  I've closed that one. Can you please review this one instead

data/cache/cache_key.py
```diff
@@ -9,48 +9,57 @@ class CacheKey(namedtuple("CacheKey", ["key", "expiration"])):
     pass
 
 
-def for_repository_blob(namespace_name, repo_name, digest, version):
+def for_repository_blob(namespace_name, repo_name, digest, version, cache_config):
```
kleesc, line 12: Is it better here to pass the whole config here or have the callers pass the value as an int or valid durations string?

data/cache/test/test_cache.py
```diff
@@ -13,6 +13,15 @@
 
 DATA = {}
 
+CACHE_CONFIG = {
```
kleesc, line 16: Can you also set the previous default values in `config.py`

kleesc: Can you add a `TEST_` prefix so that it's easier to differentiate.
We should also add mock values for other fields that are expected in the config: `engine`, `endpoint`, ...

data/cache/cache_key.py
syed: I think it's better to have TTL. I'll make this change

data/cache/test/test_cache.py
syed: 👍

data/cache/cache_key.py
syed: Right now I just pass `model_cache.cache_config` but if we are to pass TTL the caller should do something like 

`model_cache.cache_config.get('repository_blob_cache_ttl', 'default-value')` Here the caller has to know the key and also what the default value should be which shouldn't really be the caller's concern

config.py
```diff
@@ -653,6 +653,12 @@ def create_transaction(db):
     DATA_MODEL_CACHE_CONFIG = {
         "engine": "memcached",
         "endpoint": ("127.0.0.1", 18080),
+        "repository_blob_cache_ttl": "240s",
+        "catalog_page_cache_ttl": "240s",
+        "namespace_geo_restrictions_cache_ttl": "240s",
+        "active_repo_tags_cache_ttl": "240s",
```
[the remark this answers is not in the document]
syed: 👍

config.py
```diff
@@ -653,6 +653,12 @@ def create_transaction(db):
     DATA_MODEL_CACHE_CONFIG = {
         "engine": "memcached",
         "endpoint": ("127.0.0.1", 18080),
+        "repository_blob_cache_ttl": "240s",
+        "catalog_page_cache_ttl": "240s",
```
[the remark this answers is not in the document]
syed: 👍

kleesc (approved):

PR #767, opened 2021-04-21
migration: drop uploading column from imagestorage (PROJQUAY-1899)

kleesc:
Follow up to 4ad5a458. Drop deprecated uploading column from
ImageStorage table.


data/migrations/dba_operator/fdc4231fbcc1-databasemigration.yaml
```diff
@@ -0,0 +1,16 @@
...
+apiVersion: dbaoperator.app-sre.redhat.com/v1alpha1
+kind: DatabaseMigration
+metadata:
+  name: fdc4231fbcc1
+spec:
+  migrationContainerSpec:
+    command:
+    - /quay-registry/quay-entrypoint.sh
+    - migrate
+    - c91c564aad34
```
syed, line 12: should be `fdc4231fbcc1`

syed (approved):
LGTM one minor correction

data/migrations/dba_operator/fdc4231fbcc1-databasemigration.yaml
kleesc: Good catch

syed (dismissed):

kleesc:
@syed Holding off merging this. In theory, this should be done without downtime, but still remains an expensive op. Will likely need a maintenance window for this.

Ref:
- https://dev.mysql.com/doc/refman/5.7/en/innodb-online-ddl-operations.html#online-ddl-index-syntax-notes

PR #771, opened 2021-04-27
api: /v1/user/initialize to create first user (PROJQUAY-1926) 

thomasmckay:


endpoints/api/user.py
```diff
@@ -545,6 +546,78 @@ def delete(self):
         return "", 204
 
 
+@resource("/v1/user/initializexxxx")
```
jonathankingfc, line 549: do we still need this endpoint? or is the endpoint defined in `web.py` sufficient?

endpoints/web.py
```diff
@@ -897,3 +898,60 @@ def redirect_to_namespace(namespace):
...
+    Create initial user in an empty database
+    """
+
+    # Ensure that we are using database auth.
+    if app.config["AUTHENTICATION_TYPE"] != "Database":
+        response = jsonify({"message": "Cannot initialize user in a non-database auth system"})
+        response.status_code = 400
+        return response
+
+    # if model.user.get_active_user_count() != 0:
```
jonathankingfc, line 915: is this intentionally commented out? can we use the same function as in the `user.py` endpoint defined above?

jonathankingfc (changes requested):

endpoints/web.py
thomasmckay: intentional yes during my testing so i don't have to reset db constantly :)

endpoints/api/user.py
thomasmckay: still trying to figure out how this method can be included in user.py instead of web.py. one or the other will "win"

endpoints/api/user.py
```diff
@@ -545,6 +546,83 @@ def delete(self):
...
+            success, headers = common_login(new_user.uuid)
+            if not success:
+                raise request_error(message="Could not login. Failed to initialize user")
+
+            return (
+                {
+                    "username": user_data["username"],
+                    "email": user_data.get("email"),
+                    "password": user_data["password"],
+                    "encrypted_password": authentication.encrypt_user_password(
```
kleesc, line 615: **Don't return passwords**

endpoints/api/user.py
```diff
@@ -545,6 +546,83 @@ def delete(self):
...
+    def post(self):
+        """
+        Create initial user in an empty database
+        """
+
+        # Ensure that we are using database auth.
+        if app.config["AUTHENTICATION_TYPE"] != "Database":
+            raise request_error(message="Cannot initialize user in a non-database auth system")
+
+        if model.user.get_active_user_count() != 0:
```
kleesc, line 591: This is doing a full table count every time someone hits this endpoint. Especially, since anyone can hit this api, nothing prevents from constantly hitting this and exhausting db resources.

Since you don't really care about the actual count, something like `User.select().limit(1)` would be more efficient.

endpoints/api/user.py
```diff
@@ -57,6 +57,7 @@
     define_json_response,
     RepositoryParamResource,
     page_support,
+    verify_not_prod,
```
kleesc, line 60: Let's not have specific code for production vs enterprise installs

endpoints/api/__init__.py
```diff
@@ -64,7 +64,7 @@ def handle_error(self, error):
 api = ApiExceptionHandlingApi()
 api.init_app(api_bp)
 api.decorators = [
-    csrf_protect(),
+    csrf_protect(skip_paths=["/api/v1/user/initialize"]),
```
kleesc, line 67: Instead of hacking in the current decorator for one specific use case, if your api needs a different set of decorators than these ones, you should create a separate blueprint for your resource.

kleesc: Also, for this case, you can probably make the request on behalf of Quay, and use  `app.config["HTTPCLIENT"]` to go around the CRSF protection.

endpoints/api/user.py
```diff
@@ -545,6 +546,83 @@ def delete(self):
         return "", 204
 
 
+@resource("/v1/user/initialize")
+class InitializeUser(ApiResource):
```
kleesc, line 550: This api should be put behind a feature flag, and defaults to `False`

kleesc (changes requested):

endpoints/test/test_csrf.py
```diff
@@ -1,8 +1,66 @@
...
+        # POST calls are checked
+        (
+            "/api/something/else",
+            "POST",
+            ["/api/something"],
+            False,
+        ),
+    ],
+)
+def test_csrf_protect(path, method, skip_paths, success):
```
kleesc, line 53: See comment about Flask blueprints

kleesc (changes requested):

endpoints/api/__init__.py
thomasmckay: Do you have example of HTTPCLIENT and how it would be used?

endpoints/api/user.py
thomasmckay: I'll need the `encrypted_password` returned for use in follow-up API calls (that's the bearer token, right?). I can remove the `password`.

endpoints/api/user.py
thomasmckay: For what purpose? It only functions on a clean db.

kleesc:
Not sure how being able to create the first user from api will help. You'll still need to go through the auth flow through the ui to be able to create an org/oauth token anyways...

endpoints/api/user.py
kleesc: No. Passwords and tokens are different.

endpoints/csrf.py
```diff
@@ -60,13 +60,19 @@ def csrf_protect(
     request_token_name=_QUAY_CSRF_TOKEN_NAME,
     all_methods=False,
     check_header=True,
+    skip_paths=None,
```
jonathankingfc, line 63: Do we still need this file?

thomasmckay: fixed, thanks

jonathankingfc (dismissed):
LGTM

endpoints/api/test/test_user.py
```diff
@@ -40,3 +41,106 @@ def test_user_metadata_update(client):
...
+                    "web.user_initialize",
+                    url_for,
+                    "POST",
+                    {},
+                    body=metadata,
+                    expected_code=expected_code,
+                    headers={"Content-Type": "application/json"},
+                )
+
+                if expected_code == 200:
```
kleesc, line 138: An another assertion that the status code matches the expected one.

endpoints/web.py
```diff
@@ -897,3 +898,82 @@ def redirect_to_namespace(namespace):
...
+            model.oauth.create_user_access_token(
+                new_user, user_data["username"], scope, access_token=token_string
+            )
+            result["access_token"] = token_string
+
+        return (result, 200, headers)
+    except model.user.DataModelException as ex:
+        response = jsonify({"message": "Failed to initialize user: " + str(ex)})
+        response.status_code = 400
+        return response
```
kleesc, line 979: Wasn't this previously in the `user.py` or `superuser.py` module? Why did you move it to `web`?

endpoints/api/test/test_user.py
```diff
@@ -40,3 +41,106 @@ def test_user_metadata_update(client):
...
+                "password": "password",
+            },
+        ),
+        # Empty database with access token
+        (
+            0,
+            200,
+            False,
+            True,
+            {"username": "someone", "password": "password", "access_token": "true"},
```
kleesc, line 113: Not sure what case this is. Shouldn't be able to have a token without a user..

endpoints/web.py
thomasmckay: This in web.py due to the csrf and api needing to be unauthenticated. We spent a long time back and forth on this.

endpoints/api/test/test_user.py
thomasmckay: This is testing success w/o mailing feature. Not sure what comment means.

endpoints/api/test/test_user.py
thomasmckay: this assert is done in `conduct_call()`

endpoints/web.py
```diff
@@ -897,3 +898,82 @@ def redirect_to_namespace(namespace):
...
+        if user_data.get("access_token"):
+            model.oauth.create_application(
+                new_user,
+                "automation",
+                "",
+                "",
+                client_id=user_data["username"],
+                description="Application token generated via /api/v1/user/initialize",
+            )
+            token_string = random_string_generator(length=40)()
```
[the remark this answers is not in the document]
thomasmckay: done

endpoints/web.py
```diff
@@ -897,3 +898,82 @@ def redirect_to_namespace(namespace):
...
+            user_data.get("email"),
+            auto_verify=True,
+            email_required=features.MAILING,
+            is_possible_abuser=False,
+            prompts=prompts,
+        )
+        success, headers = common_login(new_user.uuid)
+        if not success:
+            response = jsonify({"message": "Could not login. Failed to initialize user"})
+            response.status_code = 400
```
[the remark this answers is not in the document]
thomasmckay: done

endpoints/web.py
```diff
@@ -897,3 +898,82 @@ def redirect_to_namespace(namespace):
...
+
+
+def has_users():
+    """
+    Return false if no users in database yet
+    """
+    return bool(list(User.select().limit(1)))
+
+
+@web.route("/api/v1/user/initialize", methods=["POST"])
```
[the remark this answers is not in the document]
thomasmckay: done

endpoints/web.py
```diff
@@ -897,3 +898,82 @@ def redirect_to_namespace(namespace):
         return redirect(url_for("web.org_view", path=namespace))
     else:
         return redirect(url_for("web.user_view", path=namespace))
+
+
+def has_users():
+    """
+    Return false if no users in database yet
+    """
+    return bool(list(User.select().limit(1)))
```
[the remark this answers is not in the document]
thomasmckay: done

endpoints/api/test/test_user.py
```diff
@@ -40,3 +41,94 @@ def test_user_metadata_update(client):
...
+                )
+
+                if expected_code == 200:
+                    assert user.json["username"] == metadata["username"]
+                    if feature_mailing:
+                        assert user.json["email"] == metadata["email"]
+                    else:
+                        assert user.json["email"] is None
+                    assert user.json.get("encrypted_password", None)
+                    if metadata.get("access_token"):
```
kleesc, line 133: Can you add an assertion that checks the opposite is also true: if access_token is not set, then there shouldn't be one in the response.

thomasmckay: done

kleesc (approved):

PR #772, opened 2021-04-27
repository: config to allow public repo create (PROJQUAY-1929)

thomasmckay:


util/config/schema.py
```diff
@@ -1162,5 +1162,10 @@
             "description": "The set of hostnames to disallow from webhooks when validating, beyond localhost",
             "x-example": ["somexternaldomain.com"],
         },
+        "REPOSITORY_CREATION_PRIVATE": {
+            "type": "boolean",
+            "description": "Whether new repositories created by push are set to private visibility. Defaults to True.",
+            "x-example": True,
+        },
```
kleesc, line 1169: Add the default value (`True`) to `config.py`

endpoints/v2/v2auth.py
```diff
@@ -255,7 +255,14 @@ def _authorize_or_downscope_request(scope_param, has_valid_auth_context):
                 # TODO: Push-to-create functionality should be configurable
                 if CreateRepositoryPermission(namespace).can() and user is not None:
                     logger.debug("Creating repository: %s/%s", namespace, reponame)
-                    found = model.repository.get_or_create_repository(namespace, reponame, user)
+                    visibility = (
+                        "private"
+                        if app.config.get("REPOSITORY_CREATION_PRIVATE", True)
```
kleesc, line 260: Can you rename the config flag to something more descriptive: `CREATE_PRIVATE_REPO_ON_PUSH` or something similar?

endpoints/v2/test/test_v2auth.py
```diff
@@ -217,3 +217,55 @@ def test_generate_registry_jwt(
...
+            ["devtable/newrepo:push,pull,*"],
+        ),
+    ],
+)
+def test_generate_registry_jwt_private(
+    scope, username, password, expected_code, expected_scopes, app, client
+):
+    try:
+        original_app.config["REPOSITORY_CREATION_PRIVATE"] = True
+        test_generate_registry_jwt(
```
kleesc, line 239: Instead of calling another test, you should parameterize it so that it tests for your new cases.

kleesc (changes requested):

endpoints/v2/test/test_v2auth.py
```diff
@@ -217,3 +217,33 @@ def test_generate_registry_jwt(
...
+)
+def test_generate_registry_jwt_visibility(push_private, visibility, app, client):
+    scope = "repository:devtable/newrepo:pull,push,*"
+    username = "devtable"
+    password = "password"
+    expected_code = 200
+    expected_scopes = ["devtable/newrepo:push,pull,*"]
+    try:
+        original_app.config["CREATE_PRIVATE_REPO_ON_PUSH"] = push_private
+        test_generate_registry_jwt(
```
kleesc, line 243: While you _can_ call another test, I would say it's generally preferable to decouple it so that it's easier to debug in the future.

kleesc (changes requested):

endpoints/v2/test/test_v2auth.py
thomasmckay: I considered it but chose this route as the clearest. I want to run the test only isolating a config change.

kleesc: The issue is I can't really tell from looking at the test what it's doing. If you're just calling another test, you could just parametrize the existing one with your constraints.

kleesc (changes requested):

endpoints/v2/test/test_v2auth.py
thomasmckay: @kleesc Moved to the existing test

kleesc (approved):
LGTM

PR #774, opened 2021-04-30
quayio: followup to 1fd8194e

kleesc:
Removes uploading filters/asertions that were not removed when cherripicking 1fd8194e.
These changes were not in the original PR to master to begin with, which is why they didn't raise a merge conflict.


syed:
LGTM

PR #775, opened 2021-04-30
db: remove transaction from empty layer upload (PROJQUAY-1946)

kleesc:
Decouples blob upload and db row creation when uploading the empty
layer. This avoids the transaction blocking on non-db/S3 related issues.


thomasmckay (approved):
LGTM

PR #780, opened 2021-05-04
db: allow peewee connection pool through configuration (PROJQUAY-1695)

kleesc:
Allows peewee's connection pooling feature to be enabled through Quay's config.
If both the environment variable and config is set, the environment variable
will take precedence.


jonathankingfc (dismissed):
LGTM

PR #782, opened 2021-05-06
fips: enforce smtp tls (PROJQUAY-1804)

jonathankingfc:
Patch CRAM-MD5 function to allow usage under FIPS but enforce TLS

### DIFF
--- a/util/fips.py
+++ b/util/fips.py
@@ -1,0 +1,62 @@
+from smtplib import SMTPException
+from smtplib import SMTPAuthenticationError
+from smtplib import SMTPNotSupportedError
+
+# This method is the same as smtplib.SMTP.login except that CRAM_MD5 has been removed from preferred auths since MD5 is not FIPS compliant.
+def login_fips_safe(self, user, password, *, initial_response_ok=True):
+    """Log in on an SMTP server that requires authentication.
+    The arguments are:
+        - user:         The user name to authenticate with.
+        - password:     The password for the authentication.
+    Keyword arguments:
+        - initial_response_ok: Allow sending the RFC 4954 initial-response
+            to the AUTH command, if the authentication methods supports it.
+    If there has been no previous EHLO or HELO command this session, this
+    method tries ESMTP EHLO first.
+    This method will return normally if the authentication was successful.
+    This method may raise the following exceptions:
+        SMTPHeloError            The server didn't reply properly to
+                                the helo greeting.
+        SMTPAuthenticationError  The server didn't accept the username/
+                                password combination.
+        SMTPNotSupportedError    The AUTH command is not supported by the
+                                server.
+        SMTPException            No suitable authentication method was
+                                found.
+    """
+
+    self.ehlo_or_helo_if_needed()
+    if not self.has_extn("auth"):
+        raise SMTPNotSupportedError("SMTP AUTH extension not supported by server.")
+
+    # Authentication methods the server claims to support
+    advertised_authlist = self.esmtp_features["auth"].split()
+
+    # Authentication methods we can handle in our preferred order:
+    preferred_auths = ["PLAIN", "LOGIN"]
+
+    # We try the supported authentications in our preferred order, if
+    # the server supports them.
+    authlist = [auth for auth in preferred_auths if auth in advertised_authlist]
+    if not authlist:
+        raise SMTPException("No suitable authentication method found.")
+
+    # Some servers advertise authentication methods they don't really
+    # support, so if authentication fails, we continue until we've tried
+    # all methods.
+    self.user, self.password = user, password
+    for authmethod in authlist:
+        method_name = "auth_" + authmethod.lower().replace("-", "_")
+        try:
+            (code, resp) = self.auth(
+                authmethod, getattr(self, method_name), initial_response_ok=initial_response_ok
+            )
+            # 235 == 'Authentication successful'
+            # 503 == 'Error: already authenticated'
+            if code in (235, 503):
+                return (code, resp)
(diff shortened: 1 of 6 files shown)


util/fips.py
```diff
@@ -0,0 +1,17 @@
...
+def fips_auth_cram_md5(self, challenge=None):
+    """Authobject to use with CRAM-MD5 authentication. Requires self.user
+    and self.password to be set."""
+    # CRAM-MD5 does not support initial-response.
+    if challenge is None:
+        return None
+    return (
+        self.user
+        + " "
+        + hmac.new(
```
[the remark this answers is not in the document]
jonathankingfc: I've included the original implementation in the comment above.

util/useremails.py
```diff
@@ -67,7 +69,13 @@ def send_email(recipient, subject, template_file, parameters, action=None):
     msg.body = plain
 
     try:
-        mail.send(msg)
+        if features.FIPS:
+            assert app.config[
+                "MAIL_USE_TLS"
+            ], "MAIL_USE_TLS must be enabled to use SMTP in FIPS mode."
+        with mock.patch("smtplib.SMTP.auth_cram_md5", fips_auth_cram_md5):
```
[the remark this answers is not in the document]
jonathankingfc: There is no way to configure the smtp library to not use MD5.

util/fips.py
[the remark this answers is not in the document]
syed: @jonathankingfc  is this specific to the fips implementation? can we use the fips implementation outside of a fips cluster would that break anything?

util/useremails.py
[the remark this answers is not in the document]
syed: @jonathankingfc are we always monkey patching regardless of if we have fips enabled or not?

util/fips.py
```diff
@@ -0,0 +1,120 @@
+from smtplib import SMTPException
+from smtplib import SMTPAuthenticationError
+from smtplib import SMTPNotSupportedError
+
+# def login(self, user, password, *, initial_response_ok=True):
```
kleesc, line 5: Remove this if it's not needed.

util/fips.py
```diff
@@ -0,0 +1,120 @@
...
+#             # 503 == 'Error: already authenticated'
+#             if code in (235, 503):
+#                 return (code, resp)
+#         except SMTPAuthenticationError as e:
+#             last_exception = e
+
+#     # We could not login successfully.  Return result of last attempt.
+#     raise last_exception
+def login_fips_safe(self, user, password, *, initial_response_ok=True):
+    """Log in on an SMTP server that requires authentication.
```
kleesc, line 65: Can you briefly describe how this implementation is different from the smtplib one?
And should we add a test for this implementation?

util/useremails.py
```diff
@@ -67,7 +69,15 @@ def send_email(recipient, subject, template_file, parameters, action=None):
     msg.body = plain
 
     try:
-        mail.send(msg)
+        if features.FIPS:
+            assert app.config[
+                "MAIL_USE_TLS"
+            ], "MAIL_USE_TLS must be enabled to use SMTP in FIPS mode."
+            with mock.patch("smtplib.SMTP.login", login_fips_safe):
```
kleesc, line 76: Same question as above: is `smtplib.SMTP.login = login_fips_safe` or is `mock` needed?

util/useremails.py
```diff
@@ -202,7 +212,12 @@ def send_invoice_email(email, contents):
     # normal template here.
     msg = Message("Quay payment received - Thank you!", recipients=[email])
     msg.html = contents
-    mail.send(msg)
+    if features.FIPS:
+        assert app.config["MAIL_USE_TLS"], "MAIL_USE_TLS must be enabled to use SMTP in FIPS mode."
+        with mock.patch("smtplib.SMTP.login", login_fips_safe):
```
kleesc, line 217: Ditto

util/useremails.py
```diff
@@ -238,4 +253,9 @@ def send_subscription_change(change_description, customer_id, customer_email, qu
     msg.html = SUBSCRIPTION_CHANGE.format(
         change_description, customer_id, customer_email, quay_username
     )
-    mail.send(msg)
+    if features.FIPS:
+        assert app.config["MAIL_USE_TLS"], "MAIL_USE_TLS must be enabled to use SMTP in FIPS mode."
+        with mock.patch("smtplib.SMTP.login", login_fips_safe):
```
kleesc, line 258: Ditto

harishsurf (commented):
Adding a test would give us some confidence. There are some review comments that needs to be addressed, otherwise lgtm!

util/useremails.py
jonathankingfc: mock is needed since we want to localize the change for only the block below (using the with keyword).

util/fips.py
```diff
@@ -0,0 +1,62 @@
+from smtplib import SMTPException
+from smtplib import SMTPAuthenticationError
+from smtplib import SMTPNotSupportedError
+
+# This method is the same as smtplib.SMTP.login except that CRAM_MD5 has been removed from preferred auths since MD5 is not FIPS compliant.
```
kleesc, line 5: For reference, https://github.com/mattupstate/flask-mail/blob/1709c70d839a7cc7b1f7eeb97333b71cd420fe32/flask_mail.py#L154 doesn't allow configuring the SMTP client other than the server and the port. So we can't do something `conn.esmtp_features['auth'] = 'PLAIN LOGIN'`

kleesc (approved):

PR #789, opened 2021-05-12
PROJQUAY-1123 (3.5.2)

kleesc:
Explicitly call abort on a mpu if no bytes were written.
Noobaa and Rados will not clean the artifacts, resulting in empty
files being stuck in an "uploading" state.


thomasmckay (approved):
LGTM

PR #790, opened 2021-05-12
chore: 3.4.5 release (PROJQUAY-2005)

thomasmckay:
3.4.5 release updates

### DIFF
--- a/config.py
+++ b/config.py
@@ -440,7 +440,7 @@ class DefaultConfig(ImmutableConfig):
     BRANDING = {
         "logo": "/static/img/RH_Logo_Quay_Black_UX-horizontal.svg",
         "footer_img": "/static/img/RedHat.svg",
-        "footer_url": "https://redhat.com",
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -737,7 +737,7 @@ def create_transaction(db):
     WEBHOOK_HOSTNAME_BLACKLIST = []
 
     # The root URL for documentation.
-    DOCUMENTATION_ROOT = "https://access.redhat.com/documentation/en-us/red_hat_quay/3.4/"
+    DOCUMENTATION_ROOT = "https://access.redhat.com/documentation/en-us/red_hat_quay/3/"
```
jonathankingfc, line 740: Should this be 3.5?

thomasmckay: `/3/`always points to the latest release so I've been moving links there so they don't need to be changed on every y-stream.

jonathankingfc (approved):
LGTM

PR #792, opened 2021-05-16
mirror: Add feature flag to allow initial user creation (PROJQUAY-1926)

jonathankingfc:
- Users can use FEATURE_INIT_USER to allow an initial user to be created.
- Credentials can be passed through `INIT_USERNAME` and `INIT_PASSWORD` otherwise defaults will be used


LiZhang19817:
Will this feature "FEATURE_INIT_USER" be exposed on quay config editor?

alecmerdler:
Agreed with @LiZhang19817, this PR should be accompanied by a subsequent PR to `config-tool`.

Makefile
```diff
@@ -195,10 +195,10 @@ local-dev-build:
 .PHONY: local-dev-up
 local-dev-up:
 	make local-dev-clean
-	docker-compose up -d redis
-	docker-compose up -d quay-db
+	docker-compose up  --build -d redis
```
alecmerdler, line 198: Can you explain what this does?

local-dev/stack/config.yaml
```diff
@@ -30,7 +30,7 @@ FEATURE_REPO_MIRROR: false
 FEATURE_REQUIRE_TEAM_INVITE: true
 FEATURE_RESTRICTED_V1_PUSH: false
 FEATURE_SECURITY_NOTIFICATIONS: true
-FEATURE_SECURITY_SCANNER: true
```
alecmerdler, line 33: Is disabling the security scanner by default an intended part of this PR?

local-dev/stack/config.yaml
```diff
@@ -69,3 +69,6 @@ USER_EVENTS_REDIS:
   host: quay-redis
   port: 6379
 USE_CDN: false
+FEATURE_INIT_USER: true
```
alecmerdler, line 72: Prefer `FEATURE_INITIAL_USER`, `INITIAL USERNAME`, and `INITIAL_PASSWORD`.

alecmerdler (changes requested):
This deserves further discussion as it involves database rows defined by `config.yaml`. 

- What happens if this user is deleted? When Quay restarts, is the user recreated?
- Should there be the same password requirements on the initial user password as on the rest of Quay's users? Will that cause issues if they are different?
- What happens if the `config.yaml` is modified to remove this initial user, but it exists previously? Is it deleted?

alecmerdler:
I think this is significant enough of a feature to deserve an [enhancement proposal](https://github.com/quay/enhancements), especially to capture the motivations and design.

data/model/user.py
```diff
@@ -111,6 +111,34 @@ def create_user(
...
+):
+    """
+    Creates an initial user, if none are in the database.
+    """
+
+    if config.app_config.get("AUTHENTICATION_TYPE") != "Database":
+        logger.debug("Cannot create init user in a non-database auth system")
+        return
+
+    if bool(list(User.select().limit(1))):
```
kleesc, line 127: We've generally favored running the query and catching the "DoesNotExist" exception when testing for existence.
e.g 
```
try:
  User.select().limit(1).get()
except UserDoesNotExist:
  logger.debug("Cannot initialize user in a non-empty database")
  return
```

data/model/user.py
```diff
@@ -111,6 +111,34 @@ def create_user(
     return created
 
 
+def create_init_user(
```
kleesc, line 114: Can we add a test for this? Just to test the new constraints we're adding to `create_user`.

kleesc (changes requested):

kleesc:
@jonathankingfc Is this a duplicate of #771 ?

kleesc:
> I think this is significant enough of a feature to deserve an [enhancement proposal](https://github.com/quay/enhancements), especially to capture the motivations and design.

@alecmerdler @jonathankingfc Agree. I'm not sure what the scope of this needs to be. Are we just looking to get access to the registry API, or also to Quay's API. AFAIK the latter also requires an application/org to be able to create an oauth token to be able to use on behalf of that first user.

jonathankingfc:
Closing in favor of https://github.com/quay/quay/pull/771

jonathankingfc:
Closing in favor of https://github.com/quay/quay/pull/771

PR #795, opened 2021-05-20
modelcache: Add read and write endpoints to Redis (PROJQUAY-1939)

syed:
Adds `read_client` and `write_client` for Redis ModelCache to seperate out read
and write calls to separate on endpoints

### DIFF
--- a/data/cache/__init__.py
+++ b/data/cache/__init__.py
@@ -38,18 +38,16 @@ def get_model_cache(config):
         return cache
 
     if engine == "redis":
-        host = cache_config.get("host", None)
-        if host is None:
-            raise Exception("Missing `host` for Redis model cache configuration")
+        primary_config = cache_config.get("primary", None)
+        replica_config = cache_config.get("replica", None)
+
+        if not primary_config or primary_config.get("host") is None:
+            raise Exception("Missing `primary_host` for Redis model cache configuration")
 
         return RedisDataModelCache(
             cache_config,
-            host=host,
-            port=cache_config.get("port", 6379),
-            password=cache_config.get("password", None),
-            db=cache_config.get("db", 0),
-            ca_cert=cache_config.get("ca_cert", None),
-            ssl=cache_config.get("ssl", False),
+            primary_config=primary_config,
+            replica_config=replica_config,
         )
 
     raise Exception("Unknown model cache engine `%s`" % engine)
--- a/data/cache/impl.py
+++ b/data/cache/impl.py
@@ -246,33 +246,44 @@ class RedisDataModelCache(DataModelCache):
     def __init__(
         self,
         cache_config,
-        host="127.0.0.1",
-        port=6379,
-        password=None,
-        db=0,
-        ca_cert=None,
-        ssl=False,
+        primary_config,
+        replica_config=None,
     ):
         super(RedisDataModelCache, self).__init__(cache_config)
-        self.client = StrictRedis(
-            host=host,
-            port=port,
-            password=password,
-            db=db,
-            ssl_ca_certs=ca_cert,
-            ssl=ssl,
+        self.write_client = StrictRedis(
+            host=primary_config.get("host"),
+            port=primary_config.get("port"),
+            password=primary_config.get("password"),
+            db=primary_config.get("db"),
+            ssl_ca_certs=primary_config.get("ca_certs"),
+            ssl=primary_config.get("ssl"),
             socket_connect_timeout=1,
             socket_timeout=2,
             health_check_interval=2,
         )
 
+        if not replica_config:
+            self.read_client = self.write_client
+        else:
+            self.read_client = StrictRedis(
+                host=replica_config.get("host"),
+                port=replica_config.get("port"),
+                password=replica_config.get("password"),
+                db=replica_config.get("db"),
+                ssl_ca_certs=replica_config.get("ca_certs"),
+                ssl=replica_config.get("ssl"),
+                socket_connect_timeout=1,
+                socket_timeout=2,
+                health_check_interval=2,
+            )
+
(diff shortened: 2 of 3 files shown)


data/cache/impl.py
```diff
@@ -244,19 +244,32 @@ class RedisDataModelCache(DataModelCache):
...
+            self,
+            cache_config,
+            read_host="127.0.0.1",
+            read_port=6379,
+            write_host="127.0.0.1",
+            write_port=6379,
+            password=None,
+            db=0,
+            ca_cert=None,
+            ssl=False,
```
kleesc, line 256: The replica may have different config value for these parameters?

data/cache/impl.py
```diff
@@ -244,19 +244,32 @@ class RedisDataModelCache(DataModelCache):
...
-        host="127.0.0.1",
-        port=6379,
-        password=None,
-        db=0,
-        ca_cert=None,
-        ssl=False,
+            self,
+            cache_config,
+            read_host="127.0.0.1",
+            read_port=6379,
```
kleesc, line 250: These new values should also be added to the default config.

kleesc (changes requested):

kleesc (changes requested):

data/cache/impl.py
syed: I don't think the replica can have a different password or DB. Not sure about CA cert but I'd don't see a use case where we will have a different CA for primary and replica

data/cache/impl.py
syed: The `config.py` has the Memcache settings. Can we add this without causing any issues?

data/cache/__init__.py
```diff
@@ -38,14 +38,17 @@ def get_model_cache(config):
         return cache
 
     if engine == "redis":
-        host = cache_config.get("host", None)
-        if host is None:
-            raise Exception("Missing `host` for Redis model cache configuration")
+        read_host = cache_config.get("read_host", None)
+        write_host = cache_config.get("write_host", None)
+        if read_host is None or write_host is None:
```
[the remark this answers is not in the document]
syed: 👍

data/cache/impl.py
syed: i've added these settings in the enhancement repo https://github.com/quay/enhancements/pull/8

data/cache/impl.py
kleesc: Not sure about passwords, but I'm assuming if a replica can be setup on a separate/completely different host, it should be able to have a separate cert for validation.
We should create a wrapper around the separate configs instead (maybe a namedtuple):
```
_RedisConfig = namedtuple("_RedisConfig", ["host", "port", "password", ...])
```

data/cache/impl.py
kleesc: If you add/change config values, defaults should be set/updated in `config.py`.

kleesc: See https://github.com/quay/quay/blob/master/config.py

kleesc: For example, with some of the host key name changes, the values in `config.py` are no longer valid.

data/cache/impl.py
syed: `redis-py` supports single host config only. We would never pass more than 2 hosts to the Redis ModelCache. It's simple if we just pass them via the constructor

data/cache/impl.py
kleesc: Nevermind, the default uses memcached

data/cache/__init__.py
```diff
@@ -38,18 +38,16 @@ def get_model_cache(config):
         return cache
 
     if engine == "redis":
-        host = cache_config.get("host", None)
-        if host is None:
-            raise Exception("Missing `host` for Redis model cache configuration")
+        primary_config = cache_config.get("primary_config", None)
+        replica_config = cache_config.get("replica_config", primary_config)
```
kleesc, line 42: nit: remove the `_config` suffix from the config key.
The final config should look like:
```
{
  "engine":  "redis",
  "primary": {
    ...
  }
  "replica": {
    ...
  }
}
```

kleesc (changes requested):

data/cache/impl.py
```diff
@@ -243,24 +243,41 @@ class RedisDataModelCache(DataModelCache):
...
+            password=replica_config.get('password'),
+            db=replica_config.get('db'),
+            ssl_ca_certs=replica_config.get('ca_certs'),
+            ssl=replica_config.get('ssl'),
+            socket_connect_timeout=1,
+            socket_timeout=2,
+            health_check_interval=2,
+        )
+
+        self.write_client = StrictRedis(
```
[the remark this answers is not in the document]
syed: 👍

data/cache/__init__.py
syed: 👍

data/cache/impl.py
```diff
@@ -243,24 +243,41 @@ class RedisDataModelCache(DataModelCache):
...
     ):
         super(RedisDataModelCache, self).__init__(cache_config)
-        self.client = StrictRedis(
-            host=host,
-            port=port,
-            password=password,
-            db=db,
-            ssl_ca_certs=ca_cert,
-            ssl=ssl,
+        self.read_client = StrictRedis(
```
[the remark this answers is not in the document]
syed: 👍

data/cache/impl.py
```diff
@@ -243,24 +243,41 @@ class RedisDataModelCache(DataModelCache):
     Implementation of the data model cache backed by a Redis service.
     """
 
+    _DEFAULT_CONFIG = {
```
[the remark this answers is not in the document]
syed: 👍

kleesc (approved):
LGTM

PR #799, opened 2021-05-26
organization: config to allow organization creation on push (PROJQUAY-1928)

thomasmckay:


config.py
```diff
@@ -770,3 +770,6 @@ def create_transaction(db):
 
     # Repos created by push default to private visibility
     CREATE_PRIVATE_REPO_ON_PUSH = True
+
+    # Create organization on push if it does not exist
+    CREATE_ORGANIZATION_ON_PUSH = False
```
alecmerdler, line 775: Would this make more sense as `CREATE_NAMESPACE_ON_PUSH`?

thomasmckay: agreed, changed

alecmerdler (approved):

PR #800, opened 2021-05-31
util: fix matching multiples in jsontemplate.py

jilleJr:
> This is my first contribution to quay. I tried my best to follow the CONTRIBUTING.md guide, but I guessed on a lot of things.

Matching multiples caused exceptions to be raised. This made it so webhook notification POST bodies could not template things like the image tags as `${tags[*]}`. This was caused by a typo in the jsonpath_rw match field that was used.

I encountered this when trying to use the notification system in quay.io and decided to dig through the code to understand why it didn't work. I did not find a relevant Jira ticket about this, so I just created this PR without a Jira ticket.

Signed-off-by: Kalle Jillheden <kalle.jillheden@iver.se>

### DIFF
--- a/util/jsontemplate.py
+++ b/util/jsontemplate.py
@@ -81,7 +81,7 @@ def _process_inline(expression, data):
             return None
 
         if len(found) > 1:
-            return [f.data for f in found]
+            return [f.value for f in found]
 
         return found[0].value
     except IndexError:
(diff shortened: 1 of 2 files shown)


kleesc (approved):
Thanks for fixing this. LGTM.

jilleJr:
I updated to fix the test and commit name (I think). Don't have a ticket ID so I added PROJQUAY-0000

kleesc (dismissed):

jilleJr:
I'm super sorry for the failing tests. I've fixed them, yet again.

kleesc (approved):
LGTM

PR #801, opened 2021-05-31
oci: add support for zstd compression (PROJQUAY-1417)

giuseppe:
allow upload of layers compressed with zstd, instead of gzip.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>


rhatdan:
We would really like to get this in, so we could start using this format.

giuseppe:
@thomasmckay @syed @kleesc @hdonnay PTAL

kleesc:
@giuseppe Technically, `application/vnd.oci.image.layer.v1.tar+zstd` is not officially part of the spec (yet).
However, there is work in progress as part of 3.6 to enable registering additional mimetypes as oci artifacts (PROJQUAY-1032).

giuseppe:
> @giuseppe Technically, `application/vnd.oci.image.layer.v1.tar+zstd` is not officially part of the spec (yet).
> However, there is work in progress as part of 3.6 to enable registering additional mimetypes as oci artifacts ([PROJQUAY-1032](https://issues.redhat.com/browse/PROJQUAY-1032)).

yes technically it is not part of the specs yet and there still an on going discussion on how to formalize its support before cutting a new release. 
The status quo though is that the Docker registry accept zstd compressed layers, so we are in the position of experimenting with the new format in our containers tools using the Docker registry but not being able to use the official Red Hat registry.

I am fine if this is added with additional mimetypes, but would it be present in the default configuration?  Do users need to add the new MIME type before being able to push zstd compressed layers to a registry?

giuseppe:
anything more I can do to move this forward?

rhatdan:
@kleesc What is the latest on this PR?

thomasmckay:
The functionality here is implemented through https://github.com/quay/quay/pull/803 , please test and comment there. Thanks!

jonathankingfc (approved):
LGTM

kleesc:
I'm fine with merging this, as long as there's no expectation that it will provide some sort of fallback to gzip for clients that will not support this.

PR #812, opened 2021-06-16
buildman: fix vm image name in startup script (PROJQUAY-2120) (#811)

kleesc:
Reference correct filename in start.sh


thomasmckay (approved):
LGTM

PR #823, opened 2021-07-07
Redirecting docker 8443 port to localhost 443

Sunandadadi:


syed:
Thanks @Sunandadadi could you add a bit more description on why we are making this change and also could you change the commit user from `parallels` to your name as well

Sunandadadi:
Docker daemon is listening on localhost port 443. So, exposed the localhost port 443, for the daemon to connect.

Sunandadadi:
> Thanks @Sunandadadi could you add a bit more description on why we are making this change and also could you change the commit user from `parallels` to your name as well

Noted, please review the changes.

PR #831, opened 2021-07-09
deploy: Add recovery endpoint deployment manifests (PROJQUAY-970)

syed:
This change adds the deployment and service manifests for the recovery endpoint
that will be used for quay.io to recover accounts which have not already been
transitioned to RH SSO


thomasmckay (approved):
LGTM

PR #840, opened 2021-07-16
API: Propagate Clair-4.2 enrichment data to Quay API and UI (PROJQUAY -2102)

Sunandadadi:
Updating Quay secscan API response to reflect enrichment data supplied by Clair-4.2 Security Scanner API


static/js/services/vulnerability-service.js
```diff
@@ -502,14 +504,14 @@ angular.module('quay').factory('VulnerabilityService', ['Config', 'ApiService',
       'description': 'This metric measures the impact on confidentiality of a successfully exploited vulnerability. Confidentiality refers to limiting information access and disclosure to only authorized users, as well as preventing access by, or disclosure to, unauthorized ones. Increased confidentiality impact increases the vulnerability score.',
       'values': [
         {
-          'id': 'C',
-          'title': 'Complete',
+          'id': 'H',
```
crozzy, line 507: Is there a document that describes these updated values that you could include in the PR?

static/js/directives/ui/manifest-vulnerability-view.js
```diff
@@ -29,7 +29,10 @@ angular.module('quay').directive('manifestVulnerabilityView', function () {
       };
 
       $scope.getDistro = function(vuln) {
-        return vuln['namespace'].split(':', 1);
+        if (vuln['namespace']) {
```
crozzy, line 32: Is this directly related to surfacing the enrichment data?

crozzy (approved):
@Sunandadadi this looks great, very clean. I just left a couple of minor comments, were you able to test with the newest Clair image?

static/js/directives/ui/manifest-vulnerability-view.js
Sunandadadi: I saw this failing on the UI when the key "namespace" is empty in the vuln object. This is part of the UI to display distribution names in manifest vulnerabilities.

Sunandadadi:
> @Sunandadadi this looks great, very clean. I just left a couple of minor comments, were you able to test with the newest Clair image?

@crozzy Thanks! Yes, @hdonnay built an image that has the latest Clair (v4) changes and I've worked and tested the code against this image.

static/js/services/vulnerability-service.js
```diff
@@ -418,7 +417,10 @@ angular.module('quay').factory('VulnerabilityService', ['Config', 'ApiService',
...
   vulnService.getVectorOptions = function(vectorString) {
     var parts = vectorString.split(':');
-    return vulnService.NVD_VECTORS[parts[0]].values;
+    if (vulnService.NVD_VECTORS[parts[0]]) {
+      return vulnService.NVD_VECTORS[parts[0]].values;
+    }
+    return null;
   };
 
   vulnService.NVD_VECTORS = {
```
hdonnay, line 426: This vectors object needs to be updated for CVSSv3: https://www.first.org/cvss/v3.1/specification-document#Vector-String

static/js/services/vulnerability-service.js
Sunandadadi: https://www.first.org/cvss/v3.1/specification-document#Vector-String
As suggested by @hdonnay

static/js/services/vulnerability-service.js
```diff
@@ -502,14 +542,14 @@ angular.module('quay').factory('VulnerabilityService', ['Config', 'ApiService',
       'description': 'This metric measures the impact on confidentiality of a successfully exploited vulnerability. Confidentiality refers to limiting information access and disclosure to only authorized users, as well as preventing access by, or disclosure to, unauthorized ones. Increased confidentiality impact increases the vulnerability score.',
       'values': [
         {
-          'id': 'C',
-          'title': 'Complete',
+          'id': 'H',
+          'title': 'High',
```
crozzy, line 546: Some of these descriptions have been modified in the v3 spec i.e. this one:
```
There is total information disclosure, resulting in all system files being revealed. The attacker is able to read all of the system\'s data (memory, files, etc.)
```
to
```
There is a total loss of confidentiality, resulting in all resources within the impacted component...
```
I assume we want to keep these aligned?

Sunandadadi: Sure, I will keep the descriptions aligned.

static/js/services/vulnerability-service.js
Sunandadadi: I updated the vectors object in the latest commit. Can you please review it?

crozzy (approved):

static/directives/manifest-vulnerability-view.html
```diff
@@ -187,15 +187,15 @@ <h3>Vulnerabilities</h3>
             <div class="severity-note" bo-if="vuln.scoreDivergence">
```
hdonnay, line 187: Is this true if the distro score is "Unknown"?

static/js/services/vulnerability-service.js
```diff
@@ -570,7 +1044,8 @@ angular.module('quay').factory('VulnerabilityService', ['Config', 'ApiService',
           'severity': 'low'
         }
       ]
-    }
+    },
+
```
hdonnay, line 1048: I don't think this trailing comma is allowed, but maybe the rules are different in javascript?

hdonnay (changes requested):
This looks good, just two points.

static/directives/manifest-vulnerability-view.html
Sunandadadi: Yes, currently this is true.

static/js/services/vulnerability-service.js
Sunandadadi: Yes, JavaScript allows trailing commas (ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas)

static/directives/manifest-vulnerability-view.html
Sunandadadi: Sorry about my previous comment. But, after some testing,  I see that if the score is “Unknown”, it will not display any note. I believe that the severity-note is display if score divergence exists as stated [here](https://github.com/quay/quay/blob/master/static/directives/manifest-vulnerability-view.html#L187). And from [here](https://github.com/quay/quay/blob/master/static/directives/manifest-vulnerability-view.html#L187), I understand that if cvss score is null then score divergence is also null.

hdonnay (approved):

PR #846, opened 2021-07-22
cache: add support for redis cluster mode (PROJQUAY-2101) (#810)

syed:
Adds support to use redis in cluster mode when using redis instead of
memcached for caching.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -149,6 +149,7 @@ PyYAML==5.1.2
 raven==6.10.0
 recaptcha2==0.1
 redis==3.3.8
+redis-py-cluster==2.1.3
 redlock==1.2.0
 reportlab==2.7
 requests-aws4auth==0.9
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -70,9 +70,10 @@ geoip2==2.9.0
 gevent==1.4.0
 gipc==1.0.1
 greenlet==0.4.15
-gunicorn==19.9.0
-hiredis==1.0.0
-html5lib==0.9999999
+grpcio==1.30.0
+gunicorn==20.0.4
+hashids==1.2.0
+html5lib==1.0.1
```
kleesc, line 76: I don't know if these libraries are compatible with python 2.
We should only import what we know we'll need (`redis-py-cluster` in this case)

requirements-dev.txt
```diff
@@ -1,6 +1,18 @@
...
+pytest-cov
+pytest-flask
+pytest-grpc
+pytest-runner
+pytest-timeout
+pytest-xdist
+python-coveralls
+pytest-sugar
+tox
+tox-docker
```
kleesc, line 15: These are probably from the `master` code base. Likely don't need these.

kleesc (approved):

PR #848, opened 2021-07-23
secscan: handle proxy model fallback to noop v2 (PROJQUAY-2289) (#847)

kleesc:
Handles the case where a V2 scanner is not configured (noop
implementation) and the proxy model tries to load security information
that has not been indexed in V4 yet. In such case, the noop
implementation of the V2 scanner should return a COULD_NOT_LOAD
status (instead of None, to comply to the secscan model interface),
and ignore the fallback to V2, and finally just return the V4
status (NOT_YET_INDEXED)


thomasmckay (approved):
LGTM

PR #863, opened 2021-08-05
PROJQUAY-2253: OCI compliance

hdonnay:
This is a stab at fixing the OCI compliance.

This first patch breaks the allowlist mechanism, which was pushed too early in the validation process.
Subsequent patches should add this mechanism back to higher layers.

### DIFF
--- a/image/oci/config.py
+++ b/image/oci/config.py
@@ -60,7 +60,6 @@ Example:
 
 import copy
 import json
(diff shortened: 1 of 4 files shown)


hdonnay:
I don't know about this PR specifically, but the issue with not being OCI compliant is outstanding and this change is needed as a part of that work.

PR #883, opened 2021-09-08
chore(Dockerfile): add local dev env target (PROJQUAY-2501)

flavianmissi:
To run a local development environment we currently use docker-compose with the custom `local-dev/Dockerfile`.
For better reuse of the improvements made to `Dockerfile`, the specific bits of `local-dev/Dockerfile` have been moved to `Dockerfile` as a build stage called `local-dev-env`, and `local-dev/Dockerfile` _has been removed_. This change also ensures that future improvements to `Dockerfile` will automatically be reused in our local dev envs.

~NOTE: `setuptools` version had to be pinned to 57.5.0. Versions >=58 break the installation of [funcparserlib](https://github.com/vlasovskikh/funcparserlib)~ fixed in https://github.com/quay/quay/pull/885


docker-compose.yaml
```diff
@@ -42,9 +42,9 @@ services:
 
   quay:
     container_name: quay-quay
-    build: 
+    build:
       context: .
-      dockerfile: ./local-dev/Dockerfile
+      target: local-dev-env
```
kleesc, line 47: Oh, I didn't know you could stop at specific stage. That's neat.

kleesc (approved):

kleesc:
@flavianmissi I think the commit check is not passing because the ticket is missing from your commit header (it's in the Github PR's title, but not in the actual git commit).

flavianmissi:
> @flavianmissi I think the commit check is not passing because the ticket is missing from your commit header (it's in the Github PR's title, but not in the actual git commit).

Ooh that makes sense! I'll fix then merge. Thx for reviewing!

flavianmissi:
@kleesc I reworded the commit so all checks would pass but now I need another approval 😅

kleesc (approved):
LGTM

PR #886, opened 2021-09-10
notification: check certs exists for webhooks (PROJQUAY-2424)

kleesc:
Make sure the Quay key/cert pair (ssl.key + ssl.cert) have been
mounted/exists whenever it tries sending a webhook request with client
cert. If that's not the case, then the webhook is sent without client
cert. This can happen when TLS is not handled by the Quay container,
in which case these certs are not required by default in the Quay
container.

### DIFF
--- a/notifications/notificationmethod.py
+++ b/notifications/notificationmethod.py
@@ -1,4 +1,5 @@
 import logging
+import os.path
 import re
 import json
 import mock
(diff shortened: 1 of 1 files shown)


notifications/notificationmethod.py
```diff
@@ -37,7 +38,9 @@ class NotificationMethodPerformException(JobException):
 
 def _ssl_cert():
     if app.config["PREFERRED_URL_SCHEME"] == "https":
-        return [OVERRIDE_CONFIG_DIRECTORY + f for f in SSL_FILENAMES]
+        cert_files = [OVERRIDE_CONFIG_DIRECTORY + f for f in SSL_FILENAMES]
+        cert_exists = all([os.path.isfile(f) for f in cert_files])
+        return cert_files if cert_exists else None
```
flavianmissi, line 43: Would it be useful to log an info or even warning when `PREFERRED_URL_SCHEME` is https but the certificate files aren't there?

flavianmissi (approved):
LGTM, with a minor question about logging.

notifications/notificationmethod.py
kleesc: I don't think so. Whether `PREFERRED_URL_SCHEME` is set and certificate files are missing or not are deployment specific, and aren't really unexpected behaviour. It would only create unnecessary noise every time a request is sent in my opinion.

flavianmissi: Alright, makes sense!

PR #887, opened 2021-09-14
Quay.io UI: Fetching severity from cvss score and removing visibility…

Sunandadadi:
… of unknown metrics (PROJQUAY-2541)

If vulnerability's severity is missing mapping severity from cvss score if present. Removed visibility if severity or distribition is Unknow

### DIFF
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -28,7 +28,7 @@ from data.registry_model import registry_model
 from util.migrate.allocator import yield_random_entries
 from util.secscan.validator import V4SecurityConfigValidator
 from util.secscan.v4.api import ClairSecurityScannerAPI, APIRequestFailure, InvalidContentSent
-from util.secscan import PRIORITY_LEVELS
+from util.secscan import PRIORITY_LEVELS, get_priority_from_cvssscore
 from util.secscan.blob import BlobURLRetriever
 from util.config import URLSchemeAndHostname
 
@@ -401,8 +401,8 @@ def features_for(report):
                 pkg["version"],
                 [
                     Vulnerability(
-                        vuln["normalized_severity"]
-                        if vuln["normalized_severity"]
+                        vuln["normalized_severity"] if vuln["normalized_severity"] and vuln["normalized_severity"] != PRIORITY_LEVELS["Unknown"]["value"]
+                        else get_priority_from_cvssscore(enrichments[vuln["id"]]["baseScore"]) if enrichments.get(vuln["id"], {}).get("baseScore", None)
                         else PRIORITY_LEVELS["Unknown"]["value"],
                         "",
                         vuln["links"],
--- a/util/secscan/__init__.py
+++ b/util/secscan/__init__.py
@@ -86,3 +86,19 @@ def get_priority_for_index(index):
             return priority
 
     return "Unknown"
+
+
+def get_priority_from_cvssscore(score):
+    try:
+        if 0 < score < 4:
+            return PRIORITY_LEVELS["Low"]["value"]
+        if 4 <= score < 7:
+            return PRIORITY_LEVELS["Medium"]["value"]
+        if 7 <= score < 9:
+            return PRIORITY_LEVELS["High"]["value"]
+        if 9 <= score < 10:
+            return PRIORITY_LEVELS["Critical"]["value"]
+    except ValueError:
+        return "Unknown"
+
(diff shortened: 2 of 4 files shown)


util/secscan/__init__.py
```diff
@@ -86,3 +86,19 @@ def get_priority_for_index(index):
             return priority
 
     return "Unknown"
+
+
+def get_priority_from_cvssscore(score):
```
kleesc, line 91: Is there an external reference to these mapping we can refer to?

kleesc:
You'll need to run `make black` to fix the formatting.

util/secscan/__init__.py
Sunandadadi: Yes, https://www.first.org/cvss/specification-document#Qualitative-Severity-Rating-Scale
Should I link this to the PR or in the code?

kleesc: Can you add it maybe as a comment, just so it's easier to find for future references.

Sunandadadi: Absolutely.

kleesc:
Since it's a UI fix, can you post images before/after the fix?

ibazulic:
I am *strictly* against this. We should *not* hide image metadata from the user in any way, shape or form. That is *not* the job of a registry. The registry's job is to store images and show correct image metadata when asked. It is up to the scanner to provide the security info which should then be shown *as is*, without modifications.

hdonnay:
@ibazulic if I'm reading this right, it's not "hiding" anything, but rather picking the most relevant information to display.

If a distributor doesn't assign a severity to their advisory, this displays the CVSS score with a note that it's additional information from the NVD. The alternative is much less useful: Alpine advisories never displaying a severity inline, but having one in an expando.

data/secscan_model/secscan_v4_model.py
```diff
@@ -401,8 +401,8 @@ def features_for(report):
                 pkg["version"],
                 [
                     Vulnerability(
-                        vuln["normalized_severity"]
-                        if vuln["normalized_severity"]
+                        vuln["normalized_severity"] if vuln["normalized_severity"] and vuln["normalized_severity"] != PRIORITY_LEVELS["Unknown"]["value"]
+                        else get_priority_from_cvssscore(enrichments[vuln["id"]]["baseScore"]) if enrichments.get(vuln["id"], {}).get("baseScore", None)
                         else PRIORITY_LEVELS["Unknown"]["value"],
```
hdonnay, line 406: This is getting a little hairy. I guess it makes sense you can chain ternary expressions like this, but it was unexpected.

hdonnay: Oh man, and the formatting that black wants to apply makes it even harder to read; I'd suggest pulling this bit of logic out.

Sunandadadi: Sure, I will pull it out.

ibazulic:
> @ibazulic if I'm reading this right, it's not "hiding" anything, but rather picking the most relevant information to display.
> 
> If a distributor doesn't assign a severity to their advisory, this displays the CVSS score with a note that it's additional information from the NVD. The alternative is much less useful: Alpine advisories never displaying a severity inline, but having one in an expando.

Yeah, it will map to CVSS score if it exists. What if it doesn't? If I am reading this correctly, those issues will not show. Again, it is not the registry's place to choose what info it will show and what it won't show. Furthermore, the fact that a vulnerability is marked as unknown does not mean that that the vulnerability is not dangerous. It is up to the image creator to decide what would be a proper way forward in that case by telling him or her "hey, we found these vulns, we have no clue what their severity is, but we are giving you the info so you can decide".

hdonnay:
> Yeah, it will map to CVSS score if it exists.

*And* if the "normalized" severity from clair is `Unknown`. If there's a severity specified, that's used. In the case that there's no severity and no CVSS score, the display is exactly the same. Nothing is hidden -- the text says the distributor severity is `Unknown` if the CVSS score is displayed as the topline, and if the severity is not `Unknown`, _that_ is displayed as the topline.

Sunandadadi:
Attaching screenshots of before and after the change:
Before: 
Screen Shot 2021-09-15 at 9 51 32 AM

After:
Screen Shot 2021-09-14 at 5 17 23 PM

Adding reference for get_priority_from_cvssscore method - https://www.first.org/cvss/specification-document#Qualitative-Severity-Rating-Scale

PR #889, opened 2021-09-15
dockerfile: Update symlink in upstream dockerfile (PROJQUAY-2550)

jonathankingfc:
- Fix symlink in upstream dockerfile


kleesc:
Fixes empty directory issue when mounting config bundle to  `/conf/stack` instead of `/quay-registry/conf/stack`

kleesc (approved):

PR #892, opened 2021-09-17
secscan: continue iterating after failure

crozzy:
If Clair returns an error the current behaviour is to
error out, thus not indexing any subsequent manifests.
This change allows the worker to continue indexing
subsequent manifests after one failure.


kleesc:
Good catch. I guess any unexpected response from Clair would block subsequent manifests to be indexed. And if that same manifest keeps showing up in the batch of candidate, it could possibly block indexing for other manifests.

kleesc (approved):

PR #895, opened 2021-09-17
secscan: continue iterating after failure (PROJQUAY-2563)

crozzy:
If Clair returns an error the current behaviour is to
error out, thus not indexing any subsequent manifests.
This change allows the worker to continue indexing
subsequent manifests after one failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### v3.5.8
+
+**Release Notes**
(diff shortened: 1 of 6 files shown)


crozzy:
@kleesc this could definitely do with a review as I needed to resolve conflicts in the Dockerfile after cherry-picking

kleesc (approved):
LGTM

PR #896, opened 2021-09-17
secscan: fix secscan api ApiRequestFailure test (PROJQUAY-2563)

kleesc:
Follow-up to 694fa2ac. Correctly set the expected value for the
pagination token when not blocking on a failed API request.


crozzy (approved):
ahh good one

PR #898, opened 2021-09-21
ci: pin oci test to a specific sha

harishsurf:
This pins the distribution-spec repo: https://github.com/opencontainers/distribution-spec to a specific sha
to prevent OCI CI failing because of a recent change in the source test: https://github.com/opencontainers/distribution-spec/pull/296/files 


kleesc:
For reference, this relates to the fallback behavior of cross mounting a blob without specifying its origin.
See https://github.com/quay/quay/blob/master/endpoints/v2/blob.py#L139

kleesc (approved):

PR #899, opened 2021-09-21
registry: update blob mount behaviour when from parameter is missing (PROJQUAY-2570)

kleesc:
Update the behaviour when attempting to mount existing blob whenever
the `from` parameter is missing. Currently, Quay throws an Bad Request
error whenever the `from parameter is missing. However, according to
the specs (Docker Registry and OCI dist-spec), we should be falling
back to the default upload behaviour.

References:
  - https://docs.docker.com/registry/spec/api/#pushing-an-image
  - https://github.com/opencontainers/distribution-spec/blob/main/spec.md#mounting-a-blob-from-another-repository
  - https://github.com/opencontainers/distribution-spec/pull/275/files

### DIFF
--- a/endpoints/v2/test/test_blob.py
+++ b/endpoints/v2/test/test_blob.py
@@ -77,21 +77,28 @@ def test_blob_caching(method, endpoint, client, app):
 
 
 @pytest.mark.parametrize(
-    "mount_digest, source_repo, username, expect_success",
+    "mount_digest, source_repo, username, include_from_param, expected_code",
     [
         # Unknown blob.
-        ("sha256:unknown", "devtable/simple", "devtable", False),
+        ("sha256:unknown", "devtable/simple", "devtable", True, 202),
+        ("sha256:unknown", "devtable/simple", "devtable", False, 202),
         # Blob not in repo.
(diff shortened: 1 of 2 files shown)


hdonnay (approved):
SGTM

endpoints/v2/test/test_blob.py
```diff
@@ -109,8 +116,9 @@ def test_blob_mounting(mount_digest, source_repo, username, expect_success, clie
     params = {
         "repository": "devtable/building",
         "mount": mount_digest,
-        "from": source_repo,
```
flavianmissi, line 112: should we cover both cases here? i.e. one test ensuring the blob is mounted correctly with `from`, and another ensuring it will also work without it.

flavianmissi (approved):
lgmt with a test suggestion

endpoints/v2/test/test_blob.py
kleesc: We are. `include_from_param` parameterize the test to include or not `from`.
See https://github.com/quay/quay/pull/899/files#diff-6015d1dddcac40a4d4acd470fe07b31e94e8ec22bf157d361b5ad67c136ec133R80

flavianmissi: ooh that flew right over my head, neat!

PR #903, opened 2021-09-27
email: fix org recovery link in email (PROJQUAY-2589)

kleesc:
Signin link was always linking ton quay.io in org recovery
emails. Updated to link to whatever the SERVER_HOSTNAME is.


harishsurf (approved):
lgtm!

PR #908, opened 2021-09-29
ldap: Lower trace level when USERS_DEBUG=1 (PROJQUAY-2540)

jonathankingfc:
- Lower trace_level from 2 to 1 when USERS_DEBUG=1 to prevent logging of user information

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -63,7 +63,7 @@ class LDAPConnection(object):
         self._conn = None
 
     def __enter__(self):
-        trace_level = 2 if os.environ.get("USERS_DEBUG") == "1" else 0
+        trace_level = 1 if os.environ.get("USERS_DEBUG") == "1" else 0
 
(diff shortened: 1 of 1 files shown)


ibazulic:
@jonathankingfc This will hinder our ability to resolve LDAP errors. From the documentation:

~~~
Possible values for `trace_level` are 0 for no logging, 1 for only logging the method calls with arguments, 2 for logging the method calls with arguments and the complete results and 9 for also logging the traceback of method calls.
~~~

We need those calls with arguments and the complete results so we know exactly where the problem lies. By only posting calls, we cannot determine whether the issue is in LDAP binding, search being wrong or something else entirely. We also have it in our docs that using that flag is not safe and should be removed immediately after debugging is done.

I would ask that this change be reverted.

ibazulic:
This PR also does not resolve the problem of catching user information as LDAP bind credentials are still stored in plain sight even when `trace_level=1`. This only affects the log verbosity.

PR #918, opened 2021-09-30
Quay: Fix for repo mirror Worker hanging (PROJQUAY-2256)

Sunandadadi:
Pushed two fixes here. Query fix where expiry time of the repo needs to be greater than the current time to be eligible to mirror the repo. retries is initialized so it is not undefined outside of the scope of the if block

### DIFF
--- a/data/model/repo_mirror.py
+++ b/data/model/repo_mirror.py
@@ -52,7 +52,7 @@ def get_eligible_mirrors():
         (RepoMirrorConfig.sync_start_date <= now)
         & (RepoMirrorConfig.sync_retries_remaining > 0)
         & (RepoMirrorConfig.sync_status == RepoMirrorStatus.SYNCING)
-        & (RepoMirrorConfig.sync_expiration_date >= now)
+        & (RepoMirrorConfig.sync_expiration_date <= now)
         & (RepoMirrorConfig.is_enabled == True)
     )
 
@@ -118,7 +118,9 @@ def release_mirror(mirror, sync_status):
     if sync_status == RepoMirrorStatus.FAIL:
         retries = max(0, mirror.sync_retries_remaining - 1)
 
-    if sync_status == RepoMirrorStatus.SUCCESS or ((retries is not None and retries < 1) or mirror.sync_retries_remaining < 1):
+    if sync_status == RepoMirrorStatus.SUCCESS or (
+        (retries is not None and retries < 1) or mirror.sync_retries_remaining < 1
+    ):
         now = datetime.utcnow()
         delta = now - mirror.sync_start_date
         delta_seconds = (delta.days * 24 * 60 * 60) + delta.seconds
--- a/data/model/repo_mirror.py
+++ b/data/model/repo_mirror.py
@@ -114,12 +114,12 @@ def release_mirror(mirror, sync_status):
     sync will be moved ahead as if it were a success. This is to allow a daily sync, for example, to
     retry the next day. Without this, users would need to manually run syncs to clear failure state.
     """
-    retries = None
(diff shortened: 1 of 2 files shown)


data/model/repo_mirror.py
```diff
@@ -52,7 +52,7 @@ def get_eligible_mirrors():
         (RepoMirrorConfig.sync_start_date <= now)
         & (RepoMirrorConfig.sync_retries_remaining > 0)
         & (RepoMirrorConfig.sync_status == RepoMirrorStatus.SYNCING)
-        & (RepoMirrorConfig.sync_expiration_date <= now)
+        & (RepoMirrorConfig.sync_expiration_date >= now)
         & (RepoMirrorConfig.is_enabled == True)
     )
```
thomasmckay, line 57: `sync_expiration_date <= now` is meant to find jobs that expired. If the expiration is in the future and it's in syncing state, it is not expired, right?

Sunandadadi: Got it, I missed this. I am reverting this change.

data/model/repo_mirror.py
```diff
@@ -114,10 +114,13 @@ def release_mirror(mirror, sync_status):
     sync will be moved ahead as if it were a success. This is to allow a daily sync, for example, to
     retry the next day. Without this, users would need to manually run syncs to clear failure state.
     """
+    retries = None
```
kleesc, line 117: Default this to 0 instead. Is it's `None` and the status is `FAIL`, the call to `max()` will throw an exception.

data/model/test/test_repo_mirroring.py
```diff
@@ -255,3 +255,21 @@ def test_repo_mirror_robot(initialized_db):
...
+    mirror = release_mirror(mirror, sync_status=RepoMirrorStatus.FAIL)
+    assert mirror.sync_retries_remaining == 2
+
+    mirror = release_mirror(mirror, sync_status=RepoMirrorStatus.FAIL)
+    assert mirror.sync_retries_remaining == 1
+
+    mirror = release_mirror(mirror, sync_status=RepoMirrorStatus.FAIL)
+    original_sync_start_date = mirror.sync_start_date
+    assert mirror.sync_retries_remaining == 3
+    assert mirror.sync_start_date == original_sync_start_date
```
kleesc, line 275: Not sure what this is testing? You assigned `original_sync_start_date = mirror.sync_start_date` a few lines above, so we would expect this to always hold anyways.

kleesc (changes requested):

data/model/repo_mirror.py
Sunandadadi: Yes, makes sense. I'll change the default initialization.

Sunandadadi: So, I am initializing 'retries' to `float('-inf')` so one can use the variable to differentiate between `RepoMirrorStatus.FAIL` and `RepoMirrorStatus.SUCCESS` and we do not run into exception on `max()`
Say, the `sync_status` passed to the method is `RepoMirrorStatus.FAIL` and defaulting `retries` to 0 allows the if block at https://github.com/quay/quay/blob/master/data/model/repo_mirror.py#L120 to execute which is not the behavior we want.

data/model/test/test_repo_mirroring.py
Sunandadadi: Yes my mistake, I meant to check `assert mirror.sync_retries_remaining == 3` for it before calling the function. I've pushed the changes.

Sunandadadi:
Closing this PR as the mirroring worker works fine with the latest code and without this change.

PR #920, opened 2021-10-01
superuser: add global readonly superuser (PROJQUAY-2604)

kleesc:
Adds the ability to grant existing superusers access to every
repository across the registry. This includes image pulls, api access,
and more. This has security implications and should only be used for
very specific use cases.

### DIFF
--- a/auth/permissions.py
+++ b/auth/permissions.py
@@ -24,6 +24,9 @@ _TeamNeed = partial(_TeamTypeNeed, "orgteam")
 _UserTypeNeed = namedtuple("userspecificneed", ["type", "username", "role"])
 _UserNeed = partial(_UserTypeNeed, "user")
 _SuperUserNeed = partial(namedtuple("superuserneed", ["type"]), "superuser")
+_GlobalReadOnlySuperUserNeed = partial(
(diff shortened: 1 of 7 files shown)


BillDett:
I think for an MVP this is fine, but having to define these users within Quay's config bundle seems pretty clunky.  Enterprise customers would expect to manage this through something like an LDAP group membership instead so you don't have to bounce the registry to add/remove users.  Maybe we can add another parameter which is the LDAP group name for these global r/o users which can be used instead of the allowlist of users (if LDAP is configured).

kleesc:
@BillDett Quay uses LDAP as an identity provider but not to manage its permissions. It really only cares about what resources a _Quay_ user has access to, and doesn't have awareness of what ldap group it is part of. 

Seems like what you're describing is more of a way to sync an LDAP group to the list of superusers, which I'm not sure is currently possible, since we don't persist that list. The superusers, and with this change, the superusers with readonly access, are loaded in memory at runtime, which is why it's required to restart the process after updating that list in `config.yaml`.

BillDett:
Right- I forgot that we use LDAP only for authentication, not authorization.  That's the proper way to do this I think- but definitely a much larger feature request.

PR #945, opened 2021-10-21
mirror: increased registry user/pass max length (PROJQUAY-2712)

kleesc:
Increases the maximum length of `external_registry_username` and
`external_registry_password` fields from `2048` to `4096`. Some registries,
e.g ECR, issue a token to be used as password, whose length is longer
than 2048 after being encrypted and base64 encoded. Since JWT's specs
do not specify a maximum length, we set the combined max length of
these fields to approximately 8K, the default header size on most
servers (including nginx), since these credentials will be passed in a
auth header during the auth flow.

### DIFF
--- a/data/migrations/versions/d94d328733e4_increase_repomirrorconfig_char_size.py
+++ b/data/migrations/versions/d94d328733e4_increase_repomirrorconfig_char_size.py
@@ -1,0 +1,45 @@
+"""Increase repomirrorconfig char size
+
+Revision ID: d94d328733e4
+Revises: b43b44996c26
+Create Date: 2022-04-12 14:16:07.660766
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = "d94d328733e4"
+down_revision = "b43b44996c26"
+
+import sqlalchemy as sa
+
+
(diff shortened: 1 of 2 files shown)


HammerMeetNail (approved):
LGTM 👍

data/migrations/versions/d94d328733e4_increase_repomirrorconfig_char_size.py
```diff
@@ -0,0 +1,45 @@
...
+
+Revision ID: d94d328733e4
+Revises: 909d725887d3
+Create Date: 2021-10-21 14:16:07.660766
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = "d94d328733e4"
+down_revision = "909d725887d3"
```
flavianmissi, line 11: I think the `down_revision` needs to be `e9f3e4dbb979` instead.

kleesc: @flavianmissi I updated it with the most recent revision.

flavianmissi (approved):

PR #947, opened 2021-10-22
Use /run for supervisord socket

dmage:
QUAYCONF may be on a file system that doesn't support unix sockets (for example, if it is mounted over network). /run seems to be a bit more traditional place for sockets.


flavianmissi (approved):
lgtm

PR #951, opened 2021-10-26
secscan: deprecate support for Clair V2 (PROJQUAY-2837)

kleesc:
Removes read support for Clair V2, along with the need to package
jwtproxy with Quay.

TODO: Drop deprecate image api + image table, remove image data model.


data/secscan_model/__init__.py
```diff
@@ -24,13 +23,8 @@ def configure(self, app, instance_keys, storage):
         except InvalidConfigurationException:
             self._model = NoopV4SecurityScanner()
 
-        try:
```
syed, line 27: Do we need the `SecurityScannerModelProxy` class? Can we just use `NoopV4SecurityScanner` instead?

kleesc: You mean `V4SecurityScanner`? `SecurityScannerModelProxy` was  just an interface to both Clair v2 and Clair v4.
Since we're deprecating v2, we don't strictly need it. It also doesn't make a difference to keep it. It's essentially just a wrapper around the v4 model at this point. Makes it easier to have separate implementations if needed though.

syed (approved):

PR #956, opened 2021-10-29
grafana: Adding Grafana dashboards (PROJQUAY-2723)

bcaton85:
Adding Grafana configuration files for graphing metrics of Quay deployments.


kleesc:
Dashboards seems fine to me. Not sure however about adding an AWS dashboard as part of the upstream repository, since it is specific to how quay.io is deployed, but that may not be the case for other deployments.

PR #967, opened 2021-11-10
blobuploadcleanupworker: Add cleanup for orphaned blobs (PROJQUAY-2313)

bcaton85:
Currently blobs leftover in the uploads directory during cancelled uploads do not get cleaned up since they are no longer tracked. This change cleans up the uploads storage directory directly.

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -3,7 +3,7 @@ import logging
 import copy
 
 from collections import namedtuple
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from io import BufferedIOBase, StringIO, BytesIO
 from itertools import chain
 from uuid import uuid4
@@ -695,6 +695,32 @@ class _CloudStorage(BaseStorageV2):
         # We have to go through and delete all of the uploaded chunks
         for chunk in self._chunk_list_from_metadata(storage_metadata):
             self.remove(chunk.path)
+
+    def clean_partial_uploads(self, deletion_date_threshold):
+        self._initialize_cloud_conn()
+        path = self._init_path("uploads")
+        paginator = self.get_cloud_conn().get_paginator(self._list_object_version)
+        for page in paginator.paginate(Bucket=self._bucket_name, Prefix=path):
+            for obj_info in page.get("Contents", []):
+                if obj_info["LastModified"] <= datetime.now(timezone.utc) - deletion_date_threshold:
+                    obj = self.get_cloud_bucket().Object(obj_info["Key"])
+                    try:
+                        obj.load()
+                        obj.delete()
+                        logger.debug(
+                            "Expired blob removed from uploads folder: %s", obj_info["Key"]
+                        )
+                    except botocore.exceptions.ClientError as s3r:
+                        if not s3r.response["Error"]["Code"] in _MISSING_KEY_ERROR_CODES:
+                            logger.exception(
+                                "Got error when attempting to clean blob with key in uploads folder: %s",
+                                obj_info["Key"],
+                                str(s3r),
+                            )
+                        else:
+                            logger.debug(
+                                "Blob not found in uploads folder with key %s", obj_info["Key"]
+                            )
 
 
 class S3Storage(_CloudStorage):
--- a/storage/distributedstorage.py
+++ b/storage/distributedstorage.py
@@ -64,6 +64,7 @@ class DistributedStorage(StoragePaths):
     validate = _location_aware(BaseStorage.validate, requires_write=True)
     get_checksum = _location_aware(BaseStorage.get_checksum)
     get_supports_resumable_downloads = _location_aware(BaseStorage.get_supports_resumable_downloads)
+    clean_partial_uploads = _location_aware(BaseStorage.clean_partial_uploads, requires_write=True)
 
     initiate_chunked_upload = _location_aware(
         BaseStorageV2.initiate_chunked_upload, requires_write=True
--- a/storage/test/test_cloud_storage.py
+++ b/storage/test/test_cloud_storage.py
@@ -1,10 +1,10 @@
 import os
+import time
 
 from io import BytesIO
 
 import pytest
 
-import moto
 import botocore.exceptions
 import boto3
 
@@ -15,11 +15,14 @@ from storage.cloud import _CloudStorage, _PartUploadMetadata
 from storage.cloud import _CHUNKS_KEY
 from storage.cloud import _build_endpoint_url
 
+from datetime import timedelta
+
 _TEST_CONTENT = os.urandom(1024)
 _TEST_BUCKET = "somebucket"
 _TEST_USER = "someuser"
 _TEST_PASSWORD = "somepassword"
 _TEST_PATH = "some/cool/path"
+_TEST_UPLOADS_PATH = "uploads/ee160658-9444-4950-8ec6-30faab40529c"
 _TEST_CONTEXT = StorageContext("nyc", None, None, None)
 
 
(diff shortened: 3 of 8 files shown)


storage/cloud.py
```diff
@@ -696,6 +696,28 @@ def cancel_chunked_upload(self, uuid, storage_metadata):
...
+                    obj = self.get_cloud_bucket().Object(obj_info["Key"])
+                    try:
+                        obj.load()
+                        obj.delete()
+                        logger.debug(
+                            "Expired blob removed from uploads folder: %s", obj_info["Key"]
+                        )
+                    except botocore.exceptions.ClientError as s3r:
+                        if not s3r.response["Error"]["Code"] in _MISSING_KEY_ERROR_CODES:
+                            logger.debug(
```
kleesc, line 715: `logger.exception` might be more appropriate here.

storage/cloud.py
```diff
@@ -696,6 +696,28 @@ def cancel_chunked_upload(self, uuid, storage_metadata):
         for chunk in self._chunk_list_from_metadata(storage_metadata):
             self.remove(chunk.path)
 
+    def clean_partial_uploads(self, deletion_date_threshold):
+        self._initialize_cloud_conn()
+        path = self._init_path("uploads")
```
kleesc, line 701: Does this need a trailing `/`?

storage/distributedstorage.py
```diff
@@ -64,6 +64,7 @@ def locations(self):
     validate = _location_aware(BaseStorage.validate, requires_write=True)
     get_checksum = _location_aware(BaseStorage.get_checksum)
     get_supports_resumable_downloads = _location_aware(BaseStorage.get_supports_resumable_downloads)
+    clean_partial_uploads = _location_aware(BaseStorage.clean_partial_uploads)
```
kleesc, line 67: This needs `requires_write=True` to be passed. Otherwise it will still run while in read-only mode.

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
```diff
@@ -28,13 +28,31 @@ def __init__(self):
...
+        if applicable.
         """
         try:
             with GlobalLock("BLOB_CLEANUP", lock_ttl=LOCK_TTL):
                 self._cleanup_uploads()
         except LockNotAcquiredException:
             logger.debug("Could not acquire global lock for blob upload cleanup worker")
-            return
+
+        if "FEATURE_CLEAN_UPLOADS_FOLDER" in app.config:
```
kleesc, line 40: You need to add this config flag in `config.py` (set it to default to `False`).
Also, any config flag prefixed with `FEATURE_` gets added to the global `features` object. So in this case it would look like:
```
if features.CLEAN_UPLOADS_FOLDER:
  ...
```

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
```diff
@@ -28,13 +28,31 @@ def __init__(self):
...
+            self._try_clean_partial_uploads()
```
kleesc, line 41: I think this should be in the same context as `BLOB_CLEANUP`, just to that it doesn't get preempted by another worker.

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
```diff
@@ -28,13 +28,31 @@ def __init__(self):
...
+
+    def _try_clean_partial_uploads(self):
+        """
+        Uploads cancelled before completion leaves the possibility of untracked blobs being
+        leftover in the uploads storage folder.
+        This function cleans those blobs older than DELETION_DATE_THRESHOLD
+        """
+        for location in storage.locations:
```
kleesc, line 49: I don't think it needs to check every storage. Just the one that it's writing to, so no need to iterate over every locations.

storage/cloud.py
```diff
@@ -696,6 +696,28 @@ def cancel_chunked_upload(self, uuid, storage_metadata):
         for chunk in self._chunk_list_from_metadata(storage_metadata):
             self.remove(chunk.path)
 
+    def clean_partial_uploads(self, deletion_date_threshold):
```
kleesc, line 699: I still think that this is something that should be handled by the storage service itself, and not Quay, since we don't really track partial uploaded blob storage, and we're essentially reimplementing a feature that's already available.

kleesc (changes requested):
Otherwise looks good.

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
kleesc: This should not be prefixed with `FEATURE_`, since it's a "subfeature" of the blobuploadcleanupworker.
Probably something just  `CLEAN_BLOB_UPLOAD_FOLDER` is enough, and then can be referred like:
```
if app.config.get("CLEAN_BLOB_UPLOAD_FOLDER", False):
   ...
```

storage/cloud.py
kleesc: Or at least not on by default.

storage/cloud.py
```diff
@@ -696,6 +696,28 @@ def cancel_chunked_upload(self, uuid, storage_metadata):
...
+        self._initialize_cloud_conn()
+        path = self._init_path("uploads")
+        paginator = self.get_cloud_conn().get_paginator(self._list_object_version)
+        for page in paginator.paginate(Bucket=self._bucket_name, Prefix=path):
+            for obj_info in page.get("Contents", []):
+                if obj_info["LastModified"] <= datetime.now(timezone.utc) - deletion_date_threshold:
+                    obj = self.get_cloud_bucket().Object(obj_info["Key"])
+                    try:
+                        obj.load()
```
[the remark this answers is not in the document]
bcaton85: This will check if the blob in the uploads folder has existed longer than `DELETION_DATE_THRESHOLD`, to prevent the clean up from deleting blobs in the uploads folder that are currently being used as a multi-part upload.

storage/cloud.py
bcaton85: Agreed

storage/cloud.py
bcaton85: I'm not actually seeing a trailing `/`, i believe i did have it in a previous commit but removed it afterword.

storage/distributedstorage.py
bcaton85: Yes good call. Added.

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
bcaton85: Understood. I updated the flag to be `CLEAN_BLOB_UPLOAD_FOLDER` and added `CLEAN_BLOB_UPLOAD_FOLDER = False` to config.py. I also updated the condition to use `if app.config.get("CLEAN_BLOB_UPLOAD_FOLDER", False):`

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
bcaton85: Due to the [_location_aware function in distributedstorage.py](https://github.com/bcaton85/quay/blob/gc-uploads/storage/distributedstorage.py#L11-L29) a location has to be passed in. Is there an easy way to get the current storage that's being written to?

storage/cloud.py
bcaton85: Yeah I agree it shouldn't be enabled by default. I can also see the usefulness in not having the operational overhead of setting up retention rules, something that works out of the box. We can discuss more with the rest of the team if you want.

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
kleesc: Just use the global `storage` object, which is already being imported at the top of the file.

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
kleesc: Under the `with GlobalLock("BLOB_CLEANUP", lock_ttl=LOCK_TTL)`

storage/cloud.py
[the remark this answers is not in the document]
kleesc: How does that work, since delete called immediately after?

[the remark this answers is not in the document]
bcaton85: To clarify, which call are you referring to? the `obj.load()`?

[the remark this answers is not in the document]
kleesc: Yes, isn't that just doing a HEAD request?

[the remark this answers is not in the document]
bcaton85: Correct, the remove logic is [ported over from the remove method](https://github.com/bcaton85/quay/blob/gc-uploads/storage/cloud.py#L424-L429) which also includes the `obj.load()`. It's used to check if the object exists through [the info provided in the metadata, this stackoverflow has more info.](https://stackoverflow.com/questions/33842944/check-if-a-key-exists-in-a-bucket-in-s3-using-boto3)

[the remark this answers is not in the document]
bcaton85: [It's only used to adjust the error output to make it a bit cleaner.](https://github.com/bcaton85/quay/blob/gc-uploads/storage/cloud.py#L714-L721)

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
bcaton85: Resolved offline, will pass `storage.preferred_locations` instead of lopping over `storage.locations`

storage/test/test_cloud_storage.py
```diff
@@ -315,3 +318,28 @@ def test_rechunked(max_size, parts):
...
+    assert storage_engine.exists(_TEST_UPLOADS_PATH)
+    assert storage_engine.get_content(_TEST_UPLOADS_PATH) == _TEST_CONTENT
+
+    # Test ensure fresh blobs are not deleted
+    storage_engine.clean_partial_uploads(timedelta(days=2))
+    assert storage_engine.exists(_TEST_UPLOADS_PATH)
+    assert storage_engine.get_content(_TEST_UPLOADS_PATH) == _TEST_CONTENT
+
+    # Test deletion of stale blobs
+    time.sleep(1)
```
kleesc, line 338: You could've also monkey patched the `LastModified` field from the response. But this should be fine for now.

kleesc (approved):

kleesc (approved):
LGTM

PR #971, opened 2021-11-11
build: Use a configtool tag in the Dockerfile instead of master (PROJ…

syed:
…QUAY-2777)

Having `master` as the config tool version causes issues where Docker
caches the layer. When there are updates to config-tool, docker won't
pull the new changes as it already has a cached version


jonathankingfc (commented):
LGTM

jonathankingfc (approved):

PR #995, opened 2021-11-25
(3.5 backport) build: Add required setup.cfg for downstream build (PROJQUAY-2713)

flavianmissi:
Downstream build uses Cachito which needs setup.cfg for
installing python dependencies

originally added on redhat-3.7, see https://github.com/quay/quay/pull/946


harishsurf (approved):
lgtm!

PR #1003, opened 2021-11-29
nginx: support client ip through x-forwarded-for header (PROJQUAY-2883)

kleesc:
Support client ip from x-forwarded header. This is needed if one of
the intermediate proxy between the client and Quay does not support
proxy protocol.

If set, FEATURE_PROXY_PROTOCOL will use the L4 feature. If not, say,
when a L7 proxy is used, fallback to using x-forwarded-for, and allow
specifying trusted ip/cidr for Nginx.

### DIFF
--- a/conf/init/nginx_conf_create.py
+++ b/conf/init/nginx_conf_create.py
@@ -124,6 +124,19 @@ def generate_rate_limiting_config(config):
     )
 
 
+def generate_http_base_config(config):
+    """
+    Generates http base config from the app config.
+    """
+    config = config or {}
+    trusted_proxy_cidr = config.get("TRUSTED_PROXY_CIDR", None)
+
+    write_config(
+        os.path.join(QUAYCONF_DIR, "nginx/http-base.conf"),
+        trusted_proxy_cidr=trusted_proxy_cidr,
+    )
+
+
 def generate_hosted_http_base_config(config):
     """
     Generates hosted http base config from the app config.
@@ -144,6 +157,7 @@ if __name__ == "__main__":
     else:
         config = None
 
+    generate_http_base_config(config)
     generate_hosted_http_base_config(config)
     generate_rate_limiting_config(config)
     generate_server_config(config)
--- a/conf/nginx/http-base.conf
+++ b/conf/nginx/http-base.conf
@@ -1,6 +1,12 @@
 # vim: ft=nginx
 
+{% if not trusted_proxy_cidr %}
 set_real_ip_from 0.0.0.0/0;
+{% else %}
+{% for proxy_ip in trusted_proxy_cidr %}
+set_real_ip_from {{ proxy_ip }}
+{% endfor %}
+{% endif %}
 real_ip_recursive on;
 log_format lb_logs '$remote_addr ($proxy_protocol_addr) '
                    '- $remote_user [$time_local] '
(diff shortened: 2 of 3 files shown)


conf/init/nginx_conf_create.py
```diff
@@ -124,6 +124,21 @@ def generate_rate_limiting_config(config):
     )
 
 
+def generate_http_base_config(config):
+    """
+    Generates http base config from the app config.
+    """
+    config = config or {}
+    feature_proxy_protocol = config.get("FEATURE_PROXY_PROTOCOL", False)
+    trusted_proxy_ips = config.get("TRUSTED_PROXY_IPS", None)
```
syed, line 133: This would be a CIDR. Maybe rename to `TRUSTED_PROXY_CIDRS`

conf/nginx/nginx.conf.jnj
```diff
@@ -36,6 +36,8 @@ http {
         # This header must be set only for HTTPS
         add_header Strict-Transport-Security "max-age=63072000; preload";
 
+        real_ip_header X-Forwarded-For;
```
syed, line 39: We will also need to patch the `7443` port as well. That's the one being used by the ELB endpoint. So I'm not sure what effect it's going to be when we remove `proxy_protocol` and replace it with `X-Forwarded-For` some proxies support the proxy protocol header

conf/init/nginx_conf_create.py
```diff
@@ -124,6 +124,21 @@ def generate_rate_limiting_config(config):
     )
 
 
+def generate_http_base_config(config):
+    """
+    Generates http base config from the app config.
+    """
+    config = config or {}
+    feature_proxy_protocol = config.get("FEATURE_PROXY_PROTOCOL", False)
```
syed, line 132: proxy protocol is always enabled. The change we want to make is to see which header is our source of truth for the IP chain, I don't think we ever need to disable this

syed (changes requested):

conf/init/nginx_conf_create.py
kleesc: Proxy protocol is not always enabled. If a load balancer doesn't support it, you should not be using Quay's  port dedicated for it (7443). I don't think that would work at all. X-Forwarded-For is done through HTTP header, while proxy protocol is TCP metadata that nginx knows how process.

conf/nginx/nginx.conf.jnj
kleesc: We'll likely need to turn it off on ELB, as nginx probably won't accept connections using a non proxy protocol server directive.

conf/init/nginx_conf_create.py
kleesc: Done. Although I think regular ip notation also works.

conf/nginx/http-base.conf
```diff
@@ -1,6 +1,12 @@
 # vim: ft=nginx
 
+{% if feature_proxy_protocol or not trusted_proxy_ips %}
```
syed, line 3:
```suggestion
{% if feature_proxy_protocol or not trusted_proxy_cidr %}
```

syed (dismissed):

syed (approved):

PR #1006, opened 2021-12-01
deploy: Add LB service with no proxy-protocol (PROJQUAY-2883)

syed:
This is required to support ALB. Without this service, the
default ELB endpoints forces the L4 proxy procotol which makes
Nginx see all the remote client IP as the IP of ALB. If the
request goes through an ELB without proxy-protocol enabled,
Nginx will parse the X-Forwarded-For header to get the remote
IP


kleesc (approved):
lgtm

PR #1008, opened 2021-12-01
Set QUAYRUN default to be backward compatible

kleesc:
This reverts commit 7fe6d506a81e9d74a09586c165436ed7a3d736c4.

The default user doesn't have permission to write in `/run`, which is why it was being written to `QUAYCONF`


dmage:
By reverting this PR you'll again break Quay for me. This isn't a collaborative work. Please address the problem instead of reverting the fix.

dmage (changes requested):
We need to make /run writable for the default user and either add mandatory approvers for Dockerfiles or create a CI test that can show that the image is broken.

syed:
@dmage  trying to understand why the was not working for you? Currently it's crashing in staging with the error

```
Error: Cannot open an HTTP server: socket.error reported errno.EACCES (13)
```

I'd like to merge this in so we are not blocked

syed (approved):

kleesc:
@dmage We generally don't support network attached storage. I've re-added the changes, but removed the default exported value for the variable you introduced, as I don't think it should be default. The previous changes as is were breaking our current deployment. Users who, for whatever reasons needs to mount their config bundle over network storage can still do so by explicitly setting the ENV var.

syed (approved):

PR #1011, opened 2021-12-01
oauth: Fix OAuth authorization code grant type (PROJQUAY-781)

dmage:
AuthorizationProvider uses explicit parameter names when it calls its methods. Therefore methods of derivative classes, i.e. the methods of the class DatabaseAuthorizationProvider, should have the same parameter names.

### DIFF
--- a/data/model/oauth.py
+++ b/data/model/oauth.py
@@ -108,9 +108,9 @@ class DatabaseAuthorizationProvider(AuthorizationProvider):
         # Make sure the token contains the given scopes (at least).
         return scopes.is_subset_string(long_scope_string, scope)
 
-    def from_authorization_code(self, client_id, full_code, scope):
-        code_name = full_code[:AUTHORIZATION_CODE_PREFIX_LENGTH]
-        code_credential = full_code[AUTHORIZATION_CODE_PREFIX_LENGTH:]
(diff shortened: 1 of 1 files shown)


dmage:
@syed is there a way to create an e2e test for the /oauth/authorize endpoint? I cannot find e2e tests in the repo.

syed:
@dmage it's going to be a bit of work. We currently only support RH SSO so we'll have to first request for a new client ID for the CI and configure it. We will also have to make sure that the CI environment gets a stable/static endpoint for OAUTH redirection to work. Since RHSSO is essentially running Keycloak, we could run an instance ourselves with the infra this would be an easier way but we would still need to spin up and configure Keycloak for testing SSO

dmage:
@syed you are talking about Quay as OAuth client, right? This PR is about Quay as an OAuth provider.

syed:
Aah sorry this is for when people create an oauth app for quay tasks. I think the challenge here was the manual creation of the oauth application. If we can automate that (maybe run it via a browser emulator)  we can have an e2e test for this

dmage:
Oh, in this case I think we can add tests later.

syed:
@dmage  yeah I think this is good to merge. We already have testers lined up to test this out :)

syed (approved):

PR #1022, opened 2021-12-07
blobuploadcleanupworker: Add  BLOBUPLOAD_DELETION_DATE_THRESHOLD (PROJQUAY-2915)

bcaton85:
Added the BLOBUPLOAD_DELETION_DATE_THRESHOLD internal property to configure the TTL on orphaned blobs.


flavianmissi (dismissed):
LGTM

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
```diff
@@ -15,7 +15,7 @@
 
 logger = logging.getLogger(__name__)
 
-DELETION_DATE_THRESHOLD = timedelta(days=2)
+DELETION_DATE_THRESHOLD = timedelta(hours=12)
```
kleesc, line 18: Can we make this configurable instead. Something like `BLOB_UPLOAD_DELETION_DATE_THRESHOLD`.
This way we can keep the current value as default, but allow it to be shorter or longer for others.

kleesc (changes requested):

workers/blobuploadcleanupworker/blobuploadcleanupworker.py
bcaton85: Added

kleesc (approved):
LGTM

PR #1033, opened 2021-12-13
healthcheck: add individual endpoints for global services (PROJQUAY-2870)

kleesc:
The status (or endtoend) health endpoint sets its status code to 503
if any of the services being checked is unavailable.

This change adds individual endpoint for each services. This allows a
service like Pingdom to ignore something less critical, like Redis.

Add cache status to instance or global health check, depending on the
endpoint used.

### DIFF
--- a/endpoints/web.py
+++ b/endpoints/web.py
@@ -62,7 +62,7 @@ from endpoints.decorators import (
     parse_repository_name,
     param_required,
 )
-from health.healthcheck import get_healthchecker
+from health.healthcheck import get_healthchecker, get_healthservices
 from util.cache import no_cache
 from util.headers import parse_basic_auth
 from util.invoice import renderInvoiceToPdf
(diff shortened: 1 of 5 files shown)


endpoints/web.py
```diff
@@ -317,6 +317,36 @@ def endtoend_health():
...
+        "%s_service_health" % service,
+        service_health_func,
+        methods=["GET"],
+    )
+
+
+for service in get_healthservices():
+    _register_healthservice(service)
+
+
```
geoberle, line 349: elegant ❤️

endpoints/web.py
```diff
@@ -317,6 +317,36 @@ def endtoend_health():
...
+        "%s_service_health" % service,
+        service_health_func,
+        methods=["GET"],
+    )
+    web.add_url_rule(
+        "/health/endtoend/%s" % service,
+        "%s_service_health" % service,
+        service_health_func,
+        methods=["GET"],
+    )
```
geoberle, line 343: what is the reason the same service_health_func is registered for two urls?

geoberle:
LGTM
thank you for those changes. this makes monitoring quay a lot easier

geoberle:
hi @kleesc, @syed - any plans to proceed with this? this would still be very valuable!

PR #1039, opened 2021-12-17
clair: cache API requests (PROJQUAY-0)

hdonnay:
This is an attempt at having the Clair client cache responses in redis.

### DIFF
--- a/config.py
+++ b/config.py
@@ -472,6 +472,10 @@ class DefaultConfig(ImmutableConfig):
     # The endpoint for the V4 security scanner.
     SECURITY_SCANNER_V4_ENDPOINT: Optional[str] = None
 
+    # A `Redis URL<https://redis-py.readthedocs.io/en/stable/connections.html#redis.client.Redis.from_url>`_ for caching
+    # security scanner API requests.
+    SECURITY_SCANNER_V4_CACHE_URL: Optional[str] = None
+
     # The number of seconds between indexing intervals in the security scanner
     SECURITY_SCANNER_INDEXING_INTERVAL = 30
 
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -5,7 +5,9 @@ from collections import namedtuple
 from datetime import datetime, timedelta
 from math import log10
 from peewee import fn, JOIN
-from enum import Enum
+from redis import Redis
+from requests import Session
+from requests_cache import CacheMixin, RedisCache
 
 from data.secscan_model.interface import SecurityScannerInterface, InvalidConfigurationException
 from data.secscan_model.datatypes import (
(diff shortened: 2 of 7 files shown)


hdonnay:
For more context, having quay cache responses allows for reducing the total number of requests made to clair and any of clair's downstream dependencies. I would expect decent latency gains on small to medium installs with redis' LFU mode.

This change may be of questionable use for a large quay instance against a single redis instance, but would still be useful against a sharded redis instance.

config.py
```diff
@@ -472,6 +472,10 @@ def create_transaction(db):
     # The endpoint for the V4 security scanner.
     SECURITY_SCANNER_V4_ENDPOINT: Optional[str] = None
 
+    # A `Redis URL<https://redis-py.readthedocs.io/en/stable/index.html#redis.Redis.from_url>`_ for caching security
+    # scanner API requests.
+    SECURITY_SCANNER_V4_CACHE_URL: Optional[str] = None
```
kleesc, line 477: Probably needs more than just the URL. SSL, password, maybe db are the ones I'm thinking of.
For other components using the Redis client, we pass the objects as is. For example:
```
BUILDLOGS_REDIS:
  host: <some endpoint>
  password: <some password>
  ssl: true
```

See `BUILDLOGS_REDIS` or `USER_EVENTS_REDIS` or `DATA_MODEL_CACHE_CONFIG`.

data/secscan_model/secscan_v4_model.py
```diff
@@ -112,9 +115,17 @@ def __init__(self, app, instance_keys, storage):
             logger.warning(msg)
             raise InvalidConfigurationException(msg)
 
+        client = app.config.get("HTTPCLIENT")
+        cache_url = app.config.get("SECURITY_SCANNER_V4_CACHE", None)
+        if cache_url is not None:
+            c = Redis.from_url(cache_url)
+            backend = RedisCache(connection=c)
+            mixin = CacheMixin(backend=backend, cache_control=True)
```
kleesc, line 123: Is this caching everything including non-200s?

kleesc: `cache_control` uses headers to set expiration. Is that expected to be set from the responses from Clair?

hdonnay: It's only caching 200 by default, and yes Clair sets `Cache-Control` headers.

util/config/schema.py
```diff
@@ -723,6 +723,12 @@
             "description": "The endpoint for the V4 security scanner",
             "x-example": "http://192.168.99.101:6060",
         },
+        "SECURITY_SCANNER_V4_CACHE_URL": {
+            "type": ["string", "null"],
+            "pattern": "^(rediss?|unix)://.+$",
```
Sunandadadi, line 728: I think there is an extra "s" in the pattern for "redis" here.

Sunandadadi, line 728: Just trying to understand the regex for the pattern. Can there be 2 consecutive "s" here? Or can it be just `(redis|unix)`?

hdonnay: yeah, there's `redis` (plaintext) and `rediss` (tls). It's weird but in the [IANA registry](https://www.iana.org/assignments/uri-schemes/prov/rediss).

kleesc:
If cache is based off a header returned in the response, does this behaviour need to documented as part of Clair's API, since the consumer (Quay in this case), is not the one defining on the caching policy used.

config.py
hdonnay: The redis-py documentation location changed since I wrote that, but according to the [docs](https://redis-py.readthedocs.io/en/stable/connections.html#redis.client.Redis.from_url) all those things can be specified in the URL

kleesc:
What's the behavior when the cache backend is unavailable. Will `requests-cache` fallback to using regular `requests`?

hdonnay:
> What's the behavior when the cache backend is unavailable. Will `requests-cache` fallback to using regular `requests`? Otherwise LGTM

Asking upstream. I would _hope_ it acts just like there's no caching configured, but I couldn't tell immediately when reading the code.

hdonnay:
Asked [here](https://github.com/reclosedev/requests-cache/issues/509). Looks like I'll have to do some additional work around the client.

hdonnay:
I thought #2245 closed this via magic, but I guess not.

PR #1044, opened 2021-12-23
auth: Fix oauth code flow (PROJQUAY-781)

syed:
OAuth Authorization Code Flow is broken in quay. Code
Flow is more secure than implicit flow and is used
by server side applications to get the access token


HammerMeetNail (approved):
LGTM 👍

PR #1048, opened 2022-01-06
QuotaManagement: Reporting (PROJQUAY-2936)

kwestpharedhat:


data/migrations/versions/e9f3e4dbb979_quota_management_reporting.py
```diff
@@ -0,0 +1,93 @@
...
+        sa.Column("percent_of_limit", sa.Integer, nullable=False),
+        sa.ForeignKeyConstraint(["quota_type_id"], ["quotatype.id"], name=op.f("fk_quotalimit_type")),
+        sa.ForeignKeyConstraint(["quota_limit_group_id"], ["quotalimitgroups.id"], name=op.f("fk_quotalimit_limit_group"))
+    )
+
+    op.create_index(
+        "quotalimits_limitgroupid", "quotalimits", ["quota_limit_group_id"], unique=False
+    )
+
+    op.create_table(
```
kleesc, line 69: Can we split this work into multiple PRs: first the reporting part (i.e anything that deals with the calculations of repo size/limits) and the actual soft//hard quota part.

Any quota implementation will need a working reporting api first anyways, and this will make it easier to track/review.

kleesc: Ideally the first PR would only include the repositorysize table, since that's the only one needed for the reporting implementation.

kwestpharedhat: Ideally for the reporting portion we would be able to report how much is in use but also how much is "allowed" and how much of the quota that has been set is being used. I think it best to keep the entire schema and provide the reporting across all of the features. The code in v2 push is currently a placeholder for a quick easy way to learn and execute the code, I will add it to the v1 reporting endpoints soon.

data/database.py
```diff
@@ -755,6 +755,26 @@ class RobotAccountToken(BaseModel):
...
+    quota_limit_group_id = ForeignKeyField(QuotaLimitGroups)
+    quota_type_id = ForeignKeyField(QuotaType)
+    percent_of_limit = IntegerField(default=0)
+
+
+class UserOrganizationQuota(BaseModel):
+    organization_id = QuayUserField(index=True, unique=True)
+    limit_bytes = BigIntegerField()
+    quota_limit_group_id = ForeignKeyField(QuotaLimitGroups)
+
```
kleesc, line 776: What are quota groups, and why do we need the `QuotaLimit` table? Seems like it's tracking the amount of storage in terms of percentage that an organization is using, but couldn't that be part of the same table as `UserOrganizationQuota`? Or even be computed from the `RepositorySize` table.

data/model/_basequery.py
```diff
@@ -75,6 +80,29 @@ def get_existing_repository(namespace_name, repository_name, for_update=False, k
     return query.get()
 
 
+def get_existing_namespace_size(namespace_name):
```
kleesc, line 83: Speaking of existing namespaces, we will need some sort of backfill, or check to make sure that existing repositories with content that doesn't yet have their row in the `RepositorySize` be populated.

kleesc: It could be something like, on pushes, if the corresponding row doesn't exist, compute the size first. This would allow the push to proceed with updating the running count. 

This would make it easier avoid any possible drifts in the repository size, as opposed if it was done in a separate worker process.

kwestpharedhat: I agree that on push we need to calculate first if a repository size does not exist then follow up with the running size. When it comes to generic reporting we will fall back to calculating from the storage table if we do not have a repo size record.

data/database.py
kwestpharedhat: I think we can remove quota groups the more I work on this and will continue to think on it for a bit. The quota limit table allows for a limit to be of type warning at 80% usage and of type error at 100% usage and provides access to other types of limits if we want to implement them at a later date.

data/database.py
```diff
@@ -951,6 +971,11 @@ class RepositorySearchScore(BaseModel):
     last_updated = DateTimeField(null=True)
 
 
+class RepositorySize(BaseModel):
+    repository = ForeignKeyField(Repository, unique=True)
+    size_bytes = BigIntegerField()
+
```
kleesc, line 977: Would it be useful to add a timestamp? Thinking if we need to ever need to recalculate this value, the last time it was updated should also be relevant. In most case, we shouldn't need to use a timestamp though, since we're keeping a running count on push.

data/registry_model/quota_registry_model.py
```diff
@@ -0,0 +1,48 @@
+from data.model import organization, repository, notification, namespacequota
```
kleesc, line 1: Just a nit, but can we move this to `data/model/quota.py`, `registry_model/` should contain only registry specific ops.

data/model/user.py
```diff
@@ -1,12 +1,12 @@
-import bcrypt
-import logging
 import json
+import logging
```
kleesc, line 2: Are these changes to this file just tiding by your formatter?

data/model/repository.py
```diff
@@ -672,3 +675,48 @@ def mark_repository_for_deletion(namespace_name, repository_name, repository_gc_
         callback(namespace_name, repository_name)
 
     return marker.id
+
+
+def get_repository_size_and_cache(repo_id: int):
+    query = (
+        RepositorySize.select(Repository.name, Repository.id, RepositorySize.size_bytes)
```
kleesc, line 682: You should just be able to use `select()`, unless you think there's a performance gain to selecting the columns explicitly.

data/model/repository.py
```diff
@@ -672,3 +675,48 @@ def mark_repository_for_deletion(namespace_name, repository_name, repository_gc_
...
+        .join(Repository)
+        .where(Repository.id == repo_id)
+    ).get()
```
kleesc, line 685: You will need to wrap this query in a try block and catch the NotExistException from peewee.

data/model/repository.py
```diff
@@ -672,3 +675,48 @@ def mark_repository_for_deletion(namespace_name, repository_name, repository_gc_
...
+
+    # TODO: this will likely fail if we have to force cache
+    return {
+        "repository_name": query.repository.name,
+        "repository_id": query.repository.id,
+        "repository_size": query.size_bytes,
+    }
+
+
+def get_size_during_upload(repo_id: int):
```
kleesc, line 698: Not sure I see where this is being used.

data/model/repository.py
```diff
@@ -672,3 +675,48 @@ def mark_repository_for_deletion(namespace_name, repository_name, repository_gc_
...
+
+    if query is None:
+        force_cache_repo_size(repo_id)
+
+    # TODO: this will likely fail if we have to force cache
+    return {
+        "repository_name": query.repository.name,
+        "repository_id": query.repository.id,
+        "repository_size": query.size_bytes,
+    }
```
kleesc, line 695: This doesn't match your type hint signature (int). Either return just the int, or the `query` itself if you need access to the parent repo.

data/model/repository.py
```diff
@@ -672,3 +676,60 @@ def mark_repository_for_deletion(namespace_name, repository_name, repository_gc_
...
+
+    size_bytes = size["repository_size"] + query.size_bytes
+
+    return size_bytes
+
+
+def force_cache_repo_size(repo_id: int):
+    try:
+        cache = (
+            Manifest.select(fn.Sum(Manifest.layers_compressed_size).alias("size_bytes")).where(
```
kleesc, line 719: You don't really need the alias.

data/model/repository.py
```diff
@@ -672,3 +676,60 @@ def mark_repository_for_deletion(namespace_name, repository_name, repository_gc_
...
+                Manifest.repository == repo_id
+            )
+        ).get()
```
kleesc, line 722: If getting the size, instead of using `get`, you can use peewee's `scalar`, which will return the actual scalar value, or None of no rows match.

data/model/repository.py
```diff
@@ -672,3 +676,60 @@ def mark_repository_for_deletion(namespace_name, repository_name, repository_gc_
...
+
+        size = cache.size_bytes
+    except Manifest.DoesNotExist:
+        size = get_size_during_upload(repo_id)
```
kleesc, line 726: This should either be 0 if the repo is empty, or the size of the already uploaded manifests. Wouldn't this be an issue for partially updated manifests (where blobs were uploaded but not the manifest)?

Dockerfile
```diff
@@ -179,6 +179,8 @@ RUN set -ex\
 
 RUN rm -Rf node_modules config_app/node_modules
 
+ENV PYDEV_DEBUG=172.17.0.1:5555
```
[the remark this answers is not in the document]
kwestpharedhat: correct

data/database.py
kwestpharedhat: Even without a timestamp we can easily recalculate without much issue at this point I suggest we leave it off.

kwestpharedhat: Yes the separate table is to help avoid adding a column and creating issues with migration

data/registry_model/quota_registry_model.py
```diff
@@ -0,0 +1,53 @@
...
+
+    for admin in admins:
+        notification.create_notification(notification_kind, admin, metadata)
+
+
+def __get_namespace_limits(namespace_name):
+    return namespacequota.get_namespace_limits(namespace_name)
+
+
+def force_cache_repo_size(repository_ref):
```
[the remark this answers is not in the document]
kwestpharedhat: This in essence just an overload to adjust the parameter type for this scope and the information that is typically given to it. Its not necessary and could easily be removed if we have any concerns

endpoints/v2/blob.py
```diff
@@ -332,6 +343,18 @@ def upload_chunk(namespace_name, repo_name, upload_uuid):
...
+        quota = quota_registry_model.verify_namespace_quota_during_upload(
+            namespace_name, repository_ref
+        )
+        if quota["severity_level"] == 2:
+            return Response(
+                status=413,
+                headers={
+                    "Range": _render_range(0),
+                },
+            )
```
kleesc, line 356: Shouldn't this check happen before anything else?
Looks like this check is needed in multiple places. You could wrap it in a decorator and apply it to the relevant endpoint. Similar to say, `@require_repo_write`

endpoints/v2/blob.py
```diff
@@ -332,6 +343,18 @@ def upload_chunk(namespace_name, repo_name, upload_uuid):
     # Upload the chunk for the blob.
     _upload_chunk(uploader)
 
+    if app.config.get("FEATURE_QUOTA_MANAGEMENT", False):
+        quota = quota_registry_model.verify_namespace_quota_during_upload(
+            namespace_name, repository_ref
+        )
+        if quota["severity_level"] == 2:
+            return Response(
+                status=413,
```
kleesc, line 352: I think either a 403 or 433 might be more appropriate here. 413 would indicate something wrong with the request itself, which is not the case.

kleesc: Any reason not using `QuotaExceeded` that you defined?

endpoints/v2/errors.py
```diff
@@ -144,6 +144,13 @@ def __init__(self, uploaded=None, max_allowed=None):
             )
 
 
+class QuotaExceeded(V2RegistryException):
+    def __init__(self):
+        super(QuotaExceeded, self).__init__(
+            "QuotaExceeded", "Quota has been exceed on namespace", {}
+        )
```
kleesc, line 151: I think as it is this will return a 400. I think a 403 or 422 is more appropriate for this use case.

endpoints/v2/manifest.py
```diff
@@ -399,6 +404,16 @@ def _write_manifest(namespace_name, repo_name, tag_name, manifest_impl):
...
 
+    if app.config.get("FEATURE_QUOTA_MANAGEMENT", False):
+        quota = quota_registry_model.verify_namespace_quota_force_cache(
+            namespace_name, repository_ref
+        )
+        if quota["severity_level"] == 1:
+            quota_registry_model.notify_organization_admins("quota_warning", namespace_name)
+        elif quota["severity_level"] == 2:
+            quota_registry_model.notify_organization_admins("quota_error", namespace_name)
+            raise QuotaExceeded()
```
kleesc, line 415: Ditto. This looks like the same check as the other endpoint.  Can probably factor this out into a single decorator for all endpoints.

data/database.py
```diff
@@ -755,6 +756,26 @@ class RobotAccountToken(BaseModel):
     fully_migrated = BooleanField(default=False)
 
 
+class QuotaLimitGroups(BaseModel):
+    group_name = CharField()
```
kleesc, line 760: I'm still not sure what "LimitGroups" are needed for.

data/database.py
```diff
@@ -755,6 +756,26 @@ class RobotAccountToken(BaseModel):
...
+
+
+class QuotaType(BaseModel):
+    description = CharField()
```
kleesc, line 764: Shouldn't this column be named "name" instead.
It will also need to be prepopulated with possible values as part of the migration.

data/database.py
```diff
@@ -755,6 +756,26 @@ class RobotAccountToken(BaseModel):
...
+
+
+class QuotaLimits(BaseModel):
+    quota_limit_group_id = ForeignKeyField(QuotaLimitGroups)
+    quota_type_id = ForeignKeyField(QuotaType)
```
kleesc, line 769: There's a wrapper for `ForeignKeyField`, `EnumField` that adds some nice-to-have methods to it,

data/database.py
kwestpharedhat: This makes sense to rename and it is already being prepopulated by the alembic upgrade.

data/database.py
kwestpharedhat: If there isnt any chance for unnecessary overhead we can swap them out but for my purposes this has everything I need.

data/database.py
kwestpharedhat: They were an abstraction I thought I needed to start with but as development went on I realized they were unnecssary they have been refactored out of the code.

data/model/namespacequota.py
```diff
@@ -0,0 +1,287 @@
...
+            )
+    else:
+        notification.create_notification(
+            notification_kind,
+            repository_ref.namespace_user,
+            {"namespace": repository_ref.namespace_user},
+        )
+
+
+def __get_namespace_limits(namespace_name):
```
dmage, line 81: why do you need this function?

data/model/namespacequota.py
```diff
@@ -0,0 +1,287 @@
...
+    return check_limits(namespace_name, namespace_size)
+
+
+def verify_namespace_quota_force_cache(namespace_name, repository_ref):
+    force_cache_repo_size(repository_ref)
+    namespace_size = get_namespace_size(namespace_name)
+    return check_limits(namespace_name, namespace_size)
+
+
+def verify_namespace_quota_during_upload(namespace_name, repository_ref):
```
dmage, line 38: Why do you count BlobUploads size only during upload?

kwestpharedhat: We want to ensure that if they are doing a large push that as the blobs add up during upload we can exit prior to a full upload.

data/model/namespacequota.py
kwestpharedhat: This is a leftover of a refactor that got overlooked. Merging of 2 files.

endpoints/v2/errors.py
```diff
@@ -144,6 +144,13 @@ def __init__(self, uploaded=None, max_allowed=None):
             )
 
 
+class QuotaExceeded(V2RegistryException):
+    def __init__(self):
+        super(QuotaExceeded, self).__init__(
+            "QuotaExceeded", "Quota has been exceed on namespace", {}, 403
```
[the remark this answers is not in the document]
kwestpharedhat: Would DENIED work here or do we need something more specific denied feels misleading

data/model/namespacequota.py
```diff
@@ -0,0 +1,371 @@
...
+    InvalidOrganizationException,
+    _basequery,
+    organization,
+    user,
+    InvalidUsernameException,
+    repository,
+    notification,
+)
+
+HUMANIZED_QUOTA_UNITS = [i.decimal.symbol for i in humanfriendly.disk_size_units] + ["bytes"]
```
kleesc, line 26: I think the any sizes should be sent as standard bytes from the api, across all endpoints. Any conversion can be done by the clients (the UI in this case).

Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN set -ex\
 		nginx \
 		openldap \
 		openssl \
-		python38 \
+		python3 \
```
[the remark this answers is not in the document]
kwestpharedhat: We will reset this file

kleesc (changes requested):

endpoints/v2/errors.py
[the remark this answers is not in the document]
kwestpharedhat: Used denied with the same message

data/model/namespacequota.py
Sunandadadi: Hmm, the reason I've added here is because it was easier to handle conversions in Python. I was not able to find a good library in angular that would convert bytes into human readable format. And since our UI will shift later, our conversion logic will remain intact in the backend. The `humanfriendly` library converts bytes to human readable format and the other way as well.

kwestpharedhat: Do we have examples of how this is done elsewhere?

Sunandadadi: Hmm, examples in our codebase or what others did? I'd need to dig for that. I just din’t want to shift any logic to the UI. In my opinion UI should only display what backend sends.

kwestpharedhat: In our code base. I do like the idea of a uniform unit within the backend code at least up until the api sends the data (allowing the API to change the unit could be a happy medium) to the UI. However I would also like to take direction from what has been done previously.

kleesc: We usually return bytes when size is concerned,

kwestpharedhat: is this worth fixing right now or can we leave it for work as we fix issues from QE?

kleesc: We should fix it.

data/migrations/versions/e9f3e4dbb979_quota_management_reporting.py
```diff
@@ -0,0 +1,88 @@
+"""quota_management_reporting
```
kleesc, line 1: We should split the migration into a separate PR. That way, we can perform the migration first, and then if there's an issue when deploying, we can revert the image without having to rollback on the migration itself.

kleesc (changes requested):

data/migrations/versions/e9f3e4dbb979_quota_management_reporting.py
kwestpharedhat: This makes sense I will get it going

Dockerfile
```diff
@@ -1,6 +1,4 @@
-# syntax=docker/dockerfile:1.2
-# Base is set up with the runtime dependencies and environment.
-FROM quay.io/centos/centos:stream8 AS base
+FROM registry.access.redhat.com/ubi8/ubi:latest AS base
```
kleesc, line 1: These changes to the Dockerfile should already be in the main branch. You might need to rebase in order to grab these changes.

kwestpharedhat: I dont understand why this is showing as a changed file in the PR when both files match.

Sunandadadi:
Looks like [util/ipresolver/aws-ip-ranges.json](https://github.com/quay/quay/pull/1048/files#diff-a7f459c3c5a5f41b7c1c65e7491808a83bdd7b53aaede87f3db4199d0f2fe028) got added in the PR

data/model/namespacequota.py
```diff
@@ -0,0 +1,380 @@
...
+        if size > limit["bytes_allowed"]:
+            if limit_bytes < limit["bytes_allowed"]:
+                limit_bytes = limit["bytes_allowed"]
+                severity_level = limit["type_id"]
+
+    return {"limit_bytes": limit_bytes, "severity_level": severity_level}
+
+
+def __gather_size_by_chunk_upload(repo_id):
+    return repository.get_size_during_upload(repo_id)
```
kleesc, line 61: nit: I initially thought this op was on a repository reference. Maybe `from data import model` instead, so that the reference is `model.repository.[..]` to make it clearer

test/testconfig.py
```diff
@@ -108,3 +108,6 @@ def create_transaction(db):
     OCI_NAMESPACE_WHITELIST: List[str] = []
 
     FEATURE_USER_INITIALIZE = True
+
+    FEATURE_QUOTA_MANAGEMENT = True
+    DEFAULT_SYSTEM_REJECT_QUOTA_BYTES = 0
```
kleesc, line 113: `None` instead of `0`. `None` makes it clearer that we don't want that restriction, whereas `0` could mean either no pushes at all, or no limits at all.

initdb.py
```diff
@@ -858,6 +862,14 @@ def populate_database(minimal=False):
     org.stripe_id = TEST_STRIPE_ID
     org.save()
 
+    QuotaType.create(name="Notify")
```
kleesc, line 865: These names don't match those created in the alembic migration. Is this intended?

data/model/_basequery.py
```diff
@@ -75,6 +80,49 @@ def get_existing_repository(namespace_name, repository_name, for_update=False, k
...
+            QuotaType.id.alias("type_id"),
+        )
+        .join(User, on=(UserOrganizationQuota.namespace_id == User.id))
+        .join(QuotaLimits, on=(UserOrganizationQuota.id == QuotaLimits.quota_id))
+        .join(QuotaType, on=(QuotaLimits.quota_type_id == QuotaType.id))
+        .where(User.username == namespace_name)
+    ).dicts()
+
+    # define limits if a system default is defined in config.py and no namespace specific limits are set
+    if config.app_config.get("DEFAULT_SYSTEM_REJECT_QUOTA_BYTES") != 0 and len(query) == 0:
```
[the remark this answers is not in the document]
kwestpharedhat: I agree with higher level sof validation however the request is for 0 = unlimited.

data/model/namespacequota.py
```diff
@@ -0,0 +1,380 @@
...
+    _basequery,
+    organization,
+    user,
+    InvalidUsernameException,
+    repository,
+    notification,
+)
+
+
+def verify_namespace_quota(namespace_name, repository_ref):
```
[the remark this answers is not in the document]
kwestpharedhat: agreed

data/model/namespacequota.py
kwestpharedhat: easy enough actually.

test/testconfig.py
kwestpharedhat: This was the requested behavior, We can re-evaluate after this PR.

data/model/namespacequota.py
```diff
@@ -0,0 +1,425 @@
...
+
+def check_limits(namespace_name, size):
+    limits = get_namespace_limits(namespace_name)
+    limit_bytes = 0
+    severity_level = None
+    for limit in limits:
+        if size > limit["bytes_allowed"]:
+            if limit_bytes < limit["bytes_allowed"]:
+                limit_bytes = limit["bytes_allowed"]
+                severity_level = limit["type_id"]
```
kleesc, line 56: What if you have multiple limits defined, what is ordering based on when having multiple threshold.
For example, I have a limit set at 50% to send a warning, and another at 90% to error (does it make sense to have an error limit at something lower than 100%?),  and I'm at 95%, which one is this going to/should return?  

I don't think `get_namespace_limits` has a specific ordering.

kwestpharedhat: this loops through all limits and finds the largest limit that has been crossed and returns that one. So if your at 95% it would only return the reject. I cant think of only 1 real reason to set the threshold below 100% and that is to account for the system overshoot if it becomes a problem. Otherwise it is just a configuration allowed currently in 3.7. THis method has been confirmed by Daniel.

data/model/namespacequota.py
```diff
@@ -0,0 +1,425 @@
...
+
+    return {"limit_bytes": limit_bytes, "severity_level": severity_level}
+
+
+def notify_organization_admins(repository_ref, notification_kind, metadata={}):
+    if repository_ref.namespace_user.organization:
```
kleesc, line 62: You might run into the same issue here, and other places where `repository_ref` is referenced. `repository_ref` is not the type you're expecting.

So you might have to change the signature of the functions in this module to take the namespace name and repo name after all, and get the actual peewee object from `data/model/repository.py`, that way you can reference the user object associated to the repo.

data/model/namespacequota.py
```diff
@@ -0,0 +1,425 @@
...
+def force_cache_repo_size(repository_ref):
+    return model.repository.force_cache_repo_size(repository_ref._db_id)
+
+
+def create_namespace_quota(name, limit_bytes):
+    user_object = user.get_namespace_user(name)
+
+    UserOrganizationQuota.create(namespace_id=user_object.id, limit_bytes=limit_bytes)
+
+    return UserOrganizationQuota
```
kleesc, line 87: You're returning the table class here. Instead of the created instance. Also need to check whether the org quota already exists or not, and handle it (do nothing, replace the existing one with the new value, ...?)

kwestpharedhat: This is true. The value returned is never used which is why this still functions. I will fix it.

data/model/namespacequota.py
kwestpharedhat: This is expecting the repository_ref object and uses the namespace_user parameter to gather the organization information if its an org.

kwestpharedhat: This is not something to fix other than explicitly defining parameter type.

kleesc: The callers of this function are using `registry_model.lookup_repository` though, which doesn't have a `namespace_user` reference. It does have a `namespace_name` reference though.

kleesc: So they're passing the repo_ref (as defined in the registry_model datatypes) and not the repo obj itself (referring to obj as peewee objects).

kwestpharedhat: Fixed.

endpoints/api/namespacequota.py
```diff
@@ -0,0 +1,328 @@
...
+        },
+    }
+
+    @nickname("getNamespaceQuota")
+    def get(self, namespace):
+        orgperm = OrganizationMemberPermission(namespace)
+        userperm = UserReadPermission(namespace)
+
+        if not orgperm.can() and not userperm.can():
+            raise Unauthorized()
```
kleesc, line 84: We have predefined scope decorator you could use to handle permissions.
For example `require_repo_read`, `require_repo_write`, `require_repo_admin` from `endpoints/api/__init__`

kleesc: Ah. I guess it's because both user and org quotas share the same endpoint.
Wondering if these should be split? Especially if we were to expand the feature set of quotas on orgs in the future. For example, quotas per member, etc. In which case it would be better for it to orgs and users to have their own quota subresource.

kleesc: I'd argue that it would be better to do it now. That way we don't introduce major api changes again in the future.
Also, currently, most of our users and orgs apis separate, so it would be consistent to manage quotas the same way.

kleesc (dismissed):

PR #1041, opened 2021-12-21
chore: download aws ip ranges via github workflow

flavianmissi:
Fixes [PROJQUAY-2941](https://issues.redhat.com/browse/PROJQUAY-2941)

---

This workflow will run at midnight, every night, and when the downloaded file changes it will create a commit such as this one https://github.com/flavianmissi/quay/commit/30e3c2e924a6b0c3525405ba7afacd01342f3af4

PR #1049, opened 2022-01-07
ipresolver: update country mmdb (PROJQUAY-3031)

kleesc:
Update the resolver's mmdb file.
TODO: We should find a way to update this at a somewhat regular interval, similar to the aws-ip-ranges file.


flavianmissi:
See https://github.com/quay/quay/pull/1041/files for details on how the auto updates to aws-ip-ranges file works.

syed (approved):

PR #1053, opened 2022-01-10
feat: pull-thru proxy cache (PROJQUAY-3030) (PROJQUAY-3033)

flavianmissi:
## --> Please read before reviewing <--

Trust me, it'll save us both time!

### Purpose

This PR covers:

* Backend and database functionality to configure a cache org (part of [PROJQUAY-3029](https://issues.redhat.com/browse/PROJQUAY-3029))
* Proxying images from an upstream registry ([PROJQUAY-3033](https://issues.redhat.com/browse/PROJQUAY-3033))
* Caching proxied images from an upstream registry ([PROJQUAY-3063](https://issues.redhat.com/browse/PROJQUAY-3063))

This PR does **_not_** cover:

* API or UI for configuring a cache org ([PROJQUAY-3029](https://issues.redhat.com/browse/PROJQUAY-3029))
* Cache flush, other than individual deletion of repositories ([PROJQUAY-3034](https://issues.redhat.com/browse/PROJQUAY-3034))
* Auto pruning of repositories from the cache organization, automatically keeping it within its quota usage limits ([PROJQUAY-3063](https://issues.redhat.com/browse/PROJQUAY-3063))
* Audit logs for image pulls from the upstream registry

### Change summary

* A new registry model (`data/registry_model/registry_proxy_model.py`) was introduced to handle creation of manifests without requiring their blobs to exist in storage. The new registry model inherits from the `OCIModel`, overriding only the methods used to retrieve manifests and blobs.
* The `registry_model` is now injected into the GET/HEAD views handling manifests and blobs v2 api endpoints, instead of relying on the imported `registry_model` from `data/registry_model/__init__.py`
* A new `proxy` module was introduced. It's a very simple client used to download manifests and blobs from other registries.
* A new configuration `FEATURE_PROXY_CACHE` has been introduced to control whether the pull through proxy cache feature is globaly enabled or not. It's on by default.
* A new model and equivalent migration were added - `ProxyCacheConfig`. It's existence in association with an organization indicates that said organization is set to function as a proxy org.
* A new pytest marker `e2e` was introduced. Tests using this marker run on a separate CI job (in the existing CI workflow), also set up on this PR.

### Limitations

* When a manifest _list_ or OCI index is cached from the upstream registry Quay will show it and all its sub-manifest objects in the UI. To avoid caching manifests that will never be used, Quay will not cache sub-manifests _unless the client specifically pulls them_. When a client encounters a manifest list or OCI index, its next operation is to pull an image by digest for the client's correct architecture (see the [content negotiation docs proposal](https://github.com/opencontainers/distribution-spec/issues/212) for details). In practice, this means that a sub-manifest may be listed in the UI, but when a user clicks on a sub-manifest that has not yet been pulled by the client, Quay will give back a 404.
* Pulling an image by digest (without ever pulling it by tag) will create a manifest with a temporary tag. Temporary tags are hidden by default and not shown in the UI. If a temporary tag is the only tag that exists in the cached repository, the repository will appear empty in the UI.

### Testing

You are encouraged to pull the code and try proxying. It's fun!

#### Proxying a whole registry

1. Start your local environment with `make local-dev-up`
2. Create an org normally via UI, let's call it "cache"
3. Login to quay's container by running `docker-compose exec quay /bin/bash`
4. Start a flask shell from inside quay's container by running `flask shell`
5. Create a proxy cache config by executing the python script in the running flask shell:
```python
from data.model.proxy_cache import create_proxy_cache_config
create_proxy_cache_config("cache", "docker.io")
```  

Exit both flask shell and container, then run:

```
docker pull localhost:8080/cache/library/postgres:14
```

#### Proxying a single namespace from a registry

1. Start your local environment with `make local-dev-up`
2. Create an org normally via UI, let's call it "cache-library" (we'll proxy the library namespace from docker hub)
3. Login to quay's container by running `docker-compose exec quay /bin/bash`
4. Start a flask shell from inside quay's container by running `flask shell`
5. Create a proxy cache config by executing the python script in the running flask shell:
```python
from data.model.proxy_cache import create_proxy_cache_config
create_proxy_cache_config("cache-library", "docker.io/library")
``` 

Exit both flask shell and container, then run:

```
docker pull localhost:8080/cache-library/busybox@sha256:62ffc2ed7554e4c6d360bce40bbcf196573dd27c4ce080641a2c59867e732dee
```

#### Testing auth

Follow the same steps above, just add the desired username and password as arguments to the `create_proxy_cache_config` function. See `data/model/proxy_cache.py` for usage details.


app.py
```diff
@@ -11,7 +11,7 @@
 from flask_login import LoginManager
 from flask_mail import Mail
 from flask_principal import Principal
-from werkzeug.contrib.fixers import ProxyFix
+from werkzeug.middleware.proxy_fix import ProxyFix
```
flavianmissi, line 14: fixes a deprecation warning.

Makefile
```diff
@@ -25,7 +25,7 @@ conf/stack/license: $(QUAY_CONFIG)/local/license
 unit-test:
 	TEST=true PYTHONPATH="." py.test \
 	--cov="." --cov-report=html --cov-report=term-missing \
-	--timeout=3600 --verbose -x \
+	--timeout=3600 --verbose -x --ignore=buildman/ \
```
flavianmissi, line 28: CI doesn't run buildman tests, so remove it from make target as well

endpoints/v2/manifest.py
```diff
@@ -47,129 +57,189 @@
...
 
-    tag = registry_model.get_repo_tag(repository_ref, manifest_ref)
-    if tag is None:
-        if registry_model.has_expired_tag(repository_ref, manifest_ref):
-            logger.debug(
-                "Found expired tag %s for repository %s/%s", manifest_ref, namespace_name, repo_name
-            )
-            msg = (
-                "Tag %s was deleted or has expired. To pull, revive via time machine" % manifest_ref
-            )
```
ibazulic, line 70: Won't removing this be a bad UX design for deleted tags that are in normal, i.e. not proxied repositories? Time machine is still a thing, we should give the user info that a tag is in time machine if the feature is enabled.

proxy/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+        # accept_ranges = resp.headers.get("Accept-Ranges", None)
+        # if accept_ranges is not None:
+        #     headers["Accept-Ranges"] = accept_ranges
+        return {
+            "response": stream_contents(),
+            "status": resp.status_code,
+            "headers": dict(resp.headers),
+        }
+
+    def _authorize(self, auth: tuple[str, str] | None = None):
```
dmage, line 93: @flavianmissi does it work for registries that use basic auth?

endpoints/v2/manifest.py
flavianmissi: Not sure I'm following. This was just moved around: https://github.com/quay/quay/pull/1053/files#diff-2487e826aa015b45ecc8a34cbf59ea7bccf235b8d51bf732fd574cf94d11eeedR132

proxy/__init__.py
flavianmissi: Not with the code as is. Are there any such registries? 🤔

flavianmissi: Should be easy enough to support it though, we just need the `parse_www_auth` to return the auth scheme, then use when adding the authentication header to the session.

dmage: there are, but they aren't public

flavianmissi: @dmesser do we care about registries that use basic auth? an option would be to list this as a known limitation for now, then if anyone needs it it's easy enough to add it.

endpoints/v2/manifest.py
ibazulic: I missed the new location, carry on :-)

flavianmissi: no worries, diffs can be confusing! I tried the unified view but it was worse, might be something worth to play with tho.

endpoints/api/test/test_security.py
```diff
@@ -5767,8 +5767,9 @@ def test_all_apis_tested(app):
     for rule in app.url_map.iter_rules():
         endpoint_method = app.view_functions[rule.endpoint]
 
-        # Verify that we have a view class for this API method.
-        if not "view_class" in dir(endpoint_method):
+        # Verify that we have a view class for this API method,
+        # and that we're not testing a v2 endpoint.
+        if not "view_class" in dir(endpoint_method) or "v2" in rule.endpoint:
```
flavianmissi, line 5772: these tests fail if any views other than the ones in `endpoints/api` are classes rather than functions. I extended the check a little to exclude views from the v2 api, since this test exclusively tests stuff inside `endpoints/api`

endpoints/v2/blob.py
```diff
@@ -51,82 +56,116 @@
...
+                media_type = request.headers.get("Accept", None)
+                resp = proxy.get_blob(digest, media_type)
+                return Response(**resp)
+            except ProxyNotSupported:
+                pass
+            raise BlobUnknown()
+
+        return self.download_blob(blob, namespace_name, digest)
+
+    def download_blob(self, blob, namespace_name, digest):
```
flavianmissi, line 127: The previous `download_blob` view became a method of the `DownloadBlob` class. No changes were made to it, other than adding the blob as a method argument.

dmage (commented):
Quick feedback, I need to try it locally

data/database.py
```diff
@@ -2048,6 +2048,21 @@ class ManifestSecurityStatus(BaseModel):
...
 
+class ProxyCacheConfig(BaseModel):
+    """
+    Represents the configuration for an organization of type proxy cache
+    """
+
+    user_id = ForeignKeyField(User)
+    creation_date = DateTimeField(default=datetime.utcnow)
+    upstream_registry = CharField(max_length=2048)
+    upstream_registry_namespace = CharField(max_length=255, null=True)
```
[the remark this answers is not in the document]
flavianmissi: > Will we ask customers to provide it as two different inputs?
Not sure I follow. The namespace is optional, and should only be provided when caching a namespace upstream. When caching the whole registry the namespace should be omited. The UI needs to make this clear somehow.

> max_length seems to be chosen arbitrarily, upstream_registry is hostname + ':' + port, so 254+1+5=260.
True. We don't have any extra validation here - users might specify protocol (if they don't we default to https) or not. I'm okay with a smaller max_length though, 2048 does seem unnecessairily exceesive. What about 270, for a round nice number? It would also allow for specifying a protocol.

> Image references cannot have more than 255 characters: [...]
Hmm even if we do take the distribution implementation as the single source of truth, from what I gather it'd be hard to know exactly how big the namespace alone can be. A valid name could have a namespace 243 chars long and it would be valid. I can't really think of a better value here.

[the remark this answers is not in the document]
dmage: I don't like the idea of splitting it into two different fields. We should ask customers to provide an image reference prefix. Like `docker.io`, or `docker.io/library`, or `quay.io/quay`. The same prefix they'll use in ImageContentSourcePolicy.
cc @dmesser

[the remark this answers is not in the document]
flavianmissi: Can you expand @dmage? I'm not seeing what the big deal is, or what's the relationship with ImageContentSourcePolicy.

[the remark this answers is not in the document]
flavianmissi: I mean, to have them separate makes sense to me because it makes it easier for us to distinguish in the code when orgs are set to cache whole registries vs single namespaces. We'd have to parse the URL to figure that out if the reference would be given as one, such as `docker.io/library`. Am I overcomplicating this?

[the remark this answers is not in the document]
flavianmissi: There's a bigger problem with using your suggestion @dmage. The `upstream_registry` is expected to be the api url of the registry, so in case of docker hub we want `registry-1.docker.io` rather than `docker.io`. We'd have to change that first in order to accept the namespace in that field.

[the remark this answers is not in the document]
dmage: It makes it easier in the code, but I want to make it easier for people. I don't see any problems, other tools (docker, podman, CRI-O, skopeo, etc.) don't ask to provide `registry-1.docker.io`, they understand `docker.io`. As it's a simple splitting by / and translating one known string into another, we can do it in UI, if you want to go this way.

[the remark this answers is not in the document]
flavianmissi: I want to make it easier for people too. If you think your suggestion gives better usability then let's do it. I'll update the code.

.github/workflows/CI.yaml
```diff
@@ -10,8 +10,8 @@ on:
     branches:
```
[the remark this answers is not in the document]
flavianmissi: it's caching the installation of Quay's requirements. It's an attempt to speed up the tests.
But yeah it should probably be moved into its own thing.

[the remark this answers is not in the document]
Sunandadadi: Yes, I agree with @kwestpharedhat on separating out the CI stuff.

endpoints/v2/proxy.py
```diff
@@ -0,0 +1,46 @@
...
+    if not app.config.get("FEATURE_PROXY_CACHE", False):
+        raise ProxyNotSupported("Proxy globally disabled.")
+
+    try:
+        config = get_proxy_cache_config_for_org(org_name)
+        return config
+    except (InvalidOrganizationException, ProxyCacheConfig.DoesNotExist):
+        # we may end up here if `org_name` either refers to a user
+        # or if the org is not set up as cache - either way proxy shouldn't work.
+        raise ProxyNotSupported("Organization is invalid or not configured as proxy cache.")
```
Sunandadadi, line 23: Do you think we can propagate the same error that was raised instead of raising a new one? Since we are catching `InvalidOrganizationException` and `ProxyCacheConfig.DoesNotExist` we propagate it to be more specific.

flavianmissi: There are multiple reasons for a `ProxyNotSupported` exception to happen:
 * the `FEATURE_PROXY_CACHE` is set to False;
 * `org_name` was a user rather than an org;
 * there is no proxy cache config for the org;
By not returning `ProxyNotSupported` exception here, the caller would have to handle 3 different exceptions to cover all cases of where proxy is not supported. Returning a single exception for all 3 situations is the way I found to make it easier for the caller to know whether proxy code needs to be executed or not, regardless of reason.

I'm open to new ideas, so long as we easily distinguish cases of when the everything looks good and the proxy code should be executed vs when the "normal" flow should take over instead.

What do you think?

data/cache/cache_key.py
```diff
@@ -9,6 +9,14 @@ class CacheKey(namedtuple("CacheKey", ["key", "expiration"])):
     pass
 
 
+def for_upstream_registry_token(registry, repo_name, expires_in):
+    """
+    Returns a cache key for an upstream registry auth token.
+    """
+    key = f"upstream_token__{registry}_{repo_name}"
+    return CacheKey(key, expires_in)
```
kleesc, line 17: Probably not a good idea to store the registry token. Any reason why you couldn't just pass this along the client in the response normally, just like any normal registry auth requests?

endpoints/v2/blob.py
```diff
@@ -51,82 +53,116 @@
...
+        disallow_for_account_recovery_mode,
+    ]
+
+    def head(self, namespace_name, repo_name, digest):
+        blob = registry_model.get_cached_repo_blob(model_cache, namespace_name, repo_name, digest)
+        if blob is None:
+            try:
+                proxy = setup_proxy(namespace_name, repo_name)
+                resp = proxy.blob_exists(digest)
+                return Response(**resp)
```
kleesc, line 72: Shouldn't this proxy check happen before anything else?

endpoints/v2/blob.py
```diff
@@ -51,82 +53,116 @@
...
+            except ProxyNotSupported:
```
kleesc, line 73: Also, this block seems to be recurring in multiple endpoints. Could you make a decorator out of it to use on the relevant endpoints?

kleesc (changes requested):

.github/workflows/CI.yaml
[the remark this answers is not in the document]
flavianmissi: removed - only relevant changes are left

data/cache/cache_key.py
flavianmissi: That's not how basic auth works[1]. The challenge[2] Quay returns on 401 responses is specific to Quay. Any third-party services Quay may talk to are not part of the equation.

A forward proxy behaves that way [3][4][5], but it isn't our goal to turn Quay into one.

Maybe we encrypt it before storing?

[1] https://datatracker.ietf.org/doc/html/rfc7617
[2] https://datatracker.ietf.org/doc/html/rfc7235#section-4.1
[3] https://datatracker.ietf.org/doc/html/rfc7235#section-3.2
[4] https://datatracker.ietf.org/doc/html/rfc7235#section-4.3
[5] https://datatracker.ietf.org/doc/html/rfc7235#section-4.4

endpoints/v2/blob.py
```diff
@@ -51,82 +53,116 @@
...
-        logger.debug("Returning direct download URL")
-        resp = redirect(direct_download_url)
-        resp.headers.extend(headers)
-        return resp
-
-    # Close the database connection before we stream the download.
-    logger.debug("Closing database connection before streaming layer data")
-    headers.update(
-        {
+class CheckBlobExists(MethodView):
```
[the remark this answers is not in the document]
flavianmissi: Agreed. I turned it into a class because I was using a mixin to inject the proxy in the views, but that is no longer the case it can probably go back to being a function.

endpoints/v2/blob.py
flavianmissi: Yes and no. Right now the proxy is quite stupid, so this is okay. When we start actually storing the blobs we'll need to revisit this, taking the staleness period into consideration.

data/cache/cache_key.py
kleesc: I wasn't talking about basic auth. Looks like you're trying to save the result of the exchange flow in Redis.
Couldn't you do something like append it to the Quay one back in the response? That way, both are available, either for auth against Quay or the upstream registry, depending on the namespace type, and you don't have to store anything.

dmage: Credentials for the upstream registry and tokens from the upstream registry should be kept secret and shouldn't be passed to Quay clients.

endpoints/v2/blob.py
flavianmissi: Yeah, true.
This block will change a lot as we work on storing the images pulled from upstream registries. It's a bit early to know what's the best way to refactor it, so I rather we wait until the code takes more shape.

endpoints/v2/blob.py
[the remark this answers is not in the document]
flavianmissi: Turned both blob views back into functions.

data/database.py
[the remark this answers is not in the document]
flavianmissi: @dmage I believe I adressed your comments in this thread - marking the conversation as resolved. Lemme know if I missed something.

data/database.py
```diff
@@ -2048,6 +2048,21 @@ class ManifestSecurityStatus(BaseModel):
...
+    upstream_registry_username = EncryptedCharField(max_length=2048, null=True)
+    upstream_registry_password = EncryptedCharField(max_length=2048, null=True)
+    staleness_period_s = IntegerField(default=0)
```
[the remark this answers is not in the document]
flavianmissi: This feature shouldn't be enabled by default on quay.io - not sure if it matters which way we go.
Relatedly - @kleesc as soon as this lands we'll need to set the feature flag por this one to False as the default is currently True.

endpoints/v2/manifest.py
```diff
@@ -47,129 +49,158 @@
...
-@parse_repository_name()
-@process_registry_jwt_auth(scopes=["pull"])
-@require_repo_read
-@anon_protect
-def fetch_manifest_by_tagname(namespace_name, repo_name, manifest_ref):
-    repository_ref = registry_model.lookup_repository(namespace_name, repo_name)
-    if repository_ref is None:
-        image_pulls.labels("v2", "tag", 404).inc()
-        raise NameUnknown()
+class FetchManifest(MethodView):
```
kleesc, line 52: Essentially the same comment as before on the blob class endpoints. I'd prefer to keep it consistent across the registry api, unless there's a reason this should be wrapped in a class.

kleesc (changes requested):

endpoints/v2/manifest.py
flavianmissi: This is not the same as for the blob class endpoints. Inheritance is in use here to share proxy code between the two manifest GET endpoints. I didn't know we had an unspoken agreement towards prefering functional programming over object oriented - if we do I can happilly adhere.

data/cache/cache_key.py
flavianmissi: FTR - we discussed on slack and decided the best way to go is to indeed keep the token in the backend. It was suggested there that we renew the token before it expires in the upstream registry, which is now done.

proxy/__init__.py
flavianmissi: basic auth works. kudos to @dmage for helping me test it 🚀

data/model/proxy_cache.py
```diff
@@ -0,0 +1,35 @@
...
+
+    return new_entry
+
+
+def get_proxy_cache_config_for_org(org_name):
+    """
+    Return the Proxy-Cache-Config associated with the given organization name, or None if it doesn't exist.
+    """
+    org = get_organization(org_name)
+    return ProxyCacheConfig.get(ProxyCacheConfig.user_id == org.id)
```
kleesc, line 35: Need to check if the given name is an org first. Unless the caller has already done so.

kleesc: Also, you need to wrap the `get()` in a try-except block, and catch `ProxyCacheConfig.DoesNotExist`, and then return None. Otherwise, an exception will be raised by Peewee

endpoints/v2/manifest.py
```diff
@@ -47,129 +49,158 @@
...
+            logger.debug(f"Skipping pull through proxy cache: {e}")
+
+        return self.get_manifest_by_ref(namespace_name, repo_name, manifest_ref)
+
+    def get_manifest_by_ref(namespace_name, repo_name, manifest_ref):
+        raise NotImplementedError()
+
+
+class FetchManifestByTag(FetchManifest):
+    def get_manifest_by_ref(self, namespace_name, repo_name, manifest_ref):
```
kleesc, line 77: Yea... The semantics of this class and the method is weird and confusing IMO. I **really** think inheritance isn't the right abstraction in this case (I'm not a fan of inheritance in general, but that's another story).

For sharing proxy logic between manifest endpoints, see my other comment about using a decorator.

endpoints/v2/manifest.py
```diff
@@ -280,14 +311,14 @@ def write_manifest_by_digest(namespace_name, repo_name, manifest_ref):
     )
 
 
-def _parse_manifest():
-    content_type = request.content_type or DOCKER_SCHEMA1_MANIFEST_CONTENT_TYPE
+def _parse_manifest(content_type, request_data):
+    content_type = content_type or DOCKER_SCHEMA1_MANIFEST_CONTENT_TYPE
```
kleesc, line 315: Why was this change needed?

test/fixtures.py
```diff
@@ -182,6 +182,7 @@ def appconfig(database_uri):
         "MAIL_SERVER": "",
         "MAIL_DEFAULT_SENDER": "admin@example.com",
         "DATABASE_SECRET_KEY": "anothercrazykey!",
+        "FEATURE_PROXY_CACHE": True,
```
kleesc, line 185: You'll also need to set it in `config.py`, defaulting to False there.
And define the schema in `util/config/schema.py`

proxy/__init__.py
```diff
@@ -0,0 +1,215 @@
+"""
+The proxy module provides the means to proxy images from other registry instances.
+Registries following the distribution spec are supported.
+"""
```
kleesc, line 4: It would probably make sense to move this into its own `proxy.py` file under the `util/` directory.

test/test_proxy.py
```diff
@@ -0,0 +1,371 @@
+from datetime import datetime
+import unittest
+import json
+from unittest import mock
```
kleesc, line 4: Related to the other comment, move this along `proxy.py`.
The `test/` directory is where all tests + registry tests used to go . Since then we've favoured having tests being stored along side its module
If you look some other module, say `util/secscan/`, the secscan tests are stored in `util/secscan/test`. It's just easier for discovery and navigating specific tests this way.

kleesc (dismissed):

proxy/__init__.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    scheme = value.split(" ", 1)[0]
+    matches = WWW_AUTHENTICATE_REGEX.findall(value)
+    parsed = dict(matches)
+    if scheme:
+        parsed["scheme"] = scheme
+    return parsed
+
+
+class Proxy:
```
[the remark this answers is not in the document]
flavianmissi: Inheritance from object was needed in Python 2.x to use new-style classes. In Python 3.x this is the default, so we don't need to inherit from object.

data/cache/cache_key.py
```diff
@@ -9,6 +9,14 @@ class CacheKey(namedtuple("CacheKey", ["key", "expiration"])):
     pass
 
 
+def for_upstream_registry_token(registry, repo_name, expires_in):
+    """
+    Returns a cache key for an upstream registry auth token.
+    """
+    key = f"upstream_token__{registry}_{repo_name}"
```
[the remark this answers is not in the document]
flavianmissi: good catch - will do.

endpoints/api/test/test_security.py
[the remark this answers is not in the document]
flavianmissi: whether a view is written as a function or class should not cause any tests to break, let alone tests in another module.

endpoints/v2/blob.py
```diff
@@ -62,6 +63,12 @@ def check_blob_exists(namespace_name, repo_name, digest):
     # Find the blob.
     blob = registry_model.get_cached_repo_blob(model_cache, namespace_name, repo_name, digest)
     if blob is None:
+        try:
+            proxy = setup_proxy(namespace_name, repo_name)
+            resp = proxy.blob_exists(digest)
+            return Response(**resp)
+        except ProxyNotSupported as e:
+            logger.debug(f"Skipping pull through proxy cache: {e}")
```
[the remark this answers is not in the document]
flavianmissi: We talked about this last week - using a decorator as you suggest assumes we _want_ all the proxy code in the same place, and I honestly can't say we do (or do not) at this point. I will not refactor code just to refactor it again using a different approach in the future. 
Can focus on the _what_ rather than on the _how_, at least for this first iteration?
Most of the view code in this PR will change in the near future, it's not worth spending either your or my time trying to find the correct abstraction at this point.

endpoints/v2/manifest.py
flavianmissi: it will be needed to parse a manifest coming from a request other than `request`, which we'll be doing when storing the manifest.

test/fixtures.py
flavianmissi: That's already done - am I missing anything?

proxy/__init__.py
flavianmissi: `util` is just a bucket for anything we don't have a specific place for. The advice here[1] is aimed at go, but applies to any language.
[1] https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

data/database.py
```diff
@@ -2048,6 +2048,32 @@ class ManifestSecurityStatus(BaseModel):
...
 
 
+class ProxyCacheConfig(BaseModel):
+    """
+    Represents the configuration for an organization of type proxy cache
+    """
+
+    user = ForeignKeyField(User)
+    creation_date = DateTimeField(default=datetime.utcnow)
+    upstream_registry = CharField(max_length=270)
```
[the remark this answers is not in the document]
flavianmissi: good question. here are some of the possible values this might take:
* `docker.io`
* `docker.io/library`
* `quay.io`
* `quay.io/projectquay`
* `my-registry.example`
I'm actually having a hard time chosing a better name. It's not a url, neither an endpoint, and definitely not the registry name.
the `registries.conf`[1] documentation calls this an image name `prefix`... which in is really what this is, but I'm not sure if referring to it like that would be interpreted in the same way by everyone reading the code.

Thoughts?

[1] https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md#choosing-a-registry-toml-table

data/model/proxy_cache.py
```diff
@@ -0,0 +1,35 @@
+from data.database import ProxyCacheConfig
+from data.model.organization import get_organization
+
+
+def create_proxy_cache_config(
+    org_name,
+    upstream_registry,
+    upstream_registry_username=None,
+    upstream_registry_password=None,
+    staleness_period_s=0,
```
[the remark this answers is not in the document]
flavianmissi: The model already set the default staleness period to `0`, this function is just reflecting that.
It's a good idea to use a constant though - added that.

data/model/proxy_cache.py
flavianmissi: > Need to check if the given name is an org first.

`get_organization` will raise an `InvalidOrganizationException` if `org_name` belongs to a user instead of an org.

> Also, you need to wrap the get() in a try-except block, and catch ProxyCacheConfig.DoesNotExist, and then return None. Otherwise, an exception will be raised by Peewee

Why would I _need_ to do that?

One thing I notice is that this function's doc string did not match it's behaviour - maybe that's why you're suggesting returning `None`? I've updated it to match the behaviour.

endpoints/v2/manifest.py
flavianmissi: This is pretty standard python code. I wouldn't aim at avoiding inheritance while coding in an object-oriented programming language, there's nothing wrong with it if we use it correctly.

Using inheritance here is, as you point out, a way of abstracting away the proxy code. And I've been arguing against us doing that prematurely - so I rolled these views back into functions, and added the proxy code within them, similar to the blob endpoint functions. Now we don't have any view related abstractions, so we're hopefully free of any bias towards classes or decorators.

test/test_proxy.py
flavianmissi: Answered in [previous comment](https://github.com/quay/quay/pull/1053#discussion_r800455617)

test/fixtures.py
flavianmissi: Noticed I had missed setting it in the schema - fixed.

data/database.py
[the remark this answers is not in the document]
flavianmissi: Seems no one have strong feelings about this one. Marking as resolved for now - please unresolve the convo if needed.

proxy/__init__.py
flavianmissi: Going to move the tests here but keep the proxy dir as is otherwise, as agreed verbally.

data/registry_model/registry_proxy_model.py
```diff
@@ -0,0 +1,620 @@
...
+            return parsed
+
+        schema1 = self.get_schema1_parsed_manifest(
+            manifest, namespace_name, repo_name, tag_name, storage
+        )
+        return schema1.convert_manifest(
+            allowed_mediatypes, namespace_name, repo_name, tag_name, None
+        )
+
+    def get_schema1_parsed_manifest(self, manifest, namespace_name, repo_name, tag_name, storage):
```
dmage, line 497: It doesn't fetch the config, right? It seems it won't provide correct environment variables, etc. after conversion.

endpoints/v2/__init__.py
```diff
@@ -127,8 +128,12 @@ def wrapped(namespace_name, repo_name, *args, **kwargs):
 
             repository = namespace_name + "/" + repo_name
             if allow_public:
-                repository_ref = registry_model.lookup_repository(namespace_name, repo_name)
-                if repository_ref is None or not repository_ref.is_public:
+                try:
+                    repository_ref = registry_model.lookup_repository(namespace_name, repo_name)
+                except RepositoryDoesNotExist:
+                    raise Unauthorized(repository=repository, scopes=scopes)
```
dmage, line 134: I suspect this is not always correct to reply with Unauthorized in this case.

image/docker/schema2/list.py
```diff
@@ -209,6 +209,13 @@ def manifest_dict(self):
     def bytes(self):
         return self._manifest_bytes
 
+    @property
+    def filesystem_layers(self):
+        """
+        Returns the file system layers of this manifest, from base to leaf.
+        """
+        return None
```
dmage, line 217: Technically that's incorrect, this information is stored in the image config. Why do you need this method?

proxy/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+        content_type = resp.headers.get("content-type")
+        return raw_manifest, content_type
+
+    def manifest_exists(self, image_ref: str, media_type: str | None = None) -> str:
+        url = f"{self.base_url}/v2/{self._repo}/manifests/{image_ref}"
+        headers = {}
+        if media_type is not None:
+            headers["Accept"] = media_type
+        resp = self.head(url, headers=headers, allow_redirects=True)
+        return resp.headers.get("docker-content-digest")
```
dmage, line 82: from OCI spec:
> the client MAY encounter Docker-specific headers, such as Docker-Content-Digest, or Docker-Distribution-API-Version. These headers are OPTIONAL and clients SHOULD NOT depend on them.
> — https://github.com/opencontainers/distribution-spec/blob/main/spec.md#definitions

proxy/__init__.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def get_blob(self, digest: str):
+        url = f"{self.base_url}/v2/{self._repo}/blobs/{digest}"
+        resp = self.get(
+            url,
+            allow_redirects=True,
+        )
+        size = resp.headers.get("content-length")
+        content_digest = sha256_digest(resp.content)
```
dmage, line 91: Will it load the entire blob in memory? We should never do it.

image/docker/schema2/list.py
flavianmissi: it's only here to satisfy the `ManifestInterface` base property, so it isn't strictly needed.
I should update the comment though, since this won't return any layers for manifests of type list.

proxy/__init__.py
flavianmissi: Yup, that's why I used `get` to retrieve it. I've changed the return type to `str | None` to reflect that (will push soon)

proxy/__init__.py
flavianmissi: yeah it will :/ 
is this something we want to address in this PR? I have a feeling that not doing so won't exactly be straight forward.

test/registry/registry_tests.py
```diff
@@ -1372,39 +1372,6 @@ def test_cancel_upload(manifest_protocol, basic_images, liveserver_session, app_
     )
 
 
-def test_blob_caching(
```
flavianmissi, line 1375: FTR: blob caching is already covered on https://github.com/quay/quay/blob/ba652e0b9879b319aabf26724a32acfbb7b74631/endpoints/v2/test/test_blob.py#L26

data/registry_model/registry_oci_model.py
```diff
@@ -793,6 +805,15 @@ def load_blacklist():
 
         return set(result)
 
+    def blob_exists(self, model_cache, namespace_name, repo_name, blob_digest):
```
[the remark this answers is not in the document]
flavianmissi: you're right, we no longer need this - will remove

proxy/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+    def __init__(self, config: ProxyCacheConfig, repository: str):
+        self._config = config
+
+        hostname = REGISTRY_URLS.get(
+            config.upstream_registry_hostname,
+            config.upstream_registry_hostname,
+        )
+        url = f"https://{hostname}"
+        if config.insecure:
+            url = f"http://{hostname}"
```
dmage, line 58: Insecure can also mean that the registry is available via `https://`, but its certificates aren't globally trusted/self-signed.

proxy/__init__.py
dmage: The caller of this function shouldn't rely on this value, as `clients SHOULD NOT depend on` it. So it's safer to always return None.

flavianmissi: You make a good point.
The thing is that we're using this digest to check whether the manifest cached in quay is up-to-date with the upstream. If the upstream doesn't return a digest then we will end up pulling the manifest again.
I'm not sure what's the best way to do this. We could do a GET instead of a HEAD to compare digests, but on docker hub a GET will count towards rate limit while a HEAD won't.
Another option would be to do a GET only if we find that the `docker-content-digest` header from the HEAD came up empty, then calculate the manifest digest from the actual manifest. I'd argue that this is a nice-to-have.
Thoughts?

proxy/__init__.py
flavianmissi: As discussed in Slack, we'll handle this as a follow-up. See https://issues.redhat.com/browse/PROJQUAY-3457

proxy/__init__.py
flavianmissi: Discussed on slack and although this is a release blocker, it'll be done in a follow-up PR. See https://issues.redhat.com/browse/PROJQUAY-3459

endpoints/v2/__init__.py
flavianmissi: Reverted back to previous code.

data/registry_model/registry_proxy_model.py
flavianmissi: tracked by https://issues.redhat.com/browse/PROJQUAY-3460

data/registry_model/registry_oci_model.py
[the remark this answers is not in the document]
flavianmissi: Removed.

proxy/__init__.py
dmage: Oh, I didn't realize `manifest_exists` returns the current digest of the tag. It might be worthy to comment this function.

flavianmissi: Good point, added comment.

dmage (approved):

PR #1055, opened 2022-01-10
builders: Make single_connection_client conifgurable (PROJQUAY-3025)

syed:
single_connection_client keeps a connection open to the pool on init
which may cause issues if the master changes

### DIFF
--- a/data/buildlogs.py
+++ b/data/buildlogs.py
@@ -32,7 +32,7 @@ class RedisBuildLogs(object):
 
         args = dict(self._redis_config)
         args.update(
-            {"socket_connect_timeout": 1, "socket_timeout": 2, "single_connection_client": True}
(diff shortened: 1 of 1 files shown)


config.py
```diff
@@ -229,7 +229,7 @@ def create_transaction(db):
     AUTHENTICATION_TYPE = "Database"
 
     # Build logs
-    BUILDLOGS_REDIS = {"host": "localhost"}
+    BUILDLOGS_REDIS = {"host": "localhost", "single_connection_client": False}
```
kleesc, line 232: Not really needed since it defaults to using the connection pool.

syed: Ah true

kleesc:
Ah nevermind, it would get overwritten by the hardcoded value

kleesc (approved):
lgtm

PR #1060, opened 2022-01-13
Use final image for development (PROJQUAY-3098)

dmage:


BillDett:
what's the context behind this- it's a pretty big change.

flavianmissi:
What do we need to make this change ready for review?

dmage:
@flavianmissi it is ready for review

docker-compose.yaml
```diff
@@ -43,26 +43,50 @@ services:
...
+    init: true
+    volumes:
+      - ".:/quay-registry"
+    working_dir: /quay-registry
+    environment:
+      npm_config_cache: /quay-registry/.npm
+    entrypoint:
+      - sh
+      - -c
+      - PYTHONPATH=$$QUAYPATH python3 -m external_libraries && npm run watch
```
dmage, line 62: @flavianmissi do we need `PYTHONPATH=$$QUAYPATH python3 -m external_libraries` here or can I remove it? We do have it right now: https://github.com/quay/quay/blob/50d2a82783da906a3a68064fe0b9ab10e956f550/local-dev/init/01_local-dev-dependencies.sh#L16

util/workers.py
```diff
@@ -28,6 +28,12 @@ def get_worker_count(worker_kind_name, multiplier, minimum=None, maximum=None):
     minimum = minimum or multiplier
     maximum = maximum or (multiplier * multiplier)
 
+    minimum_override = os.environ.get("WORKER_COUNT_UNSUPPORTED_MINIMUM")
```
dmage, line 31: @kleesc what do you think? We need only 1 worker for development, and I want to make our development env closer to production.

Makefile
```diff
@@ -163,36 +163,51 @@ black:
...
-	make local-dev-clean
-	make local-dev-build-frontend
-	docker-compose up -d redis
-	docker-compose up -d quay-db
+local-dev-up: local-dev-clean node_modules | quay-build-image
+	DOCKER_USER="$$(id -u):$$(id -g)" docker-compose up -d local-dev-frontend --force-recreate
+	docker-compose up -d redis quay-db
 	docker exec -it quay-db bash -c 'while ! pg_isready; do echo "waiting for postgres"; sleep 2; done'
-	docker-compose up -d quay
+	DOCKER_USER="$$(id -u):0" docker-compose up -d quay
```
dmage, line 200: Some directories inside the container are group writable and they expect the group 0.

.gitignore
```diff
@@ -31,14 +31,20 @@ test/dockerclients/.*
...
+conf/mitm.cert
+conf/mitm.key
+conf/nginx/hosted-http-base.conf
+conf/nginx/http-base.conf
+conf/nginx/nginx.conf
+conf/nginx/rate-limiting.conf
+conf/nginx/server-base.conf
+conf/quay.kid
+conf/quay.pem
+conf/supervisord.conf
```
kleesc, line 47: Some of these files should not be ignored by git

Dockerfile
```diff
@@ -55,8 +55,6 @@ COPY requirements.txt .
 # Note that it installs into PYTHONUSERBASE because of the '--user'
 # flag.
 RUN set -ex\
-	; python3 -m pip install --no-cache-dir --quiet\
-		--upgrade setuptools pip\
```
kleesc, line 59: Why does this needs to be removed?

util/workers.py
kleesc: Worker count is based on the number of cores running. Shouldn't need to change this. The number of gunicorn workers running in the background should not matter for a high level, when developing on Quay

conf/supervisord.conf.jnj
```diff
@@ -27,15 +27,23 @@ events = PROCESS_LOG
 [program:blobuploadcleanupworker]
 environment=
   PYTHONPATH=%(ENV_QUAYDIR)s
-command=python -m workers.blobuploadcleanupworker.blobuploadcleanupworker
+command={% if hotreload -%}
+  gunicorn --timeout=600 -b 'unix:/tmp/blobuploadcleanupworker.sock' -c %(ENV_QUAYCONF)s/gunicorn_worker.py 'workers.blobuploadcleanupworker.blobuploadcleanupworker:create_gunicorn_worker()'
```
kleesc, line 31: It doesn't really make sense to use gunicorn in these workers. These workers are not serving requests, and will end up timing out listening.

docker-compose.yaml
flavianmissi: we're storing the files that script downloads in git, so no need to run it here

.gitignore
dmage: Can you be more specific?

conf/supervisord.conf.jnj
dmage: gunicorn provides the hot reload functionality. Some Quay engineers use it in their development env, and I'd say we want to keep it.

util/workers.py
dmage: There are two problems:

1. resource consumption by the local dev env
2. lack of the common SECRET_KEY in `local-dev`

By scaling down everything to a single replica I've got the same environment as the old `local-dev`, i.e. no predefined `SECRET_KEY` and only one replica with in-memory secret. If we are ok with increased resource consumption, we can add a static SECRET_KEY to the local-dev config like we did with DATABASE_SECRET_KEY.

Makefile
```diff
@@ -163,36 +163,51 @@ black:
...
+	# Waiting until the frontend is built...
+	# Use 'docker-compose logs -f local-dev-frontend' to see the progress
+	while ! test -e ./static/build/main-quay-frontend.bundle.js; do sleep 2; done
```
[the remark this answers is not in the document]
dmage: Comments are indented, so they'll be "executed" and printed to the console.
Building frontend isn't the fastest operation, so I'd like to avoid 10-20 messages "waiting for frontend"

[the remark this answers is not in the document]
flavianmissi: oh ofc! thanks for clarifying.

flavianmissi (approved):

PR #1062, opened 2022-01-14
[WIP]: Proxy Cache Config Data Schema Implementation

Sunandadadi:


Sunandadadi:
> Great! I think maybe we also want `upstream_registry_username` and `upstream_registry_password` nullable. What you think?

Yes, they can be null. I have `null=True` for both of them in the schema.

data/migrations/versions/10d34e6536f8_create_proxy_cache_config_table.py
```diff
@@ -0,0 +1,38 @@
...
+            ["user_id"],
+            ["user.id"],
+            name=op.f("fk_proxy_cache_config_user_id"),
+        ),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_proxy_cache_config"))
+    )
+
+
+def downgrade(op, tables, tester):
+    pass
```
flavianmissi, line 38: Should this remove the table entirely? Not sure if we tend to skip the `downgrade` or not, but in this case it should be easy enough 🤔

data/model/proxy_cache.py
```diff
@@ -0,0 +1,42 @@
...
+    return new_entry
+
+
+def get_proxy_cache_config_for_org(org_name):
+    """
+        Return the Proxy-Cache-Config associated with the given organization name, or None if it doesn't exist.
+    """
+    org = get_organization(org_name)
+    if not org:
+        return None
```
flavianmissi, line 40: `get_organization` will throw an exception if no org is found with the given name, so we can remove this check.
We might want to catch the exception and throw a different one, like `ProxyCacheConfigDoesNotExist` or something.
What you think?

data/model/user.py
```diff
@@ -796,6 +796,13 @@ def get_user(username):
         return None
 
 
+def get_organization_from_name(orgname):
```
flavianmissi, line 799: This function is no longer used, so can be safely removed now.

data/model/proxy_cache.py
```diff
@@ -0,0 +1,42 @@
...
+)
+from data.model.organization import get_organization
+
+
+def create_proxy_cache_config(
+    org_name,
+    upstream_registry,
+    upstream_registry_namespace,
+    upstream_registry_username,
+    upstream_registry_password,
```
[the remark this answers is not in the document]
Sunandadadi: Yes, I made a few changes that I didn't push yet. I am pushing this in the next commit.

data/migrations/versions/10d34e6536f8_create_proxy_cache_config_table.py
Sunandadadi: [downgrade](https://sqlalchemy-migrate.readthedocs.io/en/v0.7.2/versioning.html#:~:text=creating%20a%20table).-,downgrade,-()%20should%20reverse%20changes) should reverse changes made by upgrade. I've added `drop_table` here for now.

data/model/proxy_cache.py
Sunandadadi: I think we can remove the None check and throw the same exception raised by `get_organization`. Something like `ProxyCacheConfigDoesNotExist` might imply the proxy config does not exist. But, it is actually the org that does not exist because of which we are not able to create proxy config. 
 What would you say?

flavianmissi: That was my first thought too, but I think it might be useful to differentiate the errors by using different exceptions. 
With that said, we can go with your suggestion for now and change it if needed.

Sunandadadi: Sure, sounds good.

data/model/proxy_cache.py
```diff
@@ -0,0 +1,42 @@
...
+    staleness_period_s
+):
+    """
+        Creates proxy cache configuration for the given organization name
+    """
+    org = get_organization(org_name)
+    if not org:
+        return None
+
+    new_entry = ProxyCacheConfig.create(
```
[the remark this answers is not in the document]
Sunandadadi: Yes, I added it now.

Sunandadadi:
Closing this. Raised the same against the branch that has the proof of concept for Poxy Cache

PR #1069, opened 2022-01-24
requirements-osbs.txt: remove ipython

flavianmissi:


requirements-osbs.txt
```diff
@@ -169,8 +169,6 @@ Cython==0.29.23
 Pygments==2.9.0
```
kleesc, line 169: I'm not sure we need this whole section; `pygments, matplotlib, ...`

flavianmissi: good catch, remove the section along with some other test related deps spread throughout the file.

kleesc (approved):
LGTM. 
@thomasmckay Do you recall why these were included in the osbs Dockerfile in the first place?

PR #1071, opened 2022-01-24
storage: allow arbitrary endpoint url for azure storage (PROJQUAY-891)

kleesc:
Allows Azure endpoints other than global, such as `accountname.blob.core.usgovcloudapi.net`.

### DIFF
--- a/storage/azurestorage.py
+++ b/storage/azurestorage.py
@@ -40,7 +40,9 @@ _BLOCKS_KEY = "blocks"
 _CONTENT_TYPE_KEY = "content-type"
 
 
-AZURE_STORAGE_URL_STRING = "https://{}.blob.core.windows.net"
+AZURE_STORAGE_URL_STRING = {
+    "global": "https://{}.blob.core.windows.net",
+}
 
 
 class AzureStorage(BaseStorage):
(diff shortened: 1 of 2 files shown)


HammerMeetNail:
@kleesc Where does `endpoint_url` get set? Is that coming in from somewhere else like `config-tool`?

kleesc:
@HammerMeetNail It's read from the `config.yaml`. So we should be adding it to the config tool after this is merged.

storage/azurestorage.py
```diff
@@ -52,7 +54,7 @@ def __init__(
         azure_account_name,
         azure_account_key=None,
         sas_token=None,
-        connection_string=None,
+        endpoint_url=None,
```
bcaton85, line 57: Should this be something like `azure_endpoint_url` since this is azure specific?

kleesc: @bcaton85 I don't think it matters, since this will be under the azure specific storage config:
```
DISTRIBUTED_STORAGE_CONFIG:
  azure:
  - AzureStorage
  - azure_account_key: ...
    azure_account_name: ...
    azure_container: ...
    storage_path: /quayazuredata/quayregistry
    endpoint_url: ...
```

bcaton85 (approved):
LGTM

PR #1081, opened 2022-02-02
storage: allow configuration of storage region for cloud storage (PROJQUAY-3082)

kleesc:
Boto3 behaves unexpectedly when the resource client is not set to use
the correct region. Boto3 can't seem to correctly set the
X-Amz-Credential header when generating presigned urls if the region
name is not explicitly set, and will always fall back to us-east-1.
To reproduce this:
- Create a bucket in a different region from us-east-1 (e.g
eu-northeast-1)
- Create a boto3 client/resource without specifying the region
- Generate a presigned url
   
This seems to be a DNS issue with AWS that only happens shortly after
a bucket has been created, and resolves itself eventually. 
Ref:
- https://github.com/boto/boto3/issues/2989
- https://stackoverflow.com/questions/56517156/s3-presigned-url-works-90-minutes-after-bucket-creation

To workaround this, one can specify the bucket endpoint,
    either explicitly via endpoint_url, or by setting s3_region, which
    will be used to generate the bucket's virtual address.

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -731,6 +731,7 @@ class S3Storage(_CloudStorage):
         s3_bucket,
         s3_access_key=None,
         s3_secret_key=None,
+        s3_region=None,
         # Boto2 backward compatible options (host excluding scheme or port)
         host=None,
         port=None,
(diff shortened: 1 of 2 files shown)


Sunandadadi:
Looks good, can we add tests here?

Sunandadadi (dismissed):

Sunandadadi:
LGTM

Sunandadadi (approved):

Sunandadadi:
LGTM

PR #1083, opened 2022-02-03
mirrorrepositories: Add FEATURE_MIRROR_UNSIGNED_REGISTRIES (PROJQUAY-3106)

bcaton85:
Currently when attempting to mirror a registry containing unsigned images the mirror will fail due to not finding the source signature. This is caused by the updated version of Skopeo blocking unsigned images by default. This features permits unsigned images for a given list of repositories. The Skopeo version is also now pinned.


ibazulic:
Shouldn't this be just a UI option in the repo mirror config UI? Why do we need a feature flag?

util/config/schema.py
```diff
@@ -1204,5 +1204,13 @@
...
             "x-example": False,
         },
+        "FEATURE_MIRROR_UNSIGNED_REGISTRIES": {
+            "type": "array",
+            "description": "Disable image signature verification when mirroring repositories for the list of registries",
+            "x-example": [
+                "registry.redhat.io",
+                "registry.access.redhat.com"
+            ],
+        }
```
ibazulic, line 1214: This should not be a registry wide flag. This should be a per-repo decision made by the repo owner, which means a UI option in the repo mirror configuration UI. The option should also be added to the API request with which a repo mirror configuration can be created.

ibazulic (changes requested):

util/config/schema.py
bcaton85: In order to have Skopeo copy from unsigned registries it requires changing the `policy.json` file in the quay deployment, which effects the entire registry. There does exist the `--insecure-policy` flag which may work with individual repositories. Let me look into possibly getting that working.

ibazulic: The `--policy` switch uses a path to the file as an argument. Why not just add the code to change a path to the policy file depending on the switch in the UI? A much cleaner approach.

flavianmissi:
FTR you're getting the Flat commits because you're running the [flat action](https://github.com/bcaton85/quay/actions/workflows/aws-ip-json.yaml) on your fork and it's merging stuff to your fork's master branch.
Our own [flat action](https://github.com/quay/quay/actions/workflows/aws-ip-json.yaml) is currently disabled, that's why those commits aren't on our master

util/config/schema.py
```diff
@@ -1204,5 +1204,13 @@
             "description": "Automatically clean stale blobs leftover in the uploads storage folder from cancelled uploads",
             "x-example": False,
         },
+        "FEATURE_MIRROR_UNSIGNED_REGISTRIES": {
+            "type": "array",
+            "description": "Disable image signature verification when mirroring repositories for the list of registries",
+            "x-example": [
```
HammerMeetNail, line 1210: If someone wants to enable this for all repos, could they just pass in `["*"]`?

util/config/schema.py
ibazulic: That would be even better, an even cleaner design. But that again leaves us with a per-repo settings, not global wide settings. I don't think this is something that impacts the whole registry so that it requires a feature flag. We always mirror individual repositories anyway.

bcaton85: I agree with that design. At the time it seemed using the `--insecure-policy` was less secure but it actually has the same effect as modifying the policy.json file. Let me look into having the configuration per-repo.

PR #1085, opened 2022-02-03
imagemirror: Add unsigned registries mirror option (PROJQUAY-3106)

bcaton85:
Currently when attempting to mirror a registry containing unsigned images the mirror will fail due to not finding the source signature. This is caused by the updated version of Skopeo blocking unsigned images by default. This allows users to specify the ability to pull unsigned images per-repository. The Skopeo version is also now pinned.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN set -ex\
 		openssl \
 		python38 \
 		python3-gpg \
-		skopeo \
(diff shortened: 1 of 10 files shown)


kleesc (approved):
LGTM

flavianmissi:
@bcaton85 I see the JIRA is targeting 3.6.3 - do we want to cherry-pick this into `redhat-3.6`?

bcaton85:
yes

PR #1087, opened 2022-02-04
builders: Update boto to fix signature error (PROJQUAY-2542)

syed:
There is a bug in botocore which causes issues with signing
requests with awsv4. Latest version of Boto fixes it

https://github.com/boto/botocore/issues/2358 gives more context


kleesc (approved):
LGTM

PR #1092, opened 2022-02-07
dockerfile: ubi8 requires python38, otherwise installs 3.6 by default (PROJQUAY-3148)

kleesc:
- Specify Python38
- bump Pillow to 9.0.1


bcaton85 (approved):
LGTM

syed (approved):

PR #1098, opened 2022-02-08
imagemirror: Add unsigned registries mirror option (PROJQUAY-3106)

bcaton85:
Currently when attempting to mirror a registry containing unsigned images the mirror will fail due to not finding the source signature. This is caused by the updated version of Skopeo blocking unsigned images by default. This allows users to specify the ability to pull unsigned images per-repository. The Skopeo version is also now pinned.


kleesc (approved):
LGTM

PR #1099, opened 2022-02-08
buildman: fix kubernetes not returning correct running count (PROJQUAY-3169)

kleesc:
Filter the completed or failed jobs from the kubernetes api requests.


bcaton85 (approved):
LGTM

PR #1101, opened 2022-02-09
requirements-osbs.txt: Add missing deps for Cachito (PROJQUAY-3177)

HammerMeetNail:
* Adds previously removed but required downstream dependencies

### DIFF
--- a/requirements-osbs.txt
+++ b/requirements-osbs.txt
@@ -153,3 +153,8 @@ xhtml2pdf==0.2.4
 yapf==0.29.0
 zipp==2.1.0
 zope.interface==5.1.2
+
+# Build dependencies
+Cython==0.29.23
+toml==0.10.2
+jsonpickle==1.2


flavianmissi:
I _think_ version `1.0.0a0` of `funcparserlib` isn't present in cachito, and that's why it differs from `requirements.txt`. @harishsurf do you happen to remember this?
Would be nice if we could check whether a library version is present in cachito, I don't know if that's possible 🤔

flavianmissi (approved):
Giving the green light - let's keep an eye downstream in case `1.0.0a0` isn't around.

flavianmissi:
TIL!

flavianmissi (approved):

HammerMeetNail:
@flavianmissi Updated the PR based on the latest passing build. Needed to restore a few deps that were recently removed. They're dependencies of other dependencies.

flavianmissi (approved):
Thanks for fixing @HammerMeetNail ! LGTM

PR #1115, opened 2022-02-14
ui: display manifest list manifest sizes (PROJQUAY-3196)

kleesc:
Display individual manifest size in manifest list view.
Also fixes a non-cluster Redis cache bug.

TODO: This requires making extra individiual requests to get the
children manifests. They should probably be cached.

Screen Shot 2022-02-14 at 10 01 47 AM

### DIFF
--- a/data/cache/redis_cache.py
+++ b/data/cache/redis_cache.py
@@ -27,10 +27,10 @@ class ReadEndpointSupportedRedis(object):
             )
 
     def get(self, key, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


Sunandadadi:
Do you want to separate out fixing Redis cache bug in a different PR? It will be easier if we want to revert either one of the change (UI/Redis bug).

kleesc:
@Sunandadadi can you reapprove it

Sunandadadi (approved):

PR #1121, opened 2022-02-17
deploy: Add ignore validation for py3 deployment (PROJQUAY-2542)

syed:
Add an environment variable to the deployment to disable config
validation


harishsurf (approved):
lgtm

PR #1129, opened 2022-02-21
storage: handle cn-north-1 region (PROJQUAY-3082)

kleesc:
Handle S3 `cn-north-1`'s endpoint, which has a com.cn TLD.


bcaton85:
might want to add `cn-northwest-1` into that check too, [based on this doc](https://docs.aws.amazon.com/general/latest/gr/s3.html).

bcaton85 (approved):
LGTM

PR #1132, opened 2022-02-21
Add index for backfill

kwestpharedhat:


data/migrations/versions/022aef6a2ac7_add_index_for_backfill_workder.py
```diff
@@ -0,0 +1,21 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '022aef6a2ac7'
+down_revision = '909d725887d3'
+
+import sqlalchemy as sa
+
+
+def upgrade(op, tables, tester):
+    op.create_index("manifest_compressed_size_id", "manifest", ["layers_compressed_size", "id"], unique=False)
```
kleesc, line 17: Any reason for creating the index on both the compressed_size and id and not just compressed size?

kleesc:
Why are there 33 commits to this PR?

data/migrations/versions/022aef6a2ac7_add_index_for_backfill_workder.py
kwestpharedhat: There are 2 queries 1 that has only compressed_size is null and another that does a comparison on both id and compressed_size this index and the order the columns are defined in allows this index to serve both queries.

kwestpharedhat:
When I brought quay into my fork it was unable to fast forward so it got tagged with the commits that this fork was behind.

kleesc:
@kwestpharedhat Probably need to rebase only the migration file onto upstream's master branch. There shouldn't be any changes in this PR related to the aws-ip-json file. 

@syed Also, this will run a migration when deployed. Can we make sure it will be done in place?

kwestpharedhat:
I have already fixed the ranges json issue I didnt realize that was there. The only change should be the migration.

PR #1141, opened 2022-02-23
Use $remote_addr instead of $proxy_protocol_addr for rate limiting (P…

syed:
…ROJQUAY-3283)

We are not using proxy-protocol on quay.io anymore. Change that to
remote IP instead


conf/nginx/rate-limiting.conf.jnj
```diff
@@ -3,7 +3,7 @@
 # Define two buckets: Once for http1 connections (which we force to shard across our fleet) and
 # one for http2 connections (which will all hit the same node).
 map $http2 $http1_bucket {
-  ""      $proxy_protocol_addr; # HTTP1 case: use the IP address, since shared across nodes.
+  ""      $remote_addr; # HTTP1 case: use the IP address, since shared across nodes.
```
kleesc, line 6: This might break existing users of proxy protocol.
What we could do instead is define separate buckets specifically for the listening directives 7443 and 8443.

kleesc (changes requested):

PR #1142, opened 2022-02-23
buildman: Set build token expiration to builder's lifetime (PROJQUAY-3281)

kleesc:
Make sure the build token doesn't expire before the builder instance does.


syed:
The default I see is 2 hours for the max time. Is this an optimization for that or are we fixing anything here?

kleesc:
@syed We were using the wrong value (which was shorter that the configured builder run time).
This change makes sure that the token doesn't expire for the before that, by setting it to the max that a builder can run.

syed (approved):

PR #1147, opened 2022-02-25
sescan: prioritize scanning new pushes (PROJQUAY-3287)

syed:
Currently we randomly scan for entries to be indexed. This could take a
long time if we have a lot of unindexed images. Ideally, we want to
priortize scanning of newer images before we go to the old ones.

This PR creates a smaller batch of size `batch_size` from
`max_id - batch_size` to `max_id` to priortize scanning
those

### DIFF
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -202,11 +202,23 @@ class V4SecurityScanner(SecurityScannerInterface):
 
         # TODO(alecmerdler): We want to index newer manifests first, while backfilling older manifests...
         iterator = itertools.chain(
+            # Guarantees at least one batch of most recent manifests
             yield_random_entries(
                 not_indexed_query,
                 Manifest.id,
                 batch_size,
                 max_id,
+                # Because we have multiple workers and must stay random to ensure minimal duplicative work
+                # we will instead limit the set worked to the most recent batch.
+                max_id - batch_size,
+            ),
+            # Could be a huge batch of work
+            yield_random_entries(
+                not_indexed_query,
+                Manifest.id,
+                batch_size,
+                # compensation for the above batch guarentee
+                max_id - batch_size,
                 min_id,
             ),
             yield_random_entries(


syed:
@kwestpharedhat  I didn't have permission to push to your fork so I closed your PR (#1145 ) and opened this one with your changes

data/secscan_model/secscan_v4_model.py
```diff
@@ -202,11 +202,23 @@ def needs_reindexing_query(indexer_hash):
...
+            # Guarantees at least one batch of most recent manifests
             yield_random_entries(
                 not_indexed_query,
                 Manifest.id,
                 batch_size,
                 max_id,
+                # Because we have multiple workers and must stay random to ensure minimal duplicative work
+                # we will instead limit the set worked to the most recent batch.
+                max_id - batch_size,
+            ),
```
kleesc, line 214: Won't this stop doing any work once `(max_id - batch_size) ... max_id` is done?

data/secscan_model/secscan_v4_model.py
```diff
@@ -202,11 +202,23 @@ def needs_reindexing_query(indexer_hash):
...
+            # Could be a huge batch of work
+            yield_random_entries(
+                not_indexed_query,
+                Manifest.id,
+                batch_size,
+                # compensation for the above batch guarentee
+                max_id - batch_size,
```
kleesc, line 221: Shouldn't this second iterator be left to `min_id ... max_id`. We still want to backfill the older manifests. Adding the iterator from `max_id - batch_size ... max_id` before this one should prioritize the new ones.

data/secscan_model/secscan_v4_model.py
syed: Yes but `max_id` keeps getting updated every time the worker does a `perform_indexing` call

data/secscan_model/secscan_v4_model.py
syed: The previous block was from `min_id -> max_id`  We are splitting it into two `min_id -> max_id - batch_size` and from `max_id - batch_size -> max_id` and the latter is the first one to be scanned

data/secscan_model/secscan_v4_model.py
kleesc: Isn't it the `min_id` that gets updated? 
i.e the `max_id` if the last batch becomes the new `min_id`

syed: both get updated
```
        min_id = (
            start_token.min_id
            if start_token is not None
            else Manifest.select(fn.Min(Manifest.id)).scalar()
        )
        max_id = Manifest.select(fn.Max(Manifest.id)).scalar()
```

kleesc (approved):

PR #1152, opened 2022-02-28
chore: Remove unneeded flags fromt he config schema (PROJQUAY-3309)

ibazulic:
The flag `CREATE_REPOSITORY_ON_PUSH_PUBLIC` is not referenced anywhere in the code. This PR removes it along with its entry in the schema.


kleesc (approved):
LGTM. `CREATE_PRIVATE_REPO_ON_PUSH` is the correct flag.

PR #1153, opened 2022-02-28
database: force close existing non-pooled connections before transaction (PROJQUAY-3303)

kleesc:
This hack is for handling possible MySQL closing idle
connections. Peewee/pymysql's will try to reuse the existing
connection after the MySQL server has already closed it, which will
return an InterfaceError. AFAIK, there isn't a way to actually test
for a a stale connection without actually running a query. Closing the
connection before the transaction forces peewee/pymysql to reopen a
new session to MySQL. This should only applies to non-registry
workers, as the registry workers use a pool by default, and shouldn't
have this issue. Follow-up to 58b06572 (#1148).

### DIFF
--- a/config.py
+++ b/config.py
@@ -208,6 +208,16 @@ class DefaultConfig(ImmutableConfig):
 
     @staticmethod
     def create_transaction(db):
+        # This hack is for handling possible MySQL closing idle connections.
+        # Peewee/pymysql's will try to reuse the existing connection after the MySQL server
+        # has already closed it, which will return an InterfaceError.
+        # AFAIK, there isn't a way to actually test for a a stale connection without actually
+        # running a query. Closing the connection before the transaction forces peewee/pymysql
+        # to reopen a new session to MySQL. This should only applies to non-registry workers,
+        # as the registry workers use a pool by default, and shouldn't have this issue.
+        if type(db.obj).__name__ == "ObservableRetryingMySQLDatabase":
+            db.close()
+
         return db.transaction()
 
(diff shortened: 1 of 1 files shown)


config.py
```diff
@@ -208,6 +208,16 @@ class DefaultConfig(ImmutableConfig):
...
     @staticmethod
     def create_transaction(db):
+        # This hack is for handling possible MySQL closing idle connections.
+        # Peewee/pymysql's will try to reuse the existing connection after the MySQL server
+        # has already closed it, which will return an InterfaceError.
+        # AFAIK, there isn't a way to actually test for a a stale connection without actually
+        # running a query. Closing the connection before the transaction forces peewee/pymysql
+        # to reopen a new session to MySQL. This should only applies to non-registry workers,
+        # as the registry workers use a pool by default, and shouldn't have this issue.
+        if type(db.obj).__name__ == "ObservableRetryingMySQLDatabase":
```
syed, line 218: is this problem specific to Mysql? Where is the string `ObservableRetryingMySQLDatabase` coming from? Can we use `isinstance` instead ?

kleesc: We're using the class name as we can't easily check the for the instance type as they're defined dynamically at runtime:
- https://github.com/quay/quay/blob/master/data/database.py#L380-L381
- https://github.com/quay/quay/blob/master/data/database.py#L465-L468

From what I can tell, this is is a problem specific how MySQL cleans up idle connection based on `wait_timeout`, and peewee/PyMySQL seemingly unable to handle it gracefully.

syed (approved):

PR #1156, opened 2022-02-28
secscan: make batch_size configurable (PROJQUAY-3287)

syed:
For larger databases like quay.io the default batch
size is too high to complete the job in a resonable
time. Make the batch_size configurable so we can
adjust how many scans can be in-flight for a worker


kleesc (approved):
LGTM

PR #1171, opened 2022-03-08
Fix: Adding default vault to quota parameter (PROJQUAY-0000)

Sunandadadi:


util/ipresolver/aws-ip-ranges.json
```diff
@@ -46895,4 +46895,4 @@
       "network_border_group": "us-west-2"
     }
   ]
-}
\ No newline at end of file
+}
```
kwestpharedhat, line 46898: Can we revert this file?

static/partials/org-view.html
```diff
@@ -140,7 +140,9 @@ <h3>Organization Settings</h3>
 
               <time-machine-settings organization="organization"></time-machine-settings>
 
-              <quota-management-view quay-show="Features.QUOTA_MANAGEMENT" organization="organization" disabled="true"></quota-management-view>
+              <div quay-show="Features.QUOTA_MANAGEMENT">
```
kwestpharedhat, line 143: Is there a reason to switch to a div rather than the show parameter being inside the view object?

kwestpharedhat (dismissed):

util/ipresolver/aws-ip-ranges.json
Sunandadadi: Yes, done.

kwestpharedhat (approved):

PR #1174, opened 2022-03-09
billing: use BytesIO when rendering invoice (PROJQUAY-3267)

syed:
Migrating to py3 broke invoices. xhtml2pdf returns `bytes` instead of `str`


kleesc (approved):
LGTM

PR #1178, opened 2022-03-10
Quota bug fixes

kwestpharedhat:


data/migrations/versions/e9f3e4dbb979_quota_management_reporting.py
```diff
@@ -20,6 +20,7 @@ def upgrade(op, tables, tester):
         sa.Column("id", sa.Integer, nullable=False),
         sa.Column("namespace_id", sa.Integer, nullable=False),
         sa.Column("limit_bytes", sa.BigInteger, nullable=False),
+        sa.Column("set_by_super", sa.Boolean, nullable=False, default=0),
```
kleesc, line 23: Rename this to just `superuser`. Default should be `False`

config.py
```diff
@@ -209,22 +208,6 @@ class DefaultConfig(ImmutableConfig):
...
-        # as the registry workers use a pool by default, and shouldn't have this issue.
-        if type(db.obj).__name__ == "ObservableRetryingMySQLDatabase":
-            try:
-                db.close()
-            except:
-                # Only try to close the connection. Otherwise closing connections in a nested transaction
-                # will return an OperationalError. In that case, we can just continue with the normal flow,
-                # as we know the connection is likely in use and not stale.
-                pass
-
```
kleesc, line 227: This is an unrelated fix from another PR.

data/database.py
```diff
@@ -774,6 +774,7 @@ class QuotaType(BaseModel):
 class UserOrganizationQuota(BaseModel):
     namespace_id = QuayUserField(index=True, unique=True)
     limit_bytes = BigIntegerField()
+    set_by_super = BooleanField()
```
kleesc, line 777: The alembic migration also needs to be included

kleesc (changes requested):

PR #1179, opened 2022-03-10
new column

kwestpharedhat:


data/migrations/versions/480d2fef20fc_add_quota_super_column.py
```diff
@@ -0,0 +1,23 @@
...
+revision = '480d2fef20fc'
+down_revision = 'e9f3e4dbb979'
+
+import sqlalchemy as sa
+
+
+def upgrade(op, tables, tester):
+    op.add_column(
+        "userorganizationquota", sa.Column("set_by_super", sa.Boolean, nullable=False, default=0)
+    )
```
kleesc, line 19: Can you also add a `uuid` column to the `userorganizationquota` and `quotalimit` tables. 
There currently isn't a good unique identifier to these resource, apart from the database id.

kwestpharedhat: the database ID is the primary key and a UUID would not provide any better uniqueness guarentees. The id is also linked to the user table. Adding a UUID column would only cause a slowing in data insertion.

HammerMeetNail:
@kwestpharedhat Can this be closed out?

PR #1181, opened 2022-03-11
ui: prevent manifest lists' manifests from loading indefinitely (PROJQUAY-3398)

kleesc:


static/js/directives/repo-view/repo-panel-tags.js
```diff
@@ -484,8 +484,13 @@ angular.module('quay').directive('repoPanelTags', function () {
           child_manifest = JSON.parse(resp['manifest_data']);
           manifest.layers = child_manifest["layers"];
           manifest.layers_loading = false;
-        }, ApiService.errorDisplay('Could not load manifest contents'))
-      };
+        }, ApiService.errorDisplay(
+	  'Could not load manifest contents',
```
flavianmissi, line 488: When quay fails to load multiple child manifests, this error is displayed multiple times.
What do you think of logging an error to the console instead? In practice this would mean that only sub manifests successfully loaded would have their sizes shown.
That's what ended up doing [here](https://github.com/quay/quay/pull/1053/files#diff-efcbc4c4299a4d2a733e12589414a4b2fcdb2cb1753a2f4a059cdb5ac72d264e)

flavianmissi:
This has been fixed by https://github.com/quay/quay/pull/1053, closing it

PR #1184, opened 2022-03-14
refactor: move quota logic from api endpoints

flavianmissi:
This makes it easier for the pull-thru cache code to incorporate the same logic where needed.

### DIFF
--- a/data/model/namespacequota.py
+++ b/data/model/namespacequota.py
@@ -33,6 +33,8 @@ def verify_namespace_quota(repository_ref):
 def verify_namespace_quota_force_cache(repository_ref):
     force_cache_repo_size(repository_ref)
     namespace_size = get_namespace_size(repository_ref.namespace_name)
+    if namespace_size is None:
+        namespace_size = 0
     return check_limits(repository_ref.namespace_name, namespace_size)
 
 
--- a/data/registry_model/registry_oci_model.py
+++ b/data/registry_model/registry_oci_model.py
@@ -40,6 +40,7 @@ from image.docker.schema1 import (
 )
 from image.docker.schema2 import EMPTY_LAYER_BLOB_DIGEST, EMPTY_LAYER_BYTES
 from util.timedeltastring import convert_to_timedelta
+import features
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


data/model/namespacequota.py
```diff
@@ -33,6 +33,8 @@ def verify_namespace_quota(repository_ref):
 def verify_namespace_quota_force_cache(repository_ref):
     force_cache_repo_size(repository_ref)
     namespace_size = get_namespace_size(repository_ref.namespace_name)
+    if namespace_size is None:
+        namespace_size = 0
```
kwestpharedhat, line 37: if this every happens throw error dont send 0 it will result in a non-functional hard to debug error.

data/registry_model/registry_oci_model.py
```diff
@@ -350,6 +351,14 @@ def create_manifest_and_retarget_tag(
         raise_on_error is set to True, in which case a CreateManifestException may also be
         raised.
         """
+        if features.QUOTA_MANAGEMENT:
```
kwestpharedhat, line 354: This will likely have unintented behavior due to wide spread usage of this function. Would need to have through tests built around it to ensure functionality that should work even above quota still does.

data/registry_model/registry_oci_model.py
```diff
@@ -836,6 +848,12 @@ def create_blob_upload(self, repository_ref, new_upload_id, location_name, stora
 
         If the blob upload could not be created, returns None.
         """
+        if features.QUOTA_MANAGEMENT:
```
kwestpharedhat, line 851: This is also likely to have unintended behavior

data/model/namespacequota.py
kwestpharedhat: That one is correct however the one above it is incorrect in its design. That one is not assuming that the previous line has created the sizing and does have the option of valid 0.

data/registry_model/registry_oci_model.py
flavianmissi: Can you elaborate on this function's wide spread usage?
I'm only seeing it used in 2 places (beyond tests):
* endpoints/v2/manifest.py
* endpoints/v1/tag.py (indirectly via `data.registry_model.manifestbuilder.commit_tag_and_manifest`

PR #1185, opened 2022-03-15
superuser: Fix display of build logs (PROJQUAY-3404)

bcaton85:
Build logs in the superuser admin panel do not show up. Errors out due to some syntax errors on the backend which this PR resolves.

- Checks for presence of `resource_key` before constructing archive URL
- Correctly passes a `RepositoryBuildTrigger` object to `BuildTriggerHandler.get_handler()`


kleesc (approved):
LGTM

PR #1189, opened 2022-03-18
WIP: Quay DB Double Write

kwestpharedhat:


data/readreplica.py
```diff
@@ -71,6 +71,33 @@ def execute_sql(self, sql, params=None, commit=SENTINEL):
...
+    def __getattr__(self, attribute):
+        return super(AutomaticFailoverWrapper, self).__getattr__(attribute)
+
+    def execute(self, query, commit=SENTINEL, **context_options):
+        ctx = self.get_sql_context(**context_options)
+        sql, params = ctx.sql(query).query()
+        return self.execute_sql(sql, params, commit=commit)
+
+    def execute_sql(self, sql, params=None, commit=SENTINEL):
+        result = super(AutomaticFailoverWrapper, self).execute_sql(sql, params, commit)
```
syed, line 89: Does `execute_sql` execute a full transaction or can we have cases like `START TRANSACTION` and `COMMIT` are spred over multiple `execute_sql` calls?

kwestpharedhat: I believe the transaction is outside of this for the time being. That would be something to add to allow a rollback of the primary db when we are ready to tie both database commits together.

syed: How do you make sure that the primary key IDs are same across both the database. If the sql adds a new row in DB1 how are you making sure the same ID is going to be used for DB2

kwestpharedhat: Currently we have no way to guarantee the values stay the same, thus the argument is weak, however the nature of the schema will want to stay in sync without failure writes, which in phase 2 we will guarantee if 1 write fails they both do. Secondly we only join on ID and find it with references through other means which means that in theory even if the IDs become different the reads will still succeed as they did before.

data/database.py
```diff
@@ -74,6 +74,7 @@
     "sqlite": schemedriver(SqliteDatabase, PooledSqliteDatabase),
     "postgresql": schemedriver(PostgresqlDatabase, PooledPostgresqlDatabase),
     "postgresql+psycopg2": schemedriver(PostgresqlDatabase, PooledPostgresqlDatabase),
+    "cockroachdb": schemedriver(PostgresqlDatabase, PooledPostgresqlDatabase),
```
syed, line 77: Are we doing anything special for cockroachDB? Can we treat it as a generic postgres?

data/migrations/versions/e2894a3a3c19_add_full_text_search_indexing_for_repo_.py
```diff
@@ -12,28 +12,31 @@
...
-        "repository_name__fulltext",
-        "repository",
-        ["name"],
-        unique=False,
-        postgresql_using="gin",
-        postgresql_ops={"name": "gin_trgm_ops"},
-        mysql_prefix="FULLTEXT",
-    )
+    conn = op.get_bind()
+    if conn.engine.name != "cockroachdb":
```
syed, line 21: Do we not need this for cockroach?

data/readreplica.py
```diff
@@ -71,6 +70,32 @@ def execute_sql(self, sql, params=None, commit=SENTINEL):
...
+    def execute(self, query, commit=SENTINEL, **context_options):
+        ctx = self.get_sql_context(**context_options)
+        sql, params = ctx.sql(query).query()
+        return self.execute_sql(sql, params, commit=commit)
+
+    def execute_sql(self, sql, params=None, commit=SENTINEL):
+        result = super().execute_sql(sql, params, commit)
+        # TODO: remove try when we tie primary and secondary together under transaction
+        #try:
+        self._secondary_write_db.execute_sql(sql, params, commit)
```
syed, line 91: null check for `_secondary_write_db`

data/readreplica.py
```diff
@@ -112,20 +137,21 @@ def _select_database(cls):
         # Select the master DB if read replica support is not enabled.
         read_only_config = cls._read_only_config()
         if not read_only_config.read_replicas:
-            return cls._meta.database
+            return DoubleWriteWrapper( cls._meta.database, None, cls._meta.double_write_database)
```
syed, line 140: instead of wrapping every invocation with the `DoubleWriteWrapper` manybe only do it if `cls._meta.double_write_database` at one place

data/migrations/versions/e2894a3a3c19_add_full_text_search_indexing_for_repo_.py
kwestpharedhat: Its not supported by cockroach

data/readreplica.py
kwestpharedhat: this is in the cleanup it was removed during debugging.

data/readreplica.py
kwestpharedhat: this is planned in the cleanup.

data/database.py
kwestpharedhat: unfortuantly we cannot during the migration especially as sqlalchemy gets mad.

PR #1193, opened 2022-03-22
Import ABC from collections.abc for Python 3.10 compatibility

tirkarthi:
Import ABC from `collections` was deprecated and removed in Python 3.10. Use `collections.abc` .


kleesc (approved):
LGTM. Thanks for doing this.

PR #1200, opened 2022-03-24
proxy: don't store entire blob in memory when caching

flavianmissi:
also uses blob uploader to upload the blob to storage.

[PROJQUAY-3459](https://issues.redhat.com/browse/PROJQUAY-3459)

### DIFF
--- a/proxy/__init__.py
+++ b/proxy/__init__.py
@@ -3,16 +3,13 @@ The proxy module provides the means to proxy images from other registry instance
 Registries following the distribution spec are supported.
 """
 from __future__ import annotations
-from typing import Any
 import re
-import json
 
 import requests
 
 from app import model_cache
 from data.cache import cache_key
 from data.database import ProxyCacheConfig
-from digest.digest_tools import sha256_digest
 
 
 WWW_AUTHENTICATE_REGEX = re.compile(r'(\w+)[=] ?"?([^",]+)"?')
@@ -94,12 +91,9 @@ class Proxy:
         resp = self.get(
             url,
             allow_redirects=True,
+            stream=True,
         )
-        size = resp.headers.get("content-length")
-        content_digest = sha256_digest(resp.content)
-        if digest != content_digest:
-            UpstreamRegistryError(f"digest mismatch, expected {digest}, got {content_digest}")
-        return resp.content, size
+        return resp
 
     def blob_exists(self, digest: str):
         url = f"{self.base_url}/v2/{self._repo}/blobs/{digest}"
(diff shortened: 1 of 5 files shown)


proxy/__init__.py
```diff
@@ -94,12 +91,10 @@ def get_blob(self, digest: str):
...
             url,
             allow_redirects=True,
+            stream=True,
         )
-        size = resp.headers.get("content-length")
-        content_digest = sha256_digest(resp.content)
-        if digest != content_digest:
-            UpstreamRegistryError(f"digest mismatch, expected {digest}, got {content_digest}")
-        return resp.content, size
+        size = resp.headers.get("content-length", -1)
```
kleesc, line 96: What's the content-length in this case?

flavianmissi: It's the size of the blob in bytes

proxy/__init__.py
```diff
@@ -94,12 +91,10 @@ def get_blob(self, digest: str):
         resp = self.get(
             url,
             allow_redirects=True,
+            stream=True,
```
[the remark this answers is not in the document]
flavianmissi: Good question. The connection is kept open until the body is fully read, but it's unclear to me what happens if there's an error mid-way through reading the stream... I need to do some more digging.
See https://docs.python-requests.org/en/latest/user/advanced/#body-content-workflow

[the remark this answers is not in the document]
flavianmissi: It doesn't look like an error during read would cause the stream to close. I'm thinking we can add a finally to the try/except block calling `_download_blob`, calling either `stream.close()`[1], `stream.release_conn()`[2], or both (?).
What you think?
[1] https://urllib3.readthedocs.io/en/latest/reference/urllib3.response.html#urllib3.response.HTTPResponse.close
[2] https://urllib3.readthedocs.io/en/latest/reference/urllib3.response.html#urllib3.response.HTTPResponse.release_conn

proxy/__init__.py
kleesc: Even for a stream?

proxy/__init__.py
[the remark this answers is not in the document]
kleesc: Yea, looks like you need to call close on errors. You could have a context manager to handle this for you.

proxy/__init__.py
flavianmissi: Either the `Content-Length` header is the size of the blob in bytes or it will not be present.
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#directives

kleesc (approved):

PR #1201, opened 2022-03-25
ui: Copy build logs error fix (PROJQUAY-3405)

bcaton85:
Currently when viewing the build logs and selecting `Copy Logs` the copy to clipboard will fail. This is because the content of the logs was being passed to the copy-clipboard function when it requires the ID of the element containing the text to copy. This passes the selector of the ID of the element containing the build logs.


kleesc (approved):
lgtm

PR #1206, opened 2022-03-29
endpoints/v2/errors: remove unused non-standard exceptions

flavianmissi:
removes unused exceptions that are also not listed in the distribution
specification.
see https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes
for details.

### DIFF
--- a/endpoints/v2/errors.py
+++ b/endpoints/v2/errors.py
@@ -86,13 +86,6 @@ class TagExpired(V2RegistryException):
         super(TagExpired, self).__init__("TAG_EXPIRED", message or "Tag has expired", detail, 404)
 
 
-class ManifestUnverified(V2RegistryException):
-    def __init__(self, detail=None):
-        super(ManifestUnverified, self).__init__(
-            "MANIFEST_UNVERIFIED", "manifest failed signature verification", detail
-        )
-
-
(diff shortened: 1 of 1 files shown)


ibazulic:
These do exist and are part of the Docker specification:

https://github.com/distribution/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md

```
MANIFEST_UNVERIFIED | manifest failed signature verification | During manifest upload, if the manifest fails signature verification, this error will be returned.
TAG_INVALID | manifest tag did not match URI | During a manifest upload, if the tag in the manifest does not match the uri tag, this error will be returned.
```

flavianmissi:
@ibazulic are we following the docker specification or the OCI one? I know they intentionally overlap, but I thought we were following the OCI.

flavianmissi:
@ibazulic are we following the docker specification or the OCI one? I know they intentionally overlap, but I thought we were following the OCI.

ibazulic:
As far as I know, we claim full compatibility with both Docker v2 API and OCI. To me, that means we follow both. There are subtle differences between the two and Docker is *still* the most widespread container engine around.

flavianmissi:
gotcha!

PR #1208, opened 2022-03-29
schema1: Generate kid in the correct format (PROJQUAY-3486)

bcaton85:
Clients expect signed schema 1 manifests to contain a signature with the key ID in the correct format. We currently pass the automatically generated key ID which is just a hash of characters.
This causes errors in clients that try to pull signed schema 1 manifests that are generated by Quay - when creating tags through the UI or manifest conversions.
The expected format of the Key ID can be found here: https://docs.docker.com/registry/spec/auth/jwt/


flavianmissi:
@bcaton85 looks like you need to run `make black` - that'll fix the format job.

kleesc (approved):
LGTM

kleesc (approved):

PR #1213, opened 2022-03-30
endpoints/v2: handle generic proxy related errors

flavianmissi:
also add upstream registry to proxy cache key. if a user changes their
mind about what registry to proxy, we shouldn't try to reuse the token.

----

this can be tested by setting up a proxy org and specifying the wrong credentials for the proxied registry. see https://gist.github.com/flavianmissi/48a7bd19270fd12876c488fa3332fb3b#proxy--auth-for-quayio-namespace for step-by-step setup
the client should see something like:
```
Error: initializing source docker://quay:8080/fmissi-cache/busybox-v2s1:latest: reading manifest latest in quay:8080/fmissi-cache/busybox-v2s1: unknown: Error requesting upstream registry: Failed to get token from 'https://quay.io/v2/auth?service=quay.io&scope=repository:fmisse/busybox-v2s1:pull', 429
```
(in my case I'm getting a 429 due to too many attempts to login with the wrong credentials 🤷)

### DIFF
--- a/endpoints/v2/__init__.py
+++ b/endpoints/v2/__init__.py
@@ -27,11 +27,13 @@ from endpoints.v2.errors import (
     Unsupported,
     NameUnknown,
     ReadOnlyMode,
+    InvalidRequest,
 )
 from util.http import abort
 from util.metrics.prometheus import timed_blueprint
 from util.registry.dockerver import docker_version
 from util.pagination import encrypt_page_token, decrypt_page_token
+from proxy import UpstreamRegistryError
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


endpoints/v2/__init__.py
```diff
@@ -58,6 +60,15 @@ def handle_readonly(ex):
     return response
 
 
+@v2_bp.app_errorhandler(UpstreamRegistryError)
+def handle_readonly(error):
+    e = InvalidRequest(message=str(error))
```
flavianmissi, line 65: `InvalidRequest` returns an error with `INVALID_REQUEST` code. Although that error code is not part of the distribution spec, it's used in quite a few places throughout quay's v2 api.
we can definitely use another error here, I'm just not sure which one.
suggestions are most welcome.

dmage (dismissed):

PR #1215, opened 2022-03-30
registry_proxy_model: count repository size when caching images (PROJQUAY-3464)

flavianmissi:
also check if the org is within its quota limits before caching an image
from the upstream registry.

### DIFF
--- a/data/registry_model/registry_proxy_model.py
+++ b/data/registry_model/registry_proxy_model.py
@@ -16,9 +16,11 @@ from data.database import (
 )
 from data.model import (
     oci,
+    namespacequota,
     RepositoryDoesNotExist,
     ManifestDoesNotExist,
     TagDoesNotExist,
+    QuotaExceededException,
 )
 from data.model.repository import get_repository, create_repository
 from data.model.proxy_cache import get_proxy_cache_config_for_org
@@ -249,6 +251,19 @@ class ProxyModel(OCIModel):
 
         return tag
 
+    def _recalculate_repository_size(self, repo_ref: RepositoryReference) -> None:
+        if features.QUOTA_MANAGEMENT:
+            namespacequota.force_cache_repo_size(repo_ref)
+
+    def _enforce_repository_quota(self, repo_ref: RepositoryReference) -> None:
+        if features.QUOTA_MANAGEMENT:
+            quota = namespacequota.verify_namespace_quota(repo_ref)
+            if quota["severity_level"] == "Warning":
+                namespacequota.notify_organization_admins(repo_ref, "quota_warning")
+            elif quota["severity_level"] == "Reject":
+                namespacequota.notify_organization_admins(repo_ref, "quota_error")
+                raise QuotaExceededException
+
     def _create_and_tag_manifest(
         self,
         repo_ref: RepositoryReference,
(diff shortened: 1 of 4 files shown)


data/registry_model/registry_proxy_model.py
```diff
@@ -469,6 +483,7 @@ def _download_blob(self, repo_ref: RepositoryReference, digest: str) -> int:
         Download blob from upstream registry and perform a monolitic upload to
         Quay's own storage.
         """
+        self._enforce_repository_quota(repo_ref)
```
kwestpharedhat, line 486: running this prior to actually doing the download will result in stale data

data/registry_model/test/test_registry_proxy_model.py
```diff
@@ -664,7 +664,7 @@ def test_renew_manifest_and_parent_tag_when_manifest_is_child_of_manifest_list(
         assert updated_tag.lifetime_end_ms > manifest_tag.lifetime_end_ms
         assert updated_list_tag.lifetime_end_ms > manifest_list_tag.lifetime_end_ms
 
-    def test_update_manifest_bytes_when_manifest_is_placeholder(
+    def test_update_manifest_bytes_and_layers_compressed_size_when_manifest_is_placeholder(
```
kwestpharedhat, line 667: is there a way to shorten this function name? This name by itself is over 80 characters which violates most style guides.

endpoints/v2/__init__.py
```diff
@@ -61,8 +63,19 @@ def handle_readonly(ex):
 
 
 @v2_bp.app_errorhandler(UpstreamRegistryError)
-def handle_proxy_cache_errors(error):
+def handle_proxy_cache_error(error):
```
kwestpharedhat, line 66: This and the following function are very similar perhaps make the response formatting a function and reuse that way.

data/registry_model/registry_proxy_model.py
```diff
@@ -266,6 +277,7 @@ def _create_and_tag_manifest(
         or the retrieved manifest is invalid (for docker manifest schema v1).
         """
         self._proxy.manifest_exists(manifest_ref, ACCEPTED_MEDIA_TYPES)
+        self._enforce_repository_quota(repo_ref)
```
kwestpharedhat, line 280: any form of force caching should be the final thing to do. If you need a check to ensure the quota isnt exceeded at the beginning do not use force cache for performance and server load reduction.

kwestpharedhat (changes requested):
Changes requested

data/registry_model/test/test_registry_proxy_model.py
flavianmissi: done

endpoints/v2/__init__.py
flavianmissi: good point - done

data/registry_model/registry_proxy_model.py
kwestpharedhat: that order of operations would be correct yes. 

Also yes you may be correct that the order of operations in manifest v2 may be incorrect I will re-evaluate that flow.

kwestpharedhat (approved):

PR #1226, opened 2022-04-04
secscan: add a global lock on security worker runs (PROJQUAY-3501)

kleesc:
Add a global lock on security worker iterations, based on the value of
the current pagination token. This is to  avoid multilple worker
processes possibly doing work on the same set of manifests.


workers/securityworker/securityworker.py
```diff
@@ -29,7 +30,15 @@ def __init__(self):
...
-        self._next_token = self._model.perform_indexing(self._next_token)
+        try:
+            with GlobalLock(self._next_token.normalize(), lock_ttl=DEFAULT_INDEXING_INTERVAL + 60):
+                self._next_token = self._model.perform_indexing(self._next_token)
+        except LockNotAcquiredException:
+            logger.debug(
+                "Could not acquire global lock for security worker indexing: %s",
+                self._next_token.normalize(),
+            )
+            self._next_token = self._next_token.next_page()
```
hdonnay, line 41: So the idea is that all the workers will race on the lock, and losers race on the next token on the next interval? I think that means that means no matter how many workers there are, there will only be one started every interval.

workers/securityworker/securityworker.py
```diff
@@ -29,7 +30,15 @@ def __init__(self):
         self.add_operation(self._index_in_scanner, interval)
 
     def _index_in_scanner(self):
-        self._next_token = self._model.perform_indexing(self._next_token)
+        try:
+            with GlobalLock(self._next_token.normalize(), lock_ttl=DEFAULT_INDEXING_INTERVAL + 60):
```
hdonnay, line 34: How does this TTL work? If it's only needed to adjudicate race and not while the operation is in flight, this is probably fine. If it's needed for the whole operation, this is probably too low. 90sec is probably one request out of the batch.

kleesc: Ideally it should be as long as the indexing operation lasts for that batch. Otherwise, you run into the possibility that another worker might try and do duplicative work on the same batch.

workers/securityworker/securityworker.py
kleesc: Right, so in theory, `x1 -> x2` would be reserved for one worker, the next one being `x2+1 -> x2+1+batch_size`.
Wondering about the interval using the batch size though. Say from range 100 to 1000, if the auto increment is still below the 1000 (i.e new manifests have ids lower than that) while a lock is on that range, then some manifests might just get missed (until is starts over from the beginning of the table again).

workers/securityworker/securityworker.py
hdonnay: So, with a batch size of ~30k (seen on quayio) and response time of 10 sec (eyeballing the latency information from quayio's clair) means the timeout should be about 83 hours.

syed: `DEFAULT_INDEXING_INTERVAL` is set to 30s by default.  From the above what I am understanding is that we will try to acquire a lock and if we get blocked we will wait for the TTL. If we cannot acquire that lock, we will assume that this block is being worked on by some other woker and proceed to acquire then next batch of work.

data/secscan_model/secscan_v4_model.py
```diff
@@ -196,9 +204,8 @@ def needs_reindexing_query(indexer_hash):
             )
 
         # 4^log10(total) gives us a scalable batch size into the billions.
-        batch_size = self.app.config.get(
-            "SECURITY_SCANNER_V4_BATCH_SIZE", int(4 ** log10(max(10, max_id - min_id)))
```
syed, line 200: Are you removing config param batch size from here? Where are supplying that from?

kleesc: It's being passed as a param, since the caller is already looking at `SECURITY_SCANNER_V4_BATCH_SIZE`.

workers/securityworker/securityworker.py
kleesc: @syed Yes. If a worker cannot acquire a lock, it should try on the next "page". 
@hdonnay I updated the lock library, to one that supports auto renewal, as long as the context hasn't finished.

data/secscan_model/secscan_v4_model.py
hdonnay: good change, as this was confusing when reading this the other day.

hdonnay (commented):
SGTM

syed (approved):

PR #1228, opened 2022-04-05
chore: Pin to an older version of setuptools (PROJQUAY-3184)

HammerMeetNail:
* Attempt to fix downstream builds


Sunandadadi (approved):
LGTM

PR #1239, opened 2022-04-11
quayio: Add export compliance service to Red Hat SSO (PROJQUAY-2056)

jonathankingfc:
- Add FEATURE_EXPORT_COMPLIANCE to config schema
- Add call to export compliance service during OAuth workflow when feature is enabled

### DIFF
--- a/endpoints/oauth/login.py
+++ b/endpoints/oauth/login.py
@@ -1,14 +1,18 @@
 import logging
 import time
 import recaptcha2
+import os
 
 from collections import namedtuple
 from flask import request, redirect, url_for, Blueprint, abort, session
 from peewee import IntegrityError
+import requests
+
 
 import features
 
 from app import app, analytics, get_app_url, oauth_login, authentication, url_scheme_and_hostname
+from _init import CONF_DIR
 from auth.auth_context import get_authenticated_user
 from auth.decorators import require_session_login
 from data import model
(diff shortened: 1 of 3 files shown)


endpoints/oauth/login.py
```diff
@@ -297,6 +301,26 @@ def attach_func():
         if result.error_message is not None:
             return _get_response(result)
 
+        # Conduct RedHat Export Compliance if enabled
+        if features.EXPORT_COMPLIANCE:
+            logger.debug("Attempting to hit export compliance service")
+            result = requests.post(
+                "https://export-compliance.dev.api.redhat.com/v1/screening",
```
syed, line 308: make this a const/config?

syed (approved):

kleesc:
@jonathankingfc Are there any references to the external api that this is hitting?

endpoints/oauth/login.py
```diff
@@ -297,6 +301,26 @@ def attach_func():
         if result.error_message is not None:
             return _get_response(result)
 
+        # Conduct RedHat Export Compliance if enabled
+        if features.EXPORT_COMPLIANCE:
+            logger.debug("Attempting to hit export compliance service")
+            result = requests.post(
+                app.config.get("EXPORT_COMPLIANCE_ENDPOINT"),
```
jonathankingfc, line 308: @kleesc This is the endpoint, it is set as a config variable which is in the staging config since we don't want to expose it publicly.

endpoints/oauth/login.py
```diff
@@ -297,6 +301,26 @@ def attach_func():
         if result.error_message is not None:
             return _get_response(result)
 
+        # Conduct RedHat Export Compliance if enabled
+        if features.EXPORT_COMPLIANCE:
+            logger.debug("Attempting to hit export compliance service")
+            result = requests.post(
```
syed, line 307: @jonathankingfc is there a timeout here? I think the default action is to wait indefinitely

syed (changes requested):

syed (approved):

PR #1240, opened 2022-04-12
proxy: make upstream related error message more actionable

flavianmissi:
Example trying to pull a private image after having configured a proxy org without any credentials:
image

fixes [PROJQUAY-3553](https://issues.redhat.com//browse/PROJQUAY-3553)

### DIFF
--- a/proxy/__init__.py
+++ b/proxy/__init__.py
@@ -23,7 +23,10 @@ REGISTRY_URLS = {"docker.io": "registry-1.docker.io"}
 
 class UpstreamRegistryError(Exception):
     def __init__(self, detail):
-        msg = f"Error requesting upstream registry: {detail}"
+        msg = (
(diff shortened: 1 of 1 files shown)


proxy/__init__.py
```diff
@@ -23,7 +23,10 @@
 
 class UpstreamRegistryError(Exception):
     def __init__(self, detail):
-        msg = f"Error requesting upstream registry: {detail}"
+        msg = (
+            "the requested image may not exist in the upstream registry, or you do not"
```
dmage, line 27: `you` is misleading there, the client who will read this error may not know how the Quay organization is configured.

flavianmissi: updated as discussed on slack.

dmage (approved):

PR #1242, opened 2022-04-12
UI: Quota UI enhancements (PROJQUAY-0000)

Sunandadadi:
This PR improves the Quota UI with the following changes:
1. When no Quota limit is set, default "Reject - 100%" is populated
2. Frontend validations for quota value <= 0
3. If "Reject" quota limit does not exist, warning message is shown.
4. There is a "Remove" button that deletes the set quota.
5. If there is not quota limit, warning message is shown.
6. If `FEATURE_QUOTA_MANAGEMENT` is not enabled, there is no `QUOTA CONSUMPTION` column that tracks storage of an org and repo
7.  If `FEATURE_QUOTA_MANAGEMENT` is enabled and quota is not set, storage for org and repo are reported.
8. If `FEATURE_QUOTA_MANAGEMENT` is enabled and quota is set, storage with percent consumed for org and repo are reported.
9. Throw error messages if multiple reject limit types exist.
10. Throw error messages if duplicate limit types exist. 
11. Org view - editable by organization admin if quota is not previously set. If set, only super user can make updates/delete set quota.


Sunandadadi:
Screen Shot 2022-04-13 at 9 57 35 AM
Screen Shot 2022-04-13 at 9 58 10 AM
Screen Shot 2022-04-13 at 10 00 19 AM

Sunandadadi:
Screen Shot 2022-04-13 at 10 02 30 AM

data/model/namespacequota.py
```diff
@@ -279,20 +279,16 @@ def get_repo_quota_for_view(namespace_name, repo_name):
         return None
 
     quotas = get_namespace_quota_list(repository_ref.namespace_user.username)
-    if not quotas:
-        return {
-            "quota_bytes": None,
-            "configured_quota": None,
-        }
```
kleesc, line 286: The list could be empty, which would cause an exception if not checked for.

endpoints/api/namespacequota.py
```diff
@@ -154,9 +151,9 @@ def get(self, orgname, quota_id):
     def put(self, orgname, quota_id):
         orgperm = AdministerOrganizationPermission(orgname)
 
+        # Only super users can update quota
         if not features.SUPER_USERS or not SuperUserPermission().can():
-            if not orgperm.can():
-                raise Unauthorized()
+            raise Unauthorized()
```
kleesc, line 156: Superusers have their own endpoints for non-GET requests in `endpoints/api/superuser.py`
I might've missed making the ui's api calls in the previous PR though.

Sunandadadi: So, do we want to make different API calls depending on the view it is called from?

kleesc: Not strictly necessary right away, since you've added the checks, just something to note, we should differentiate the calls from the superuser panel and org admin panel at some point. They're essentially the same thing, on different endpoints; it's easier to manage permissions.

endpoints/api/namespacequota.py
```diff
@@ -76,7 +76,7 @@ class OrganizationQuotaList(ApiResource):
     @nickname("listOrganizationQuota")
     def get(self, orgname):
         orgperm = OrganizationMemberPermission(orgname)
-        if not orgperm.can():
+        if not orgperm.can() and not SuperUserPermission().can():
```
kwestpharedhat, line 79: this endpoint wont be called by super users it should now call the super users endpoint.

static/directives/quota-management-view.html
```diff
@@ -1,48 +1,52 @@
...
-      <td>
-        Quota Management:
+      <td class="row-alignment">
+        Set storage quota:
       </td>
       <td>
 	<table class="co-table">
 	  <td>
-            <input class="margin-2 form-control width-440" type="number" name="quota-limit" min="1" ng-model="currentQuotaConfig['quota']">
+            <input class="margin-2 form-control width-440" type="number" name="quota-limit" ng-model="currentQuotaConfig['quota']">
```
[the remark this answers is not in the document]
Sunandadadi: Because adding a check [here](https://github.com/quay/quay/pull/1242/files#diff-97350d41b191619358d2e04565d8f58061faafcf99224960fe1cd38bc9d79846R323). If I let `min=1` be, it returns undefined and I cannot validate against it to throw appropriate error msg.

endpoints/api/namespacequota.py
Sunandadadi: The super users endpoint only have `put` and `delete` methods as [here](https://github.com/quay/quay/blob/master/endpoints/api/superuser.py#L274)

static/js/directives/ui/quota-management-view.js
```diff
@@ -20,51 +20,67 @@ angular.module('quay').directive('quotaManagementView', function () {
...
+            $scope.prevQuotaConfig['id'] = quota["id"];
+            $scope.currentQuotaConfig['id'] = quota["id"];
+
+            for (const i in quota['limits']) {
+              const limitId = quota['limits'][i]['id'];
+              $scope.prevQuotaConfig['limits'][limitId] = $.extend({}, quota["limits"][i]);
+              $scope.currentQuotaConfig['limits'][limitId] = $.extend({}, quota["limits"][i]);
             }
+
+            let {result, byte_unit} = normalizeLimitBytes(quota["limit_bytes"]);
```
kwestpharedhat, line 65: Why did this get moved outside the if block. quota will not be guarenteed to be not null here.

static/js/directives/ui/quota-management-view.js
```diff
@@ -76,139 +92,338 @@ angular.module('quay').directive('quotaManagementView', function () {
...
-      $scope.disableSaveQuota = function() {
+      let displayErrorAlert = function (message) {
+        bootbox.alert(message);
+      }
+
+      let handleError = function (resp) {
+        if (resp.status == 403) {
+          displayErrorAlert("You do not have sufficient permissions to perform the action.");
+        } else {
+          displayErrorAlert(resp.data.error_message);
```
kwestpharedhat, line 232: should the possibility of a 500 error show up in a alert message or redirect here?

Sunandadadi: All 500s get redirected to a different page (similar to - https://quay.io/500). So, they won't show up here.

kwestpharedhat: what about other error types perhaps 500 was the wrong  example

Sunandadadi: They will be alerted based on the `error_message` returned.

static/js/directives/ui/quota-management-view.js
Sunandadadi: It is inside the `resp.length > 0` if block. Similar to as it was before. Only changes here are indentations.

endpoints/api/namespacequota.py
Sunandadadi: Fixed in the recent commit.

endpoints/api/namespacequota.py
```diff
@@ -151,9 +151,9 @@ def get(self, orgname, quota_id):
     def put(self, orgname, quota_id):
         orgperm = AdministerOrganizationPermission(orgname)
 
-        # Only super users can update quota
         if not features.SUPER_USERS or not SuperUserPermission().can():
-            raise Unauthorized()
+            if not orgperm.can():
```
kwestpharedhat, line 155: I would suggest we remove super user permissions from within this and secure it using @require_scope(scopes.SUPERUSER) to be removed at a later time when the code is smarter

kleesc (approved):

PR #1244, opened 2022-04-13
mirror: Get all tags during rollback (PROJQUAY-3146)

bcaton85:
During a rollback the mirror worker checks for new tags that were created in the repository in the time that the mirror operation has been running. If it encounters older tags that have been updated it will attempt to create a new tag that will point to the previous manifest. Currently for large lists of tags this will fail since we only retrieve the 100 latest tags. The mirror worker will never reach the tags that have been updated and will never recreate them, leading to the behavior of deleting tags during a rollback.


workers/repomirrorworker/__init__.py
```diff
@@ -31,6 +31,9 @@
     "number of repositories in the database that have not yet been mirrored",
 )
 
+# Used only for testing - should not be set in production
+TAG_ROLLBACK_PAGE_COUNT = app.config.get("REPO_MIRROR_TAG_ROLLBACK_PAGE_COUNT", 100)
```
kleesc, line 35: I don't see where it's being used in your tests. Do we need it?
You could also use `IS_TESTING`, I think the configuration is, for test specific values.

bcaton85: Yeah i have this set for QE to use. Since the problem only occurs when tags reach the limit I was going to ask them to set the limit to be low and confirm the rollback still works.

kleesc: Ok. Small nit: can you rename it page `SIZE` instead of page count?

bcaton85: Done

kleesc (approved):

PR #1246, opened 2022-04-14
builders: Set backoffLimit to 1 (PROJQUAY-3587)

bcaton85:
Setting the backoffLimit to 1 for kubernetes and kubernetesPodman builds. Prevents subsequent attempts from failing due to the token expiring. Having the job recreate pods is unnecessary as the build manager already has the retry logic.


kleesc (approved):
lgtm

PR #1247, opened 2022-04-14
secscan: optimize deduplicating manifests for indexing in securityworker (PROJQUAY-3501)

kleesc:
Acquire lock on recent manifests batch to prevent multiple workers
from working on the same set.

### DIFF
--- a/requirements-osbs.txt
+++ b/requirements-osbs.txt
@@ -98,6 +98,7 @@ python-gitlab==2.0.0
 python-keystoneclient==3.22.0
 python-ldap==3.4.0
 python-magic==0.4.15
+python-redis-lock==3.7.0
 python-swiftclient==3.8.1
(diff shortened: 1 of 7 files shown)


requirements-osbs.txt
```diff
@@ -141,10 +141,10 @@ zope.event==4.5.0
 zope.interface==5.4.0
 
 # Build dependencies
-Cython==3.0a9
+Cython==0.29.23
```
hdonnay, line 144: Seems weird to move off the prerelease train in this PR.

kleesc: Must've changed it by accident. . Changed it back to `Cython==3.0a9`

hdonnay (commented):
SGTM

hdonnay (approved):

PR #1248, opened 2022-04-15
nginx: block v1/tag for helium miner curl calls (PROJQUAY-3594)

syed:
There are scripts calling the v1/tags API for helium miner
causing an DB load


Sunandadadi (approved):

Sunandadadi:
LGTM

PR #1249, opened 2022-04-18
Quota UI: Showing percent consumed if quota is configured on an organization (PROJQUAY-0000)

Sunandadadi:


Sunandadadi:
Showing Repository size along with percent of storage consumed as shown: Screen Shot 2022-04-18 at 2 35 59 PM

PR #1263, opened 2022-04-19
Quota Management: Quota settings on Organization view needs to be read only (PROJQUAY-3622)

Sunandadadi:


endpoints/api/namespacequota.py
```diff
@@ -94,11 +94,8 @@ def post(self, orgname):
         """
         Create a new organization quota.
         """
-        orgperm = AdministerOrganizationPermission(orgname)
```
kwestpharedhat, line 97: this change should just be @requirescope super user no need to change anything else.

endpoints/api/namespacequota.py
```diff
@@ -151,15 +148,12 @@ def get(self, orgname, quota_id):
     @show_if(features.SUPER_USERS)
     @validate_json_request("UpdateOrgQuota")
     def put(self, orgname, quota_id):
-        orgperm = AdministerOrganizationPermission(orgname)
```
kwestpharedhat, line 154: change is probably not required since its already require scope super user

kwestpharedhat (changes requested):

kwestpharedhat (approved):

PR #1270, opened 2022-04-20
secscan: fix check for end of table (PROJQUAY-3501)

kleesc:
Correctly check for the end of the manifest table, for when the worker
needs to start over. Also add missing token to lock key.


syed:
LGTM

hdonnay (approved):

PR #1271, opened 2022-04-20
API: Removing internal only decorator from exposed Super user endpoints (PROJQUAY-0000)

Sunandadadi:


endpoints/api/superuser.py
```diff
@@ -103,7 +103,6 @@ def get(self, parsed_args):
 
 
 @resource("/v1/superuser/logs")
-@internal_only
```
kleesc, line 106: Why are we making this one not internal?

kleesc (changes requested):

endpoints/api/superuser.py
Sunandadadi: It is plainly for docs. Gabriel requested for it. Currently they need to remember to add in this one manually every time. It would make a big difference as they can auto generate all the API docs with this change.

kleesc (approved):

PR #1272, opened 2022-04-20
Quota UI: Display Error when decimal values entered from UI (PROJQUAY-3627)

Sunandadadi:
Screen Shot 2022-04-20 at 1 13 08 PM
Screen Shot 2022-04-20 at 1 14 53 PM
Screen Shot 2022-04-20 at 1 15 14 PM

### DIFF
--- a/static/js/directives/ui/quota-management-view.js
+++ b/static/js/directives/ui/quota-management-view.js
@@ -308,6 +308,14 @@ angular.module('quay').directive('quotaManagementView', function () {
         return false
       }
 
+      var isOfTypeDecimal = function (value) {
+        if (value % 1 != 0) {
+          $scope.errorMessage = $scope.errorMessagesObj['decimalEntryError'];
+          return true;
+        }
+        return false;
+      }
+
       $scope.disableAddQuotaLimit = function () {
         // Cannot add limits without configuring quota
         if (!$scope.currentQuotaConfig.id) {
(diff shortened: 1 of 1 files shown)


static/js/directives/ui/quota-management-view.js
```diff
@@ -312,6 +314,16 @@ angular.module('quay').directive('quotaManagementView', function () {
           return true;
         }
 
+        if ($scope.newLimitConfig['limit_percent'] != null && isNaN(parseInt($scope.newLimitConfig['limit_percent']))) {
```
kwestpharedhat, line 317: parseInt here may not work if the value is decimal. check for decimals first then parse

static/js/directives/ui/quota-management-view.js
```diff
@@ -339,6 +351,11 @@ angular.module('quay').directive('quotaManagementView', function () {
           return false;
         }
 
+        if (limit_percent % 1 != 0) {
```
kwestpharedhat, line 354: can this be functionalized? I see the same code in 3 or 4 places.

kwestpharedhat (changes requested):

static/js/directives/ui/quota-management-view.js
Sunandadadi: This is not a check for decimals. This is to check if the input is an integer.

kwestpharedhat: yes but this is a if statement what happens if parseInt is given a decimal. In other languages it would fail and throw an exception here your if statement may not evaluate.

Sunandadadi: So, In javascript, [parseInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) converts a string type to integer. parseInt(-12.23) gives -12. I am using this to catch exponential values like `"e12.23"` which should return an error. parseInt("e12.23") returns Nan (Not a number) so using `isNan` to raise an error here.

kwestpharedhat (approved):

PR #1281, opened 2022-04-22
secscan: split indexing of recent manifest into separate background operation (PROJQUAY-3501)

kleesc:
Index recent manifests in a separate background process, allowing the
main process to correctly select random slabs from the entire table
set and marking them completed in the allocator (rbtree). This avoids
the worker having to start iterating from the beginning of the table
whenever it is restarted.


workers/securityworker/securityworker.py
```diff
@@ -28,9 +28,22 @@ def __init__(self):
...
+        self._next_token = self._model.perform_indexing(self._next_token, batch_size)
+
+    def _index_recent_manifests_in_scanner(self):
+        batch_size = self.app.config.get("SECURITY_SCANNER_V4_BATCH_SIZE", 0)
+
+        try:
+            with GlobalLock(
+                "SECURITYWORKER_INDEX_RECENT_MANIFEST", lock_ttl=300, auto_renewal=True
+            ):
+                self._next_token = self._model.perform_indexing_recent_manifests(batch_size)
```
syed, line 44: is `self._next_token` being used in both `_index_in_scanner` and `_index_recent_manifests_in_sanner` would it cause a race condition?

syed (approved):

PR #1294, opened 2022-04-28
storage: Add Cloudflare as a CDN provider for an S3 backed storage (PROJQUAY-3699)

syed:
This adds CloudFlare as a CDN provider for quay for any storage backed
by S3. This requires a worker script that needs to be setup seperately
on CloudFlare. More details on the worker at
https://github.com/quay/quay-cloudflare-cdn-worker

### DIFF
--- a/storage/cloudflarestorage.py
+++ b/storage/cloudflarestorage.py
@@ -1,0 +1,100 @@
+import base64
+import logging
+import urllib.parse
+from datetime import datetime, timedelta
+from functools import lru_cache
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization, hashes
+from cryptography.hazmat.primitives.asymmetric import padding
+
+logger = logging.getLogger(__name__)
+
+from storage.cloud import S3Storage
+
+
+class CloudFlareS3Storage(S3Storage):
+    """
+    CloudFlare CDN backed by S3 storage
+    """
+
+    def __init__(
+            self,
+            context,
+            cloudflare_domain,
+            cloudflare_privatekey_filename,
+            storage_path,
+            s3_bucket,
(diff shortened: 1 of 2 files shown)


storage/cloudflarestorage.py
```diff
@@ -0,0 +1,97 @@
...
+        super(CloudFlareS3Storage, self).__init__(context, storage_path, s3_bucket, *args, **kwargs)
+
+        self.cloudflare_domain = cloudflare_domain
+        self.cloudflare_privatekey = self._load_private_key(cloudflare_privatekey_filename)
+
+    def get_direct_download_url(
+        self, path, request_ip=None, expires_in=60, requires_cors=False, head=False
+    ):
+        # If CloudFront could not be loaded, fall back to normal S3.
+        s3_presigned_url = super(CloudFlareS3Storage, self).get_direct_download_url(
```
kleesc, line 40: Put this inside the `if` block, that way it doesn't have to sign _every_ request.

kleesc (changes requested):

storage/cloudflarestorage.py
kleesc: Just saw that this is still used even if you're able to sign with CloudFlare keys.
Then the comment `If CloudFront could not be loaded, fall back to normal S3.` doesn't really hold right?

syed: This is basically a graceful fallback. We would always set `cloudflare_privatekey`. I could instead throw an error here. I was just following how CloudFront has implemented it

kleesc: Ok. I guess I was confused since you were building the cloudflare_url based off the s3 one...

kleesc (approved):

PR #1295, opened 2022-04-29
chore: various small changes to fix exceptions, remove unused code

kleesc:
Addresses some uncaught exceptions seen on Sentry.

### DIFF
--- a/endpoints/v2/manifest.py
+++ b/endpoints/v2/manifest.py
@@ -69,7 +69,7 @@ def fetch_manifest_by_tagname(namespace_name, repo_name, manifest_ref, registry_
         )
     except RepositoryDoesNotExist as e:
         image_pulls.labels("v2", "tag", 404).inc()
-        raise NameUnknown(str(e))
+        raise NameUnknown("repository not found")
 
     try:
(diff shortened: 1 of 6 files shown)


Sunandadadi (approved):

Sunandadadi:
LGTM

endpoints/v2/manifest.py
```diff
@@ -69,7 +69,7 @@ def fetch_manifest_by_tagname(namespace_name, repo_name, manifest_ref, registry_
         )
     except RepositoryDoesNotExist as e:
         image_pulls.labels("v2", "tag", 404).inc()
-        raise NameUnknown(str(e))
+        raise NameUnknown("repository not found")
```
flavianmissi, line 72: When the proxy code raises a `RepositoryDoesNotExist` this will override the message it returns, making it considerably difficult for clients to understand what happened. I suggest that you change `lookup_repository` in the `OCIModel` so that it raises `RepositoryDoesNotExist("repository not found")`, then continue using the previous `str(e)` as argument to `NameUnknown` here.

PR #1296, opened 2022-04-29
Quota API: Add super user permissions on Organization endpoints (PROJQUAY-3742)

Sunandadadi:


kleesc (changes requested):

endpoints/api/namespacequota.py
```diff
@@ -108,15 +108,13 @@ def get(self, orgname):
     @nickname("createOrganizationQuota")
     @validate_json_request("NewOrgQuota")
     @require_scope(scopes.SUPERUSER)
+    @show_if(features.SUPER_USERS)
```
kleesc, line 111: Might as well just put this at the class level, since it doesn't really make sense to have quota enabled, but not superusers in this case.

Likewise for the other resource classes.

Sunandadadi: Ok, makes sense

kleesc (approved):

PR #1303, opened 2022-05-05
quota: address possible integrity exception if computing size concurrently (PROJQUAY-3752)

kleesc:


kwestpharedhat (approved):

data/model/repository.py
```diff
@@ -739,17 +739,18 @@ def force_cache_repo_size(repo_id: int):
     except Manifest.DoesNotExist:
         size = 0
 
-    repo = get_repository_size(repo_id)
-
     if size is None:
         size = 0
 
-    if repo is not None:
```
kwestpharedhat, line 747: I dont think this is going to fix what your trying to fix. If this is a race condition as you explain in the ticket this PR is linked too this will not fix that as 2 executions reaching the get_repo_size function at the same time will still error.

kwestpharedhat (changes requested):

data/model/repository.py
```diff
@@ -739,17 +739,21 @@ def force_cache_repo_size(repo_id: int):
...
+        repo_size_ref = get_repository_size(repo_id)
+        try:
+            if repo_size_ref is not None:
+                update = RepositorySize.update(size_bytes=size).where(
+                    RepositorySize.repository_id == repo_id
+                )
+                update.execute()
+            else:
+                RepositorySize.create(repository_id=repo_id, size_bytes=size)
+        except IntegrityError:
```
kwestpharedhat, line 755: can we add a comment here to explain that the only known way to reach this expection is due to race condition and the assumed behavior is that both with have the same value to cache?

kwestpharedhat (changes requested):

data/model/repository.py
kleesc: Done.

PR #1305, opened 2022-05-09
builders: Add dnsPolicy option (PROJQUAY-3755)

bcaton85:
Some clusters use `dnsPolicy: ClusterFirst` while we have the field hardcoded to `Default`. This change allows the field to be configurable.


kleesc (approved):
LGTM

PR #1315, opened 2022-05-10
v2auth: Check for user before creating org (PROJQUAY-3766)

bcaton85:
Currently when Quay is set to automatically create organizations it will check the user table for a matching username with organization set to true. This causes conflicts when a user already exists with the same username, where the check will fail and Quay will attempt to create an organization with the same name as the user. This change checks for only a matching username.

### DIFF
--- a/endpoints/v2/v2auth.py
+++ b/endpoints/v2/v2auth.py
@@ -261,20 +261,20 @@ def _authorize_or_downscope_request(scope_param, has_valid_auth_context):
                 else:
                     logger.debug("No permission to modify repository %s/%s", namespace, reponame)
             else:
-                if app.config.get("CREATE_NAMESPACE_ON_PUSH", False):
+                if (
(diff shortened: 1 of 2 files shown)


endpoints/v2/v2auth.py
```diff
@@ -261,20 +261,20 @@ def _authorize_or_downscope_request(scope_param, has_valid_auth_context):
                 else:
                     logger.debug("No permission to modify repository %s/%s", namespace, reponame)
             else:
-                if app.config.get("CREATE_NAMESPACE_ON_PUSH", False):
+                if (
+                    app.config.get("CREATE_NAMESPACE_ON_PUSH", False)
+                    and model.user.get_namespace_user(namespace) is None
+                ):
```
kleesc, line 267: Can you add a test for this?

bcaton85: Yes

kleesc (approved):
LGTM

PR #1316, opened 2022-05-11
Quota Management: Adding default quota check for image push (PROJQUAY-3789)

Sunandadadi:


data/model/namespacequota.py
```diff
@@ -189,10 +189,11 @@ def check_limits(namespace_name, size):
     namespace_user = model.user.get_user_or_org(namespace_name)
     quotas = get_namespace_quota_list(namespace_user.username)
     if not quotas:
-        if config.app_config.get("DEFAULT_SYSTEM_REJECT_QUOTA_BYTES") != 0:
+        default_size = config.app_config.get("DEFAULT_SYSTEM_REJECT_QUOTA_BYTES")
+        if default_size != 0:
```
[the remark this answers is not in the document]
Sunandadadi: Agreed, pushed in the latest commit

data/model/namespacequota.py
```diff
@@ -189,13 +189,13 @@ def check_limits(namespace_name, size):
...
-            }
-        else:
-            return {"limit_bytes": 0, "severity_level": None}
+        default_size = config.app_config.get("DEFAULT_SYSTEM_REJECT_QUOTA_BYTES")
+        return {
+            "limit_bytes": default_size,
+            "severity_level": None
+            if (default_size == 0 or size < default_size)
+            else QuotaTypes.REJECT,
+        }
```
kleesc, line 198: What if `DEFAULT_SYSTEM_REJECT_QUOTA_BYTES` is not set?

Sunandadadi: By default, `DEFAULT_SYSTEM_REJECT_QUOTA_BYTES` is set to 0 in [config.py](https://github.com/quay/quay/blob/master/config.py#L822). So, if `DEFAULT_SYSTEM_REJECT_QUOTA_BYTES` is set not set (as in it is 0), we do not reject a push.

kwestpharedhat (approved):

PR #1317, opened 2022-05-11
sso: Add test for RHSSO OAuth service (PROJQUAY-2056)

jonathankingfc:
- Add test for RHSSOOAuthService class

### DIFF
--- a/oauth/loginmanager.py
+++ b/oauth/loginmanager.py
@@ -28,7 +28,7 @@ class OAuthLoginManager(object):
                     if custom_service.login_enabled(config):
                         self.services.append(custom_service)
                 else:
-                    prefix = key.rstrip("_LOGIN_CONFIG").lower()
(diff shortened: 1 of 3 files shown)


oauth/loginmanager.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, config, client=None):
                     prefix = key.rstrip("_LOGIN_CONFIG").lower()
                     if prefix in PREFIX_BLACKLIST:
                         raise Exception("Cannot use reserved config name %s" % key)
-                    if prefix == "rhsso":
+                    if prefix == "rhss":
```
kleesc, line 34: This should be `rhsso`. Instead of using `rstrip()`, use `removesuffix()`.

kleesc (changes requested):

kleesc (approved):
LGTM

PR #1330, opened 2022-05-16
builders: Set imagePullPolicy to always (PROJQUAY-3507)

bcaton85:
Currently imagePullPolicy is set to IfNotPresent. This will ignore updates to the builder image.

### DIFF
--- a/buildman/manager/executor.py
+++ b/buildman/manager/executor.py
@@ -552,7 +552,7 @@ class KubernetesExecutor(BuilderExecutor):
 
         container = {
             "name": "builder",
-            "imagePullPolicy": "IfNotPresent",
+            "imagePullPolicy": self.executor_config.get("IMAGE_PULL_POLICY", "Always"),
             "image": self.image,
             "securityContext": {"privileged": True},
             "env": [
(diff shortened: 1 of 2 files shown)


oauth/services/rhsso.py
```diff
@@ -34,7 +34,10 @@ def exchange_code_for_login(self, app_config, http_client, code, redirect_suffix
 
                 # 200 => Endpoint was hit successfully and user was found
                 # 400 => Endpoint was hit successfully but no user was found
-                if result.status_code == 200 and result.json().get("result", "") == "ERROR_EXPORT_CONTROL":
+                if (
```
syed, line 37: can you rebase for this

buildman/manager/executor.py
```diff
@@ -552,7 +552,7 @@ def _build_job_containers(self, token, build_uuid):
 
         container = {
             "name": "builder",
-            "imagePullPolicy": "IfNotPresent",
+            "imagePullPolicy": "Always",
```
syed, line 555: As @kleesc  suggested, can we make this a config value?

syed (changes requested):

oauth/services/rhsso.py
bcaton85: it looks like i'm already on top of the latest commit - b21400b90d46ff1d6a989c2b3008769420bbd608

buildman/manager/executor.py
bcaton85: yes, updated to be configurable.

kleesc (approved):
LGTM

syed (approved):

PR #1250, opened 2022-04-18
chore: Bump python dependencies (PROJQUAY-2175)

HammerMeetNail:
Bumps:
* boto3
* botocore
* oslo.utils
* pypdf2
* requests
* urllib3

Resolves the following CVEs

* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-26137
* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-0718
* https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-33503

PR #1340, opened 2022-05-24
secscan: update https proxy scheme

kleesc:
Update the https proxy scheme from "https" to "http". The scheme was
ignored prior to urllib3 1.26, which is why it was working.

https://github.com/psf/requests/issues/5943#issuecomment-926615360
https://github.com/quay/quay/pull/1250/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R131


crozzy (approved):

crozzy:
Damn, good find

PR #1348, opened 2022-05-27
database: remove foreign key constraints from deprecated tables

kleesc:


data/model/oci/shared.py
```diff
@@ -1,16 +1,15 @@
-from data.database import Manifest, ManifestLegacyImage, Image
+from data.database import Manifest, ManifestSecurityStatus
 
 
-def get_legacy_image_for_manifest(manifest_id):
+def get_manifestsecuritystatus_for_manifest(manifest_id):
     """
     Returns the legacy image associated with the given manifest, if any, or None if none.
```
flavianmissi, line 6: comment needs update as well

data/registry_model/registry_oci_model.py
```diff
@@ -545,16 +543,18 @@ def get_security_status(self, manifest_or_legacy_image):
         """
         # TODO: change from using the Image row once we've moved all security info into MSS.
```
flavianmissi, line 544: time to remove this TODO?

data/model/oci/shared.py
```diff
@@ -1,16 +1,15 @@
-from data.database import Manifest, ManifestLegacyImage, Image
+from data.database import Manifest, ManifestSecurityStatus
 
 
-def get_legacy_image_for_manifest(manifest_id):
+def get_manifestsecuritystatus_for_manifest(manifest_id):
```
flavianmissi, line 4: nit: the word "manifest" is repeated in the function name, what about calling it something like `get_manifest_securitystatus` instead?

kleesc: It's the name of the actual table, so I'll just keep it as is to follow convention.

kleesc:
@syed ping

syed (approved):

PR #1355, opened 2022-06-02
Dockerfile: use nodejs ubi8 image to build frontends

flavianmissi:
Also add nginx module before installing it, without this dnf won't find it.

Note that we're using ubi8/node-10, which is deprecated - but this is
the version previously installed (before the nodejs package disappeared
on us).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ ENV QUAYPATH $QUAYDIR
 ENV PYTHONUSERBASE /app
 ENV PYTHONPATH $QUAYPATH
 RUN set -ex\
+	; dnf -y module enable nginx:1.20 \
 	; dnf -y -q --setopt=tsflags=nodocs --setopt=skip_missing_names_on_install=False install\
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -67,9 +65,10 @@ RUN set -ex\
 	;
 
 # Build-static downloads the static javascript.
-FROM build-python AS build-static
-COPY --chown=0:0 static/  ./static/
-COPY --chown=0:0 *.json *.js  ./
+FROM registry.access.redhat.com/ubi8/nodejs-10 AS build-static
+WORKDIR /opt/app-root/src
+COPY --chown=1001:0 static/  ./static/
+COPY --chown=1001:0 *.json *.js  ./
```
flavianmissi, line 71: see https://catalog.redhat.com/software/containers/ubi8/nodejs-10/5c839aa3d70cc51dd4c425d9?container-tabs=dockerfile for background on ownership change

kleesc (approved):

PR #1356, opened 2022-06-02
init: ensure a newline is present before appending certs (PROJQUAY-3881)

kleesc:
Make sure the script still works when mounting certs as a read-only fs.


bcaton85 (approved):
LGTM

PR #1358, opened 2022-06-03
UI: Show settings icon for super user under Superuser Users Panel (PROJQUAY-3905)

Sunandadadi:
Screen Shot 2022-06-03 at 12 04 52 PM

### DIFF
--- a/static/directives/manage-users-tab.html
+++ b/static/directives/manage-users-tab.html
@@ -73,7 +73,7 @@
         </td>
         <td style="text-align: center;">
                   <span class="cor-options-menu"
-                        ng-if="user.username != current_user.username && !current_user.super_user && !inReadOnlyMode">
+                        ng-if="user.username != current_user.username && !inReadOnlyMode">
                     <span class="cor-option" option-click="showChangeEmail(current_user)"
                           quay-show="Config.AUTHENTICATION_TYPE == 'Database' || Config.AUTHENTICATION_TYPE == 'AppToken'">
                       <i class="fa fa-envelope-o"></i> Change E-mail Address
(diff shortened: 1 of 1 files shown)


static/directives/manage-users-tab.html
```diff
@@ -73,32 +73,37 @@
...
                       <i class="fa fa-times"></i> Delete User
                     </span>
-                    <span class="cor-option" option-click="askDisableUser(current_user)">
+                    <span class="cor-option" option-click="askDisableUser(current_user)"
+                          ng-if="!current_user.super_user">
                       <i class="fa" ng-class="current_user.enabled ? 'fa-circle-o' : 'fa-check-circle-o'"></i> <span
                         ng-if="current_user.enabled">Disable</span> <span ng-if="!current_user.enabled">Enable</span> User
                     </span>
                     <span class="cor-option" option-click="askTakeOwnership(current_user)"
-                          ng-if="user.username != current_user.username && !current_user.super_user">
```
kleesc, line 97: Shouldn't this still be the case? We want to show it only if the user is not the current one.

Sunandadadi: I've restructured this to - `((current_user.super_user && user.username == current_user.username) || !current_user.super_user)` because in the previous logic, say we have 2 super users: admin1 and admin2, if admin1 is logged in, they can edit the settings of admin2. Similarly if admin2 is logged in, they can edit the settings of admin1. But in the current change, admin1 can only change admin1's settings and admin2 can only change admin2's settings.
And finally any super user can edit any non-super users settings.

kleesc (approved):
lgtm

PR #1362, opened 2022-06-06
oauth: Fix oauth code flow when user has not authorized the app (PROJQUAY-3648)

syed:
Fixes an issue where the code param is not passed to the app redirect
URI if the user has not authorized the app before

### DIFF
--- a/endpoints/web.py
+++ b/endpoints/web.py
@@ -642,11 +642,27 @@ def authorize_application():
 
     provider = FlaskAuthorizationProvider()
     redirect_uri = request.form.get("redirect_uri", None)
+    response_type = request.form.get("response_type", "code")
     scope = request.form.get("scope", None)
     state = request.form.get("state", None)
 
     # Add the access token.
-    return provider.get_token_response("token", client_id, redirect_uri, scope=scope, state=state)
+    if response_type == "token":
+        return provider.get_token_response(
+            response_type,
+            client_id,
+            redirect_uri,
+            scope=scope,
+            state=state,
+        )
+    else:
+        return provider.get_authorization_code(
+            response_type,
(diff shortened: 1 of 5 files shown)


endpoints/web.py
```diff
@@ -690,6 +690,7 @@ def request_authorization_code():
     redirect_uri = request.args.get("redirect_uri", None)
     scope = request.args.get("scope", None)
     state = request.args.get("state", None)
+    code = request.args.get("code", None)
```
kleesc, line 693: The code shouldn't be included as part of the authorize request step of oauth. It's what we the Quay app would return as part of the uri when redirecting. The `code` is already be included, based on `oauth/provider.py`

kleesc (changes requested):

kleesc (dismissed):

requirements-dev.txt
```diff
@@ -13,6 +13,7 @@ ipython
 mock==3.0.5
 mypy==0.910
 moto==2.0.1
+parameterized==0.8.1
```
kleesc, line 16: Is this required?

test/test_endpoints.py
```diff
@@ -396,22 +397,26 @@ def test_redirect_to_namespace(self):
 
 
 class OAuthTestCase(EndpointTestCase):
-    def test_authorize_nologin(self):
+    @parameterized.expand(["token", "code"])
```
kleesc, line 400: Could use pytest's existing parameterize?

syed: can't use `pytest.parameterize` because the test class is based on `unittest.Testcase` which doesn't support it

kleesc (approved):

PR #1366, opened 2022-06-08
fix: enable non-admins to cache images via pull-thru (PROJQUAY-3806)

flavianmissi:
introduces a check of `CreateRepositoryPermission` for pulls
against a proxy org.

if the proxy cache feature is disabled, or the org is not a proxy org
the check is not performed and Quay will behave normally.

the check does not mean pulls will work transparently though -
non-admin users need to be added to a team in the proxy org with the
creator role, and default read and write permissions need to be given to
that team so that non-admin users can pull and update the cache for
images they do not own (the user who first pulls an image ends up
owning the repository since that is when the repo gets created).


endpoints/v2/v2auth.py
```diff
@@ -296,8 +296,12 @@ def _authorize_or_downscope_request(scope_param, has_valid_auth_context):
                     logger.debug("No permission to create repository %s/%s", namespace, reponame)
 
     if "pull" in requested_actions:
-        # Grant pull if the user can read the repo or it is public.
-        if ReadRepositoryPermission(namespace, reponame).can() or repo_is_public:
+        can_pullthru = False
+        if features.PROXY_CACHE and model.proxy_cache.has_proxy_cache_config(namespace):
+            can_pullthru = (
+                CreateRepositoryPermission(namespace).can() and get_authenticated_user() is not None
```
dmage, line 302: why this check is needed?

flavianmissi: Users will only have `ReadRepositoryPermission` on a repository that already exists in Quay (given the correct team and default permission settings), which is not the case on first pulls thru a proxy org. 
And since first pulls thru a proxy org do end up creating a repository, it made sense to me that users would need to have the `CreateRepositoryPermission`.

flavianmissi: There aren't many great alternatives here AFAIK. With Quay's current permission model it's not possible to give all users in a team read or write permissions on an organization basis. Read and write permissions are given on a repository basis, so the repository needs to exist for it to be given.

dmage: `CreateRepositoryPermission` means the client can actively create repositories with whatever name they want. I don't want to give my clients that permission. I want to allow them to pull images and I expect that they'll be able to pull images from the proxy org.

flavianmissi: I guess we could disable pushes to proxy orgs (we said we would but I didn't have the time and it ended up not getting done). I think that would ensure that repositories are only created via pull, thus no arbitrary naming could be used. But that would be out of the scope of this change IMO.

flavianmissi: Another (unrelated) weird thing is that when I, as a non-admin member of a proxy org (with creator role), pull an image for the first time (and thus trigger a repository creation), I will end up being the owner of that repository, and thus removing me from the creator team will not remove my permissions to the repository(ies) I created, since I own them.
I can't claim to fully understand all of Quay's quirks, but I have tested the above scenario. Not sure if this is something we want to change generally, but it feels kinda weird when it comes to proxy orgs.

dmage: can we create a repo without an owner?

flavianmissi: It does seem to work, although I only tested the happy path.

flavianmissi: well right now for users to be able to pull they need to have admin, with the changes here they will only need to be creator along with having read/write on other peoples repositories, so if you ask me we're making it better, even if not resolving the issue entirely.
and yes, the idea is to backport it to 3.7.

flavianmissi: We have a PM sync later today, let's discuss it there?

flavianmissi:
Changed creator role check to member role check instead, as per our agreement in the PM sync.
@dmage can you take another look?

dmage (dismissed):

dmage (approved):

flavianmissi:
manual cherry pick: https://github.com/quay/quay/pull/1379

PR #1370, opened 2022-06-09
Reduce memory usage

dmage:
Reduce notificationworker's memory usage from 110 MB to 75 MB.

### DIFF
--- a/app.py
+++ b/app.py
@@ -3,13 +3,9 @@ import json
 import logging
 import os
 
-from functools import partial
-
 from authlib.jose import JsonWebKey
-from cryptography.hazmat.primitives import serialization
-from flask import Flask, request, Request
+from flask import request, Request
 from flask_login import LoginManager
-from flask_mail import Mail
 from flask_principal import Principal
 from werkzeug.middleware.proxy_fix import ProxyFix
 from werkzeug.exceptions import HTTPException
@@ -17,15 +13,12 @@ from werkzeug.exceptions import HTTPException
 import features
 
 from _init import (
-    config_provider,
-    CONF_DIR,
     IS_KUBERNETES,
     IS_TESTING,
     OVERRIDE_CONFIG_DIRECTORY,
     IS_BUILDING,
 )
 
-from avatars.avatars import Avatar
 from buildman.manager.buildcanceller import BuildCanceller
 from data import database
 from data import model
@@ -35,7 +28,6 @@ from data.billing import Billing
 from data.buildlogs import BuildLogs
 from data.cache import get_model_cache
 from data.model.user import LoginWrappedDBUser
-from data.queue import WorkQueue
 from data.userevent import UserEventsBuilderModule
 from data.userfiles import Userfiles
 from data.users import UserAuthentication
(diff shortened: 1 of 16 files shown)


app.py
```diff
@@ -70,62 +59,34 @@
...
-
-    logger.debug("Loading test config.")
-    app.config.from_object(TestConfig())
-else:
-    from config import DefaultConfig
-
-    logger.debug("Loading default config.")
-    app.config.from_object(DefaultConfig())
+if not is_testing:
     app.teardown_request(database.close_db_filter)
```
dmage, line 84: Any ideas why we don't need it in testing?

util/ssl.py
```diff
@@ -0,0 +1 @@
+SSL_FILENAMES = ["ssl.cert", "ssl.key"]
```
dmage, line 1: After the great cleanup (#1384) `SSL_FILENAMES` will be only in `notifications/notificationmethod.py`.

HammerMeetNail (approved):
I like the use of singletons, adds a nice structure and makes it clear which things should only have one instance.

singletons/avatar.py
```diff
@@ -0,0 +1,4 @@
+from avatars.avatars import Avatar
+from singletons.app import _app
```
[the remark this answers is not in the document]
flavianmissi: rather `from flask import current_app as app` https://flask.palletsprojects.com/en/1.1.x/api/#flask.current_app

workers/notificationworker/notificationworker.py
```diff
@@ -48,15 +49,18 @@ def create_gunicorn_worker():
 
     utilizing this method will enforce a 1:1 quay worker to gunicorn worker ratio.
     """
+    from flask import Flask
+
     note_worker = NotificationWorker(
         notification_queue, poll_period_seconds=10, reservation_seconds=30, retry_after_seconds=30
     )
+    app = Flask(__name__)
```
flavianmissi, line 57: hmmm `create_gunicorn_worker` creates this flask app, and `from singletons.config import app_config` creates another one - do we really need both? furthermore, what kind of side effects could this cause? 🤔

flavianmissi (commented):
Left some comments, but looks good overall.

Do you think it's time we consider using an app factory? https://flask.palletsprojects.com/en/1.1.x/patterns/appfactories/

singletons/avatar.py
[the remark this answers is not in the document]
dmage: `_app` _is_ our Flask object. You can write:

```
from singletons.app import _app as app1
from app import app as app2
print(app1 is app2)
```

and it'll print you `True`. `current_app` is a smart proxy object that afaik only works in the context of an HTTP handler.

workers/notificationworker/notificationworker.py
dmage: This is just to make gunicorn happy. On next iterations most likely I'll remove `app` from GunicornWorker and replace it with some basic stub like `start_response("200 OK"); return []`. We don't expect it to serve any HTTP requests anyway, we only want it to watch code changes in local-dev.

dmage:
App factory is a good way to eliminate global variables, maybe in 2023/2024 😄

PR #1374, opened 2022-06-10
fix several warnings

flavianmissi:
mostly invalid escape sequences

### DIFF
--- a/buildtrigger/customhandler.py
+++ b/buildtrigger/customhandler.py
@@ -84,7 +84,7 @@ class CustomBuildTrigger(BuildTriggerHandler):
             "ref": {
                 "type": "string",
                 "description": "git reference for a git commit",
-                "pattern": "^refs\/(heads|tags|remotes)\/(.+)$",
+                "pattern": "^refs/(heads|tags|remotes)/(.+)$",
             },
             "default_branch": {
                 "type": "string",
--- a/image/docker/schema2/config.py
+++ b/image/docker/schema2/config.py
@@ -15,9 +15,9 @@ Example:
     "OpenStdin": false,
     "StdinOnce": false,
     "Env": [
-      "HTTP_PROXY=http:\/\/localhost:8080",
-      "http_proxy=http:\/\/localhost:8080",
-      "PATH=\/usr\/local\/sbin:\/usr\/local\/bin:\/usr\/sbin:\/usr\/bin:\/sbin:\/bin"
+      "HTTP_PROXY=http://localhost:8080",
+      "http_proxy=http://localhost:8080",
+      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
     ],
     "Cmd": [
       "sh"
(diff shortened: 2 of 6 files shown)


buildtrigger/customhandler.py
```diff
@@ -84,7 +84,7 @@ class CustomBuildTrigger(BuildTriggerHandler):
             "ref": {
                 "type": "string",
                 "description": "git reference for a git commit",
-                "pattern": "^refs\/(heads|tags|remotes)\/(.+)$",
+                "pattern": "^refs/(heads|tags|remotes)/(.+)$",
```
dmage, line 87: note that python interprets `"foo\/bar"` as `"foo\\/bar"`, so maybe it needs to be `r"^refs\/(heads|tags|remotes)\/(.+)$"`

image/docker/schema2/config.py
```diff
@@ -64,11 +64,11 @@
   "history": [
     {
       "created": "2018-04-03T18:37:09.284840891Z",
-      "created_by": "\/bin\/sh -c #(nop) ADD file:9e4ca21cbd24dc05b454b6be21c7c639216ae66559b21ba24af0d665c62620dc in \/ "
+      "created_by": "/bin/sh -c #(nop) ADD file:9e4ca21cbd24dc05b454b6be21c7c639216ae66559b21ba24af0d665c62620dc in / "
     },
     {
       "created": "2018-04-03T18:37:09.613317719Z",
-      "created_by": "\/bin\/sh -c #(nop)  CMD [\"sh\"]",
+      "created_by": "/bin/sh -c #(nop)  CMD ["sh"]",
```
dmage, line 71: `\"` seems to be legit, perhaps it should be `\\"` or the whole string should be `r"..."`

buildtrigger/customhandler.py
flavianmissi: I've tested both `r"^refs\/(heads|tags|remotes)\/(.+)$"` and `r"^refs/(heads|tags|remotes)/(.+)$"` and they seem to work the same way.

```python
import jsonschema
from pprint import pprint

schema = {
    "type": "object",
    "properties": {
        "ref": {
            "type": "string",
            "description": "git reference for a git commit",
            "pattern": r"^refs\/(heads|tags|remotes)\/(.+)$",
        },
    },
}

def check(obj, validator):
    pprint(obj)
    result = "VALID" if validator.is_valid(obj) else "INVALID"
    print(f"=> {result}")

validator = jsonschema.Draft4Validator(schema)
check({"ref": "refs/heads/master"}, validator)
# valid
check({"ref": "refs/tags"}, validator)
# invalid

# now without the escape
schema = {
    "type": "object",
    "properties": {
        "ref": {
            "type": "string",
            "description": "git reference for a git commit",
            "pattern": r"^refs/(heads|tags|remotes)/(.+)$",
        },
    },
}

validator = jsonschema.Draft4Validator(schema)
check({"ref": "refs/heads/master"}, validator)
# valid
check({"ref": "refs/tags"}, validator)
# invalid
```

image/docker/schema2/config.py
flavianmissi: good catch!

flavianmissi: fixed.

buildtrigger/customhandler.py
dmage: nice!

dmage (approved):

PR #1378, opened 2022-06-13
quota: Fix join on caching repo size (PROJQUAY-3889)

bcaton85:
Method for calculating size of repo joined on repo id instead of manifest id. This causes deleted manifests to be counted. Change updates call to join on manifest id.

### DIFF
--- a/data/model/repository.py
+++ b/data/model/repository.py
@@ -733,16 +733,17 @@ def force_cache_repo_size(repo_id: int):
     try:
         now_ms = get_epoch_timestamp_ms()
         subquery = (
-            Tag.select(Tag.repository_id)
+            Tag.select(Tag.manifest)
             .where(Tag.hidden == False)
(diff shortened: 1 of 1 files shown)


data/model/repository.py
```diff
@@ -733,16 +733,16 @@ def force_cache_repo_size(repo_id: int):
     try:
         now_ms = get_epoch_timestamp_ms()
         subquery = (
-            Tag.select(Tag.repository_id)
+            Tag.select(Tag.manifest)
             .where(Tag.hidden == False)
             .where((Tag.lifetime_end_ms >> None) | (Tag.lifetime_end_ms > now_ms))
```
kwestpharedhat, line 738: Its missing on both sides but lets add a repository_id where clause here in the subquery as well since we are here and I think that will make it more efficient

kwestpharedhat (changes requested):

PR #1379, opened 2022-06-14
 [redhat-3.7] fix: enable non-admins to cache images via pull-thru (PROJQUAY-3806)

flavianmissi:
introduces a check of the OrganizationMemberPermission for pulls
against a proxy org.

if the proxy cache feature is disabled, or the org is not a proxy org
the check is not performed and Quay will behave normally.

this check does not mean pulls will work transparently though -
non-admin users need to be added to a team in the proxy org with the
member role, and default read and write permissions need to be given to
that team so that non-admin users can pull and update the cache for
images they do not own (the user who first pulls an image ends up
owning the repository since that is when the repo gets created).


HammerMeetNail (approved):
LGTM 👍

PR #1382, opened 2022-06-14
chore: Use Python 3.9

dmage:


flavianmissi (commented):
this looks great!
how about updating the required python version on our getting started docs as well? https://github.com/quay/quay/blob/master/docs/getting-started.md#requirements

flavianmissi (approved):
LGTM 🚀

HammerMeetNail:
Did the base64 decodestring import get fixed? I think that was removed in python 3.9

dmage:
`elasticsearch==7.6.0` doesn't use decodestring anymore.

PR #1387, opened 2022-06-15
storage: optimize large azure chunked upload (PROJQUAY-3753)

kleesc:
- Increase nginx send timeout on blobs endpoints
- Reduce Azure blob block size

TODO: Spread block uploads over multiple worker threads.

### DIFF
--- a/storage/azurestorage.py
+++ b/storage/azurestorage.py
@@ -11,6 +11,7 @@ import uuid
 import copy
 import time
 
+import concurrent.futures
 from datetime import datetime, timedelta
 
 from azure.core.exceptions import AzureError, ResourceNotFoundError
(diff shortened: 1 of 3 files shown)


storage/azurestorage.py
```diff
@@ -232,41 +233,41 @@ def stream_upload_chunk(self, uuid, offset, length, in_fp, storage_metadata, con
         total_bytes_written = 0
 
         while True:
-            current_length = length - total_bytes_written
             max_length = (
-                min(current_length, self._max_block_size)
+                min(length - total_bytes_written, self._max_block_size)
```
bcaton85, line 237: should `self._max_block_size` here also be set to the smaller block size (`1024 * 1024 * 128`)?

kleesc: Updated to use the minimum of that value or what the api supports.

bcaton85 (approved):
LGTM

PR #1389, opened 2022-06-16
Enable some flake8 checks in CI

dmage:


util/config/validators/validate_secscan.py
```diff
@@ -54,5 +54,5 @@ def validate(cls, validator_context):
             message = str(last_exception)
             raise ConfigValidationException("Could not ping security scanner: %s" % message)
         elif not response.get("state"):
-            message = "Invalid indexer state" % (response.status_code, response.text)
+            message = "Invalid indexer state"
```
dmage, line 57: @kleesc you authored this line, is my change correct?

data/appr_model/channel.py
```diff
@@ -19,7 +19,7 @@ def get_channel_releases(repo, channel, models_ref):
             Channel.repository == repo,
             Channel.name == channel_name,
             Channel.tag_kind == tag_kind_id,
-            Channel.lifetime_end != None,
+            Channel.lifetime_end.is_null(False),
```
kleesc, line 22: Is that a change with made by Flake8?

image/docker/schema2/test/test_manifest.py
```diff
@@ -293,8 +287,8 @@ def test_get_schema1_manifest():
                 {"created": "2018-04-12T18:37:09.284840891Z", "created_by": "bar"},
             ],
         },
-        "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7",
-        1885,
+        CONFIG_DIGEST,
+        CONFIG_SIZE,
```
kleesc, line 291: Were these changes inferred by Flake8? Neat if so.

kleesc (commented):
Otherwise LGTM

data/appr_model/channel.py
kleesc: Isn't lifetime_end just an integer?

kleesc:
Do we still need to run `black` as part of CI if this is included, or do they intersect in what they do?

data/appr_model/channel.py
dmage: flake8 complained about `!= None`, and @flavianmissi pointed that `.is_null` is the right way to do it.

https://github.com/quay/quay/blob/d9dbbd88dbb9a1da95ef46195bcfb5458db6a7b8/data/database.py#L1733
If that's the right model, lifetime_end can indeed be null.

image/docker/schema2/test/test_manifest.py
dmage: Almost, flake8 complained about `\/` in the context where it should be just `/` (as the original json has `\/` that is interpreted as `/`). I could incorrectly replace it with `\\/`, but I decided to fix it, so it has changed its size and digest.

data/appr_model/channel.py
kleesc: Does it complain on `is not None`?

kleesc: Ah, it's peewee's wrapper for  numbers.

dmage: It doesn't complain on `is not None`, but `is not None` has different semantics and breaks tests.

kleesc (approved):

PR #1391, opened 2022-06-16
quota: Fix calculating org size (PROJQUAY-3889)

bcaton85:
Fixing database call which will now match on manifest vs repository to correctly calculate org size.


dmage:
Looks good, can cover it with a test?

kwestpharedhat (approved):

kwestpharedhat (dismissed):

kwestpharedhat (approved):

bcaton85:
@dmage As of now there isn't an easy way to add a test for this. I'll follow up with tests in a separate commit.

PR #1396, opened 2022-06-17
Initialize database and log_model on their import

dmage:
This fixes notificationworker as it no longer depends on `app.py`.

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -47,6 +47,7 @@ from data.text import match_mysql, match_like
 from data.encryption import FieldEncrypter
 from data.readreplica import ReadReplicaSupportedModel, ReadOnlyConfig, disallow_replica_use
 from data.estimate import mysql_estimate_row_count, normal_row_count
+from singletons.config import app_config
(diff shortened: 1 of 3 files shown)


data/database.py
```diff
@@ -2149,3 +2150,5 @@ def upstream_registry_hostname(self) -> str:
 
 is_model = lambda x: inspect.isclass(x) and issubclass(x, BaseModel) and x is not BaseModel
 all_models = [model[1] for model in inspect.getmembers(sys.modules[__name__], is_model)]
+
+configure(app_config)
```
flavianmissi, line 2154: won't this run everytime this module is imported?

dmage: it'll be run only once on the first import and its results will be cached in sys.modules, subsequent imports reuse the cached object

dmage: the same thing happens today with `from app import app`, `app` is executed only once even though it is imported into many places

flavianmissi (approved):

PR #1397, opened 2022-06-17
Fix unit tests on macOS

dmage:
Tests that rely on multiprocessing to spawn liveserver fail on macOS:

    Can't pickle local object 'liveserver_app.<locals>._spawn_live_server.<locals>.<lambda>'

Changing the start method to "fork" addresses the problem.


flavianmissi (approved):
💯

PR #1402, opened 2022-06-20
chore: Fix startup script

ibazulic:
The startup script still references the `certs_create.sh` file due to which config-tool startup fails.
This PR removes the reference.


kleesc (approved):
lgtm

PR #1410, opened 2022-06-23
Update type hints for supervisord_conf_create.py

dmage:
I wanted to update this script so that it doesn't start workers for disabled features, but apparently all features are enabled by default, so by default it'll start all workers anyway. Therefore here is it just my preparation cleanup work.

### DIFF
--- a/conf/init/supervisord_conf_create.py
+++ b/conf/init/supervisord_conf_create.py
@@ -1,7 +1,8 @@
-from typing import Union, List
 import os
 import os.path
 import sys
+from typing import Dict, Optional, TypedDict
+
 import jinja2
 
 QUAYPATH = os.getenv("QUAYPATH", ".")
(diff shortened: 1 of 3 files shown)


features/__init__.py
```diff
@@ -10,6 +12,10 @@ def import_features(config_dict):
             _FEATURES[feature_name] = globals()[feature_name] = FeatureNameValue(
                 feature_name, feature_val
             )
+    for key, val in os.environ.items():
+        if key.startswith("QUAY_FEATURE_"):
```
dmage, line 16: In some cases we don't need FEATURE_SECURITY_SCANNER in our local-dev, so I added ability to enable/disable features via environment variables.

Makefile
```diff
@@ -204,7 +204,7 @@ local-dev-up: local-dev-clean node_modules | build-image-quay
 	DOCKER_USER="$$(id -u):$$(id -g)" $(DOCKER_COMPOSE) up -d --force-recreate local-dev-frontend
 	$(DOCKER_COMPOSE) up -d redis quay-db
 	$(DOCKER) exec -it quay-db bash -c 'while ! pg_isready; do echo "waiting for postgres"; sleep 2; done'
-	DOCKER_USER="$$(id -u):0" $(DOCKER_COMPOSE) up -d quay
+	DOCKER_USER="$$(id -u):0" QUAY_FEATURE_SECURITY_SCANNER=false $(DOCKER_COMPOSE) up -d quay
```
syed, line 207: isni't this controlled by `local-dev/stack/config.yaml` don't think we need to disable via env variable

dmage: We have `local-dev-up` and `local-dev-up-with-clair`, so far they both use the same config, but the security scanner worker cannot work without Clair and generates noise in logs. Either we need a new config file, or disabling features via environment variables.

PR #1412, opened 2022-06-27
Initialize model.config on its import (PROJQUAY-4006)

dmage:
It prevents model.config from being used uninitialized.


app.py
```diff
@@ -201,9 +200,6 @@ def _request_end(resp):
 model_cache = get_model_cache(app.config)
 login_manager = LoginManager(app)
 prometheus = PrometheusPlugin(app)
```
syed, line 202: Is the idea to eventually move these other objects to singletons as well?

syed (approved):
LGTM

app.py
dmage: the idea is to avoid all-in-one packages like `app`. Moving these objects into their own packages is the easiest way to achieve that.

PR #1415, opened 2022-06-28
Allow https + insecure pull-thru proxy support (PROJQUAY-3457)

harishsurf:

### DIFF
--- a/proxy/__init__.py
+++ b/proxy/__init__.py
@@ -55,15 +55,32 @@ class Proxy:
             config.upstream_registry_hostname,
             config.upstream_registry_hostname,
         )
-        url = f"https://{hostname}"
-        if config.insecure:
-            url = f"http://{hostname}"
 
-        self.base_url = url
         self._session = requests.Session()
+
+        self.base_url = self._get_valid_upstream_registry(hostname)
         self._repo = repository
         self._authorize(self._credentials(), force_renewal=self._validation)
-        # flag used for validating Proxy cache config before saving to db
+        # flag used for validating Proxy cache config before saving to db    
+
+    def _get_valid_upstream_registry(self, hostname: str) -> str :
+        url = f"https://{hostname}"
+        if self._config.insecure:
+            # ignore verifying the SSL certificate
+            self._session.verify = False
+        try:
+            resp = self._safe_request(self._session.get, f"{url}/v2")
+            if resp.ok or resp.status_code == 401:
+                return url
+        except requests.exceptions.SSLError:
+            raise UpstreamRegistryError(f"SSL certificate verification failed for upstream registry")
+        
+        # https + insecure failed, so fallback to http
+        url = f"http://{hostname}"
+        resp = self._safe_request(self._session.get, f"{url}/v2")
+        if not resp.ok and (resp.status_code == 404 or resp.status_code == 429):
+            raise UpstreamRegistryError(f"Connection to upstream registry failed with status code {resp.status_code}")
+        return url
 
     def get_manifest(
         self, image_ref: str, media_types: list[str] | None = None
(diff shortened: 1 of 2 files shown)


proxy/__init__.py
```diff
@@ -55,15 +55,38 @@ def __init__(self, config: ProxyCacheConfig, repository: str, validation: bool =
...
+                self._session.verify = False
+                resp = self._session.get(f"{url}/v2")
+                if resp.ok:
+                    return url
+        
+        # https + insecure failed, so fallback to http
+        url = f"http://{hostname}"
+        try:
+            resp = self._session.get(f"{url}/v2")
+        except requests.exceptions.SSLError as e:
```
flavianmissi, line 84: `requests` only raises a `SSLError` when SSL certificate validation fails[1] - there's no need to catch this exception here since the request at this point is using http, in which case there is no certificate to verify.

Here's the relevant bits from the docs:
> [...] and Requests will throw a SSLError if it’s unable to verify the certificate

[1] https://requests.readthedocs.io/en/latest/user/advanced/#ssl-cert-verification

proxy/__init__.py
```diff
@@ -55,15 +55,38 @@ def __init__(self, config: ProxyCacheConfig, repository: str, validation: bool =
...
-        self.base_url = url
         self._session = requests.Session()
+
+        self.base_url = self._get_valid_upstream_registry(hostname)
         self._repo = repository
         self._authorize(self._credentials(), force_renewal=self._validation)
-        # flag used for validating Proxy cache config before saving to db
+        # flag used for validating Proxy cache config before saving to db    
+
+    def _get_valid_upstream_registry(self, hostname: str) -> str :
```
flavianmissi, line 66: neat with typing! 💯

flavianmissi (commented):
Great work, Harish!
Left some comments.

proxy/__init__.py
```diff
@@ -55,15 +55,38 @@ def __init__(self, config: ProxyCacheConfig, repository: str, validation: bool =
...
+        url = f"https://{hostname}"
+        try:
+            resp = self._session.get(f"{url}/v2")
```
flavianmissi, line 69: use `self._safe_request` everywhere in this function. there's an example of it being used in `_authorize`

proxy/__init__.py
```diff
@@ -55,15 +55,38 @@ def __init__(self, config: ProxyCacheConfig, repository: str, validation: bool =
...
+            if resp.ok:
```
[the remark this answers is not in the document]
harishsurf: good call, added a check!

proxy/__init__.py
```diff
@@ -55,15 +55,32 @@ def __init__(self, config: ProxyCacheConfig, repository: str, validation: bool =
...
-        # flag used for validating Proxy cache config before saving to db
+        # flag used for validating Proxy cache config before saving to db    
+
+    def _get_valid_upstream_registry(self, hostname: str) -> str :
+        url = f"https://{hostname}"
+        if self._config.insecure:
+            # ignore verifying the SSL certificate
+            self._session.verify = False
+        try:
+            resp = self._safe_request(self._session.get, f"{url}/v2")
```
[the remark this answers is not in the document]
harishsurf: > It'll cause Quay to send a second request on every incoming requests
 
how do you mean? if the upstream registry supports `https`, it'll return after the first `get` req on #72.  A second request is made when the upstream registry only supports `http`

> Can we ask the user to provide whether it is http or https+insecure?

Do you mean having one more checkbox? We have 3 possible workflows here (described in the [jira](https://issues.redhat.com/browse/PROJQUAY-3457)). The `insecure` checkbox in the current UI is to decide whether Quay should verify the SSL cert or not (when checked). As per the Jira it was to do a `https` first and on error try `http` as a fallback before returning.

PR #1429, opened 2022-06-30
chore: Add flit_core to deps (PROJQUAY-4033)

HammerMeetNail:
* Required by `typing-extensions`
* Should resolve downstream error:
```
2022-06-30 19:19:13,357 - atomic_reactor.plugins.imagebuilder - INFO -   ERROR: Could not find a version that satisfies the requirement flit_core<4,>=3.4 (from versions: none)
2022-06-30 19:19:13,357 - atomic_reactor.plugins.imagebuilder - INFO -   ERROR: No matching distribution found for flit_core<4,>=3.4
```

PR #1430, opened 2022-06-30
Check that requirements.txt has all dependencies

dmage:
It could help avoid #1428 (and maybe #1429).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -130,7 +130,7 @@ text-unidecode==1.3
 tldextract==2.2.2
 toposort==1.5
 tzlocal==2.0.0
-urllib3==1.26.9 
+urllib3==1.26.9
 webencodings==0.5.1
 WebOb==1.8.6
 websocket-client==0.57.0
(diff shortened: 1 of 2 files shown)


HammerMeetNail:
This is interesting! I'd been using `pip show` to try and find the deps before adding to `requirements.txt`, but it didn't show any build deps. Even after install of `azure-core`, there was still no reference in `pip freeze` or anything to `flit_core`.

dmage:
I excluded pip (`s/^pip==.*/pip/`) because it stays 22.0.4 despite requirements.txt

HammerMeetNail:
@dmage How would this have caught `flit-core`? I think it's cross checking against `pip freeze`, but `flit-core` comes as part of `typing-extensions` and doesn't show in `pip freeze`.

requirements.txt
```diff
@@ -143,8 +143,8 @@ zope.event==4.5.0
 zope.interface==5.4.0
 
 # Build dependencies
-Cython==3.0a9
-flit_core==3.7.1
+Cython==3.0.0a9
+flit-core==3.7.1
```
HammerMeetNail, line 147: Is this right? I'd confirmed that `pip install flit_core==3.7.1` installs. https://pypi.org/project/flit_core/

HammerMeetNail (approved):
LGTM 👍

dmage:
We've got a new version of `pip` in master, so I updated this PR.

HammerMeetNail (approved):
LGTM 👍

PR #1432, opened 2022-07-01
dev: Add pre-commit script to run black (PROJQUAY-4039)

syed:
Automatically formats python files before commit


syed:
@HammerMeetNail  added a target for installing the pre-commit hook

PR #1434, opened 2022-07-05
secscan: Don't delete manifest security status on error (PROJQUAY-4060) 

crozzy:
Currently if Clair returns errors Quay will delete the security
status for that manifest, meaning it will revert to a "Queued"
status and needs to be reindexed. This in-turn means that the
functionality is not immediately recovered when Clair becomes
healthy. It is also generally a bad idea (except in specific
use-cases to have a data modification in a read path).

### DIFF
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -154,11 +154,6 @@ class V4SecurityScanner(SecurityScannerInterface):
         try:
             report = self._secscan_api.vulnerability_report(manifest_or_legacy_image.digest)
         except APIRequestFailure as arf:
-            try:
-                status.delete_instance()
-            except ReadOnlyModeException:
-                pass
-
             return SecurityInformationLookupResult.for_request_error(str(arf))
 
         if report is None:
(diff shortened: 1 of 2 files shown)


data/secscan_model/secscan_v4_model.py
```diff
@@ -154,11 +154,6 @@ def load_security_information(self, manifest_or_legacy_image, include_vulnerabil
         try:
             report = self._secscan_api.vulnerability_report(manifest_or_legacy_image.digest)
         except APIRequestFailure as arf:
-            try:
-                status.delete_instance()
-            except ReadOnlyModeException:
-                pass
-
```
kleesc, line 161: Are there any case where we _should_ be requeuing the manifest for indexing, depending on Clair's response?
e.g a 5XX vs 4XX status code

crozzy: I don't believe so, if the indexing was successful the vulnerability report should be available, if it isn't it's a transient error. @hdonnay can you think of a reason we'd re-queue a manifest when the vulnerability report request is now successful?

hdonnay: No, I don't think so. We do need to be able to indicate permanent/transient errors back to quay, but we don't have that yet.

kleesc (approved):

PR #1438, opened 2022-07-07
[WIP] Add new ui (PROJQUAY-4124)

HammerMeetNail:
* Adds new patternfly ui to quay container


Dockerfile
```diff
@@ -66,15 +67,24 @@ RUN set -ex\
...
+# Build-static downloads the legacy Quay UI.
+FROM registry.access.redhat.com/ubi8/nodejs-10 AS build-legacy-ui
 WORKDIR /opt/app-root/src
 COPY --chown=1001:0 static/  ./static/
 COPY --chown=1001:0 *.json *.js  ./
-RUN set -ex\
-	; npm install --quiet --no-progress --ignore-engines\
-	; npm run --quiet build\
-	;
+RUN set -ex &&\
```
syed, line 75: `quay.io/projectquay/quay-ui:latest` already has built assets. Why don't we just copy that?

Dockerfile
```diff
@@ -146,7 +158,7 @@ RUN set -ex\
 
 RUN rm -Rf node_modules config_app/node_modules
 
-EXPOSE 8080 8443 7443 9091 55443
+EXPOSE 8080 8443 7443 9000 9091 55443
```
syed, line 161: Why expose port 9000, new UI should be routed via the same port, there could be a header or a cookie switch between new and old ui

util/request.py
```diff
@@ -27,22 +27,25 @@ def crossorigin(anonymous=True):
     def decorate(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
-            cors_origin = app.config.get("CORS_ORIGIN", "*")
+            cors_origins = app.config.get("CORS_ORIGINS", "*")
```
syed, line 30: if you are taking an input as list then this would be `["*"]` no?

util/request.py
```diff
@@ -27,22 +27,25 @@ def crossorigin(anonymous=True):
...
             # request
-            if not anonymous and cors_origin == "*":
+            if not anonymous and cors_origins == "*":
                 return func(*args, **kwargs)
 
             credentials = False
-            if cors_origin != "*":
+
+            origin = request.environ.get("HTTP_ORIGIN")
+            if "*" not in cors_origins and origin in cors_origins:
```
syed, line 42: not sure I understand the need for the `HTTP_ORIGIN` environment variable

syed (changes requested):

syed:
We can just add `CORS_ORIGIN: domain1, domain2, etc` instead of list, that's how the underlying CORS library expects the input

Dockerfile
HammerMeetNail: @syed How would the header or cookie be set? If a user is trying to navigate to the old ui at `/` and the new ui at `/`, how would we set the header?

util/request.py
HammerMeetNail: yep, good catch

util/request.py
HammerMeetNail: In the current implementation we rely on the `cors_origin` to be whatever is set in `CORS_ORIGIN`, but now that `CORS_ORIGIN` could be multiple origins, we need to get the requested origin and check if it's in the list of supported `CORS_ORIGIN`

Dockerfile
syed: and in the new UI we can have another banner saying to go back to the old UI click here. We could have this till we have feature parity. Then we just remove the old UI. For this to work, both the old UI and new UI should run on the same port

util/request.py
syed: I'd say leave it as it is and just use a comma separated list in the input

HammerMeetNail:
> We can just add `CORS_ORIGIN: domain1, domain2, etc` instead of list, that's how the underlying CORS library expects the input

Huh, I was trying to find docs on that but didn't have any luck. I tried with `CORS_ORIGIN: "http://localhost:9000,http://localhost:8081"` and that didn't work. I'll go back and see what I can pass to `crossdomain`

Dockerfile
HammerMeetNail: Ok, I like that. It'll also mean we don't need to change anything related to `cors_origin`

PR #1439, opened 2022-07-08
quayio: Redirect user to billing page when starting free trial (PROJAUAY-4130)

jonathankingfc:
- When a user clicks 'Start Free Trial', they should be redirected to the user billing page after signing in with RHSSO

### DIFF
--- a/static/js/services/plan-service.js
+++ b/static/js/services/plan-service.js
@@ -69,7 +69,8 @@ function(KeyService, UserService, CookieService, ApiService, Features, Config, $
         if (planService.isOrgCompatible(plan)) {
           $location.path('/organizations/new').search('plan', planId);
         } else {
-          $location.path('/user').search('plan', planId);
+          const username = UserService.currentUser().username;
+          $location.path(`/user/${username}/billing`);
         }
       });
     });


static/js/services/plan-service.js
```diff
@@ -69,7 +69,8 @@ function(KeyService, UserService, CookieService, ApiService, Features, Config, $
         if (planService.isOrgCompatible(plan)) {
           $location.path('/organizations/new').search('plan', planId);
         } else {
-          $location.path('/user').search('plan', planId);
+          const username = UserService.currentUser().username;
+          $location.path(`/user/${username}/billing`);
```
Sunandadadi, line 73: Just want to confirm if this the correct endpoint as I was not able to find a flask route for it.

jonathankingfc: This isn't an API route. It is a frontend route. You can check the frontend route here `https://quay.io/user/{your_username}/billing`

Sunandadadi: Oh! Got it. Thanks!

Sunandadadi (approved):

Sunandadadi:
lgtm

PR #1445, opened 2022-07-15
api: fix CORS headers, use concat instead of extend (PROJQUAY-4163)

syed:
Fixes issue where we send large header on CORS


HammerMeetNail (approved):
LGTM

PR #1454, opened 2022-07-22
Add Feature flag to whitelist users from recaptcha security check (PROJQUAY-3697)

harishsurf:
Add Feature flag to whitelist users from recaptcha security check(PROJQUAY-3697)

### DIFF
--- a/config.py
+++ b/config.py
@@ -592,6 +592,9 @@ class DefaultConfig(ImmutableConfig):
     RECAPTCHA_SITE_KEY: Optional[str] = None
     RECAPTCHA_SECRET_KEY: Optional[str] = None
 
+    # List of users allowed to pass through recaptcha security check to enable org/user creation via API
+    RECAPTCHA_WHITELISTED_USERS: List[str] = []
+
     # Server where TUF metadata can be found
     TUF_SERVER = None
 
--- a/endpoints/api/user.py
+++ b/endpoints/api/user.py
@@ -516,13 +516,16 @@ class User(ApiResource):
 
         # If recaptcha is enabled, then verify the user is a human.
         if features.RECAPTCHA:
-            recaptcha_response = user_data.get("recaptcha_response", "")
-            result = recaptcha2.verify(
-                app.config["RECAPTCHA_SECRET_KEY"], recaptcha_response, get_request_ip()
-            )
+            user = get_authenticated_user()
+            # check if the user is whitelisted to bypass recaptcha security check
+            if user is None or (user.username not in app.config["RECAPTCHA_WHITELISTED_USERS"]):
+                recaptcha_response = user_data.get("recaptcha_response", "")
+                result = recaptcha2.verify(
+                    app.config["RECAPTCHA_SECRET_KEY"], recaptcha_response, get_request_ip()
+                )
 
-            if not result["success"]:
-                return {"message": "Are you a bot? If not, please revalidate the captcha."}, 400
+                if not result["success"]:
+                    return {"message": "Are you a bot? If not, please revalidate the captcha."}, 400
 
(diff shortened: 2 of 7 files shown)


config.py
```diff
@@ -592,6 +592,9 @@ def create_transaction(db):
     RECAPTCHA_SITE_KEY: Optional[str] = None
     RECAPTCHA_SECRET_KEY: Optional[str] = None
 
+    # List of users allowed to pass through recaptcha security check to enable org/user creation via API
+    RECAPTCHA_WHITELIST_USERS: List[str] = []
```
kleesc, line 596: RECAPTCHA_WHITELISTED_NAMESPACES

endpoints/api/user.py
```diff
@@ -480,6 +480,7 @@ def put(self):
 
     @show_if(features.USER_CREATION)
     @show_if(features.DIRECT_LOGIN)
+    @require_user_admin
```
kleesc, line 483: Why is this added?

kleesc (commented):
Otherwise lgtm

endpoints/api/user.py
harishsurf: the `get_authenticated_user()` was returning a `NoneType` exception without it

config.py
syed: should it be `NAMESPACES` instead of `USERS` ?

endpoints/api/user.py
kleesc: Which is expected if this is called by an anonymous user.

kleesc (changes requested):

endpoints/api/user.py
kleesc: Say when a new user is registering for the first time.

test/test_api_usage.py
```diff
@@ -814,6 +814,11 @@ def captcha_endpoint(url, request):
                 details["recaptcha_response"] = "somecode"
                 self.postResponse(User, data=details, expected_code=200)
 
+    def test_recaptcha_whitelist_users(self):
+        with (self.toggleFeature("RECAPTCHA", True)):
+            app.config["RECAPTCHA_WHITELIST_USERS"] = ["bobby"]
```
kleesc, line 819: Can you reference "bobby" as `NEW_USER_DETAILS["username"]` instead?

kleesc (changes requested):

config.py
harishsurf: wouldn't `Namespaces` include orgs? However this list would only take in `users` as input?

kleesc: Right, RECAPTCHA_WHITELISTED_USERS should be fine then.

endpoints/api/user.py
harishsurf: ah, yes, I've updated it. PTAL

endpoints/api/user.py
```diff
@@ -516,13 +516,16 @@ def post(self):
 
         # If recaptcha is enabled, then verify the user is a human.
         if features.RECAPTCHA:
-            recaptcha_response = user_data.get("recaptcha_response", "")
-            result = recaptcha2.verify(
-                app.config["RECAPTCHA_SECRET_KEY"], recaptcha_response, get_request_ip()
-            )
+            user = get_authenticated_user()
+            # check if the user is whitelisted to bypass recaptcha security check
+            if user is None or (user is not None and user.username not in features.RECAPTCHA_WHITELISTED_USERS):
```
kleesc, line 521: You don't need `user is not None` since it's a disjunction. Just `user is None or user.username not in features.RECAPTCHA_WHITELISTED_USERS`

kleesc (changes requested):

kleesc:
LGTM otherwise

kleesc (approved):
LGTM

PR #1457, opened 2022-07-26
oauth: fix oauth token generation when using dangerous scopes (PROJQUAY-4042)

syed:
Fixes issue where we are not passing the response type to oauth
with dangerous scopes


kleesc (approved):
lgtm

PR #1479, opened 2022-08-11
registry: correctly bump tag expiration when tag changes upstream (PROJQUAY-3976)

flavianmissi:
fixes [PROJQUAY-3976](https://issues.redhat.com//browse/PROJQUAY-3976)

### DIFF
--- a/data/registry_model/registry_proxy_model.py
+++ b/data/registry_model/registry_proxy_model.py
@@ -1,10 +1,13 @@
 from __future__ import annotations
+import logging
 
 from typing import Callable
+from peewee import Select, fn
 
 import features
 from app import app, storage
 from data.database import (
+    db,
     get_epoch_timestamp_ms,
     db_disallow_replica_use,
     db_transaction,
@@ -52,6 +55,9 @@ from image.docker.schema1 import (
 )
 from proxy import Proxy, UpstreamRegistryError
 from util.bytes import Bytes
+
+
+logger = logging.getLogger(__name__)
 
 
 ACCEPTED_MEDIA_TYPES = [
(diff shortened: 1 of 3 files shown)


data/registry_model/registry_proxy_model.py
```diff
@@ -185,13 +191,36 @@ def lookup_manifest_by_digest(
...
+                    .order_by_extend(TagTable.lifetime_end_ms.desc())
                 )
-                parent_ids = [link.manifest_id for link in links]
+                items = (
+                    Select(columns=[subq.c.rnk, subq.c.id, subq.c.name, subq.c.lifetime_end_ms])
+                    .from_(subq)
+                    .where(subq.c.rnk == 1)
+                    .bind(db)
+                )
+                tag_ids = [item["id"] for item in items]
```
flavianmissi, line 220: this query turned out way more complicated than I wanted it to be. I started using `DISTINCT ON` but sqlite doesn't support that (only `DISTINCT` without `ON`), so this is how I managed to make the same query work on both postgres and sqlite.

flavianmissi: ugh typical
https://github.com/quay/quay/runs/7790348396?check_suite_focus=true#step:5:1181

data/registry_model/registry_proxy_model.py
```diff
@@ -185,13 +191,36 @@ def lookup_manifest_by_digest(
...
+                    .join(ManifestChild, on=(TagTable.manifest_id == ManifestChild.manifest_id))
+                    .where(
+                        (ManifestChild.repository_id == repository_ref.id)
+                        & (ManifestChild.child_manifest_id == wrapped_manifest.id)
+                    )
+                    .order_by(TagTable.name)
+                    .order_by_extend(TagTable.lifetime_end_ms.desc())
                 )
-                parent_ids = [link.manifest_id for link in links]
+                items = (
```
dmage, line 214: Will this query use indexes?

flavianmissi: I'm not sure about the order by name tbh, but that's there for troubleshooting purposes so I'll remove it. the rest should be using the indexes.

HammerMeetNail (approved):
LGTM 👍

HammerMeetNail (approved):

PR #1483, opened 2022-08-15
oci: Disable active_repo_tags_cache for tests (PROJQUAY-3872)

dmage:
The OCI Content Discovery tests requires Quay to immediately reflect tag changes in tag lists when a tag is created/deleted, so to make tests green we should disable active_repo_tags_cache.

### DIFF
--- a/util/expiresdict.py
+++ b/util/expiresdict.py
@@ -50,7 +50,7 @@ class ExpiresDict(object):
         # Otherwise the key has expired or was not found. Rebuild the cache and check it again.
         items = self._rebuild()
         found_item = items.get(key)
-        if found_item is None:
+        if found_item is None or found_item.expired:
             return default_value
 
         return found_item.value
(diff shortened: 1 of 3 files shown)


util/expiresdict.py
```diff
@@ -50,7 +50,7 @@ def get(self, key, default_value=None):
         # Otherwise the key has expired or was not found. Rebuild the cache and check it again.
         items = self._rebuild()
         found_item = items.get(key)
-        if found_item is None:
+        if found_item is None or found_item.expired:
```
bcaton85, line 53: Was this required for tests to pass?

bcaton85 (approved):
LGTM

util/expiresdict.py
dmage: yes, otherwise the in-memory cache can incorrectly return an expired item

PR #1485, opened 2022-08-17
Makefile: use non-standard port for postgres test container

flavianmissi:
this way the tests can be run in parallel with a quay instance running locally.


bcaton85 (approved):
LGTM

PR #1486, opened 2022-08-17
storage: Add username field to requests on CloudFront (PROJQUAY-3511)

syed:
This is to corelate the a user with CDN traffic in CloudFront logs

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -230,7 +230,7 @@ class _CloudStorage(BaseStorageV2):
         return True
 
     def get_direct_download_url(
-        self, path, request_ip=None, expires_in=60, requires_cors=False, head=False
+        self, path, request_ip=None, expires_in=60, requires_cors=False, head=False, user=None
     ):
         self._initialize_cloud_conn()
         path = self._init_path(path)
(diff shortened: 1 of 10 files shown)


storage/cloud.py
```diff
@@ -1009,12 +1010,12 @@ def __init__(
         self.cloudfront_privatekey = self._load_private_key(cloudfront_privatekey_filename)
 
     def get_direct_download_url(
-        self, path, request_ip=None, expires_in=60, requires_cors=False, head=False
+        self, path, request_ip=None, expires_in=60, requires_cors=False, head=False, user=None
```
kleesc, line 1013: nit: Unless we need the full context, I would just pass the username as a `str`, or `**kwargs`, so that in the future we more parameters are needed, they can just be passed arbitrarily.

storage/cloud.py
```diff
@@ -1026,10 +1027,13 @@ def get_direct_download_url(
         logger.debug("Resolved IP information for IP %s: %s", request_ip, resolved_ip_info)
         if resolved_ip_info and resolved_ip_info.provider == "aws":
             return super(CloudFrontedS3Storage, self).get_direct_download_url(
-                path, request_ip, expires_in, requires_cors, head
+                path, request_ip, expires_in, requires_cors, head, user
             )
 
         url = "https://%s/%s" % (self.cloudfront_distribution_domain, path)
+        if user and user.username:
+            url += "?user=%s" % user.username
```
kleesc, line 1035: Then here you could just do something like
``` 
if kwargs:
   url +=  "?{}".format(urllib.parse.urlencode(kwargs))
```

kleesc (approved):
LGTM

PR #1490, opened 2022-08-22
mirror: Default mirror rollback to false (PROJQUAY-4296)

bcaton85:
Adds the `REPO_MIRROR_ROLLBACK` option to specify whether the mirror will rollback the state of the repo on failure of any one of the tags. Defaults to false. Adds additional `PARTIAL_SYNC` error status which logs the tags that failed to sync to the console.

### DIFF
--- a/workers/repomirrorworker/__init__.py
+++ b/workers/repomirrorworker/__init__.py
@@ -154,6 +154,7 @@ def perform_mirror(skopeo, mirror):
     # Sync tags
     now_ms = database.get_epoch_timestamp_ms()
     overall_status = RepoMirrorStatus.SUCCESS
+    failed_tags = []
     try:
         try:
             username = (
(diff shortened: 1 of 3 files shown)


workers/repomirrorworker/__init__.py
```diff
@@ -257,14 +259,33 @@ def perform_mirror(skopeo, mirror):
         return
     finally:
         if overall_status == RepoMirrorStatus.FAIL:
+            log_tags = []
+            log_message = "'%s' with tag pattern '%s'"
+
+            # Handle the case where not all tags were synced and state will not be rolled back
+            if (
+                len(failed_tags) != len(tags)
```
HammerMeetNail, line 267: Could this be simplified by just checking if there's `failed_tags`? Is there a situation where `failed_tags` and `tags` would be equal?

bcaton85: If `failed_tags` is equal to `tags` it isn't considered a partial failure, since all tags have failed instead of only some.

HammerMeetNail (approved):
LGTM

PR #1493, opened 2022-08-24
proxy: Evict LRU images when proxy cache org reaches limit (PROJQUAY-3063)

harishsurf:
Adds ability to automatically prune images from a proxy org when it's quota limit is reached 
- The tags marked for delete are eventually garbage collected by a GC worker

### DIFF
--- a/data/registry_model/registry_proxy_model.py
+++ b/data/registry_model/registry_proxy_model.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 import logging
 
 from typing import Callable
@@ -56,9 +57,7 @@ from image.docker.schema1 import (
 from proxy import Proxy, UpstreamRegistryError
 from util.bytes import Bytes
 
-
 logger = logging.getLogger(__name__)
-
 
 ACCEPTED_MEDIA_TYPES = [
     OCI_IMAGE_MANIFEST_CONTENT_TYPE,
@@ -131,6 +130,49 @@ class ProxyModel(OCIModel):
             state=repo.state if repo is not None else None,
         )
 
+    def _check_image_upload_possible_or_prune(
+        self, repo_ref: RepositoryReference, upstream_manifest: ManifestInterface
+    ) -> None:
+        """
+        Checks whether the given image fits within the quota size for the namespace
+        the repository is part of. If it doesn't, it prunes older tags in the namespace
+        by marking them expired which is eventually garbage collected by the gc worker.
+
+        Raises QuotaExceededException if the given tag is larger than the max quota
+        allotted for the namespace or if there are not enough tags to prune to free up space.
+        """
+        if upstream_manifest.is_manifest_list:
+            return
+
+        curr_ns_size = namespacequota.get_namespace_size(repo_ref.namespace_name)
+
+        # if quota limit is not set for the namespace, skip auto pruning of images
+        quotas = namespacequota.get_namespace_quota_list(repo_ref.namespace_name)
+        if quotas:
+            # currently only one quota per namespace is supported
+            ns_quota_limit = quotas[0].limit_bytes
+        else:
+            logger.info("No quota configured")
+            return
+
+        image_size = upstream_manifest.layers_compressed_size
+        if image_size > ns_quota_limit:
+            raise QuotaExceededException
+
+        if (curr_ns_size + image_size) <= ns_quota_limit:
+            return
+
+        tags = oci.tag.get_tag_with_least_lifetime_end_for_ns(repo_ref.namespace_name)
+        if tags is not None:
+            for tag in tags:
+                oci.tag.delete_tag(tag.repository_id, tag.name)
+                repository.force_cache_repo_size(tag.repository_id)
+                curr_ns_size = namespacequota.get_namespace_size(repo_ref.namespace_name)
+                if (curr_ns_size + image_size) <= ns_quota_limit:
+                    return
+        # if we got here, then there aren't enough tags in the namespace to expire, so we raise an exception
+        raise QuotaExceededException
+
     def lookup_manifest_by_digest(
         self,
         repository_ref,
@@ -145,6 +187,9 @@ class ProxyModel(OCIModel):
 
         If a manifest with the digest does not exist, fetches the manifest upstream
         and creates it with a temp tag.
+
+        Raises QuotaExceededException if the given tag is larger than the max quota
+        allotted for the namespace or if there are not enough tags to prune.
         """
         wrapped_manifest = super().lookup_manifest_by_digest(
             repository_ref, manifest_digest, allow_dead=True, require_available=False
@@ -171,8 +216,8 @@ class ProxyModel(OCIModel):
                 manifest_digest,
                 self._create_manifest_with_temp_tag,
             )
-        except ManifestDoesNotExist as e:
-            raise e
+        except ManifestDoesNotExist:
+            raise
         except UpstreamRegistryError:
             # when the upstream fetch fails, we only return the tag if
             # it isn't yet expired. note that we don't bump the tag's
@@ -225,6 +270,9 @@ class ProxyModel(OCIModel):
         from upstream, and creates them both.
         If tag and manifest exists and the manifest is a placeholder, pull the
         upstream manifest and save it locally.
+
+        Raises QuotaExceededException if the given tag is larger than the max quota
+        allotted for the namespace or if there are not enough tags to prune.
         """
         db_tag = oci.tag.get_current_tag(repository_ref.id, tag_name)
         existing_tag = Tag.for_tag(db_tag, self._legacy_image_id_handler)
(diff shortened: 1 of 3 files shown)


data/registry_model/registry_proxy_model.py
```diff
@@ -156,6 +215,8 @@ def lookup_manifest_by_digest(
                 )
             except (UpstreamRegistryError, ManifestDoesNotExist) as e:
                 raise ManifestDoesNotExist(str(e))
+            except QuotaExceededException as e:
+                raise e
```
flavianmissi, line 219: consider not catching just to raise it again. and if you do re-raise, use `raise` only (without the `e`)

data/registry_model/registry_proxy_model.py
```diff
@@ -171,7 +232,7 @@ def lookup_manifest_by_digest(
                 manifest_digest,
                 self._create_manifest_with_temp_tag,
             )
-        except ManifestDoesNotExist as e:
+        except (ManifestDoesNotExist, QuotaExceededException) as e:
             raise e
```
flavianmissi, line 236: could you remove the `e` from this line as well? we don't want to mess with the stacktrace.

data/registry_model/registry_proxy_model.py
```diff
@@ -233,7 +294,7 @@ def get_repo_tag(self, repository_ref, tag_name, raise_on_error=True):
                 _, tag = self._create_and_tag_manifest(
                     repository_ref, tag_name, self._create_manifest_and_retarget_tag
                 )
-            except (UpstreamRegistryError, ManifestDoesNotExist) as e:
+            except (UpstreamRegistryError, ManifestDoesNotExist, QuotaExceededException) as e:
```
flavianmissi, line 297: this changes behavior compared to the pull by tag - we want to propagate quota exceptions to the end user when they happen. there's no catch `QuotaExceededException` here, it'll be propagate to the end user and treated by api exception handlers.

data/registry_model/registry_proxy_model.py
```diff
@@ -246,7 +307,7 @@ def get_repo_tag(self, repository_ref, tag_name, raise_on_error=True):
                 tag_name,
                 self._create_manifest_and_retarget_tag,
             )
-        except ManifestDoesNotExist as e:
+        except (ManifestDoesNotExist, QuotaExceededException) as e:
```
flavianmissi, line 310: same as above

data/registry_model/registry_proxy_model.py
```diff
@@ -131,6 +130,66 @@ def lookup_repository(
...
+            logger.info(f"tags: given manifest is a list")
+            return True
+
+        upstream_blob_size = upstream_manifest.layers_compressed_size
+        logger.info(f"tags:upstream_manifest size: %s", upstream_blob_size)
+        logger.info(f"tags: curr_ns_size:  %s", curr_ns_size)
+        logger.info(f"tags: max_ns_quota:  %s", max_ns_quota)
+
+        if upstream_blob_size > max_ns_quota:
+            raise QuotaExceededException
```
flavianmissi, line 164: on this function you'll want to choose to either raise an exception or to return a boolean. the boolean and the exception are really about the same thing: whether the operation is allowed or not.

data/registry_model/registry_proxy_model.py
```diff
@@ -346,13 +407,14 @@ def _update_manifest_for_tag(
...
         placeholder = manifest.internal_manifest_bytes.as_unicode() == ""
+
         if up_to_date and not placeholder:
             return tag, False
 
         if upstream_manifest is None:
             upstream_manifest = self._pull_upstream_manifest(repo_ref.name, manifest_ref)
 
-        self._enforce_repository_quota(repo_ref)
+        # self._enforce_repository_quota(repo_ref)
```
flavianmissi, line 417: I think this is no longer called anywhere, if that's the case please remove it 💥

flavianmissi (commented):
first round, great work so far! 💯 🎉 🌮

data/registry_model/registry_proxy_model.py
harishsurf: removed `QuotaExceededException`

data/registry_model/registry_proxy_model.py
harishsurf: done

data/registry_model/test/test_registry_proxy_model.py
```diff
@@ -1213,3 +1213,107 @@ def test_doesnt_bump_tag_expiration_when_upstream_is_dead(
...
+        )
+        proxy_model = ProxyModel(
+            self.orgname,
+            self.upstream_repository,
+            self.user,
+        )
+        assert proxy_model._check_blob_upload_and_prune(repo_ref, input_manifest) is None
+
+    @patch("data.registry_model.registry_proxy_model.Proxy", MagicMock())
+    def test_auto_pruning_skipped_for_manifest_list(self, create_repo, initialized_db):
```
dmage, line 1251: I wouldn't recommend covering buggy behaviour with tests

data/registry_model/registry_proxy_model.py
```diff
@@ -131,6 +130,51 @@ def lookup_repository(
...
+        upstream_blob_size = upstream_manifest.layers_compressed_size
+        if upstream_blob_size > max_ns_quota:
+            raise QuotaExceededException
+
+        # check if upstream image fits within the current namespace size
+        if (curr_ns_size + upstream_blob_size) <= max_ns_quota:
+            return
+
+        # fetch tags with least lifetime_end_ms and mark them expired
+        tags = oci.tag.get_tag_with_least_lifetime_end_for_ns(repo_ref.namespace_name)
```
flavianmissi, line 168: You can save a db lookup if you give `repo_ref.id` to `get_tag_with_least_lifetime_end_for_ns` instead of `namespace_name`

data/registry_model/registry_proxy_model.py
```diff
@@ -131,6 +130,51 @@ def lookup_repository(
...
+        quotas = namespacequota.get_namespace_quota_list(repo_ref.namespace_name)
+        if quotas:
+            # currently only one quota per namespace is supported
+            max_ns_quota = quotas[0].limit_bytes
+        else:
+            logger.info("No quota configured")
+            return
+
+        # if given manifest_ref is a manifestList, we skip checking quota limit
+        if upstream_manifest.is_manifest_list:
```
flavianmissi, line 156: we probably want to run this check before doing anything in this function. the way it currently is now we're making two queries to the database only to return early when `upstream_manifest` is a manifest list

data/registry_model/registry_proxy_model.py
```diff
@@ -131,6 +130,51 @@ def lookup_repository(
...
+            raise QuotaExceededException
+
+        # check if upstream image fits within the current namespace size
+        if (curr_ns_size + upstream_blob_size) <= max_ns_quota:
+            return
+
+        # fetch tags with least lifetime_end_ms and mark them expired
+        tags = oci.tag.get_tag_with_least_lifetime_end_for_ns(repo_ref.namespace_name)
+        if tags is not None:
+            for tag in list(tags):
```
flavianmissi, line 170: when you cast `tags` into a `list` you're loading all objects into memory - is there a reason why we need to do that here? `get_tag_with_least_lifetime_end_for_ns` returns an iterator, which lazy loads each tag into memory as the code iterates through them.

data/registry_model/registry_proxy_model.py
```diff
@@ -446,6 +492,18 @@ def _create_manifest_with_temp_tag(
...
+        Creates a manifest in the given repository. Also creates placeholders for the
+        objects referenced by the manifest. For manifest lists, it creates
+        sub manifests entries attached to the manifest list along with a temporary tag.
+
+        Also checks whether the given manifest size is within the quota limit
+        of the namespace the repository is part of. If not, it prunes older tags.
+        Raises QuotaExceededException if there are not enough tags to prune.
+        """
+        # check if the given manifest size is within quota limit
+        self._check_blob_upload_and_prune(repository_ref, manifest)
```
flavianmissi, line 505: is there a way to communicate what your comment is saying by improving the function name?

data/registry_model/registry_proxy_model.py
flavianmissi: `repo_ref.id` and `namespace_name` actually refer to different tables/objects - thx @harishsurf for pointing that out.

data/registry_model/test/test_registry_proxy_model.py
```diff
@@ -1213,3 +1213,107 @@ def test_doesnt_bump_tag_expiration_when_upstream_is_dead(
...
+        input_manifest = parse_manifest_from_bytes(
+            Bytes.for_string_or_unicode(testdata.PYTHON_LATEST["manifest"]),
+            DOCKER_SCHEMA2_MANIFESTLIST_CONTENT_TYPE,
+        )
+        proxy_model = ProxyModel(
+            self.orgname,
+            self.upstream_repository,
+            self.user,
+        )
+        limit_bytes = 2048
```
flavianmissi, line 1262: let's set this to 0? this way we ensure that pulling a manifest list requires no space free in the namespace.

flavianmissi (commented):
LGTM after the last comments are resolved.

data/registry_model/registry_proxy_model.py
harishsurf: agreed! moved it to the beginning of the func

data/registry_model/registry_proxy_model.py
harishsurf: removed `list()` wrapping

data/registry_model/test/test_registry_proxy_model.py
harishsurf: removed setting quota size all together since the `upstream_manifest.is_manifest_list` moved to the very beginning of the func. I added an `assert` to verify ns_size doesn't change when a `manifest_list` is passed

data/registry_model/registry_proxy_model.py
flavianmissi: that's better!

data/registry_model/test/test_registry_proxy_model.py
flavianmissi: Good thinking!

HammerMeetNail (approved):

PR #1495, opened 2022-08-24
feature: Update nginx to support Patternfly (PROJQUAY-4124)

HammerMeetNail:
Note: There will be a follow up PR to clean up the nginx config related to setting cookies. Expect cookie values to change.  

* Removes legacy `Dockerfile.osbs`
* Updates `Dockerfile` to build new Quay UI based on Patternfly
* Adds new `make` commands for building static assets locally

**Testing Instructions**:
1. Remove all running instances of Quay
2. Run `make local-dev-up-static`
3. Wait for Quay to fully come online. This could be a few minutes after the container starts. 
4. Navigate to [http://localhost:8080/](http://localhost:8080/) to load the current UI
5. Navigate to [http://localhost:8080/react](http://localhost:8080/react) to load the new UI
6. Delete the cookie or navigate to [http://localhost:8080/angular](http://localhost:8080/angular) to revert to the old UI.


conf/nginx/server-base.conf.jnj
```diff
@@ -102,7 +137,7 @@ location /cnr {
     {% endif %}
 }
 
-location /api/ {
```
syed, line 105: accidental spacing added

conf/nginx/server-base.conf.jnj
```diff
@@ -18,7 +18,42 @@ proxy_redirect off;
...
+    }
+
+    root {{static_dir}}/patternfly/;
+    index index.html;
+
+    # Show new UI if patternfly cookie is set
+    if ($cookie_patternfly) {
+        # catch new static paths and direct to /index.html
+        # TODO: figure out how to match on $uri
+        rewrite ^(/organizations|/repositories|/signin) /index.html break;
```
syed, line 47: so what happens if I open a link in the new UI? let's say I click `<domain>/repositories/myrepo` in the new UI, will I always be redirected to `/` seems like bad UX

HammerMeetNail: Nah, you won't hit the backend. It's an internal route that should stay client side.

conf/nginx/server-base.conf.jnj
HammerMeetNail: fixed

syed (approved):
LGTM

PR #1496, opened 2022-08-26
mirror: Rollback failed tags (PROJQUAY-4322)

bcaton85:
Currently Quay creates tags for Docker V2 schema 1 manifests in manifest lists. This makes it appear a tag was mirrored successfully when it had actually failed. This change rolls back those failed tags when the sync fails.

### DIFF
--- a/workers/repomirrorworker/test/test_repomirrorworker.py
+++ b/workers/repomirrorworker/test/test_repomirrorworker.py
@@ -311,8 +311,8 @@ def test_successful_mirror_verbose_logs(run_skopeo_mock, initialized_db, app, mo
 @pytest.mark.parametrize(
     "rollback_enabled, expected_delete_calls, expected_retarget_tag_calls",
     [
-        (True, ["deleted", "updated", "created"], ["updated"]),
-        (False, ["deleted"], []),
+        (True, ["deleted", "zzerror", "updated", "created"], ["updated"]),
+        (False, ["deleted", "zzerror"], []),
     ],
 )
 @disable_existing_mirrors
@@ -421,6 +421,8 @@ def test_rollback(
                 _create_tag(repo, "updated")
             elif args[1] == "copy" and args[8].endswith(":created"):
                 _create_tag(repo, "created")
+            elif args[1] == "copy" and args[8].endswith(":zzerror"):
+                _create_tag(repo, "zzerror")
 
             return skopeo_call["results"]
         except Exception as e:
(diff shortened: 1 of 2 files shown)


data/model/oci/tag.py
```diff
@@ -156,9 +156,9 @@ def list_repository_tag_history(
     repository_id,
     page,
     page_size,
-    specific_tag_name=None,
     active_tags_only=False,
     since_time_ms=None,
+    tag_names=None,
 ):
     """
     Returns a tuple of the full set of tags found in the specified repository, including those that
```
kleesc, line 164: Update the docstring

data/model/oci/tag.py
```diff
@@ -184,8 +184,8 @@ def list_repository_tag_history(
         .offset(page_size * (page - 1))
     )
 
-    if specific_tag_name is not None:
-        query = query.where(Tag.name == specific_tag_name)
+    if tag_names is not None:
+        query = query.where(Tag.name << tag_names)
```
kleesc, line 188: Will this make use of the indices. We're changing the query from `... WHERE X = Y` to `... WHERE X in SOMELIST`.

kleesc: When this query is executed, is it of the same type as before? Can you add a small test for this.

bcaton85: Talked to @kwestpharedhat about the where query. There's a possibility of performance impact. For mirroring the `tag_names` list can be in the thousands so i'm rethinking adding this. We may want to go with the safer option of not modifying this query and instead filtering on the returned list.

data/model/oci/tag.py
bcaton85: No longer relevant

data/model/oci/tag.py
bcaton85: updated

workers/repomirrorworker/test/test_repomirrorworker.py
```diff
@@ -311,8 +311,8 @@ def skopeo_test(args, proxy):
 @pytest.mark.parametrize(
     "rollback_enabled, expected_delete_calls, expected_retarget_tag_calls",
     [
-        (True, ["deleted", "updated", "created"], ["updated"]),
-        (False, ["deleted"], []),
+        (True, ["deleted", "zzerror", "updated", "created"], ["updated"]),
+        (False, ["deleted","zzerror"], []),
```
kleesc, line 315: What's zzerror?

kleesc (approved):
LGTM otherwise

workers/repomirrorworker/test/test_repomirrorworker.py
bcaton85: Handles the case where the tag was still created for the errored image. In this case tag `zzerror` is a `skopeo copy` that fails.

kleesc (approved):

PR #1498, opened 2022-08-29
bug: Fix schema discovery on basic auth (PROJQUAY-4362)

ibazulic:
Fixes [PROJQUAY-4362](https://issues.redhat.com/browse/PROJQUAY-4362).


flavianmissi (approved):
Thanks for the fix @ibazulic!

PR #1500, opened 2022-08-29
task: remove obsolete logrotate.conf file (PROJQUAY-4364)

ibazulic:
Removes the obsolete `logrotate.conf` file which was used in Quay 2.9 and below when nginx logs were stored under `/var/log/nginx` inside the container instead of being redirected to `/dev/stdout`. This file is no longer needed.


kleesc (approved):
lgtm

PR #1507, opened 2022-08-30
healthcheck: Use db_kwargs in health check (PROJQUAY-4222)

jonathankingfc:
- The database health check is currently not using the db_kwargs and not using ssl settings
- This is causing health check failures for MySQL behind SSL


HammerMeetNail (approved):
LGTM

PR #1510, opened 2022-08-31
bug: Increase column size in logentry3 table (PROJQUAY-4305)

ibazulic:
We increase the size of `metadata_json` column in `logentry3` table from `TEXT` to `MEDIUMTEXT` on MySQL deployments to prevent the UI from erroring out when reading and decoding the audit log. This only touches MySQL, PostgreSQL is unaffected.


oauth/oidc.py
```diff
@@ -202,6 +202,7 @@ def exchange_code_for_login(self, app_config, http_client, code, redirect_suffix
         else:
             email_address = user_info.get("email") if user_info.get("email_verified") else None
 
+        logger.debug("Printing out decoded token: `%s`", user_info)
```
kleesc, line 205: What's this for?

ibazulic: Was needed for debugging a different issue. Guess I just forgot to remove it by mistake. Will remove it now.

kleesc (approved):

PR #1513, opened 2022-08-31
ui: Force /signin to use current UI (PROJQUAY-4345)

HammerMeetNail:
* Updates nginx to always route `/signin` to the current UI

This is the server side of the work. A second PR in `quay-ui` will be created to handle client side routing.

### DIFF
--- a/conf/nginx/server-base.conf.jnj
+++ b/conf/nginx/server-base.conf.jnj
@@ -18,6 +18,11 @@ proxy_redirect off;
 
 proxy_set_header Transfer-Encoding $http_transfer_encoding;
 
+# Temporarily force signin for old and new UI to route to web app
+location /signin {
+    proxy_pass   http://web_app_server/;
+}
+
 location /angular {
     # Expire cookie and switch to old UI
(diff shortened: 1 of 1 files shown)


conf/nginx/server-base.conf.jnj
```diff
@@ -18,6 +18,11 @@ proxy_redirect off;
 
 proxy_set_header Transfer-Encoding $http_transfer_encoding;
 
+# Temporarily force signin for old and new UI to route to web app
+location /signin {
+    proxy_pass   http://web_app_server/;
```
HammerMeetNail, line 23: Adding a trailing `/` is required at the moment, we have a redirect to `/signin` somewhere and it ultimately leads to a bad redirect where the port is dropped.

syed: does nginx treat `/signin` and `/signin/` the same?

HammerMeetNail: It does. Here's my local test. I add a custom header of `X-Signin` if it hits the `/signin` location. 

image
image

syed: nice looks good 👍

syed (approved):
LGTM

PR #1515, opened 2022-09-02
ui: Don't show UI toggle on signin page (PROJQUAY-4402)

HammerMeetNail:
* Hide toggle when on signin page

### DIFF
--- a/static/directives/header-bar.html
+++ b/static/directives/header-bar.html
@@ -44,7 +44,7 @@
 
         <!-- Normal -->
         <ul class="nav navbar-nav navbar-right hidden-xs" ng-switch on="user.anonymous">
-          <li class="new-ui-toggle" ng-if="Features.UI_V2">
+          <li class="new-ui-toggle" ng-if="Features.UI_V2 && !checkElementByID('signin-username')">
               <new-ui-toggle></new-ui-toggle>
           </li>
           <li>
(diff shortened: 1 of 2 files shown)


static/directives/header-bar.html
```diff
@@ -44,7 +44,7 @@
 
         <!-- Normal -->
         <ul class="nav navbar-nav navbar-right hidden-xs" ng-switch on="user.anonymous">
-          <li class="new-ui-toggle" ng-if="Features.UI_V2">
+          <li class="new-ui-toggle" ng-if="Features.UI_V2 && checkElementByID('signin-username') !== true">
```
syed, line 47: you don't need the `!== true` it's already a boolean just the return value will suffice

static/js/directives/ui/header-bar.js
```diff
@@ -145,6 +145,14 @@ angular.module('quay').directive('headerBar', function () {
           'namespace': namespace
         };
       };
+
+      $scope.checkElementByID = function(elementID) {
+        var results = false;
```
syed, line 150: `results` suggest a multiple value like an array. `result` or even `exists` should work

static/directives/header-bar.html
HammerMeetNail: Fixed

static/js/directives/ui/header-bar.js
HammerMeetNail: Fixed

syed (approved):
LGTM

PR #1516, opened 2022-09-02
api: feature to limit org creation to superusers (PROJQUAY-1245)

kleesc:


HammerMeetNail:
Maybe another way to ask, is there an api call that returns all orgs or repos that can be implemented in the new ui?

endpoints/v2/__init__.py
```diff
@@ -128,44 +128,79 @@ def callback(results, response):
...
-                    msg = (
-                        "This repository is for managing %s and not container images."
-                        % repository_ref.kind
-                    )
-                    raise Unsupported(detail=msg)
-
-                if repository_ref.is_public:
-                    if not features.ANONYMOUS_ACCESS:
-                        raise Unauthorized(repository=repository, scopes=scopes)
+    def _require_permission(allow_for_superuser=False, disallow_for_restricted_users=False):
```
Sunandadadi, line 131: NIT: Can we split this decorator? Feels little long.

endpoints/v2/__init__.py
```diff
@@ -128,44 +128,79 @@ def callback(results, response):
...
+                        repository_ref = registry_model.lookup_repository(namespace_name, repo_name)
+                        if repository_ref is None or not repository_ref.is_public:
+                            raise Unauthorized()
+
+                        if not features.ANONYMOUS_ACCESS:
+                            raise Unauthorized()
+
+                        return func(namespace_name, repo_name, *args, **kwargs)
+
+                    context = get_authenticated_context()
```
Sunandadadi, line 156: NIT: If we define context outside the `if` block, we don't have to define it again in the superuser if condition block. Also, if context is `None`, how do we handle it?

Sunandadadi (commented):
Minor comments, otherwise neatly done.

HammerMeetNail:
When logged in as an admin with `FEATURE_SUPERUSERS_FULL_ACCESS: True` set, I can read orgs/repos/tags for non-root users, but I can't push to a repo I am not a member of:

```
$ docker push localhost:8080/demo/alpine:private
The push refers to repository [localhost:8080/demo/alpine]
994393dc58e7: Layer already exists 
unauthorized: access to the requested resource is not authorized
```

kleesc:
@HammerMeetNail Giving global read access can be argued for auditing purposes, but is there ever a case where one would need to be able to _write_ to another namespace he doesn't own?

endpoints/v2/__init__.py
kleesc: I didn't putting it outside the if block in order to avoid making a database call for the user if the features were not enabled. If the context is None, then the regular anonymous flow should still apply.

HammerMeetNail:
@kleesc Only edge cases. Could see a scenario where a team loses access (maybe a team was dissolved or a team member with access left) to push something that needs to go out immediately and someone non-technical reaches out to their support team to handle the push on their behalf. There's definitely other ways to tackle the problem, like granting ownership to someone else to perform the action, but this could be a last resort option.

kleesc:
@HammerMeetNail Last resort in what case? In that scenario you would _need_ to take ownership of that namespace anyways.

HammerMeetNail:
It's not so much about need as it is about removing obstacles. There are other ways to achieve the same result, but more steps required. I think it's similar to taking ownership to do something. We could take ownership using existing mechanisms, but it's just an extra step for a user that should effectively be root and be able to do anything directly.

BillDett:
I think the expected behavior for a superuser is that they can write to any repository. True, you can achieve it by first taking Ownership of an Org, but a lack of 'push anywhere' seems odd for a super user. Like if you had to do a special chmod dance before letting root write to a user's home directory...would be weird (to me at least).

HammerMeetNail (approved):

PR #1520, opened 2022-09-07
ui: Invalidate browser cache on new ui toggle (PROJQUAY-4423)

HammerMeetNail:
Second part of https://github.com/quay/quay-ui/pull/82

* Prevents browser from caching nginx redirect to `/` by passing query args in replace request.


bcaton85:
What's the best way to test this? When switching the toggle from the old ui to the new ui with `make local-dev-up` i'm always redirected to a 404 page:
image

bcaton85:
Clearing the cookie brings me back to the old ui

HammerMeetNail:
@bcaton85 Now that the other quay-ui PR is merged, the best way to test is to do `make local-dev-up-static`. That should spin up both UIs on localhost:8080

bcaton85 (approved):
LGTM

PR #1521, opened 2022-09-07
UI: Reset switch if user doesn't navigate to new UI (PROJQUAY-4389)

Sunandadadi:


HammerMeetNail:
@Sunandadadi It's a small thing, but the toggle no longer changes when clicked. It'll show the modal and stay grey.

HammerMeetNail:
Ah didn't even notice that. All good.

PR #1525, opened 2022-09-09
net: Add feature flags for IPv6 (PROJQUAY-131)

HammerMeetNail:
* Provides a toggle for enabling IPv6, helps resolve issue with nginx failing to start in an IPv4 only environment

### DIFF
--- a/conf/init/nginx_conf_create.py
+++ b/conf/init/nginx_conf_create.py
@@ -75,6 +75,11 @@ def generate_nginx_config(config):
     ssl_protocols = config.get("SSL_PROTOCOLS", SSL_PROTOCOL_DEFAULTS)
     ssl_ciphers = config.get("SSL_CIPHERS", SSL_CIPHER_DEFAULTS)
 
+    # Enable IPv4 and/or IPv6. Valid values are IPv4, IPv6 or dual-stack.
+    ip_version = config.get("FEATURE_LISTEN_IP_VERSION", "IPv4")
+    use_ipv4 = True if ip_version.lower() != "ipv6" else False
+    use_ipv6 = True if ip_version.lower() in ["ipv6", "dual-stack"] else False
+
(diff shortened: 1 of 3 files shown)


conf/init/nginx_conf_create.py
```diff
@@ -74,6 +74,11 @@ def generate_nginx_config(config):
     enable_rate_limits = config.get("FEATURE_RATE_LIMITS", False)
     ssl_protocols = config.get("SSL_PROTOCOLS", SSL_PROTOCOL_DEFAULTS)
     ssl_ciphers = config.get("SSL_CIPHERS", SSL_CIPHER_DEFAULTS)
+    
+    # Enable IPv4 and/or IPv6. Valid values are IPv4, IPv6 or dual-stack. 
+    ip_version = config.get("FEATURE_LISTEN_IP_VERSION", "IPv4")
+    use_ipv4 = True if ip_version.lower() != "ipv6" else False
```
Sunandadadi, line 80: nit: Here, if `ip_version` is `test` or something random, it will default to use_ipv4. Should we make it a strict check - Something like how you check below - `ip_version.lower() in ["ipv4", "dual-stack"]`? Don't have a strong opinion on this, what you have works too.

HammerMeetNail: Yep, this is intentional. We need to default to something or nginx won't start. So if a user fat fingers the value or puts in something clearly wrong, we just default to starting Quay normally on IPv4 instead of forcing them to correct their mistake. This should catch things like `ip4`, `cheese`, `""`. The only way we disable IPv4 is if the user explicitly says they want only IPv6.

Sunandadadi: Ok, sounds good.

syed (approved):
👍  LGTM

PR #1530, opened 2022-09-15
Export Compliance: Fetching quay user data from federated login username (PROJQUAY-0000)

Sunandadadi:
The function `get_quay_user_from_federated_login_name` is used by the Self Service Tool to fetch quay.io users based on RH SSO username. RH SSO username is stored as a JSON blob with key `service_username`.


kleesc (approved):
LGTM

PR #1538, opened 2022-09-21
net: Update nginx to support all old UI endpoints (PROJQUAY-4420)

HammerMeetNail:
* Add remaining `url_prefix` from web.py


syed (approved):
LGTM

PR #1539, opened 2022-09-21
storage: return S3 url only for ip ranges in the same region (PROJQUAY-4498)

syed:
This optimization ensures that we return the direct S3 URL for CloudFront storage only for requests from the same region. This ensures we don't get charged for cross-region traffic to S3

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -22,6 +22,7 @@ from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import padding
 from prometheus_client import Counter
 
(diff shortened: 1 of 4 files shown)


storage/cloud.py
```diff
@@ -998,13 +999,15 @@ def __init__(
         cloudfront_privatekey_filename,
         storage_path,
         s3_bucket,
+        s3_region,
```
kleesc, line 1002: I think this will require explicitly specifying region in the config.yaml.

kleesc (approved):
LGTM

PR #1546, opened 2022-10-04
nginx: mark beginning of string in location match

kleesc:
To prevent nginx from preemtivly handle a route the application should.


ibazulic (approved):
LGTM!

PR #1547, opened 2022-10-04
secscan: Generate key to reduce vulnerabilities (PROJQUAY-4562)

crozzy:
Currently Quay is displaying the Clair response with no interpretation meaning when Clair reports on vulns per repo they appear to be duplicated (RHEL based images with multiple repos and packages existing in more than one). The correct way to fix this is via an interpretation layer in Quay, this change is a stop-gap to improve readability for the user.

### DIFF
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -445,12 +445,24 @@ def features_for(report):
     Quay Security scanner response.
     """
     features = []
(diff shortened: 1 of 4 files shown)


crozzy:
Before:
image


After:
image

crozzy:
It'd probably be good to review the key and if it's potentially over-eagar

kleesc (approved):
LGTM

PR #1548, opened 2022-10-05
Quay: Exporting image.oci package to Quay module (PROJQUAY-0000)

Sunandadadi:
This PR fixes -  `ModuleNotFoundError: No module named 'image.oci'` in the Quay module used in Service tool.


bcaton85 (approved):
LGTM

PR #1553, opened 2022-10-11
ui: Fix font size in superuser page (PROJQUAY-4407)

bcaton85:
Recent CSS changes caused the font size in the superuser panel to increase. The `active` and `inactive` CSS classes were added and overrode the existing `active` and `inactive` definitions.


Sunandadadi (approved):
lgtm

PR #1557, opened 2022-10-11
chore: Add server side assembly of chunked metadata for RADOSGW driver (PROJQUAY-4592)

ibazulic:
RadosGW did not support multipart copying from keys so we needed to do a local join and reupload of the whole blob. This creates issues for very big blobs where timeouts can occur. Since the issue was fixed in 2015. on the Rados side, we no longer need this part of legacy code.

See [here](https://github.com/ceph/ceph/pull/5139) for more information.


kleesc (approved):
LGTM

PR #1562, opened 2022-10-13
cleanup: Remove old validation code (PROJQUAY-4606)

jonathankingfc:
- The old validation code is no longer used and can be removed from Quay


flavianmissi (approved):
LGTM!

PR #1567, opened 2022-10-14
config: Add conftest mediatypes to default Quay configuration (PROJQUAY-4614)

Marcusk19:
Allow conftest policies to be stored on Quay


bcaton85 (approved):
LGTM

PR #1575, opened 2022-10-18
buildman: add SLO metric for time spent in queue for build jobs (PROJQUAY-4487)

Marcusk19:
This adds a metric that calculates the time a build job spends in the queue.

### DIFF
--- a/buildman/manager/ephemeral.py
+++ b/buildman/manager/ephemeral.py
@@ -59,6 +59,11 @@ build_ack_duration = Histogram(
     "seconds taken for the builder to acknowledge a queued build",
     labelnames=["executor"],
 )
+build_queue_duration = Histogram(
+    "quay_build_queue_duration_seconds",
+    "seconds taken for a build job to be scheduled from the queue",
+    labelnames=["schedule_success"],
+)
 build_duration = Histogram(
     "quay_build_duration_seconds",
     "seconds taken for a build's execution",
@@ -224,6 +229,7 @@ class EphemeralBuilderManager(BuildStateInterface):
         max_expiration = datetime.utcnow() + timedelta(seconds=self.machine_max_expiration)
         build_metadata["max_expiration"] = calendar.timegm(max_expiration.timetuple())
         build_metadata["last_heartbeat"] = None
+        build_metadata["created_at"] = time.time()
 
         job_key = self._job_key(build_id)
         try:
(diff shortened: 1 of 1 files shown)


buildman/manager/ephemeral.py
```diff
@@ -59,6 +59,9 @@
     "seconds taken for the builder to acknowledge a queued build",
     labelnames=["executor"],
 )
+build_queue_duration = Histogram(
+    "quay_build_queue_duration_seconds", "seconds build exists in the queue", labelnames=["build"]
```
kleesc, line 63: The build id is unique, so you're metric's cardinality will be equal to the number of build ids that exists. i.e You'll end up with too many time series. You should use the executor type instead.

kleesc (changes requested):

buildman/manager/ephemeral.py
kleesc: Also, you might wanna record **unsuccessful** job schedule requests instead. It's usually easier to take actions based on errors, vs successful requests. i.e We know something is wrong if we see an elevated rate of unsuccessful job schedules, vs just successful ones.

Marcusk19: Understood, thanks. I made the changes requested.

syed: @kleesc The idea here was to capture the avg time a build is in a "queued" state ie. time since someone submits a build till the time it lands on an executor. I agree with you that we should also capture the #of unsuccessful schedules as well

kleesc: That's fine. But the current change is only collecting metrics on unsuccessful schedules.
We should also remove the label associated with that metric. The executor is picked post scheduling, so it's not really applicable in this case. And again, using the build id will result in too many time series / cardinality.

Marcusk19: @kleesc I edited the code so now it should be tracking both successful and unsuccessful schedules, also got rid of the labelname.

kleesc: @Marcusk19 Can we use the result of the schedule call as a label (`success` or `failure`). As it is now, I don't think we differentiate between successful and unsuccessful schedules on any data point.

Marcusk19: @kleesc Just pushed my changes, does using a `schedule_success` label with a value of either `True` or `False` work in this case?

kleesc (approved):
LGTM

PR #1578, opened 2022-10-19
scripts: push to ecr backup when building quay.io (PROJQUAY-3273)

Marcusk19:
Change order to push to ecr first when deploying quay.io - fixes the problem of needing a newer build when deploying quay.io but cannot get it because quay.io itself is down.


syed (approved):
LGTM

PR #1580, opened 2022-10-20
Update auth flow to add "access_token" response field (PROJQUAY-4431)

harishsurf:
Azure registry uses OAuth 2.0 for registry authentication which will also accept token under the name `access_token`. The change adds this field

### DIFF
--- a/proxy/__init__.py
+++ b/proxy/__init__.py
@@ -197,6 +197,10 @@ class Proxy:
 
         resp_json = resp.json()
         token = resp_json.get("token")
+        if token is None:
+            # For OAuth2.0 compatability, "access_token" can also used and is equivalent to "token" "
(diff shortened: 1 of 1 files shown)


ibazulic (commented):
Maybe put this in a `try` block? Error out when it's not `token` or `access_token`. And when it errors out, add some debugging logs so we see what went wrong. It was difficult to figure out what the problem was since no debug information was printed.

flavianmissi (approved):

flavianmissi:
wouldn't hurt to add some logs when no token is found, as per Ivan's suggestion. You can do it in a follow-up PR if that feels easier.
to raise an exception we need to make sure it's properly handled in the api layer so that users don't end up with a generic 500 - probably better done as a follow-up to this PR.

PR #1585, opened 2022-10-24
Util: Adding git pre-commit checks (PROJQUAY-0000)

Sunandadadi:


syed (approved):
LGTM

HammerMeetNail:
@Sunandadadi Can you add a `make` target that installs the dependency and then does the pre-commit install?

Sunandadadi:
> @Sunandadadi Can you add a `make` target that installs the dependency and then does the pre-commit install?

Can I add that in another PR? I want to clean up existing pre-commit checks ([black](https://github.com/quay/quay/blob/master/Makefile#L71)).

PR #1587, opened 2022-10-25
Util: Clean up and adding make target to install pre-commit hooks(PROJQUAY-000)

Sunandadadi:


HammerMeetNail:
@Sunandadadi Will this also install `black` in the new hook?

Sunandadadi:
> @Sunandadadi Will this also install `black` in the new hook?

Yes, it will install as mentioned [here](https://github.com/quay/quay/blob/master/.pre-commit-config.yaml#L12) in the pre-commit config file.

HammerMeetNail (approved):
LGTM

PR #1596, opened 2022-10-27
config: Remove whitespace from config (PROJQUAY-4666)

bcaton85:
White space in the mediatype config is preventing conftest policies containing data from being pushed.


Marcusk19 (approved):
lgtm

PR #1598, opened 2022-10-28
Quay: Cloudfront multi domain (PROJQUAY-4506)

kwestpharedhat:


syed:
I don't see any schema for `cloudfront_distribution_org_overrides` is it possible to add it as a comment or even better create a typed class or named tuple?

storage/cloud.py
```diff
@@ -1032,6 +1034,15 @@ def get_direct_download_url(
                 path, request_ip, expires_in, requires_cors, head, **kwargs
             )
 
+        if kwargs:
+            namespace = kwargs.get("namespace")
+            if namespace is not None:
+                for key in self.cloudfront_distribution_org_overrides:
```
syed, line 1040: if this is a `dict` then you don't need to loop over the keys, you can directly index the name like `cloudfront_distribution_org_overrides[namespace]`

storage/test/test_cloudfront.py
```diff
@@ -73,6 +73,7 @@ def test_direct_download(
         engine = CloudFrontedS3Storage(
             context,
             "cloudfrontdomain",
+            {},
```
syed, line 76: is it possible to add a test which tests this config? i.e pass in a `namespace` to the `get_direct_downaload_url` function and expect a certain response

kwestpharedhat:
I certainly could add schema etc for the org overrides, however it seemed that the config-tool contained all of that and provided checks, I didnt see any other schema definition at least.

storage/cloud.py
kwestpharedhat: I will rework this.

storage/cloud.py
```diff
@@ -1032,6 +1034,14 @@ def get_direct_download_url(
                 path, request_ip, expires_in, requires_cors, head, **kwargs
             )
 
+        if kwargs:
+            namespace = kwargs.get("namespace")
+            if namespace is not None:
```
syed, line 1039: this check can be removed, the next check also makes sure that if `namespace` is `None` it doesn't go in the if

storage/cloud.py
```diff
@@ -1032,6 +1034,14 @@ def get_direct_download_url(
...
+                if namespace in self.cloudfront_distribution_org_overrides:
+                    self.cloudfront_distribution_domain = (
```
syed, line 1041: you are updating a class variable here. What happens if there are other requests at the same time which may depend on this variable and it may wrongly be set. There's only one object of `CloudFrontedS3Storage` so all requests will use that same variable. A better approach here would be to use a local variable for the domain

storage/cloud.py
```diff
@@ -1008,6 +1010,7 @@ def __init__(
     def get_direct_download_url(
         self, path, request_ip=None, expires_in=60, requires_cors=False, head=False, **kwargs
     ):
+        domain = self.cloudfront_distribution_domain
```
syed, line 1013: minor nit. Can you move this line to 1038 that way all the `domain` logic is in one place

storage/test/test_cloudfront.py
```diff
@@ -98,6 +99,44 @@ def test_direct_download(
             # get back an S3 URL.
             assert "s3.amazonaws.com" in engine.get_direct_download_url(_TEST_PATH, "1.2.3.4")
 
+        engine = CloudFrontedS3Storage(
+            context,
+            "failure",
```
syed, line 104: instead of calling this `failure` can you call this `defaultdomain` and the other one `overridedomain`

syed (approved):
LGTM

PR #1535, opened 2022-09-19
arch: add ppc64le support to quay (PROJQUAY-4595)

snehakpersistent:
Add ppc64le support to quay image

PR #1602, opened 2022-10-30
arch: Map aarch64 to arm64 in ARCH variable (PROJQUAY-4679)

jnohlgard:
Building the image on Fedora arm64 fails without this change.

`uname -m` prints `aarch64` on my machine. I have never seen `arm_64` (with an underscore) anywhere before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,7 +115,7 @@ ENV OS=linux
 ARG PUSHGATEWAY_VERSION=1.0.0
 RUN set -ex\
 	; ARCH=$(uname -m) ; echo $ARCH \
-	; if [ "$ARCH" == "x86_64" ] ; then ARCH="amd64" ; elif [ "$ARCH" == "arm_64" ] ; then ARCH="arm64" ; fi \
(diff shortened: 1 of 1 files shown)


jnohlgard:
For future reference: The issue was introduced in #1535

HammerMeetNail:
@jnohlgard Can you confirm how you're building?

HammerMeetNail:
We might want this check to be for both `aarch64` and `arm64`. Seeing this from my M1 Mac:

```
$ uname -m
arm64
```

Thoughts?

jnohlgard:
> We might want this check to be for both `aarch64` and `arm64`. Seeing this from my M1 Mac:
> 
> ```
> $ uname -m
> arm64
> ```
> 
> Thoughts?

The arch variable is used to construct a download url and the file is named -arm64 for arm64, so there is no translation needed on Darwin.

jnohlgard:
> @jnohlgard Can you confirm how you're building?

I don't have the log here with me, but I think I just did `podman build -t quay .`

HammerMeetNail:
Gotcha, I think this looks good. Let's merge.

HammerMeetNail (approved):
LGTM

PR #1603, opened 2022-10-31
proxy: Modify registry auth url for proxy cache validation (PROJQUAY-4585)

harishsurf:
When authenticating against a registry, if the `www-authenticate` header doesn't specify a `service` parameter, it should be excluded from the registry auth url.

Eg:
```
$  curl -I -L https://registry.ci.openshift.org/v2/
HTTP/2 401
content-type: application/json; charset=utf-8
docker-distribution-api-version: registry/2.0
www-authenticate: Bearer realm="https://registry.ci.openshift.org/openshift/token"
```

the `www-authenticate` header doesn't specify a `service` parameter. Prior to this change, the malformed url request failed with a `401`
```
$ curl https://registry.ci.openshift.org/openshift/token\?service\=None\&scope\=repository:something-totally-fake:pull
{"details":"repository name \"something-totally-fake\" invalid: it must be of the format \u003cproject\u003e/\u003cname\u003e"} 
```

### DIFF
--- a/endpoints/api/organization.py
+++ b/endpoints/api/organization.py
@@ -992,7 +992,7 @@ class ProxyCacheConfigValidation(ApiResource):
             existing = model.organization.get_organization(orgname)
             config.organization = existing
 
-            proxy = Proxy(config, "something-totally-fake", True)
+            proxy = Proxy(config, validation=True)
             response = proxy.get(f"{proxy.base_url}/v2/")
             if response.status_code == 200:
                 return "Valid", 202
--- a/proxy/__init__.py
+++ b/proxy/__init__.py
@@ -4,6 +4,7 @@ Registries following the distribution spec are supported.
 """
 from __future__ import annotations
 import re
+from urllib.parse import urlencode
 
 import requests
 from requests.exceptions import RequestException
@@ -11,7 +12,6 @@ from requests.exceptions import RequestException
 from app import model_cache
 from data.cache import cache_key
 from data.database import ProxyCacheConfig
-
 
 WWW_AUTHENTICATE_REGEX = re.compile(r'(\w+)[=] ?"?([^",]+)"?')
 TOKEN_VALIDITY_LIFETIME_S = 60 * 60  # 1 hour, in seconds - Quay's default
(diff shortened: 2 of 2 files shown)


ibazulic:
I am not sure we should allow this. The [authentication specification](https://docs.docker.com/registry/spec/auth/token/) that Docker and all registries indicates that the field is mandatory.

According to the [RFC 6750](https://www.rfc-editor.org/rfc/rfc6750#page-7), which defines the Oauth protocol, the parameter `realm` is apparently optional, but the [RFC-2617](https://www.rfc-editor.org/rfc/rfc2617) defines it as mandatory. Specifically:

> The authentication parameter realm is defined for all authentication
>    schemes:
> 
>       realm       = "realm" "=" realm-value
>       realm-value = quoted-string
> 
>    The realm directive (case-insensitive) is required for all
>    authentication schemes that issue a challenge. The realm value
>    (case-sensitive), in combination with the canonical root URL (the
>    absoluteURI for the server whose abs_path is empty; see section 5.1.[2](https://www.rfc-editor.org/rfc/rfc2617#ref-2)
>    of [2]) of the server being accessed, defines the protection space.
>    These realms allow the protected resources on a server to be
>    partitioned into a set of protection spaces, each with its own
>    authentication scheme and/or authorization database. The realm value
>    is a string, generally assigned by the origin server, which may have
>    additional semantics specific to the authentication scheme. Note that
>    there may be multiple challenges with the same auth-scheme but
>    different realms.

In other words, the `realm` or `service` **must be** returned by the authentication endpoint, after which we **may** include it in our own challenge.

harishsurf:
@ibazulic , are you referring to the `realm` parameter? This change doesn't remove this parameter. It excludes `service` parameter from the url if the `www-authenticate` header doesn't specify it. 
I couldn't find any related material for `service` usage in the RFC doc. I did check with @dmage whether this is a bug on the openshift token endpoint, may be he can say more..

ibazulic:
Yeah, but in this case I think that the `service` parameter plays the role of the `realm`. If my logic is wrong here, then this discussion doesn't matter, of course :-)

flavianmissi (dismissed):
@ibazulic the realm is sent regardless of the service being present or not, the condition is whether to use the service (+scope) along with the realm, or to just use the realm when no service is present.

@harishsurf I see that the scope is also being removed when no service is sent - have you tested this on a few registries? An easy way to test it is to configure a proxy org. Before creating a proxy cache config the backend will try to get an auth token from the registry and return an error if it fails.
I think we should also notify QE to do a full regression test after this one goes in, just to be on the safe side.

dmage (changes requested):
Lack of service doesn't mean that we need to reject scope. Let's not mix them.

flavianmissi (approved):
LGTM

PR #1605, opened 2022-11-02
buildman: allow fallback on non-exception build failures (PROJQUAY-4609)

kleesc:
i.e Failed api calls caused by the external builder servic (expired certificate, network issues, ...)

### DIFF
--- a/buildman/manager/ephemeral.py
+++ b/buildman/manager/ephemeral.py
@@ -922,11 +922,11 @@ class EphemeralBuilderManager(BuildStateInterface):
                 logger.debug("Build job %s scheduled.", job_id)
             else:
                 logger.warning(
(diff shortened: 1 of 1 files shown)


buildman/manager/ephemeral.py
```diff
@@ -922,11 +922,11 @@ def _work_checker(self):
                 logger.debug("Build job %s scheduled.", job_id)
             else:
                 logger.warning(
-                    "Unsuccessful schedule. Build ID: %s. Retry restored.",
+                    "Unsuccessful schedule. Build ID: %s. Check build executor service.",
```
syed, line 925: @kleesc  looks good. If we get a lot of warnings because of this we might have to think about adding a retry counter here

PR #1609, opened 2022-11-03
buildman: add SLO metric that calculates build success rate (PROJQUAY-4486)

Marcusk19:
Added a counter metrics to track total number of build jobs and total number of failed builds. These can be used to calculate the build success % in Grafana.


kleesc (approved):
lgtm

PR #1621, opened 2022-11-08
users: prevent CREATE_NAMESPACE_ON_PUSH is restricted (PROJQUAY-4702)

kleesc:
Prevent creating namespaces/orgs on pushes (CREATE_NAMESPACE_ON_PUSH) if user is restricted.


HammerMeetNail (approved):
LGTM

PR #1623, opened 2022-11-09
UI: Fix redirection to user/org page (PROJQUAY-4667)

Sunandadadi:
The previous implementation of org check uses `UserService.getOrganization` which checks for logged in users orgs and returns true only if org matches to users orgs as [here](https://github.com/quay/quay/blob/master/static/js/services/user-service.js#L99-L113)
This PR fixes to use the `is_organization` flag returned by the repository response which is irrespective of user belonging to the org. 

Before:
Screen Shot 2022-11-09 at 10 36 37 AM
After:
Screen Shot 2022-11-09 at 10 36 12 AM

Here `Fedora123` org is created by a different user but admin has superuser perms and can view it.

### DIFF
--- a/static/js/pages/repo-view.js
+++ b/static/js/pages/repo-view.js
@@ -194,9 +194,5 @@
     $scope.getImages = function(callback) {
       loadImages(callback);
     };
-
-    $scope.isOrganization = function(namespace) {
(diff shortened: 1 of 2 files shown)


HammerMeetNail (approved):
LGTM

PR #1629, opened 2022-11-21
WIP: Adding OCI Artifact type and referrers API

bcaton85:
NOTICE: Work in progress. Done for shift week so some implementations may be hacky.

What this does:
- Adds support for the OCI Artifact manifest spec
- Adds the `/referrers` API
- Adds garbage collection where:
    - If any manifests point to the manifest currently being GC'd, attempt to GC those too
    - If the manifest currently points to an existing manifest, do not GC that manifest
- Adds the `refers` field to the previous manifests

Concerns to address:
- Need to conclude that we're GC'ing the way we want it too
- Currently adds a column to the manifest table that contains the digest of the manifest it points to. This was done instead of ID's since the manifest being pointed too may not exist. Is this how we want to relate manifests?
- Need to ensure the current manifest schema implementations are complete
- Do we permit cycles to appear between manifests? How will GC handle it?
- Implementing tests
- Implementing types

### DIFF
--- a/deploy/openshift/quay-py3-app.yaml
+++ b/deploy/openshift/quay-py3-app.yaml
@@ -154,6 +154,17 @@ objects:
           secret:
             secretName: ${{QUAY_APP_CONFIG_SECRET}}
         serviceAccountName: ${{NAME}}
+        affinity:
+          podAntiAffinity:
+            preferredDuringSchedulingIgnoredDuringExecution:
+              - podAffinityTerm:
+                  labelSelector:
+                    matchExpressions:
+                    - key: ${{QUAY_APP_COMPONENT_LABEL_KEY}}
+                      operator: In
+                      values:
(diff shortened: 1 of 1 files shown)


flavianmissi:
> Currently adds a column to the manifest table that contains the digest of the manifest it points to. This was done instead of ID's since the manifest being pointed too may not exist. Is this how we want to relate manifests?

I don't think that adding a column to the manifest table is something we want to be doing on quay.io. If we decide we will do this we probably want to try this out on a copy of production first to see what kind of migration time we're looking at.

bcaton85:
Yeah makes sense. I suppose the better approach is to create another table similar to manifest child? I remember we talked about reusing the manifestChild table but it's used often in the context of manifests lists which might cause some issues. @syed Do you have any thoughts on this as well?

PR #1630, opened 2022-11-21
storage: Add S3 Region to CloudFlare URL (PROJQUAY-3512)

syed:
This will be used by the CloudFlare worker to redirect to the correct S3 origin bucket


kleesc (approved):
lgtm

PR #1637, opened 2022-11-23
storage: Add storage driver for VeritasObjectAccess Server(PROJQUAY-4754)

DilipRenkila:
This adds veritas storage driver for quay. Veritas Object Access provides an implementation of the simple object access service. The supported APIs are compatible with Amazon S3 with only mentioned parameters. More documentation On veritas can be found at https://www.veritas.com/content/support/en_US/doc/object_access_api_80

**Background**
by using rados driver to Veritas S3, we can push images but not pull because Veritas S3 doesn't support Signature S2 version .

### DIFF
--- a/storage/__init__.py
+++ b/storage/__init__.py
@@ -26,6 +26,7 @@ STORAGE_DRIVER_CLASSES = {
     "AzureStorage": AzureStorage,
     "RHOCSStorage": RHOCSStorage,
     "CloudFlareStorage": CloudFlareS3Storage,
+    "VeritasObjectAccessStorage": VeritasObjectAccessStorage,
 }
 
 


flavianmissi:
Hi @DilipRenkila, thanks for you contribution!
Before diving into code reviewing, could you please make sure you're using 4 spaces per indentation level (instead of 8) as per [pep8](https://peps.python.org/pep-0008/#indentation)?

dmesser:
Should we maybe just add SigV4 support to the Ceph/RADOS driver and ensure there is a graceful fallback to Sigv2 when the storage does not support it? Ceph is definitely easier to test as part of CI.

DilipRenkila:
it makes sense to add sigv4 support to rados driver.

DilipRenkila:
can we just add a Signature version parameter in RADOSGW  driver ?

flavianmissi:
> can we just add a Signature version parameter in RADOSGW driver ?

Let's go with that @DilipRenkila

bcaton85:
> can we just add a Signature version parameter in RADOSGW driver ?
@DilipRenkila This PR has been open for a long time. Do you have plans to make these changes? Are they still required?

PR #1644, opened 2022-11-25
build(deps): reduce CVEs in dependencies and runtime environment (PROJQUAY-4777)

dmesser:
- switch to ubi-minimal to get rid of subscription-manager and associated old urllib3 versions
- remove older pip version part of python39 appstream
- bump python dependencies

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi:latest AS base
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS base
 # Only set variables or install packages that need to end up in the
 # final container here.
 ENV PATH=/app/bin/:$PATH \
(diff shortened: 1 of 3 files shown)


flavianmissi (approved):
LGTM.

we need a plan to propagate these changes downstream - if I remember correcty that doesn't happen automatically. cc @HammerMeetNail

HammerMeetNail:
Yeah, the `requirements.txt` might be good for downstream, but the `Dockerfile` changes need to be pulled in. We can take a look at this Monday.

dmage:
Why can't we use pip from python39 appstream?

dmesser:
Causes false positives with Clair scanning and is generally older than upstream.

flavianmissi:
@dmesser looks like this needs a rebase

Dockerfile
```diff
@@ -45,21 +48,22 @@ RUN set -ex\
...
         file \
         make \
         libjpeg-turbo \
         libjpeg-turbo-devel \
-		wget\
-	; dnf -y -q clean all
+		wget \
+	; microdnf -y clean all
 WORKDIR /build
+RUN python3 -m ensurepip --upgrade
```
dmage, line 66: Downstream build system doesn't allow us to download unpinned versions. Most likely it'll be removed when we reconcile upstream and downstream Dockerfile (ideally we should be able to generate downstream Dockerfile with `sed`)

dmesser: Isn't ensurepip part of python and hence the pip version is aligned to the python version?

dmage: TIL (from the front documentation page):

> This module (ensurepip) does not access the internet. All of the components needed to bootstrap pip are included as internal parts of the package.

dmage (approved):

PR #1645, opened 2022-11-25
users: default to true if LDAP_RESTRICTED_USER_FILTER is not set (PROJQUAY-4776)

kleesc:
When LDAP is used and FEATURE_RESTRICTED_USERS is set, if LDAP_RESTRICTED_USER_FILTER is set, then, by default, all ldap users are restricted.


syed (approved):
LGTM

PR #1649, opened 2022-12-01
oci/index.py: support docker schema2 sub-manifest types

flavianmissi:


dmesser (approved):
/lgtm

syed:
We should eventually allow all OCI media types

PR #1651, opened 2022-12-01
cors: Adding missing method type (PROJQUAY-4800)

bcaton85:
The PUT method has not been added to the list of cors methods, causing PUT requests to fail.


kleesc (approved):
LGTM

PR #1653, opened 2022-12-02
ui: Show UI toggle on quay.io only to redhat users (PROJQUAY-4582)

syed:
New UI preview on quay.io is being released as a preview only to RedHat users before being GA


HammerMeetNail (approved):
LGTM

PR #1654, opened 2022-12-05
ui: Add script for adobe analytics for quay.io in angular UI (PROJQUAY-4582)

syed:
This will be used for usage analytics on quay.io


HammerMeetNail (approved):
LGTM

PR #1658, opened 2022-12-07
ui: Remove FEATURE_UI_V2 from analytics scripts (PROJQUAY-4582)

syed:
the variable `feature_set` has features with the `FEATURE_` prefix removed. Fix the jinja check for this


HammerMeetNail (approved):
LGTM

PR #1663, opened 2022-12-07
secscan: garbage collect manifests (PROJQUAY-4126)

kleesc:
Garbage collect manifests no longer referenced in Quay from the
security scanner service.

Also moved quota related code from data/registry_model/ to data/model/
to avoid circular dependencies.


BillDett:
Does this address https://issues.redhat.com/browse/PROJQUAY-4126?

syed (approved):

PR #1668, opened 2022-12-09
deploy: fix api panels in dashboard (PROJQUAY-4485)

Marcusk19:
Api success rate was previously calculated by only looking at `2XX`s as successful and everything else as an error. This adjusts the query so only `5XX`s are considered errors and `4XX`s are not being incorrectly grouped as issues with the service.


syed (approved):
LGTM

PR #1670, opened 2022-12-12
Nginx: Minor update to fix toggling issue on Safari (PROJQUAY-4527)

Sunandadadi:


HammerMeetNail (approved):
Confirmed this fixes things locally, LGTM

PR #1674, opened 2022-12-13
secscan: add metric for scan results (PROJQUAY-4488)

Marcusk19:
Add histogram metric that calculates time taken to get security scan results after pushing image to quay.

### DIFF
--- a/util/metrics/prometheus.py
+++ b/util/metrics/prometheus.py
@@ -65,6 +65,10 @@ secscan_index_request_duration = Histogram(
     "seconds taken to make an index request to the secscan service",
 )
 
+secscan_result_duration = Histogram(
+    "quay_secscan_result_duration",
+    "how long it takes to receive scan results after pushing an image",
+)
 
 PROMETHEUS_PUSH_INTERVAL_SECONDS = 30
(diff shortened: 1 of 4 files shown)


util/metrics/prometheus.py
```diff
@@ -65,6 +65,10 @@
     "seconds taken to make an index request to the secscan service",
 )
 
+# secscan latency
+secscan_result_duration = Histogram(
+    "quay_secscan_result_duration", "seconds taken to retrieve secscan results from service"
```
kleesc, line 70: You're currently measuring the time between a manifest gets created and the time the security worker sends it for indexing.  So you're not really getting the time the service takes to actually create the report (which I think we can get from the current Clair metrics available). Wondering if having the number of outstanding manifests to be indexed by the securityworker would be a better measure?

kleesc: Or even the throughput of the securityworker?

Marcusk19: Ah ok, @syed thoughts on this? I think the original goal was to have a way to track how long it takes to see vulnerabilities reported after pushing an image. Maybe that should be time from manifest to security worker + time to create a report (from Clair metrics)?

kleesc (approved):

PR #1675, opened 2022-12-13
chore: Update Dockerfile to reduce size of image (PROJQUAY-4837)

dmage:
`RUN rm -rm` doesn't delete data from previous layers, but it creates a new layer that marks which files should be deleted. So it increases the size of the image.

`RUN chmod` creates a new layer with a copy of the files with new permissions.

To reduce the size of the final image, these commands are moved into intermediate images.


harishsurf (approved):
nice, the image size is down to 909mb from 1.58gb :rocket: Verified that the config tool ui works fine. LGTM!

PR #1676, opened 2022-12-15
chore: remove jsonpickle from dependencies (PROJQUAY-4865)

dmage:
jsonpickle was required by aws-xray-sdk that we don't use anymore.


dmage:
Replaced by #1644

PR #1693, opened 2023-01-04
chore: remove yapf (PROJQUAY-4865)

dmage:
Yapf is a formatter for Python files that is no longer used by Quay. It was replaced by black.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,18 +151,6 @@ clean:
 	rm -rf screenshots
 
 
-yapf-all:
-	yapf -r . -p -i
-
-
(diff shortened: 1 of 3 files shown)


bcaton85 (dismissed):
LGTM - do we also want to remove this from the [bill-of-materials.json](https://github.com/quay/quay/blob/master/bill-of-materials.json#L642-L646) too?

bcaton85:
I don't believe so. @HammerMeetNail Can you confirm whether the bill-of-materials.json file is autogenerated or not?

PR #1696, opened 2023-01-06
login: Use the correct username for export compliance (PROJQUAY-4844)

syed:
SSO username can be different from quay username. Use the subject returned from the token and parse the SSO username from the subject


bcaton85 (approved):
LGTM

PR #1697, opened 2023-01-08
database: add creation date for robot account token (PROJQUAY-4863)

dmitrii-stolbikov:
it will allow users to rotate robot account tokens more easily

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -766,6 +766,7 @@ class RobotAccountMetadata(BaseModel):
 class RobotAccountToken(BaseModel):
     robot_account = QuayUserField(index=True, allows_robots=True, unique=True)
     token = EncryptedCharField(default_token_length=64)
+    creation_date = DateTimeField(default=datetime.utcnow)
     fully_migrated = BooleanField(default=False)
 
 


dmitrii-stolbikov:
I am an admin but not a Red Hat Quay instance but our own instance inside of our company. So, I do not think that RH Quay admins can verify this patch. But we need this patch due to security reasons.

kleesc:
There's already a creation date in the user table. And this should already being returned by the robot api endpoints: `/api/v1/user/robots/...`

dmitrii-stolbikov:
It is a creation date of a robot, but the PR is about the creation date of a robot account token.

kleesc:
It's the datetime at which a robot was created, so it shouldn't be changing on updates.

dmitrii-stolbikov:
Yes, I know, but you are talking about another creation date. I mentioned the creation date of a robot token, which must be rotated repeatedly, so we really need it. Could you please reopen the PR?

jankazim:
> There's already a creation date in the user table. And this should already be returned by the robot api endpoints: `/api/v1/user/robots/...`. There should be a `created` field as part of the response.

Actually it is not the same. Creation of Robot and creation of token is not the same thing.
When the robot is created, then also the token for him, but when the token is regenerated, the date for Robot creation is not updated.
So there is no timestamp when the token was generated when using this:
image

So the request is still valid.

kleesc:
My bad, I didn't catch you were referring to token generation. We _could_ add it if really necessary, but like @BillDett said, this requires a migration + a backfill of existing robots.

Note that any token regeneration are logged as part of Quay's action logs. So you could audit user/org for robot regenerations through the logs api: `/api/v1/user/logs` or `/api/v1/<org>/logs`, and filter the results for `regenerate_robot_token` specifically.

PR #1700, opened 2023-01-09
deploy: add podAntiAffinity to deployment (PROJQUAY-3684)

Marcusk19:
Adds pod anti-affinity in `quay-py3-app.yaml` according to https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

Trying to address alert received for quay.io where pod anti-affinity is not set for deployment.


syed (approved):
LGTM

PR #1707, opened 2023-01-16
logs: validate date range is within configuration (PROJQUAY-4959)

kleesc:


endpoints/api/logs.py
```diff
@@ -103,14 +104,17 @@ def _get_aggregate_logs(
     (start_time, end_time) = _validate_logs_arguments(start_time, end_time)
```
syed, line 104: we could fail early by adding validation logic in this function instead of down the code

syed (approved):
LGTM

PR #1710, opened 2023-01-17
chore: Use GitHub Actions cache for Docker Build jobs (PROJQUAY-4970)

dmage:
It takes >2 hours to build a Quay image for linux/arm64 by GitHub Actions. Caching should help us speed things up a bit.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -154,7 +154,7 @@ jobs:
 
   docker-amd64:
     name: Docker Build amd64
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
 
     - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


dmage:
It also should help us with `npm install` failures on arm64 as its results often can be taken from cache.

syed:
@dmage there's also `Dockerfile.static` does that need to be updated as well?

dmage:
@syed I cannot find Dockerfile.static, perhaps it was deleted?

syed:
sorry the other file was `Dockerfile.dev` not sure if we need that anymore. If not, we should remove it, keep only one Dockerfile

dmage:
I see only one Dockerfile in the repo.

syed (approved):
LGTM

PR #1711, opened 2023-01-17
deploy: add weight to anti-affinity (PROJQUAY-3684)

Marcusk19:
fixing error on deploy:
`spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].weight: Invalid value: 0: must be in the range 1-100`


syed:
@Marcusk19  can you also add this `podAntiAffinity` to the file `quay-py3-cloudflare-app.yaml` as well

Marcusk19:
@syed should be in `quay-py3-cloudflare-app.yaml` now too

syed (approved):

PR #1713, opened 2023-01-18
deploy: decrease query max points (PROJQUAY-4974)

Marcusk19:
limits the number of data points returned by secscan_result_duration query to 15 this is to fix error in grafana where secscan latency shows a 502 or "query processing would load too many samples into memory in query execution"


Marcusk19:
screenshot of the panel:
Screenshot 2023-01-18 at 10 45 51 AM

I think the buckets also need to be changed, the default is for api calls that are in the ms which is why all the values in the heatmap are at the top.

PR #1728, opened 2023-01-27
chore: Remove appr dependencies (PROJQUAY-4992)

dmage:
A follow-up of #1718. We can remove dependencies that are not needed anymore.


bcaton85 (approved):
LGTM

PR #1731, opened 2023-01-30
config: clean upload folder by default (PROJQUAY-4395)

dmesser:
If quay doesn't clean up temporary files under `uploads` folder by default, that means it is very possible for customer to pay more money for storage usage.  A lot of  important customers are money-sensitive,  so it doesn't make sense to set CLEAN_BLOB_UPLOAD_FOLDER as false by default.


bcaton85 (approved):
LGTM

PR #1733, opened 2023-01-30
secscan: add option to limit manifests with layer size too large to index (PROJQUAY-4957)

kleesc:


data/secscan_model/secscan_v4_model.py
```diff
@@ -307,6 +314,21 @@ def mark_manifest_unsupported(manifest):
                     metadata_json={},
                 )
 
+        def mark_manifest_layer_too_large(manifest):
+            with db_transaction():
+                ManifestSecurityStatus.delete().where(
+                    ManifestSecurityStatus.manifest == manifest._db_id,
```
syed, line 320: isin't `manifest_id` unique across? Do we need the extra `AND` with the `repository_id`?

util/secscan/v4/api.py
```diff
@@ -198,7 +224,16 @@ def _join(first, second):
 
         body = {
             "hash": manifest.digest,
-            "layers": [
+            "layers": [],
+        }
+
+        for l in layers:
+            layer_compressed_size = l.layer_info.compressed_size
+            secscan_index_layer_size.observe(layer_compressed_size)
```
syed, line 232: Do we want to limit on each layer size or the whole manifest size? (sum of all compressed layer sizes)

kleesc: I put a limit on the layer size since that was the limiting factor from Clair's side. The thing I didn't want was not being able to index a manifest that had to 8GB layers when we put a 10GB limit, for example.

data/secscan_model/secscan_v4_model.py
kleesc: It _should_ be. I didn't change the existing query we had for unsupported manifests, so I think it should be fine. I don't think the extra AND clause should have a performance impact on the overall query.

syed (approved):
LGTM

PR #1747, opened 2023-02-07
storage: Add MultiCDN storage provider (PROJQUAY-5048)

syed:
This storage provider can route to different underlying sub-providers based on a critiera. Currently supported filters are source_ip and namespace.

Example Config:

```    
    - MultiCDNStorage
    - providers:
        TargetName1:
            - ProviderName1
            - porviderConfig1
        Targetname2:
            - ProviderName2
            - ProviderConfig2
      default_provider: TargetName1
      rules:
      - namespace: test
        continent: APAC
        target: TargetName2
```

### DIFF
--- a/storage/__init__.py
+++ b/storage/__init__.py
@@ -1,3 +1,4 @@
+from storage.multicdnstorage import MultiCDNStorage
 from storage.cloudflarestorage import CloudFlareS3Storage
 from storage.local import LocalStorage
 from storage.cloud import (
@@ -9,24 +10,16 @@ from storage.cloud import (
 )
 from storage.fakestorage import FakeStorage
 from storage.distributedstorage import DistributedStorage
+from storage.storagecontext import StorageContext
+from storage.storagedriverclasses import STORAGE_DRIVER_CLASSES
 from storage.swift import SwiftStorage
 from storage.azurestorage import AzureStorage
 from storage.downloadproxy import DownloadProxy
-from util.ipresolver import NoopIPResolver
 
 TYPE_LOCAL_STORAGE = "LocalStorage"
 
-STORAGE_DRIVER_CLASSES = {
-    "LocalStorage": LocalStorage,
-    "S3Storage": S3Storage,
-    "GoogleCloudStorage": GoogleCloudStorage,
-    "RadosGWStorage": RadosGWStorage,
-    "SwiftStorage": SwiftStorage,
-    "CloudFrontedS3Storage": CloudFrontedS3Storage,
-    "AzureStorage": AzureStorage,
-    "RHOCSStorage": RHOCSStorage,
-    "CloudFlareStorage": CloudFlareS3Storage,
-}
+# Adding multi-CDN storage provider here to avoid circular import
+STORAGE_DRIVER_CLASSES["MultiCDNStorage"] = MultiCDNStorage
 
 
 def get_storage_driver(location, chunk_cleanup_queue, config_provider, ip_resolver, storage_params):
@@ -39,14 +32,6 @@ def get_storage_driver(location, chunk_cleanup_queue, config_provider, ip_resolv
     driver_class = STORAGE_DRIVER_CLASSES.get(driver, FakeStorage)
     context = StorageContext(location, chunk_cleanup_queue, config_provider, ip_resolver)
     return driver_class(context, **parameters)
-
-
-class StorageContext(object):
-    def __init__(self, location, chunk_cleanup_queue, config_provider, ip_resolver):
-        self.location = location
-        self.chunk_cleanup_queue = chunk_cleanup_queue
-        self.config_provider = config_provider
-        self.ip_resolver = ip_resolver or NoopIPResolver()
 
 
 class Storage(object):
--- a/storage/basestorage.py
+++ b/storage/basestorage.py
@@ -7,6 +7,10 @@ from util.registry.filelike import READ_UNTIL_END
 logger = logging.getLogger(__name__)
 
 
+class InvalidConfigurationException(Exception):
+    pass
+
+
 class StoragePaths(object):
     shared_images = "sharedimages"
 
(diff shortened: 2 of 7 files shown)


storage/multicdnstorage.py
```diff
@@ -0,0 +1,195 @@
...
+
+logger = logging.getLogger(__name__)
+
+from storage.storagecontext import StorageContext
+from storage.storagedriverclasses import STORAGE_DRIVER_CLASSES
+
+VALID_RULE_KEYS = ["namespace", "continent", "target"]
+
+
+class MultiCDNStorage(BaseStorageV2):
```
kleesc, line 15: Could you subclass S3Storage instead and create an interface for your CDNs (abstract `get_direct_download_url` method)?
That would save you from redefining a bunch of method that's not relevant to the CDN.
Then you would just need to initialize your providers which implement that interface.

util/ipresolver/__init__.py
```diff
@@ -19,11 +19,15 @@
 from util.abchelpers import nooper
 
 ResolvedLocation = namedtuple(
-    "ResolvedLocation", ["provider", "service", "sync_token", "country_iso_code", "aws_region"]
+    "ResolvedLocation",
+    ["provider", "service", "sync_token", "country_iso_code", "aws_region", "continent"],
```
kleesc, line 23: "region" might be more accurate than "continent".

util/ipresolver/__init__.py
```diff
@@ -196,3 +206,6 @@ def _parse_amazon_ranges(ranges):
                 all_amazon[region].add(IPNetwork(service_description["ip_prefix"]))
 
         return all_amazon
+
+    def get_ip_region(self, ip):
```
kleesc, line 210: Not implemented, and I don't see it being called anywhere.

storage/basestorage.py
```diff
@@ -7,6 +7,10 @@
 logger = logging.getLogger(__name__)
 
 
+class InvalidConfigurationException(Exception):
+    pass
```
kleesc, line 11: That's a very generic name. Can we rename this to something more relevant to this particular use case (relevant to storage/cdn spcifically)?

kleesc (changes requested):

storage/multicdnstorage.py
```diff
@@ -0,0 +1,195 @@
+import logging
+from typing import Dict
+
+from storage.basestorage import BaseStorageV2, InvalidConfigurationException
+from util.ipresolver import GEOIP_CONTINENT_CODES
+
+logger = logging.getLogger(__name__)
+
+from storage.storagecontext import StorageContext
+from storage.storagedriverclasses import STORAGE_DRIVER_CLASSES
```
kleesc, line 10: I don't think you need the full list of storage drivers here, just the CDN specific ones, assuming you subclass the S3Storage class.

storage/__init__.py
```diff
@@ -9,24 +10,16 @@
...
-    "GoogleCloudStorage": GoogleCloudStorage,
-    "RadosGWStorage": RadosGWStorage,
-    "SwiftStorage": SwiftStorage,
-    "CloudFrontedS3Storage": CloudFrontedS3Storage,
-    "AzureStorage": AzureStorage,
-    "RHOCSStorage": RHOCSStorage,
-    "CloudFlareStorage": CloudFlareS3Storage,
-}
+# Adding multi-CDN storage provider here to avoid circular import
+STORAGE_DRIVER_CLASSES["MultiCDNStorage"] = MultiCDNStorage
```
kleesc, line 22: Should be able to add it to the set, and avoid the circular import since you don't really need that in your multicdn file. See my other comments.

storage/multicdnstorage.py
kleesc: It'll be also easier to write tests if you interface it. That way we can just test the routing logic, instead of having to deal directly with the implementations for CloudFlare and CloudFront.

syed: I don't want to subclass `S3Storage` because that would mean I'll have to add things like bucket, access and secret key in the multiCDN provider config. I want all of that to be transparent to the MultiCDN provider as it's just a proxy and everything else should be handled by the underlying sub-providers.

In this case, I'm using the default provider for other operations but in the future that could just be another config value.

util/ipresolver/__init__.py
syed: I chose `continent` because that is the most consistent across the board. `region` is different for different people. For aws regions are like `us-east-1` and for RedHat regions are like APAC, NASA etc. Continent codes are pretty much standard.

util/ipresolver/__init__.py
syed: Removed

storage/multicdnstorage.py
syed: added classes just for the multiCDN provider

storage/__init__.py
syed: fixed

storage/multicdnstorage.py
kleesc: You still need S3 for other storage ops other than blob downloads though. How are you configuring your storage from the snippet in the description?

If it's proxy then it shouldn't be subclassing BaseStorage IMO, since it's not really a storage class. It should have its own separate abstract class/interface, which the existing storage implementations would reference.

storage/storagecontext.py
```diff
@@ -0,0 +1,9 @@
+from util.ipresolver import NoopIPResolver
+
+
+class StorageContext(object):
```
kleesc, line 4: Why did this need to move?

syed: because it's being imported in `multicdnprovider.py` and `__init__.py` and  `multicdnprovider.py` is also being imported in `__init__.py` so I had to move it otherwise there will be an cyclic import

kleesc (approved):

PR #1750, opened 2023-02-08
registry: add option to ignore unknown mediatypes (PROJQUAY-5018)

kleesc:


bcaton85 (approved):
LGTM - tested locally, works well

PR #1752, opened 2023-02-13
storage: add option to validate all configured storages (PROJQUAY-5074)

kleesc:
Add config option to validate all configured storages, and not just the first preferred one during healthcheck.


ibazulic (approved):
LGTM!

PR #1753, opened 2023-02-14
deploy: add annotation for disabling DVO check (PROJQUAY-3680)

Marcusk19:
disable min replicas check for staging environment where we only have a single replica

### DIFF
--- a/deploy/openshift/quay-py3-app.yaml
+++ b/deploy/openshift/quay-py3-app.yaml
@@ -148,6 +148,8 @@ objects:
           ${{QUAY_APP_COMPONENT_LABEL_KEY}}: ${{QUAY_APP_COMPONENT_LABEL_VALUE}}
         annotations:
           ${{QUAY_APP_COMPONENT_ANNOTATIONS_KEY}}: ${{QUAY_APP_COMPONENT_ANNOTATIONS_VALUE}}
+          ${{QUAY_DISABLE_MIN_REPLICAS_CHECK}}: ${{QUAY_DISABLE_MIN_REPLICAS_REASON}}
+
       spec:
         volumes:
         - name: configvolume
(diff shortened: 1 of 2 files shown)


deploy/openshift/quay-py3-app.yaml
```diff
@@ -148,6 +148,8 @@ objects:
           ${{QUAY_APP_COMPONENT_LABEL_KEY}}: ${{QUAY_APP_COMPONENT_LABEL_VALUE}}
         annotations:
           ${{QUAY_APP_COMPONENT_ANNOTATIONS_KEY}}: ${{QUAY_APP_COMPONENT_ANNOTATIONS_VALUE}}
+          ignore-check.kube-linter.io/minimum-replicas: ${{QUAY_DISABLE_MIN_THREE_REPLICAS_CHECK}}
```
syed, line 151: can you rename to `QUAY_DISABLE_MIN_REPLICAS_CHECK`

deploy/openshift/quay-py3-app.yaml
```diff
@@ -393,6 +395,9 @@ parameters:
   - name: QUAY_APP_READINESS_PROBE_TIMEOUT_SECONDS
     value: "10"
     displayName: quay app readiness probe timeout
+  - name: QUAY_DISABLE_MIN_THREE_REPLICAS_CHECK
+    value: ""
```
syed, line 399: one thing I'm not sure is if having an empty value keeps the check enabled. We need to validate this in stage before we deploy in prod.

syed (changes requested):

deploy/openshift/quay-py3-app.yaml
syed: or we can even evaluate this in our dev cluster

Marcusk19: @syed sounds good to me, fwiw [the doc](https://github.com/stackrox/kube-linter/blob/main/docs/configuring-kubelinter.md#ignoring-violations-for-specific-cases) I was looking at suggests that an empty value will still disable the check

syed: ok that might be an issue here. If that's the case, can you also make a param of the annotation key as well? `ignore-check.kube-linter.io/minimum-replicas` will become a param with default empty value

Marcusk19: done 🙂

syed:
can you also add this to `quay-py3-cloudflare-app.yaml` file

Marcusk19:
it's in `quay-py3-cloudflare-app.yaml` too

syed (approved):
LGTM

PR #1756, opened 2023-02-15
ui: Initial configuration for exposing UI as a dynamic plugin (PROJQAY-3203)

syed:
Adds a new entrypoint component for the UI
without topnav and sidenav. Also adds webpack
config to build dynamic plugin

### DIFF
--- a/web/src/routes/PluginMain.tsx
+++ b/web/src/routes/PluginMain.tsx
@@ -18,8 +18,9 @@ import SiteUnavailableError from 'src/components/errors/SiteUnavailableError';
 import NotFound from 'src/components/errors/404';
 import {useCurrentUser} from 'src/hooks/UseCurrentUser';
 import {InfoCircleIcon} from '@patternfly/react-icons';
-import {GlobalAuthState} from '../resources/AuthResource';
-import {IsPluginState} from '../atoms/QuayConfigState';
+import {GlobalAuthState} from 'src/resources/AuthResource';
+import {IsPluginState} from 'src/atoms/QuayConfigState';
+import axios from 'axios';
 
 const NavigationRoutes = [
   {
@@ -56,6 +57,14 @@ function PluginMain() {
     console.log('chrome auth token', token);
     GlobalAuthState.bearerToken = token;
   });
+
+  if (chrome?.isProd()) {
+    axios.defaults.baseURL = 'https://quay.io';
+  } else if (chrome?.isDemo()) {
+    axios.defaults.baseURL = 'http://localhost:8080';
+  } else {
+    axios.defaults.baseURL = 'https://stage.quay.io';
+  }
 
   useEffect(() => {
     if (quayConfig?.config?.REGISTRY_TITLE) {
--- a/scripts/app_sre_build_deploy_frontend.sh
+++ b/scripts/app_sre_build_deploy_frontend.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+# Local build script for front ends
+# This script is used to build the front end locally
+# Use this to test the frontend build system locally when making changes
+
+# Usage:
+# 3. Change the IMAGE var at the top of this script to the value from build_deploy.sh
+# 3. Run the script
+# 4. Run the container podman run -p 8000:8000 localhost/edge:5316bd7
+# 5. Open the app in your browser at http://localhost:8000/apps/edge 
+#
+# Note: You can find the image name and tag by looking at the output of the script
+# or by running `podman images`. Also, fill in the app name in the URL with the 
+# app you are testing
+
+# --------------------------------------------
+# Export vars for helper scripts to use
+# --------------------------------------------
+
+set -exv
+CURRENT_DIR=$(dirname $0)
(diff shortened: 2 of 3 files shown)


web/webpack.plugin.js
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * A host application typically provides some modules to its plugins. If an application
+ * provided module is configured as an eager singleton, we suggest using `import: false`
+ * to avoid bundling a fallback version of the module when building your plugin.
+ *
+ * Plugins may provide additional shared modules that can be consumed by other plugins.
+ *
+ * @see https://webpack.js.org/plugins/module-federation-plugin/#sharing-hints
+ */
+const pluginSharedModules = {
```
Hyperkid123, line 24: For console dot, you also want to share the  `react-router-dom` dependency.

web/src/routes/PluginMain.tsx
```diff
@@ -0,0 +1,92 @@
...
+              to provide feedback on your experience
+            </FlexItem>
+          </Flex>
+        </Banner>
+        <Routes>
+          <Route index element={<Navigate to="/organization" replace />} />
+          {NavigationRoutes.map(({path, Component}, key) => (
+            <Route path={path} key={key} element={Component} />
+          ))}
+          <Route path="*" element={<NotFound />} />
```
Hyperkid123, line 86: Just FYI. Routing will have to be eventually adjusted based on the final location of the plugin.

web/webpack.plugin.js
```diff
@@ -0,0 +1,207 @@
...
+    entryScriptFilename: isProd ? 'plugin-entry.[fullhash].min.js' : 'plugin-entry.js',
+  }),
+];
+
+module.exports = {
+  mode: isProd ? 'production' : 'development',
+  entry: {}, // Plugin container entry is generated by DynamicRemotePlugin
+  output: {
+    path: pathTo('dist'),
+    publicPath: 'http://localhost:9001/',
```
Hyperkid123, line 47: Something for the future, the public path needs to be adjusted to match the actual deployment once we get there.

web/webpack.plugin.js
```diff
@@ -0,0 +1,207 @@
...
+            loader: 'ts-loader',
+            options: {
+              transpileOnly: true,
+              experimentalWatchApi: true,
+            },
+          },
+        ],
+      },
+      {
+        test: /\.(svg|ttf|eot|woff|woff2)$/,
```
Hyperkid123, line 77: You might want to consider webpack asset modules for images,fonts,etc.
https://webpack.js.org/guides/asset-modules/#resource-assets

web/package.json
```diff
@@ -4,6 +4,8 @@
   "private": true,
   "homepage": ".",
   "dependencies": {
+    "@openshift-console/dynamic-plugin-sdk": "^0.0.18",
```
Hyperkid123, line 7: I don't think you need the "@openshift-console/dynamic-plugin-sdk" dependency.

web/package.json
```diff
@@ -4,6 +4,8 @@
...
+    "@openshift/dynamic-plugin-sdk-webpack": "^2.0.0",
```
Hyperkid123, line 8: Just a nit pick, this should be a dev dependency.

Hyperkid123 (commented):
I think this is a good start.

What we can also do, is use a shared webpack config that is used for most of consoledot UI modules. We are just one PR away from using the dynamic plugin SDK internally.

web/plugin-metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "quay-ui-plugin",
+  "version": "0.0.1",
+  "dependencies": {
+    "quay-ui": "^0.1.0"
+  },
+  "exposedModules": {
+    "QuayPluginMain": "./src/routes/PluginMain"
+  }
```
karelhala, line 9: You are missing an `extensions` list. Without it we won't know what to render.

@Hyperkid123 we should discuss how we want to handle multi route plugin. This plugin exposes similiar thing to what CRC app is exposing, however true plugin should expose just the routes and chroming service should handle the rendering.

karelhala (commented):
Looking good! With this approach we'd be able to load quay app in CRC without any problems.

web/plugin-metadata.json
Hyperkid123: You don't need any extensions list to render in HCC. Scalprum is looking for the exposedModules directly and will be for some time. We can migrate to "true" plugin routing anyways until every HCC app is a plugin. This will suffice for some time.

web/webpack.plugin.js
syed: Added

web/plugin-metadata.json
syed: I'm gonna leave this as-is for now then

web/package.json
syed: fixed

web/package.json
syed: fixed

syed:
Merged as part of #1799

PR #1759, opened 2023-02-16
deploy: Allow for dynamic service names (PROJQUAY-5103)

bcaton85:
Allows for creation of separate services for multiple deployments.

### DIFF
--- a/deploy/openshift/quay-py3-deploy-template.yaml
+++ b/deploy/openshift/quay-py3-deploy-template.yaml
@@ -1,0 +1,485 @@
+---
+apiVersion: v1
+kind: Template
+metadata:
+  name: quay-py3-deploy-template
+objects:
+- apiVersion: rbac.authorization.k8s.io/v1
+  kind: Role
+  metadata:
+    name: ${{NAME}}
+  rules:
+  - apiGroups:
+    - ""
+    resources:
+    - secrets
+    verbs:
+    - get
+    - patch
+    - update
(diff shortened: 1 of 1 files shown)


syed:
can you make this change in the `quay-py3-app.yaml`  and delete the `quay-py3-cloudflare-app.yaml` file we can also remove `quay-app.yaml` and `clair-backfiller.yaml` that was for the py2 deployment.

deploy/openshift/quay-py3-app.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Template
 metadata:
-  name: quay-py3
+  name: quay-py3-cloudflare
```
syed, line 5: can we rename this to `quay-py3-deploy-template`

syed: same with the file as well

deploy/openshift/quay-py3-app.yaml
```diff
@@ -286,7 +285,7 @@ objects:
             value: ${QUAY_WORKER_CONNECTION_COUNT_REGISTRY}
 parameters:
   - name: NAME
-    value: "quay-py3"
+    value: "quay-py3-cloudflare"
```
syed, line 288: let's keep the default to `quay-py3`

syed (approved):
LGTM

PR #1764, opened 2023-02-21
PROJQUAY-4993: Allow action logs to be forwarded to Splunk

harishsurf:
This PR adds Splunk logging model to forward user action logs to a splunk instance

### DIFF
--- a/data/logs_model/logs_producer/splunk_logs_producer.py
+++ b/data/logs_model/logs_producer/splunk_logs_producer.py
@@ -1,0 +1,44 @@
+import logging
+
+from splunklib import client
+
+from data.logs_model.logs_producer import LogSendException
+from data.logs_model.logs_producer.interface import LogProducerInterface
+
+logger = logging.getLogger(__name__)
+
+
+class SplunkLogsProducer(LogProducerInterface):
+    """
+    Log producer for writing log entries to Splunk
+    This implementation writes directly to Splunk without a streaming/queueing service.
+    """
+
+    def __init__(
+        self,
+        host,
+        port,
+        bearer_token,
+        url_scheme="https",
+        verify_ssl=True,
+        index_prefix=None,
+    ):
+        try:
+            service = client.connect(
+                host=host, port=port, token=bearer_token, scheme=url_scheme, verify=verify_ssl
+            )
+        except Exception as ex:
+            logger.exception("Failed to connect to Splunk instance %s", ex)
+            raise ex
+        try:
+            self.index = service.indexes[index_prefix]
+            logger.info("splunk index %s", self.index)
+        except KeyError:
+            self.index = service.indexes.create(index_prefix)
+
+    def send(self, log):
+        try:
+            self.index.submit(log, sourcetype="access_combined", host="quay")
+        except Exception as e:
+            logger.exception("SplunkLogsProducer exception sending log to Splunk: %s", e)
+            raise LogSendException("SplunkLogsProducer exception sending log to Splunk: %s" % e)
--- a/data/logs_model/splunk_logs_model.py
+++ b/data/logs_model/splunk_logs_model.py
@@ -1,0 +1,144 @@
+import json
+import logging
+
+from datetime import datetime
+
+from data import model
+from data.logs_model.interface import ActionLogsDataInterface
+from data.logs_model.logs_producer import LogProducerProxy, LogSendException
+from data.logs_model.logs_producer.splunk_logs_producer import SplunkLogsProducer
+from data.logs_model.shared import SharedModel
+from data.model import config
+from data.model.log import ACTIONS_ALLOWED_WITHOUT_AUDIT_LOGGING
+
+logger = logging.getLogger(__name__)
+
+
+class SplunkLogsModel(SharedModel, ActionLogsDataInterface):
+    """
+    SplunkLogsModel implements model for establishing connection and sending events to Splunk cluster
+    """
+
+    def __init__(self, producer, splunk_config, should_skip_logging=None):
+        self._should_skip_logging = should_skip_logging
+        self._logs_producer = LogProducerProxy()
+        if producer == "splunk":
+            self._logs_producer.initialize(SplunkLogsProducer(**splunk_config))
+        else:
+            raise Exception("Invalid log producer: %s" % producer)
+
+    def log_action(
+        self,
+        kind_name,
+        namespace_name=None,
+        performer=None,
+        ip=None,
+        metadata=None,
+        repository=None,
+        repository_name=None,
+        timestamp=None,
+        is_free_namespace=False,
+    ):
+
+        if self._should_skip_logging and self._should_skip_logging(
+            kind_name, namespace_name, is_free_namespace
+        ):
+            return
+
+        if repository_name is not None:
+            if repository is None or namespace_name is not None:
+                raise ValueError(
+                    "Incorrect argument provided when logging action logs, namespace name should not be "
+                    "empty"
+                )
+            repository = model.repository.get_repository(namespace_name, repository_name)
+
+        if timestamp is None:
+            timestamp = datetime.today()
+
+        account_id = None
+        performer_id = None
+        repository_id = None
+
+        if namespace_name is not None:
+            ns_user = model.user.get_namespace_user(namespace_name)
+            if ns_user is not None:
+                account_id = ns_user.id
+
+        if performer is not None:
+            performer_id = performer.id
+
+        if repository is not None:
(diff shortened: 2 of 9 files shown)


config.py
```diff
@@ -680,6 +680,7 @@ def create_transaction(db):
     # Log model
     LOGS_MODEL = "database"
     LOGS_MODEL_CONFIG: Dict[str, Any] = {}
+    SPLUNK_LOGS_MODEL_CONFIG: Dict[str, Any] = {}
```
dmage, line 683: Shouldn't it reuse LOGS_MODEL_CONFIG?

cc @kleesc

docker-compose.yaml
```diff
@@ -81,6 +81,7 @@ services:
     ports:
       - "8080:8080" 
       - "8443:8443"
+      - "8089:8089"
```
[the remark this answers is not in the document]
harishsurf: It's not required - it was only for development work

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -0,0 +1,40 @@
...
+                self.index = service.indexes[index_prefix]
+                logger.info("splunk index %s", self.index)
+            except Exception as e:
+                self.index = service.indexes.create(index_prefix)
+
+    def send(self, log):
+        try:
+            self.index.submit(log, sourcetype="access_combined", host="quay")
+        except Exception as e:
+            logger.exception("SplunkLogsProducer exception sending log to Splunk: %s", e)
```
flavianmissi, line 40: Same question here. When the `submit` call above raises an exception you are logging it and nothing else.
If we are using the elastic logs producer as reference, when it encounters an exception it both logs it and raises another exception of it's own (`LogSendException`).

data/logs_model/splunk_logs_model.py
```diff
@@ -0,0 +1,85 @@
...
+            "repository": repository_id,
+            "ip": ip,
+            "metadata_json": metadata_json or {},
+            "datetime": timestamp,
+        }
+
+        try:
+            self._logs_producer.send(json.dumps(log_data, indent=None, sort_keys=True, default=str))
+        except LogSendException as lse:
+            logger.exception("log_action failed", lse)
```
flavianmissi, line 85: Unless you change `SplunkLogsProducer.send` to raise a `LogSendException`, this code will never execute.

data/logs_model/splunk_logs_model.py
```diff
@@ -0,0 +1,85 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SplunkLogsModel(SharedModel):
+    """
+    SplunkLogsModel implements model for establishing connection and sending events to Splunk cluster
+    """
+
+    def __init__(self, should_skip_logging=None, producer=None, splunk_config=None):
```
flavianmissi, line 19: Is there a point to setting all these arguments to be `None` by default? I understand that makes sense for `should_skip_logging`, but `producer` and `splunk_config` should never be `None`. In fact if either are `None` an exception will occur.

data/logs_model/splunk_logs_model.py
```diff
@@ -0,0 +1,85 @@
...
+
+from data import model
+from data.logs_model.logs_producer import LogProducerProxy, LogSendException
+from data.logs_model.logs_producer.splunk_logs_producer import SplunkLogsProducer
+from data.logs_model.shared import SharedModel
+
+logger = logging.getLogger(__name__)
+
+
+class SplunkLogsModel(SharedModel):
```
flavianmissi, line 14: I think you'll need to implement at least _some_ of the methods in `ActionLogsDataInterface`. I found that `lookup_logs` for example is called by `endpoints/api/superuser.py` and `endpoints/api/logs.py`. `workers/exportactionlogsworker.py` calls `yield_logs_for_export`. 

These are only the ones I checked - I think you need to grep around the code and see which functions from `ActionLogsDataInterface` are called and which ones aren't. Then you'll need to make a choice on whether the splunk logs model needs them or not (and what happens if you don't write them?). I would assume that the superuser and logs apis need to function when customers use splunk logs, but I don't know for sure. Same question applies to other parts of quay.

flavianmissi (commented):
this is a good start!

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -0,0 +1,40 @@
...
+    Log producer for writing log entries to Splunk
+    This implementation writes directly to Splunk without a streaming/queueing service.
+    """
+
+    def __init__(
+        self,
+        host,
+        port,
+        bearer_token=None,
+        use_ssl=True,
```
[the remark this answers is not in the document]
harishsurf: changed variable name to `verify_ssl` and added `url_scheme` option too

data/logs_model/splunk_logs_model.py
harishsurf: makes sense. I see that elastic search does this too [here](https://github.com/quay/quay/blob/master/data/logs_model/elastic_logs.py#L101-L106)

data/logs_model/splunk_logs_model.py
harishsurf: ah, you're right!

data/logs_model/splunk_logs_model.py
harishsurf: except for `log_action` every other method in `ActionLogsDataInterface` does a look up of logs from splunk. I confirmed with PM that for this release we only support forwarding logs to splunk. I have added a generic message for the unimplemented methods

flavianmissi: What does it look like from the perspective of a user of the api? I think raising a `NotImplementedError` will blow up a 500 unless you treat that exception somehow.

dmage: What features won't be available to customers if they use Splunk?

data/logs_model/splunk_logs_model.py
```diff
@@ -0,0 +1,141 @@
...
+    ):
+
+        if self._should_skip_logging and self._should_skip_logging(
+            kind_name, namespace_name, is_free_namespace
+        ):
+            return
+
+        if repository_name is not None:
+            assert repository is None
+            assert namespace_name is not None
```
flavianmissi, line 50: Please don't do this. `assert` is for testing and debugging. If this code ever errors, customers will see an `AssertionError` in their logs, and that means very little to them. It's better to raise a generic `Exception` with a meaningful message instead.

Better even (although that would require a bigger refactor, which is out of the scope of this PR) would be change this function's arguments so that if one gives a repository name, then they _have_ to also give the namespace name. Having to do this kind of argument validation is a smell. But as I said, it's out of scope for you to handle to root cause, so just do the best you can with what you have.

flavianmissi (commented):
Almost there!

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -0,0 +1,46 @@
...
+            service = client.connect(host=host, port=port,  token=bearer_token, scheme=url_scheme, verify=verify_ssl)
+        except Exception as ex:
+            logger.exception("Failed to connect to Splunk instance %s", ex)
+            raise ex
+        else:
+            try:
+                self.index = service.indexes[index_prefix]
+                logger.info("splunk index %s", self.index)
+            except Exception as e:
+                self.index = service.indexes.create(index_prefix)
```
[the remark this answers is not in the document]
harishsurf: agreed. `else` was added initially since I wasn't re-raising the exception in my previous try/except block

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            service = client.connect(host=host, port=port,  token=bearer_token, scheme=url_scheme, verify=verify_ssl)
+        except Exception as ex:
+            logger.exception("Failed to connect to Splunk instance %s", ex)
+            raise ex
+        else:
+            try:
+                self.index = service.indexes[index_prefix]
+                logger.info("splunk index %s", self.index)
+            except Exception as e:
```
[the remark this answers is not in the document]
harishsurf: changed it to `KeyError`.

flavianmissi (approved):
LGTM with a note on tests.

The automated test coverage here is _very_ basic. I won't stop you from merging this because I trust you have tested this manually, but you need to work on these automated tests in a follow-up PR. Make sure you cover at least a couple of error cases. `log_action` has quite some logic in it, none of which is covered by tests. Let me know if you need help with this.

PR #1767, opened 2023-03-01
security: Change error messages in UI during LDAP login (PROJQUAY-4845)

ibazulic:
Previously, on installations where LDAP is used, we were telling users whether the username or password was failing when login attempts were made. This might pose a security risk, a malicious user could, via the returned message, identify which users have access to Quay and which don't. 

With this change, we return a general message saying the user used wrong credentials instead of providing any details.


bcaton85 (approved):
LGTM

PR #1790, opened 2023-03-16
config: Add config support for IPv6 (PROJQUAY-272)

ibazulic:
This PR adds config support for choosing the IP stack that Quay will run on. The property is marked as required. Defaults to IPv4, unless specified otherwise.


util/config/schema.py
```diff
@@ -105,6 +105,7 @@
         "DISTRIBUTED_STORAGE_PREFERENCE",
         "DEFAULT_TAG_EXPIRATION",
         "TAG_EXPIRATION_OPTIONS",
+        "FEATURE_LISTEN_IP_VERSION",
```
syed, line 108: this is already added in the schema here: https://github.com/quay/quay/blob/300aed145099a5918e77a82e54b7c549f4ebcb07/util/config/schema.py#L1231

maybe you could add possible values in the schema which are "IPv6", "IPv4" and "dual-stack"

syed (changes requested):

ibazulic:
Closing this PR as it already exists as per Syed's comment.

PR #1791, opened 2023-03-16
api: Allow setting multiple CORS_ORIGIN (PROJQUAY-5213)

syed:
multiple CORS Origin are required for integration with console.redhat

### DIFF
--- a/util/request.py
+++ b/util/request.py
@@ -29,7 +29,7 @@ def crossorigin(anonymous=True):
     def decorate(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
-            cors_origin_list = app.config.get("CORS_ORIGIN", None)
+            cors_origin_list = app.config.get("CORS_ORIGIN", [])
             cors_origin = "*"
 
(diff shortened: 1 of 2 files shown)


util/request.py
```diff
@@ -30,15 +30,27 @@ def decorate(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
             cors_origin = app.config.get("CORS_ORIGIN", "*")
+
+            # if there are multiple CORS_ORIGIN set, then use
+            # the Origin header from the request to set the
+            # correct Allow-Origin
+            if isinstance(cors_origin, list):
```
kleesc, line 37: Is this a list, or do you need to split on `,` first?

kleesc: I'd say, either make everything a string, or everything a list. The change of from string to list when there's only one policy vs multiple is awkward. 

Otherwise LGTM

syed: yeah that's what I was thinking as well. I'll make everything a list

kleesc (approved):

PR #1799, opened 2023-03-20
ui: add support for exposing quay UI as a dynamic plugin (PROJQUAY-3203)

syed:
* Introduces a new SSO JWT based auth for client side Oauth
* Adds a new entrypoint component for the UI without topnav and sidenav for plugin
* Adds webpack config to build dynamic plugin


auth/oauth.py
```diff
@@ -27,6 +37,63 @@ def validate_bearer_auth(auth_header):
 
 
 def validate_oauth_token(token):
+    if is_jwt(token):
+        return validate_sso_oauth_token(token)
+    else:
+        return validate_app_oauth_token(token)
+
+
+def validate_sso_oauth_token(token):
```
kleesc, line 46: Is this different from an oauth token we would normally validate?

auth/test/test_oauth.py
```diff
@@ -1,5 +1,8 @@
+import jwt
 import pytest
+from httmock import urlmatch, HTTMock
 
+from app import oauth_login
```
kleesc, line 5: Are these new imports being used?

auth/oauth.py
syed: yes our oauth token has a different format that a standard jwt token

Makefile
```diff
@@ -231,7 +231,7 @@ endif
 local-dev-up-with-clair: local-dev-up
 	$(DOCKER_COMPOSE) up -d clair-db
 	$(DOCKER) exec -it clair-db bash -c 'while ! pg_isready; do echo "waiting for postgres"; sleep 2; done'
-	$(DOCKER_COMPOSE) up -d clair
+	DOCKER_USER="$$(id -u):0" $(DOCKER_COMPOSE) up -d clair
```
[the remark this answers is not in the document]
syed: This might be specific to mac. If you don't add this then when you mount a config volume it causes a permission denied error as the user inside the container is different from the user who owns the files in the volume. This is already  being used when starting the quay container https://github.com/quay/quay/blob/master/Makefile#L208

oauth/login_utils.py
```diff
@@ -0,0 +1,243 @@
...
+def is_jwt(token):
+    try:
+        [h_b64, p_b64, s_b64] = token.split(".")
+        headers_json = base64.b64decode(h_b64 + "==")
+        headers = json.loads(headers_json)
+        return headers.get("typ", "").lower() == "jwt"
+    except (json.JSONDecodeError, ValueError):
+        pass
+
+    return False
```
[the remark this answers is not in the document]
syed: this is parsing the header. `pyjwt` requires the algorithm to be passed in order to decode + verify the JWT token. The header needs to be parsed outside pytjwt to extract the algorithm

auth/test/test_oauth.py
syed: removed unused imports

oauth/login_utils.py
```diff
@@ -0,0 +1,243 @@
...
+    """
+    Extract the issuer from the JWT token.
+    The passed token is assumed to be a valid
+    JWT token
+    """
+    [h_b64, p_b64, s_b64] = token.split(".")
+    payload_json = base64.b64decode(p_b64 + "==")
+    payload = json.loads(payload_json)
+
+    return payload.get("iss", None)
```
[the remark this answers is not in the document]
syed: updated to use the JWT library

oauth/login_utils.py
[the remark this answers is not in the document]
syed: updated to use the JWT library

app.py
```diff
@@ -65,6 +67,7 @@
 from util.label_validator import LabelValidator
 from util.metrics.prometheus import PrometheusPlugin
 from util.repomirror.api import RepoMirrorAPI
+from util.security import jwtutil
```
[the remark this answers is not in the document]
syed: removed import

endpoints/oauth/test/test_login.py
```diff
@@ -113,7 +126,13 @@ def test_flagged_user_creation(
         with patch("features.INVITE_ONLY_USER_CREATION", invite_only):
             # Conduct login.
             result = _conduct_oauth_login(
-                internal_auth, login_service, login_service_lid, login_service_lid, email
+                app,
+                analytics,
```
kleesc, line 130: Do we need to pass this analytics objects around, are we actually using it?

oauth/utils.py
```diff
@@ -5,6 +5,7 @@
 import urllib.parse
 from random import SystemRandom
 
+
```
kleesc, line 8: Extra space

auth/decorators.py
```diff
@@ -57,6 +58,10 @@ def wrapper(*args, **kwargs):
                 if result.error_message is not None:
                     # Log the failure.
                     authentication_count.labels(result.kind, False).inc()
+                    # Do we only need to abort for JWT based errors?
+                    if result.kind == AuthKind.ssojwt:
+                        abort(401, message=result.error_message)
```
kleesc, line 63: Where is this actually being validated? I don't see changes made to the existing cookie decorator.

app.py
```diff
@@ -2,6 +2,7 @@
 import json
 import logging
 import os
+import jwt
```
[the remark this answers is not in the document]
syed: removed

app.py
```diff
@@ -34,6 +35,7 @@
 from data.billing import Billing
 from data.buildlogs import BuildLogs
 from data.cache import get_model_cache
+from data.database import User
```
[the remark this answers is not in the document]
syed: removed

auth/decorators.py
syed: This is used in https://github.com/quay/quay/blob/master/auth/decorators.py#L72 which gets added to all API calls in https://github.com/quay/quay/blob/master/endpoints/api/__init__.py#L72

oauth/utils.py
syed: removed

endpoints/oauth/test/test_login.py
syed: yes, the analytics object is used to count successful creations of a federated login. Not sure where we actually hook it into though ... https://github.com/quay/quay/blob/5defae13285881e582b2fa69eab3084d853bc38e/oauth/login_utils.py#L224

auth/decorators.py
syed: This decorator get's called on the oauth flow and this oauth token is not in the cookie but rather passed as a header. hence validation is here

kleesc: What are we validating it against? Normally, we would compare the token against the one generated in the database. In this case, is it just a matter of whether the signature is valid or not?

syed: summarizing the discussion here. I'm passing `verify_signature: True` to the `ODICLoginService` which causes it to verify the jwt token with the public key from the SSO sever

kleesc (approved):

auth/decorators.py
kleesc: ack. did not see  the that it was verifying against the jwks.

kleesc (approved):

PR #1803, opened 2023-03-21
chore: Downgrade cryptography to 3.3.2 (PROJQUAY-5120)

dmage:
Quay downstream build system cannot build new versions of the package `cryptography` as they use Rust. Unless downstream is fixed, we should use the latest available version that does not depend on Rust.


flavianmissi (approved):
LGTM

PR #1808, opened 2023-03-24
ldap: Don't convert dashes to underscores in usernames (PROJQUAY-5253)

ibazulic:
When `FEATURE_USER_CONFIRMATION` is set to true and an LDAP username that has dashes inside is used, Quay will automatically change dashes to underscores. This breaks `LDAP_RESTRICTED_USER_FILTER`  since namespaces in Quay do not have to be the same as LDAP namespaces. The change can only impact ancient Docker versions prior to version 1.5 where support for dashes in namespaces has been introduced.


ibazulic:
> how will this behave on current installations where the replacement has already happened?

It should behave exactly as it is right now because we're not changing any username logic apart from not changing dashes to underscores automatically.

PR #1811, opened 2023-03-27
cors: check for request_origin being set (PROJQUAY-5213)

syed:
Some browsers might not set the Origin header
in the request. Ignore the origin check in such
cases


Sunandadadi (approved):
lgtm

PR #1813, opened 2023-03-28
secscan: send notifications for new indexes (PROJQUAY-4659)

Marcusk19:
On a new push, send a notification if vulnerabilities are found after indexing.

### DIFF
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -2,7 +2,6 @@ import logging
 import functools
 import itertools
 import random
-import json
 
 from collections import namedtuple
 from datetime import datetime, timedelta
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -2,6 +2,7 @@ import logging
 import functools
 import itertools
 import random
+import json
 
 from collections import namedtuple
 from datetime import datetime, timedelta
@@ -55,7 +56,7 @@ logger = logging.getLogger(__name__)
 
 
 DEFAULT_SECURITY_SCANNER_V4_REINDEX_THRESHOLD = 86400  # 1 day
-
+FIRST_INDEX = 0
 
 IndexReportState = namedtuple("IndexReportState", ["Index_Finished", "Index_Error"])(  # type: ignore[call-arg]
     "IndexFinished", "IndexError"
@@ -257,6 +258,8 @@ class V4SecurityScanner(SecurityScannerInterface):
         )
 
         end_index = Manifest.select(fn.Max(Manifest.id)).scalar()
+        if end_index is None:
+            end_index = 0
         start_index = max(end_index - batch_size, 1)
 
         iterator = self._get_manifest_iterator(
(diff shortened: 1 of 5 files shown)


kleesc:
@Marcusk19 There should be a way to turn this on/off.

data/secscan_model/secscan_v4_model.py
```diff
@@ -401,6 +404,45 @@ def should_skip_indexing(manifest_candidate):
...
+                            vulnerability_report = None
+
+                        try:
+                            found_vulnerabilities = vulnerability_report["vulnerabilities"]
+                        except TypeError:
+                            found_vulnerabilities = None
+
+                        if found_vulnerabilities is not None:
+                            keys = list(found_vulnerabilities)
+                            vuln = found_vulnerabilities[keys[FIRST_INDEX]]
```
kleesc, line 424: Any reason to only add the first one?

util/config/schema.py
```diff
@@ -1260,5 +1260,10 @@
             "description": "Whitelisted users to exclude when FEATURE_RESTRICTED_USERS is enabled",
             "x-example": ["devtable"],
         },
+        "FEATURE_NOTIFY_NEW_PUSH_VULNS": {
```
kleesc, line 1263: Can you rename this `FEATURE_SECURITY_SCANNER_NOTIFY_ON_NEW_INDEX` or something similar to keep it consistent with the other flags?

notifications/notificationevent.py
```diff
@@ -211,6 +212,9 @@ def should_perform(self, event_data, notification_data):
         if VulnerabilityFoundEvent.CONFIG_LEVEL not in event_config:
             return True
 
+        if VulnerabilityFoundEvent.NEW_PUSH in event_data:
```
kleesc, line 215: nit: I would rename this `VULNERABLE_INDEX_REPORT_CREATED`

data/secscan_model/secscan_v4_model.py
```diff
@@ -401,6 +404,45 @@ def should_skip_indexing(manifest_candidate):
...
+                            }
+
+                            import notifications
+
+                            logger.debug("Attempting to add vulnerability notifications to queue")
+                            notifications.spawn_notification(
+                                manifest.repository, "vulnerability_found", event_data
+                            )
+                        else:
+                            logger.debug("No net new vulnerabilities found")
```
kleesc, line 444: The `else` block isn't really needed.

data/secscan_model/secscan_v4_model.py
```diff
@@ -401,6 +404,45 @@ def should_skip_indexing(manifest_candidate):
                     dur_ms = get_epoch_timestamp_ms() - manifest.created_at
                     dur_sec = dur_ms / 1000
                     secscan_result_duration.observe(dur_sec)
+
+                    if features.NOTIFY_NEW_PUSH_VULNS:
+                        # if there are "new" vulnerabilities found, take the first one and add it to the notification queue
```
kleesc, line 409: Comment not really needed.

kleesc (changes requested):

data/secscan_model/secscan_v4_model.py
Marcusk19: I was thinking it might spam notifications if the index report contains a lot of vulnerabilities, but looking back I think we might as well add them all since the user can turn this on and off now.

PR #1817, opened 2023-04-03
reconciler: add worker (PROJQUAY-5320)

Marcusk19:
For marketplace integration for quay.io, adds a reconciliation worker.
The worker should periodically retrieve Red Hat subscriptions for all active users and store/update them in the `RedHatSubscriptions` table.


data/database.py
```diff
@@ -2121,6 +2122,34 @@ def upstream_registry_hostname(self) -> str:
...
+
+    MW00584MO = 5
+    MW00585MO = 10
+    MW00586MO = 20
+    MW00587MO = 50
+    MW00588MO = 125
+    MW00589MO = 250
+    MW00590MO = 500
+    MW00591MO = 1000
+    MW00592MO = 2000
```
kleesc, line 2150: This should go into `data/billing.py` in my opinion. Also, can you create a dict or something similar mapping these SKUs to their equivalent Stripe plan.

data/marketplace.py
```diff
@@ -0,0 +1,155 @@
+import json
+import logging
+import requests
+
+logger = logging.getLogger(__name__)
+
+REQUEST_TIMEOUT = 200
```
kleesc, line 7: 200 seconds seems like a lot for what should be a simple api request.

data/marketplace.py
```diff
@@ -0,0 +1,155 @@
...
+
+MARKETPLACE_FILE = "/conf/stack/auth/quay-marketplace-api-stage.crt"
+MARKETPLACE_SECRET = "/conf/stack/auth/quay-marketplace-api-stage.key"
+
+
+class UserAPI:
```
kleesc, line 13: Can you prefix this name with RH (i.e `RHUserAPI`).  User api is very generic.

data/marketplace.py
```diff
@@ -0,0 +1,155 @@
...
+        logger.debug(r.content)
+        info = json.loads(r.content)
+        if not info:
+            return None
+        # gathering customer id from response
+        CustomerId = info[0]["accountRelationships"][0]["account"]["ebsAccountNumber"]
+        return CustomerId
+
+
+class MarketplaceAPI:
```
kleesc, line 58: Same, prefix this with "RH", since this is Redhat specific.

data/marketplace.py
```diff
@@ -0,0 +1,155 @@
...
+            json=request_body_dict,
+            verify=False,
+            timeout=REQUEST_TIMEOUT,
+        )
+
+        logger.debug(r.content)
+        info = json.loads(r.content)
+        if not info:
+            return None
+        # gathering customer id from response
```
kleesc, line 53: Don't really need this comment.

workers/reconciliationworker.py
```diff
@@ -0,0 +1,187 @@
...
+        return False
+
+    def _perform_reconciliation(self, user_api, marketplace_api):
+        """
+        Gather all entitlements from internal marketplace api and store in quay db
+        Create new entitlements for stripe customers if needed
+        """
+        logger.info("Reconciliation worker is gathering entitlements...")
+
+        users = model.user.get_active_users()
```
kleesc, line 63: This will loop through every user on each reconciliation, even if no changes were made. Is there any way to do this only when required? If so, does this need to happen both on Stripe change or RH marketplace changes?

workers/reconciliationworker.py
```diff
@@ -0,0 +1,187 @@
...
+            self._reconcile_entitlements,
+            app.config.get("RECONCILIATION_FREQUENCY", RECONCILIATION_FREQUENCY),
+        )
+
+    def _check_stripe_matches_sku(self, user, sku):
+        """
+        Check if user's stripe plan matches with RH sku
+        """
+        stripe_customer = stripe.Customer.retrieve(user.stripe_id)
+        logger.debug("stripe customer = %s", stripe_customer)
```
kleesc, line 44: Don't really need this debug statement.

workers/reconciliationworker.py
```diff
@@ -0,0 +1,187 @@
...
+                    logger.debug(
+                        "Adding new subscription %s to database for user %s",
+                        rh_sku,
+                        user.username,
+                    )
+                    entitlements.save_subscription(
+                        user.id, subscription_id, ebsAccountNumber, endDateDatetime, rh_sku
+                    )
+                elif local_subscription.subscription_end_date < endDateDatetime:
+                    entitlements.update_subscription_end_date(subscription_id, endDateDatetime)
```
kleesc, line 122: Could we just keep track of the account number: what the purpose of persisting these subscriptions in the database?

workers/reconciliationworker.py
```diff
@@ -0,0 +1,187 @@
...
+        Check if user's stripe plan matches with RH sku
+        """
+        stripe_customer = stripe.Customer.retrieve(user.stripe_id)
+        logger.debug("stripe customer = %s", stripe_customer)
+        if stripe_customer is None:
+            logger.debug("user %s has no valid subscription on stripe", user.username)
+            return False
+
+        plan = get_plan(stripe_customer.subscription.plan.id)
+        if plan["privateRepos"] == sku.value:
```
kleesc, line 50: This should be matched based on some id. See above comment about having a mapping between the SKUs and Stripe plans.

workers/reconciliationworker.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def _reconcile_entitlements(self, skip_lock_for_testing=True):
+        """
+        Performs reconciliation for user entitlements
+        """
+        internal_user_api = marketplace.UserAPI(app.config)
+        internal_marketplace_api = marketplace.MarketplaceAPI(app.config)
+        # generate random wait
+        random_wait = random.randint(1, 100)
+        time.sleep(random_wait)
```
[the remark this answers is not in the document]
Marcusk19: From the original design it was specified the worker should wait a random amount of time before proceeding. I'm not 100% why, I think it has something to do with locking?

workers/reconciliationworker.py
Marcusk19: I'm not sure if there's a way to do it. I think we'll have to keep polling this to make sure entitlements consistently stay in sync. If there was a way, it would only need to happen on RH marketplace changes.

workers/reconciliationworker.py
Marcusk19: My understanding here is that we would look at the database every time we check stripe as well. That way we don't have to call the internal api too often.

util/marketplace.py
```diff
@@ -0,0 +1,170 @@
...
+import requests
+
+from data.billing import RH_SKU_STRIPE_IDS, RH_SKUS, get_plan
+from datetime import datetime
+
+from data.model import entitlements
+
+logger = logging.getLogger(__name__)
+
+REQUEST_TIMEOUT = 60
```
kleesc, line 12: I would set a lower default. These requests shouldn't ever take over 60 sec.

util/marketplace.py
```diff
@@ -0,0 +1,170 @@
...
+            cert=self.cert,
+            json=request_body_dict,
+            verify=True,
+            timeout=REQUEST_TIMEOUT,
+        )
+
+        info = json.loads(r.content)
+        if not info:
+            return None
+        CustomerId = info[0]["accountRelationships"][0]["account"]["ebsAccountNumber"]
```
kleesc, line 67: Any particular reason this is CamelCased?

data/billing.py
```diff
@@ -542,6 +567,16 @@ def retrieve(cls, payment_method_id):
...
+    class Charge(AttrDict):
+        @staticmethod
+        def list(limit):
+            return AttrDict(
+                {
+                    "has_more": False,
+                    "data": [{"id": 1}],
+                }
+            )
+
```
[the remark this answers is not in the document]
Marcusk19: This was left over from when I was trying to mock the stripe api - removing it

util/marketplace.py
Marcusk19: No reason, I'm going to rename this to account_number for consistency

data/billing.py
```diff
@@ -338,6 +348,19 @@
...
+    "MW00585MO": 10,
+    "MW00586MO": 20,
+    "MW00587MO": 50,
+    "MW00588MO": 125,
+    "MW00589MO": 250,
+    "MW00590MO": 500,
+    "MW00591MO": 1000,
+    "MW00592MO": 2000,
+    "FakeSKU": 0,  # mock sku for unit test
+}
```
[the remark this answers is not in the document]
Marcusk19: I'd like to keep at least a list of the sku ids since I iterate through them in a couple functions. I can get rid of the private repo count though?

endpoints/api/billing.py
```diff
@@ -40,6 +40,19 @@
 import uuid
 import json
 
+MILLISECONDS_IN_SECONDS = 1000
+
+
+def check_internal_api_for_subscription(namespace_user):
+    """
+    If Reconciler is enabled, RH marketplace should also be checked
+    for valid subscriptions
```
kleesc, line 49: Is this comment still valid, or is this independent of the reconciler?

workers/reconciliationworker.py
```diff
@@ -0,0 +1,151 @@
...
+    def _perform_reconciliation(self, user_api, marketplace_api):
+        """
+        Gather all entitlements from internal marketplace api and store in quay db
+        Create new entitlements for stripe customers if needed
+        """
+        logger.info("Reconciliation worker looking to create new subscriptions...")
+
+        users = model.user.get_active_users()
+
+        for user in users:
```
kleesc, line 77: Do you need to iterate every user, or only those with a stripe id?

endpoints/api/billing.py
Marcusk19: Should be independent, changed it to be more accurate

workers/reconciliationworker.py
Marcusk19: We can only look at stripe users, the only benefit I think going over all users is we can save account numbers for people that aren't using stripe with the reconciler

kleesc (approved):

PR #1818, opened 2023-04-05
billing: update Stripe checkout to support 3DS (PROJQUAY-5129)

kleesc:
Update Stripe checkout in order to support auth requirements from EU/India banks.


Sunandadadi (approved):
lgtm

PR #1820, opened 2023-04-07
locking: change log severity (PROJQUAY-5221)

Marcusk19:
When a worker fails to acquire the lock, make sure a log message is sent as a warning for better visibility when troubleshooting.

### DIFF
--- a/util/locking.py
+++ b/util/locking.py
@@ -74,7 +74,7 @@ class GlobalLock(object):
             logger.debug("Acquired lock %s", self._lock_name)
             return True
         except RedisError as re:
-            logger.debug("Could not connect to Redis for lock %s: %s", self._lock_name, re)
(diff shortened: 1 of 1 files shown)


util/locking.py
```diff
@@ -74,7 +74,7 @@ def acquire(self):
             logger.debug("Acquired lock %s", self._lock_name)
             return True
         except RedisError as re:
-            logger.debug("Could not connect to Redis for lock %s: %s", self._lock_name, re)
+            logger.warn("Could not connect to Redis for lock %s: %s", self._lock_name, re)
```
bcaton85, line 77: warn has been deprecated, [use warning instead.](https://github.com/python/cpython/blob/3.3/Lib/logging/__init__.py#L1252-L1253) They end up calling the same function but better to be up to date.

bcaton85 (approved):
LGTM

PR #1822, opened 2023-04-11
logs: Add audit logs for login/logout to Quay (PROJQUAY-2344)

harishsurf:
Adds audit events for : 
- login/ logout of quay UI
- login via docker/podman CLI

Note: No call is made to docker registry API during a logout from CLI, hence no audit events are logged
Testing: verified that the above audit events are generated via a local splunk instance

### DIFF
--- a/data/migrations/versions/0246c2d0e750_add_login_logout_logentrykind.py
+++ b/data/migrations/versions/0246c2d0e750_add_login_logout_logentrykind.py
@@ -1,0 +1,33 @@
+"""add login logout logentrykind
+
+Revision ID: 0246c2d0e750
+Revises: b2d1e4b95fc2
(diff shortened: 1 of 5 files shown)


data/migrations/versions/0246c2d0e750_add_login_logout_logentrykind.py
ibazulic, line 1: We don't need a new table. We just need an event that would trigger on successful login and write itself down in the `logentry3` table. The counter of invalid login attempts already exists in the `user` table.

dmage:
E2E Registry Tests have failed, and it looks like a legit failure because of `endpoints/v1/index.py` changes.

harishsurf:
Replaced with #1866

PR #1825, opened 2023-04-13
chore: Ensure use of HTTP 1.1 when proxying storage (PROJQUAY-5140)

ibazulic:
We were not enforcing the use of `HTTP 1.1` when storage proxy was concerned. This causes problems in certain complex scenarios.


bcaton85 (approved):
LGTM

ibazulic:
@syed do you have any objections on this change? In regards to Quay.io, as far as I know we don't use `_storage_proxy` there at all so it should not matter.

LiZhang19817:
@dmage not seeing this fix was cherrypick to Quay 3.9, pls help confirm

PR #1832, opened 2023-04-25
storagereplication: add retry logic without exhausting queue retries (PROJQUAY-4793)

kleesc:
Allow the replication worker to retry immediately without having to wait and exhaust the queueitem's retries. This addresses transient unreliable storage service issues.


Sunandadadi (approved):
lgtm

PR #1834, opened 2023-04-25
build: disable push to backup repo for quayio-frontend (PROJQUAY-5229)

syed:
The backup ECR needs correct IAM rules for pushing. Disabling it until we get that


Sunandadadi (approved):
lgtm

PR #1839, opened 2023-04-26
API/UI: Filtering of tags API through query parameter (PROJQUAY-5362)

Sunandadadi:


data/model/oci/tag.py
```diff
@@ -159,6 +159,7 @@ def list_repository_tag_history(
     specific_tag_name=None,
     active_tags_only=False,
     since_time_ms=None,
+    query_param=None,
```
syed, line 162: can we call it `filter_tag_name`

endpoints/api/tag.py
```diff
@@ -68,6 +68,7 @@ class ListRepositoryTags(RepositoryParamResource):
     @disallow_for_app_repositories
     @parse_args()
     @query_param("specificTag", "Filters the tags to the specific tag.", type=str, default="")
+    @query_param("query", "Filters the tags that match the query", type=str, default="")
```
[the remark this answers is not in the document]
syed: another option is to have `?filter_name=<op>:<tagname>` where `op` can be `eq` or `like`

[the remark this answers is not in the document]
Sunandadadi: Added -  `?name_filter=<op>:<tagname>`

data/model/oci/tag.py
```diff
@@ -184,6 +185,13 @@ def list_repository_tag_history(
         .offset(page_size * (page - 1))
     )
 
+    if filter_tag_name is not None:
+        operation, value = filter_tag_name.split(':')
```
jonathankingfc, line 189: will this have a runtime error if the filter does not contain a ":" ?

jonathankingfc (changes requested):
implementation LGTM, just some tests need fixing

data/model/oci/tag.py
Sunandadadi: Good catch. Added try except to return 400: Bad response when syntax doesn't match.

jonathankingfc (dismissed):
LGTM

data/model/oci/tag.py
```diff
@@ -183,6 +184,16 @@ def list_repository_tag_history(
         .limit(page_size + 1)
         .offset(page_size * (page - 1))
     )
+    try:
+        if filter_tag_name is not None:
+            operation, value = filter_tag_name.split(':')
```
syed, line 189: consider using `split(':', 1)` so you only have 2 possible entries. if someone passes a `:` in the query then you'll have 3 entries in the split which will cause an exception here

endpoints/api/test/test_tag.py
```diff
@@ -112,3 +112,30 @@ def test_list_repo_tags(repo_namespace, repo_name, client, query_count, app):
...
+
+    with client_with_identity("devtable", client) as cl:
+        with assert_query_count(query_count):
+            params["filter_tag_name"] = "eq:prod"
+            tags = conduct_api_call(cl, ListRepositoryTags, "get", params).json["tags"]
+        assert len(tags) == 1
+
+    with client_with_identity("devtable", client) as cl:
+        params["filter_tag_name"] = "random"
+        resp = conduct_api_call(cl, ListRepositoryTags, "get", params, None, expected_code=400)
```
syed, line 141: Thanks for adding tests :)

endpoints/api/tag.py
```diff
@@ -68,6 +69,7 @@ class ListRepositoryTags(RepositoryParamResource):
     @disallow_for_app_repositories
     @parse_args()
     @query_param("specificTag", "Filters the tags to the specific tag.", type=str, default="")
+    @query_param("filter_tag_name", "Syntax: <op>:<name> Filters the tag names based on the operation", type=str, default="")
```
syed, line 72: can you add possible values for `op` are `like` and `eq`

data/model/oci/tag.py
```diff
@@ -183,6 +184,16 @@ def list_repository_tag_history(
...
+            if operation == 'like':
+                query = query.where(Tag.name.contains(value))
+            elif operation == 'eq':
+                query = query.where(Tag.name == value)
+    except ValueError:
+        raise ValueError("UnSupported syntax for filter_tag_name. Expected <operation>:<tag_name> where <operation>"
```
syed, line 195: captalization typo?

data/model/oci/tag.py
Sunandadadi: Nice one! I've updated to reflect the same.

data/model/oci/tag.py
Sunandadadi: Yes, fixed it.

syed (approved):
lgtm

PR #1840, opened 2023-04-27
task: Update pyopenssl and cryptography packages to newer versions (PROJQUAY-5393)

ibazulic:
Our `pyopenssl` and `cryptography` packages were very old.


dmage:
We cannot merge it as it'll break downstream. New cryptography needs Rust dependencies to be vendored downstream. See https://github.com/quay/quay/pull/1803 and other PRs linked to the ticket for details.

PR #1841, opened 2023-04-27
chore: deprecate image table support

kleesc:
Remove code related to deprecated image table


data/model/gc.py
```diff
@@ -302,19 +250,6 @@ def garbage_collect_repo(repo):
         _run_garbage_collection(context)
         had_changes = True
 
-    # TODO: Remove once we've removed the foreign key constraints from RepositoryTag and Image.
```
bcaton85, line 305: Have these constraints been already removed? Is this required?

bcaton85 (approved):
LGTM

PR #1843, opened 2023-04-27
ui: Refresh auth token for plugin flow on 401 (PROJQUAY-5390)

syed:
For standalone flow, redirect to /signin


Sunandadadi (approved):
lgtm

PR #1844, opened 2023-04-28
ui: use route location instead of window.location (PROJQUAY-5392)

syed:
window.location causes issues when running in beta or preview on console

### DIFF
--- a/web/src/routes/RepositoriesList/RepositoriesList.tsx
+++ b/web/src/routes/RepositoriesList/RepositoriesList.tsx
@@ -14,13 +14,13 @@ import {
   Tbody,
   Td,
 } from '@patternfly/react-table';
-import {useRecoilState, useRecoilValue} from 'recoil';
+import {useRecoilState} from 'recoil';
 import {IRepository} from 'src/resources/RepositoryResource';
 import {ReactElement, useState} from 'react';
 import {Link, useLocation} from 'react-router-dom';
 import CreateRepositoryModalTemplate from 'src/components/modals/CreateRepoModalTemplate';
 import {getRepoDetailPath} from 'src/routes/NavigationPath';
-import {selectedReposState, searchRepoState} from 'src/atoms/RepositoryState';
+import {selectedReposState} from 'src/atoms/RepositoryState';
 import {formatDate, formatSize} from 'src/libs/utils';
 import {BulkDeleteModalTemplate} from 'src/components/modals/BulkDeleteModalTemplate';
 import {RepositoryToolBar} from 'src/routes/RepositoriesList/RepositoryToolBar';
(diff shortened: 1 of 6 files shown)


web/src/routes/RepositoriesList/RepositoriesList.tsx
```diff
@@ -69,6 +69,9 @@ export default function RepositoriesList(props: RepositoriesListProps) {
   const [makePublicModalOpen, setmakePublicModal] = useState(false);
   const [makePrivateModalOpen, setmakePrivateModal] = useState(false);
   const [err, setErr] = useState<string[]>();
+  const location = useLocation();
+
+  console.log('current route location', location.pathname);
```
Sunandadadi, line 74: Can you remove the console.

web/src/routes/RepositoriesList/RepositoriesList.tsx
```diff
@@ -414,3 +429,7 @@ interface RepoListTableItem {
   size: number;
   last_modified: number;
 }
+
+interface RepositoriesListProps {
+  organizationName: string;
```
Sunandadadi, line 434: Is there a reason for adding the props here? I don't see the `organizationName` getting used in the PR.

web/src/routes/TagDetails/TagDetailsTabs.tsx
```diff
@@ -20,6 +20,7 @@ function getTabIndex(tab: string) {
 export default function TagTabs(props: TagTabsProps) {
   const [activeTabKey, setActiveTabKey] = useState<TabIndex>(TabIndex.Details);
   const navigate = useNavigate();
+  const location = useLocation();
```
Sunandadadi, line 23: `location` here doesn't seem to getting used in the PR.

web/src/routes/RepositoriesList/RepositoriesList.tsx
syed: I see `RepositoriesListProps` used in `src/routes/RepositoriesList/RepositoriesList.tsx:65` but it's not defined so I'm defining it

web/src/routes/RepositoriesList/RepositoriesList.tsx
syed: removed

web/src/routes/TagDetails/TagDetailsTabs.tsx
syed: it's used in `src/routes/TagDetails/TagDetailsTabs.tsx:35` but not defined in the context which causes issues

Sunandadadi (approved):
lgtm

PR #1853, opened 2023-05-01
ui: Adding option to permanently delete tags past time machine window (PROJQUAY-5303)

bcaton85:
Adds additional option in new UI `Permanently Delete` for individual alive tags where the tag's `lifetime_end_ms` will be set outside the time machine window.


Sunandadadi (approved):
lgtm

PR #1861, opened 2023-05-03
UI: Filtering security report vulnerabilities (PROJQUAY-5401)

Sunandadadi:


syed:
@Sunandadadi  there's also another search input in Packages list. Can you update that as well

syed (approved):
lgtm

PR #1864, opened 2023-05-04
UI: Adding functionality to update organization settings (PROJQUAY-5402)

Sunandadadi:


web/src/routes/OrganizationsList/Organization/Tabs/Settings/Settings.tsx
```diff
@@ -95,8 +153,14 @@ const GeneralSettings = (props: GeneralSettingsProps) => {
           justifyContent={{default: 'justifyContentFlexEnd'}}
           width={'100%'}
         >
-          <Button variant="primary">Save</Button>
-          <Button variant="link">Cancel</Button>
+          <Button
+            variant="primary"
+            type="submit"
+            onClick={onSave}
+            isDisabled={!checkForChanges()}
```
jonathankingfc, line 160: Small optimization: We may want to put this inside a useEffect that updates a state variable, otherwise this function will be called on every re-render

jonathankingfc (changes requested):

web/src/routes/OrganizationsList/Organization/Tabs/Settings/Settings.tsx
Sunandadadi: Hmm, we want to call it on every re-render to make sure the button is enabled/disabled based on the changes in the form.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/Settings.tsx
```diff
@@ -7,47 +7,158 @@ import {
...
+      else {
+        setValidated('error');
+        setError('Please enter a valid email address');
+      }
+    }
+  }
+
+  const checkForChanges = () => {
+    if (namespaceEmail != emailFormValue) {
+      if (validated == 'success') {
```
bcaton85, line 115: can be simplified to `return validated == 'success'`. Same with L122

web/src/hooks/UseOrganizationSettings.ts
```diff
@@ -0,0 +1,46 @@
+import {useState} from 'react';
+import {updateOrgSettings} from 'src/resources/OrganizationResource';
+import {useMutation, useQueryClient} from '@tanstack/react-query';
+
+export function useOrganizationSettings({name, onSuccess, onError}) {
+  const queryClient = useQueryClient();
+  const [namespace, setNamespace] = useState(name);
```
bcaton85, line 7: Seems like `setNamespace` is never called. Do we need this here?

web/src/routes/OrganizationsList/Organization/Tabs/Settings/Settings.tsx
```diff
@@ -7,47 +7,158 @@ import {
...
 import {useLocation} from 'react-router-dom';
 import {useCurrentUser} from 'src/hooks/UseCurrentUser';
 import {useOrganization} from 'src/hooks/UseOrganization';
+import {useOrganizationSettings} from 'src/hooks/UseOrganizationSettings';
+import {IOrganization} from 'src/resources/OrganizationResource';
+import {humanizeTimeForExpiry, getSeconds, isValidEmail} from 'src/libs/utils';
+import {addDisplayError} from 'src/resources/ErrorHandling';
+
+type validate = 'success' | 'warning' | 'error' | 'default';
+const timeMachineOptions = {
```
bcaton85, line 28: This is actually configurable from the backend via `TAG_EXPIRATION_OPTIONS`. You should be able to read in that value from the quay config

web/src/routes/OrganizationsList/Organization/Tabs/Settings/Settings.tsx
```diff
@@ -7,47 +7,158 @@ import {
...
+    },
+  });
+
+  const [alerts, setAlerts] = useState<ReactNode[]>([]);
 
   // Time Machine
-  const timeMachineOptions = ['1 week', '1 month', '1 year', 'Never'];
   const [timeMachineFormValue, setTimeMachineFormValue] = useState(
-    timeMachineOptions[0],
+    timeMachineOptions['0 s'],
```
bcaton85, line 66: `0s` is actually disabled on quay.io. This default should be read in from the quay config

web/src/routes/OrganizationsList/Organization/Tabs/Settings/Settings.tsx
```diff
@@ -7,47 +7,158 @@ import {
...
+        <FormAlert>
+          <Alert variant="danger" title={error} aria-live="polite" isInline />
+        </FormAlert>
+      )}
       <FormGroup
         isInline
         label="Organization"
         fieldId="form-organization"
-        helperText={'Orgnization names cannot be changed once set.'}
+        helperText={'Organization names cannot be changed once set.'}
```
bcaton85, line 161: May want to rename this to `namespace` since it could also be a user

bcaton85 (approved):
LGTM

PR #1866, opened 2023-05-05
logs: add audit log events for login/logout to Quay (PROJQUAY-2344)

dmesser:
Based on @harishsurf 's work in #1822 this is adding audit events for :

- login/ logout of quay (e.g. via the Web UI)
- login via registry docker v2/OCI protocol (e.g. via docker/podman CLI)  with metadata covering robots and app-specific tokens

It also adds visualization in the current log view UI. The behavior is on by default and can be controlled by the `ACTION_LOG_AUDIT_LOGINS` config flag.

Note: No call is made to docker registry API during a logout from docker/podman CLI, hence no audit events are logged
Testing: verified with a local instance and database audit logging

Screenshot 2023-05-14 at 22 33 03

Screenshot 2023-05-05 at 18 47 17

### DIFF
--- a/config.py
+++ b/config.py
@@ -436,6 +436,9 @@ class DefaultConfig(ImmutableConfig):
     LOG_ARCHIVE_LOCATION = "local_us"
     LOG_ARCHIVE_PATH = "logarchive/"
 
+    # Action logs configuration for advanced events
+    ACTION_LOG_AUDIT_LOGINS = True
+
     # Action logs archive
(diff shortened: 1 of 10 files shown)


harishsurf (dismissed):
lgtm! the visualization is a nice touch :)

harishsurf (approved):

dmage (changes requested):
`amd64` is failing, `make update-testdata` is required

dmage (approved):

PR #1870, opened 2023-05-09
marketplace: fix path to api cert (PROJQUAY-5409)

Marcusk19:
remove `/auth` from path to marketplace cert
secrets are mounted to `/conf/stack` by default


kleesc (approved):
lgtm

PR #1876, opened 2023-05-10
superuser: paginating superuser organization list (PROJQUAY-4297)

bcaton85:
Paginates calls to the `/v1/superuser/organizations/` endpoint. Allows the superuser organization list to load when quota is enabled.

### DIFF
--- a/endpoints/api/superuser.py
+++ b/endpoints/api/superuser.py
@@ -223,9 +223,11 @@ class SuperUserOrganizationList(ApiResource):
         if SuperUserPermission().can():
             if parsed_args["all"] and parsed_args["limit"] > 100:
                 raise InvalidRequest("Page limit cannot be above 100")
-            
+
             if parsed_args["all"]:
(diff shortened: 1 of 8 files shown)


endpoints/api/superuser.py
```diff
@@ -204,13 +204,32 @@ class SuperUserOrganizationList(ApiResource):
     @require_fresh_login
     @verify_not_prod
     @nickname("listAllOrganizations")
+    @parse_args()
+    @query_param(
+        "all",
+        "Returns all organizations within the registry with no pagination",
+        type=truthy_bool,
+        default=True,
+    )
```
kleesc, line 213: I don't think we need this anymore. The caller should just go through every pages until no token is returned.

bcaton85: I'm thinking we may want this for backwards compatibility. Is it possible user's have written automation that expect all orgs to be returned?

bcaton85: updated

kleesc (approved):

PR #1877, opened 2023-05-10
api: allow manifests to be pulled by digest (PROJQUAY-5467)

Marcusk19:
Prevent manifests pushed by digest from being hidden. Lets newer versions of conftest push to quay, since it expects to be able to pull the policy by digest when performing a push.

### DIFF
--- a/endpoints/v2/manifest.py
+++ b/endpoints/v2/manifest.py
@@ -142,7 +142,9 @@ def fetch_manifest_by_digest(namespace_name, repo_name, manifest_ref, registry_m
 
     try:
         manifest = registry_model.lookup_manifest_by_digest(
-            repository_ref, manifest_ref, raise_on_error=True
+            repository_ref,
+            manifest_ref,
+            raise_on_error=True,
         )
     except ManifestDoesNotExist as e:
         image_pulls.labels("v2", "manifest", 404).inc()
@@ -284,7 +286,7 @@ def write_manifest_by_digest(namespace_name, repo_name, manifest_ref):
 
     expiration_sec = app.config["PUSH_TEMP_TAG_EXPIRATION_SEC"]
     manifest = registry_model.create_manifest_with_temp_tag(
-        repository_ref, parsed, expiration_sec, storage
+        repository_ref, parsed, expiration_sec, storage, hidden=False
     )
 
     if manifest is None:
--- a/endpoints/v2/manifest.py
+++ b/endpoints/v2/manifest.py
@@ -286,7 +286,7 @@ def write_manifest_by_digest(namespace_name, repo_name, manifest_ref):
 
     expiration_sec = app.config["PUSH_TEMP_TAG_EXPIRATION_SEC"]
     manifest = registry_model.create_manifest_with_temp_tag(
-        repository_ref, parsed, expiration_sec, storage, hidden=False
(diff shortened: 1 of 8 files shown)


flavianmissi:
The need behind this change is unclear in this PR's description - would you mind expanding a little more on that @Marcusk19 ?

endpoints/v2/manifest.py
```diff
@@ -142,7 +142,11 @@ def fetch_manifest_by_digest(namespace_name, repo_name, manifest_ref, registry_m
 
     try:
         manifest = registry_model.lookup_manifest_by_digest(
-            repository_ref, manifest_ref, raise_on_error=True
+            repository_ref,
+            manifest_ref,
+            raise_on_error=True,
+            allow_dead=True,
+            require_available=True,
```
dmage, line 149: allow_dead isn't exactly what you need. It'll return expired manifests, which is incorrect. Quay should always pretend that they are deleted.

Why do you need `require_available`? What happens if `require_available` isn't set?

dmage (changes requested):
We need tests that confirm that this feature works as expected and Quay doesn't accidentally distribute expired images.

Marcusk19:
@flavianmissi Yeah, this is supposed to address an error you get when [pushing an OPA policy to quay with conftest](https://www.conftest.dev/sharing/#pushing-to-an-oci-registry):
```
Error: push bundle: tagging: 
sha256:fae4613127bc35df711682c11e462dd9db2d2716dca45553c00669457eda91b1: not found 
```
It looks like this error occurs because manifests pushed by their digest are hidden by default, and conftest expects to be able to get that manifest to verify that it exists. This check was introduced in newer versions of conftest, which is why we didn't run into this issue in the past.

Marcusk19:
@dmage Would the already existing registry tests cover this? It looks like there's [test_basic_push_by_manifest_digest](https://github.com/quay/quay/blob/d49dbd1515b3a17d21127b881461d333045552a5/test/registry/registry_tests.py#L204) that can be used to test pulling by digest and [test_attempt_pull_by_manifest_digest_for_deleted_tag](https://github.com/quay/quay/blob/d49dbd1515b3a17d21127b881461d333045552a5/test/registry/registry_tests.py#L2834) for testing expired images aren't returned

flavianmissi:
good job! lgtm but let's wait for others' feedback before landing this

dmage:
It makes temporary tags for manifests pushed by digest visible in UI.

Screenshot 2023-05-24 at 12 41 20

flavianmissi:
> @dmage I saw the discussion on slack mentioning using the digest to tag manifests pushed by digest. Should we do that instead of using these temp tags?

I like this solution.

@Marcusk19 I see you added a new flag for tagging by digest, is there a reason to not ditch the $temp-.* stuff completely? Will we break anything by doing that? 🤔

flavianmissi:
Unrelatedly, it looks like some of your tests are failing, please investigate if the failure is related to your change as it might affect the solution 🙏🏼

Marcusk19:
@flavianmissi I think we're going back to the original approach, but instead of `allow_dead=True` a new boolean called `allow_hidden` is used to return the hidden but not expired tags. Should prevent the UI from becoming noisy with all the temp tags.

dmage (approved):

flavianmissi (approved):

PR #1878, opened 2023-05-11
logs: Add SSL cert support and test coverage for splunk logging (PROJQUAY-4993)

harishsurf:
This PR adds SSL cert support or splunk and additional test coverage

### DIFF
--- a/data/logs_model/logs_producer/splunk_logs_producer.py
+++ b/data/logs_model/logs_producer/splunk_logs_producer.py
@@ -1,4 +1,5 @@
 import logging
+import ssl
 
 from splunklib import client
 
@@ -22,19 +23,45 @@ class SplunkLogsProducer(LogProducerInterface):
         url_scheme="https",
         verify_ssl=True,
         index_prefix=None,
+        ssl_ca_path=None,
     ):
-        try:
-            service = client.connect(
-                host=host, port=port, token=bearer_token, scheme=url_scheme, verify=verify_ssl
-            )
-        except Exception as ex:
-            logger.exception("Failed to connect to Splunk instance %s", ex)
-            raise ex
+        connect_args = {
+            "host": host,
+            "port": port,
+            "token": bearer_token,
+            "scheme": url_scheme,
+            "verify": verify_ssl,
+        }
+        if ssl_ca_path:
+            # Create an SSLContext object for cert validation
+            context = ssl.create_default_context()
+            context.check_hostname = True
+            context.verify_mode = ssl.CERT_REQUIRED
+            if not verify_ssl:
+                context.check_hostname = False
+                context.verify_mode = ssl.CERT_NONE
+            try:
+                context.load_verify_locations(cafile=ssl_ca_path)
+            except ssl.SSLError as e:
+                raise Exception("SSL cert is not valid %s" % e)
+            except FileNotFoundError as e:
+                raise Exception("Path to cert file is not valid %s" % e)
+            connect_args["context"] = context
+        elif verify_ssl:
+            raise ValueError("SSL cert path needs to be specified")
+
+        # establish connection with splunk
+        service = client.connect(**connect_args)
         try:
             self.index = service.indexes[index_prefix]
-            logger.info("splunk index %s", self.index)
+            logger.info("using existing splunk index %s", self.index)
         except KeyError:
             self.index = service.indexes.create(index_prefix)
+            logger.info("Created new splunk index %s", self.index)
+        except ConnectionRefusedError as e:
+            raise Exception(
+                "Connection to Splunk refused, check if Splunk instance is available: %s" % e
+            )
 
     def send(self, log):
         try:
--- a/data/logs_model/splunk_logs_model.py
+++ b/data/logs_model/splunk_logs_model.py
@@ -46,7 +46,7 @@ class SplunkLogsModel(SharedModel, ActionLogsDataInterface):
             return
 
         if repository_name is not None:
-            if repository is None or namespace_name is not None:
+            if repository is not None or namespace_name is None:
                 raise ValueError(
                     "Incorrect argument provided when logging action logs, namespace name should not be "
                     "empty"
@@ -56,30 +56,28 @@ class SplunkLogsModel(SharedModel, ActionLogsDataInterface):
         if timestamp is None:
             timestamp = datetime.today()
 
-        account_id = None
-        performer_id = None
-        repository_id = None
+        username = None
+        performer_name = None
+        repo_name = None
 
         if namespace_name is not None:
             ns_user = model.user.get_namespace_user(namespace_name)
             if ns_user is not None:
-                account_id = ns_user.id
+                username = ns_user.username
 
-        if performer is not None:
-            performer_id = performer.id
+        if performer is not None and performer.username:
+            performer_name = performer.username
 
-        if repository is not None:
-            repository_id = repository.id
-
-        kind_id = model.log._get_log_entry_kind(kind_name)
+        if repository is not None and repository.name:
+            repo_name = repository.name
 
         metadata_json = metadata or {}
 
         log_data = {
-            "kind": kind_id,
-            "account": account_id,
-            "performer": performer_id,
-            "repository": repository_id,
+            "kind": kind_name,
+            "account": username,
+            "performer": performer_name,
+            "repository": repo_name,
             "ip": ip,
             "metadata_json": metadata_json or {},
             "datetime": timestamp,
--- a/data/logs_model/test/test_splunk.py
+++ b/data/logs_model/test/test_splunk.py
@@ -1,4 +1,9 @@
-from unittest.mock import MagicMock
+import os
+import ssl
+import tempfile
+from ssl import SSLError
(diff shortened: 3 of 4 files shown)


data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -22,11 +23,33 @@ def __init__(
...
+        ssl_ca_path=None,
     ):
+        connect_args = {
+            "host": host,
+            "port": port,
+            "token": bearer_token,
+            "scheme": url_scheme,
+            "verify": verify_ssl,
+        }
+        if verify_ssl:
```
flavianmissi, line 35: I think you need to create the context if `ssl_ca_path` is set (to something other than `""`), independent of `verify_ssl`.
Users who will configure a custom CA will usually want SSL verification, but in theory a user could set a custom CA and still skip SSL verification. In other words, the SSL context creation should happen in connection to the user telling Quay to use a custom CA when talking to Splunk.

Furthermore, I think you want to investigate the meaning of [`context.verify_mode`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.verify_mode). From what I gather, `verify_ssl = False` is equivalent to `context.verify_mode = CERT_NONE`, and `verify_ssl = True` is equivalent to `context.verify_mode = ssl.CERT_REQUIRED`.

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -22,11 +23,33 @@ def __init__(
...
+            connect_args["context"] = context
+        else:
+            connect_args["verify"] = False
+
         try:
-            service = client.connect(
-                host=host, port=port, token=bearer_token, scheme=url_scheme, verify=verify_ssl
-            )
+            service = client.connect(**connect_args)
+        except TypeError as ex:
```
flavianmissi, line 47: when should we expect a `TypeError`? 🤔

flavianmissi (commented):
Haven't gone through the tests yet - it'll be next week's work.

data/logs_model/logs_producer/splunk_logs_producer.py
flavianmissi: Interesting. In that case, maybe we should also treat the `FileNotFoundError` here?
I think this syntax should work but please double check:
```
except TypeError, FileNotFoundError as ex:
```

Also, a nitpick: in python we usually call an exception `e` (instead of `ex`), for example:
```
except TypeError as e:
```

data/logs_model/splunk_logs_model.py
```diff
@@ -46,7 +46,7 @@ def log_action(
             return
 
         if repository_name is not None:
-            if repository is None or namespace_name is not None:
+            if repository is not None or namespace_name is None:
```
flavianmissi, line 49: is this a bugfix? Just curious 🤔

data/logs_model/test/test_splunk.py
```diff
@@ -28,29 +53,161 @@ def splunk_logs_model_config():
...
+
+    # Clean up the temporary file
+    os.unlink(cert_path)
+
+
+@pytest.mark.parametrize(
+    """
+  kind_name, namespace_name,
+  performer, ip, metadata, repository, repository_name, timestamp, throws
+  """,
```
flavianmissi, line 80: nit: please use the same amount of spaces on every line. the ones I selected use 2 while the rest uses 4

data/logs_model/test/test_splunk.py
```diff
@@ -28,29 +53,161 @@ def splunk_logs_model_config():
...
+        throws,
+        logs_model,
+        splunk_logs_model_config,
+        mock_db_model,
+        initialized_db,
+        cert_file_path,
+):
+    with patch(
+            "data.logs_model.logs_producer.splunk_logs_producer.SplunkLogsProducer.send"
+    ) as mock_send, patch("splunklib.client.connect", MagicMock()):
```
flavianmissi, line 125: how is this:
```
patch("splunklib.client.connect", MagicMock())
```
different than this:
```
patch("splunklib.client.connect")
```
?

flavianmissi (commented):
Nicely done with the tests!
We're close to getting this in, just need to look into that verify thing I commented about - let me know if you need help!

data/logs_model/splunk_logs_model.py
harishsurf: I had incorrect conditional before. I'm following what elastic search does [here](https://github.com/quay/quay/blob/74d6d82748df7fc9b5dcb5f1e712fd3be1b7b38f/data/logs_model/document_logs_model.py#L479-L480)

data/logs_model/test/test_splunk.py
harishsurf: nothing, replaced with `patch("splunklib.client.connect")`

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -22,19 +23,50 @@ def __init__(
...
+                context.check_hostname = False
+                context.verify_mode = ssl.CERT_NONE
+            try:
+                context.load_cert_chain(certfile=ssl_ca_path)
+            except ssl.SSLError as e:
+                logger.exception("SSL certificate error %s", e)
+                raise e
+            except FileNotFoundError as e:
+                logger.exception("Path to cert file is not valid %s", e)
+                raise e
```
dmage, line 50: Usually it's a bad idea to do both: log an error and propagate it to the caller. How does the caller handle these errors?

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -22,19 +23,50 @@ def __init__(
...
+
         try:
             self.index = service.indexes[index_prefix]
             logger.info("splunk index %s", self.index)
         except KeyError:
             self.index = service.indexes.create(index_prefix)
+        except ConnectionRefusedError as e:
+            raise Exception(
+                "Connection to Splunk refused, check if Splunk instance is available %s", e
+            )
```
dmage, line 69: You replace a detailed exception with a generic one, that caller won't be able to distinguish it from any other error. Why do you want to do it?

util/config/schema.py
```diff
@@ -1183,6 +1183,12 @@
                             "description": "Splunk's index prefix",
                             "x-example": "splunk_logentry_",
                         },
+                        "ssl_ca_path": {
+                            "type": "string",
+                            "description": "*Absolute container path* to a single .pem file containing the private "
```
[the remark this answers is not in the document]
harishsurf: I am not 100% familiar with what user would send as part of SSL, but from the description of [SSLContext.load_cert_chain](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain), the `certfile` param allows private key to be bundled with the certificate. 
Is user sending a cert without a private key a valid use case? - If so, then I will update the description to say user can _optionally_ include the key

util/config/schema.py
```diff
@@ -1183,6 +1183,12 @@
...
+                            "key followed by a CA certificate bundled together for SSL verification",
+                            "x-example": "conf/stack/ssl-ca-cert.pem",
```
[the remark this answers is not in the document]
harishsurf: `conf/stack/ssl-ca-cert.pem` works, changing it to relative path

util/config/schema.py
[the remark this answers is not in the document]
dmage: Clients should use `load_verify_locations` when they want to load a bundle of trusted CAs. Private keys for a CA are never available to clients.

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -22,19 +23,47 @@ def __init__(
...
-        except Exception as ex:
-            logger.exception("Failed to connect to Splunk instance %s", ex)
-            raise ex
+            service = client.connect(**connect_args)
+        except Exception as e:
+            raise Exception("Failed to connect to Splunk instance %s" % e)
+
         try:
             self.index = service.indexes[index_prefix]
             logger.info("splunk index %s", self.index)
```
dmage, line 60: Is it for debugging?

dmage (dismissed):

dmage (dismissed):

data/logs_model/logs_producer/splunk_logs_producer.py
harishsurf: yes, removed it

dmage (approved):

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -22,19 +23,45 @@ def __init__(
         url_scheme="https",
         verify_ssl=True,
         index_prefix=None,
+        ssl_ca_path=None,
```
LiZhang19817, line 26: What's the path of this parameter, could you pls give an example?

harishsurf: `ssl_ca_path` is not mandatory but is required along with `verify_ssl: True`. You would set it to the relative container path where your cert is stored. Ex: `conf/stack/<cert-name>`

PR #1879, opened 2023-05-11
quota: calculating registry size (PROJQUAY-5476)

bcaton85:
Allows superusers to trigger a calculation of the deduplicated registry size. A superuser can go to the organization panel of the superuser page and select `Calculate` to queue a calculation of the registry total. The total will only be calculated when requested. Includes warning to user of increase of database load when running calculation.


data/database.py
```diff
@@ -1012,6 +1012,13 @@ class QuotaRepositorySize(BaseModel):
     backfill_complete = BooleanField(null=False, default=False)
 
 
+class QuotaRegistrySize(BaseModel):
+    size_bytes = BigIntegerField(null=False, default=0)
+    running = BooleanField(null=False, default=False)
+    queued = BooleanField(null=False, default=False)
```
kleesc, line 1018: Why do you need a `queued` state if  there's already a `running` state? i.e what's the difference between use cases between the two.

kleesc (commented):
.

data/database.py
```diff
@@ -1012,6 +1012,13 @@ class QuotaRepositorySize(BaseModel):
...
+    completed_ms = BigIntegerField(null=True)
```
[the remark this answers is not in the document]
bcaton85: the timestamp of when it was completed

data/database.py
kleesc: Couldn't you do that with only one state, with either `running` set or not?

bcaton85: well if running is false it doesn't mean it should be ran. There needs to be a field saying whether the calculation is currently running or not so no other job picks it up and another field saying that the calculation is queued to be ran. If both are false nothing should happen. So really there's 3 states - do nothing, queued, running. I don't think we can represent that as a single boolean.

bcaton85: this isn't a queue worker, it's a regular worker that periodically checks the table. This way when attempting to queue if it's already been queued it's just a single field check whether `queued == true`.

kleesc (approved):

PR #1880, opened 2023-05-11
chore: Delete outdated k8 manifests (PROJQUAY-5490)

crozzy:
The manifests contained in this dir are very outdated and not maintained, this has led to some confusion. As there are other examples of how to deploy Quay, this feels more of a liability.


ibazulic (approved):
LGTM.

PR #1887, opened 2023-05-16
api: add permanently delete tag usage log (PROJQUAY-5496)

bcaton85:
Now logging the permanently delete tag as a usage log.


kleesc (approved):
lgtm

PR #1888, opened 2023-05-16
chore: Remove Docker Build jobs (PROJQUAY-5519)

dmage:


bcaton85:
Is there any negative impact to removing these tests? I suppose we would know about it at some point. Also shouldn't `Docker Build` be in the CI tests for this PR?

dmage:
@bcaton85 `Docker Build` will be replaced by `Cypress Tests`, once this PR is merged - `Cypress Tests` will become required, and `Docker Build amd64` will be removed completely.

amd64 builds are verified twice anyway - by OCI Distribution tests and by Cypress tests.
Other platforms will be verified by the post-merge job that build `3.9-unstable` for amd64/arm64/s390x/ppc64le.

bcaton85 (approved):
LGTM

dmage:
Negative impact - we can accidentally break arm64/s390x/ppc64le, and we'll know about it only after the change is merged. Good thing - we'll be able to use our own builders, so these builds should become more reliable.

PR #1891, opened 2023-05-18
marketplace: fixing allowed repo count (PROJQUAY-5513)

Marcusk19:
* Allowed number of private repos should be a sum of stripe plans and marketplace subscriptions.

* Make sure subscriptions are not expired when fetching from marketplace.

* Check subscription returned from stripe api is non-null value when performing reconciliation.

### DIFF
--- a/endpoints/api/billing.py
+++ b/endpoints/api/billing.py
@@ -50,9 +50,9 @@ def check_internal_api_for_subscription(namespace_user):
     """
     user_account_number = rh_user_api.get_account_number(namespace_user)
     if user_account_number:
-        user_subscription = rh_marketplace_api.find_stripe_subscription(user_account_number)
-        return user_subscription
-    return None
+        user_subscriptions = rh_marketplace_api.find_stripe_subscription(user_account_number)
+        return user_subscriptions
+    return []
 
 
 def get_namespace_plan(namespace):
@@ -64,10 +64,6 @@ def get_namespace_plan(namespace):
         return None
 
     if not namespace_user.stripe_id:
-        # if entitlement reconciliation is enabled
-        # we want to check marketplace too
-        if features.RH_MARKETPLACE:
-            return check_internal_api_for_subscription(namespace_user)
         return None
 
     # Ask Stripe for the subscribed plan.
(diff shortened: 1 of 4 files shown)


endpoints/api/billing.py
```diff
@@ -86,14 +83,22 @@ def lookup_allowed_private_repos(namespace):
...
+    repos_allowed = 0
     current_plan = get_namespace_plan(namespace)
-    if current_plan is None:
-        return False
+
+    if features.RH_MARKETPLACE:
+        namespace_user = model.user.get_namespace_user(namespace)
+        marketplace_subscriptions = check_internal_api_for_subscription(namespace_user)
+        for subscription in marketplace_subscriptions:
+            repos_allowed = repos_allowed + subscription["privateRepos"]
```
kleesc, line 93: nit: can you use `+=`

endpoints/api/user.py
```diff
@@ -642,11 +642,15 @@ def get(self):
...
             user_account_number = rh_user_api.get_account_number(user)
-            user_subscription = rh_marketplace_api.find_stripe_subscription(user_account_number)
-            if user_subscription:
-                repos_allowed = user_subscription["privateRepos"]
+            if user_account_number:
+                marketplace_subscriptions = rh_marketplace_api.find_stripe_subscription(
+                    user_account_number
+                )
+                for user_subscription in marketplace_subscriptions:
+                    repos_allowed = user_subscription["privateRepos"] + repos_allowed
```
kleesc, line 653: nit: can you use `+=`

util/marketplace.py
```diff
@@ -102,7 +105,13 @@ def lookup_subscription(self, ebsAccountNumber, skuId):
         except json.decoder.JSONDecodeError:
             return None
 
-        return subscription
+        if subscription:
+            end_date = subscription["effectiveEndDate"]
+            now = time.time() * MILLISECONDS
```
kleesc, line 110: nit: instead of MILLISECONDS, i would indicate it in the variable's name: `now_ms = time.time() * 1000`.
Also, do you need to round the value.

util/marketplace.py
```diff
@@ -160,11 +169,12 @@ def create_entitlement(self, customerId, sku):
...
         """
-        Returns the stripe plan for a given account number
+        Returns the stripe plan/s for a given account number
         """
+        stripe_plans = []
         for sku in RH_SKUS:
             user_subscription = self.lookup_subscription(account_number, sku)
             if user_subscription is not None:
-                return get_plan_using_rh_sku(sku)
+                stripe_plans.append(get_plan_using_rh_sku(sku))
```
kleesc, line 178: Do these plans include RH subscriptions? Because a user should only ever have a single _Stripe_ plan.

util/marketplace.py
Marcusk19: "effectiveEndDate" is in milliseconds, which is why I rounded it here.

kleesc (approved):
lgtm

PR #1894, opened 2023-05-19
quota: moving resetting of child manifest temporary tags to delete endpoint (PROJQUAY-5512)

bcaton85:
The tagged JIRA contains most of the details. Moves the resetting of child manifest temporary tags to happen on deletion instead of on push/pull. Resetting child manifest temporary tags caused issues in other portions of the code like proxy cache where temporary tags were deleted too early.


dmage (approved):

dmage:
@bcaton85 there is a failing cypress test, apparently it's flaky.

PR #1897, opened 2023-05-19
quota: removing QUOTA_TOTAL_DELAY_SECONDS option (PROJQUAY-5503)

bcaton85:
Initially introduced as a solution to the race conditions of the running total on first deployment. It delays the backfill until the deployment has finished to prevent the running total from running at the same time as the deployment. Has the potential to be incorrectly used and result in incorrect totals if set to low.

### DIFF
--- a/config.py
+++ b/config.py
@@ -795,10 +795,6 @@ class DefaultConfig(ImmutableConfig):
     FEATURE_QUOTA_MANAGEMENT = False
     # default value for all organizations to reject by default. 0 = no configuration
     DEFAULT_SYSTEM_REJECT_QUOTA_BYTES = 0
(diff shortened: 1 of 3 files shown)


dmage (approved):
What is the new solution?

bcaton85:
The new solution is to deploy first with `FEATURE_QUOTA_MANAGEMENT: true, QUOTA_BACKFILL: false`. During this time all totals will be set to 0. Then deploy again with `FEATURE_QUOTA_MANAGEMENT: true, QUOTA_BACKFILL: true` which will run the backfill which enables the running total.

kleesc (approved):

PR #1903, opened 2023-05-23
logs: add failure logging for login, push, pull and delete events (PROJQUAY-5411)

dmesser:
This adds the ability to log the following events:

- failed logins via Oauth tokens, Robot tokens, app-specific tokens and regular user accounts (due wrong credentials, expiry, disabled accounts etc)
- via metadata login events are differentiated between logins to the Quay UI/API or to the OCI/Docker v2 API 
- failed pushes due to insufficient privileges or non-existent orgs or repositories
- failed pulls due to insufficient privileges or non-existent orgs and repos

There are new config tunables implemented to control whether any of these audit events are logged (`ACTION_LOG_AUDIT_LOGIN_FAILURES`, `ACTION_LOG_AUDIT_PULL_FAILURES`, `ACTION_LOG_AUDIT_PUSH_FAILURES` and `ACTION_LOG_AUDIT_DELETE_FAILURES`), the default is false. For audit-ability as much information of the request as possible is logged and stored in event metadata.
Failed pushes create multiple audit log entries due to the way the docker protocol works (blobs are attempted to be pushed first before the manifest).
Also all unit tests are run by default with extended auditing of login and failure events to increase code coverage.

Screenshot 2023-05-14 at 22 22 20
Screenshot 2023-05-14 at 21 58 52
Screenshot 2023-05-14 at 21 52 17
Screenshot 2023-05-14 at 21 51 52


bcaton85:
@dmesser I can start reviewing this, can you rebase this onto master? The alembic upgrade is most likely out of order now

bcaton85 (commented):
LGTM but i think some of the comments need to be addressed. Also the migration has been updated in master so it will have to be updated in this branch.

data/model/user.py
```diff
@@ -403,7 +404,8 @@ def get_robot_and_metadata(robot_shortname, parent):
 
 def lookup_robot(robot_username):
     try:
-        robot_username.encode("ascii")
+        if robot_username and isinstance(robot_username, str):
+            robot_username.encode("ascii")
```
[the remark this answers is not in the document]
dmesser: There is a specific unit test where swedish letters are used as a user name and that fails to store in the database.

endpoints/api/__init__.py
```diff
@@ -401,6 +401,87 @@ def wrapped(self, *args, **kwargs):
 require_user_admin = require_user_permission(UserAdminPermission, scopes.ADMIN_USER)
 
 
+def log_unauthorized(audit_event):
```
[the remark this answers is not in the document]
dmesser: Yeah, I think we definitely would give guidance that you would not enable this in a large, or even public registry instance.

config.py
```diff
@@ -441,6 +441,11 @@ def create_transaction(db):
     # Action logs configuration for advanced events
     ACTION_LOG_AUDIT_LOGINS = True
 
+    # Action logs configuration for failure tracking
+    ACTION_LOG_AUDIT_LOGIN_FAILURES = False
+    ACTION_LOG_AUDIT_PULL_FAILURES = False
+    ACTION_LOG_AUDIT_PUSH_FAILURES = False
```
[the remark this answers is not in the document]
dmesser: yes, good catch!

bcaton85 (approved):
LGTM

data/model/user.py
```diff
@@ -789,34 +794,69 @@ def validate_reset_code(token):
 
 
 def find_user_by_email(email):
+    # Make sure we didn't get any unicode for the username.
+    try:
+        if email and isinstance(email, str):
+            email.encode("ascii")
+    except UnicodeEncodeError:
+        return None
```
kleesc, line 802: Why do we need to do this conversion? We shouldn't be allowing non ascii characters in the first place.
You end up dealing with a n instance of bytes as opposed to a string after this encoding.

auth/credentials.py
```diff
@@ -83,10 +120,90 @@ def validate_credentials(auth_username, auth_password_or_token):
         logger.debug("Found credentials header for robot %s", auth_username)
         try:
             robot = model.user.verify_robot(auth_username, auth_password_or_token)
+
+            robot_user, robot_name = parse_robot_username(auth_username)
+
+            owner = model.user.get_nonrobot_user(robot_user)
```
kleesc, line 126: The same calls are already being made when verifying the username/pass. 
And these will essentially result in significantly more DB calls.

data/model/user.py
dmesser: IIRC, there was a unit test testing a user with Swedish characters in their username and email which was failing without this.

data/model/user.py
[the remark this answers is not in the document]
dmesser: actually, I reverted this, I think you are right

auth/credentials.py
dmesser: Refactored the implementation to be based on specific exception subclasses. Except in one case I now also avoid an additional DB lookup when logging the audit event about the auth failure.

data/model/user.py
dmesser: Found it, it was the `test_signin_unicode` in `test_api_usage.py` that tried to login with Unicode characters. In endpoints/api/user.py I am using `model.user.find_user_by_email()` there to attempt to detect the performing user. This tests fails when the registry unit tests are run on a MySQL database. The call to `find_user_by_email(()` wasn't there before but `get_nonrobot_user()` in the user model has the same check, most likely for the same reason.

auth/credentials.py
```diff
@@ -83,10 +120,100 @@ def validate_credentials(auth_username, auth_password_or_token):
...
+        except model.DeactivatedRobotOwnerException as dre:
+            robot_owner, robot_name = parse_robot_username(auth_username)
+
+            try:
+                performer = model.user.lookup_robot(auth_username)
+            except User.DoesNotExist:
+                performer = None
+
+            logger.debug(
+                "Tried to use the robot %s for a disabled user: %s", (robot_name, robot_owner)
```
bcaton85, line 136: Parens need to be removed here. It's passing in a single tuple for the first placement and nothing for the second.

auth/credentials.py
```diff
@@ -83,10 +120,100 @@ def validate_credentials(auth_username, auth_password_or_token):
...
+                ValidateResult(AuthKind.credentials, error_message=str(dre)),
+                CredentialKind.robot,
+            )
+        except model.InvalidRobotCredentialException as ire:
             logger.warning("Failed to validate credentials for robot %s: %s", auth_username, ire)
+
+            robot_owner, _ = parse_robot_username(auth_username)
+
+            try:
+                performer = model.user.lookup_robot(auth_username)
```
bcaton85, line 164: Is this adding a DB call on failed logins that's not behind a feature flag? Same for the other DB calls in the other exception cases.

bcaton85: it looks like they should be in the `if app.config.get("ACTION_LOG_AUDIT_LOGIN_FAILURES"):` block

dmesser:
@bcaton85, thanks for the comments. The last commit should address them all and bring the logging levels in sync. Sorry for the mess!

bcaton85 (approved):
LGTM - @kleesc are there any other changes you want to see?

auth/test/test_basic.py
```diff
@@ -13,6 +12,7 @@
 )
 from auth.validateresult import AuthKind, ValidateResult
 from data import model
+from test.fixtures import *
```
kleesc, line 15: Is this extra import used?

test/testconfig.py
```diff
@@ -116,5 +116,10 @@ def create_transaction(db):
     FEATURE_RH_MARKETPLACE = True
 
     FEATURE_AUTO_PRUNE = True
+    ACTION_LOG_AUDIT_LOGINS = True
```
kleesc, line 119: Do you need both `ACTION_LOG_AUDIT_LOGINS` and `ACTION_LOG_AUDIT_LOGIN_FAILURES`?

dmesser: Not necessarily but I was missing this from https://github.com/quay/quay/pull/1866/ and thought it would be good to have it enabled during testing to catch any kind of side effects, even though our current unit tests are explicitly testing for audit events.

dmesser: Not necessarily but I was missing this from https://github.com/quay/quay/pull/1866 and thought it would be good to have it enabled during testing to catch any kind of side effects, even though our current unit tests are explicitly testing for audit events.

auth/test/test_basic.py
dmesser: Yes, otherwise the test will fail with errors about fixtures not being found. Other tests in this directories import the fixtures the same way.

bcaton85 (approved):
LGTM

PR #1905, opened 2023-05-24
authentication(LDAP): allow LDAP referrals to not be followed (PROJQUAY-5291) 

michaelalang:
LDAP referrals grant LDAP servers to distribute load and or migrate parts of the directory to another LDAP Server. 
The LDAP protocol does not specify how authentication is passed onwards and by default an `anonymous bind` is send. 
Depending on the configuration of the referral LDAP Server we see either an Authentication exception or nothing is returned (ACI based)

The patch provides the possibility to set, if we want to follow referrals or not by specifying the config.yaml variable: 

- LDAP_FOLLOW_REFERRALS=0  # do not follow
- LDAP_FOLLOW_REFERRALS=1 # to follow (default)

with the default set to what was hard coded, this patch is not expected to break any behavior

### DIFF
--- a/data/users/__init__.py
+++ b/data/users/__init__.py
@@ -67,6 +67,7 @@ def get_users_handler(config, _, override_config_dir):
         ldap_user_filter = config.get("LDAP_USER_FILTER", None)
         ldap_superuser_filter = config.get("LDAP_SUPERUSER_FILTER", None)
         ldap_restricted_user_filter = config.get("LDAP_RESTRICTED_USER_FILTER", None)
+        ldap_referrals = int(config.get("LDAP_FOLLOW_REFERRALS", True))
 
         allow_tls_fallback = config.get("LDAP_ALLOW_INSECURE_FALLBACK", False)
         return LDAPUsers(
@@ -85,6 +86,7 @@ def get_users_handler(config, _, override_config_dir):
             ldap_user_filter=ldap_user_filter,
             ldap_superuser_filter=ldap_superuser_filter,
             ldap_restricted_user_filter=ldap_restricted_user_filter,
+            ldap_referrals=ldap_referrals,
         )
 
     if authentication_type == "JWT":
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -15,6 +15,8 @@ _DEFAULT_NETWORK_TIMEOUT = 10.0  # seconds
 _DEFAULT_TIMEOUT = 10.0  # seconds
 _DEFAULT_PAGE_SIZE = 1000
 
+# setting env FEATURE_FOLLOW_REFERRALS=0 disabled referral lookups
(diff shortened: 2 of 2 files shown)


data/users/externalldap.py
```diff
@@ -33,14 +35,15 @@ def __init__(
         self._timeout = timeout
         self._network_timeout = network_timeout
 
-    def get_connection(self):
+    def get_connection(self, referral=1):
```
kleesc, line 38: I don't think following referrals should be on by default, since the assumption that the same credentials will work on a separate server is kinda broken.

michaelalang:
@kleesc patched it to utilize the config.yaml file as suggested, still I do not want to break the behavior we currently have so I kept the `do referrals` behavior as default.
If you still consider it the right approach to change that, I'll adjust the _DEFAULT_REFERRALS variable accordingly

data/users/__init__.py
```diff
@@ -67,6 +67,7 @@ def get_users_handler(config, _, override_config_dir):
         ldap_user_filter = config.get("LDAP_USER_FILTER", None)
         ldap_superuser_filter = config.get("LDAP_SUPERUSER_FILTER", None)
         ldap_restricted_user_filter = config.get("LDAP_RESTRICTED_USER_FILTER", None)
+        ldap_referrals = int(config.get("LDAP_FOLLOW_REFERRALS", True))
```
kleesc, line 70: Should this be set by default. In my opinion, this should be set to False, and if required, explicitly set.

kleesc (commented):
Small comment. Otherwise LGTM

kleesc (commented):
Small comment. Otherwise LGTM

dmage:
Please fix formatting before we can merge it - https://github.com/quay/quay/actions/runs/5078550493/jobs/9264087635?pr=1905

data/users/__init__.py
HammerMeetNail: @kleesc Since we follow ldap referrals today, albeit erroneously in most cases, wouldn't it make sense to preserve existing behavior and keep this set to `True` so we don't unintentionally break people on upgrade? I think the alternative is to risk breaking someone's existing config and forcing a callout/warning in the release notes. 

If we're confident that ldap referrals are genuinely broken, maybe it does make sense to change this to be default `False`?

michaelalang: for the renuinely broken referrals ... we have one use case where referrals are working:

if for example the People Base DN `ou=people,dc=example,dc=com` is the referral itself, than our code will receive an ReferralException and handle it accordingly.

If the referral object is inside the search tree like `ou=External,ou=People,dc=example,dc=com` than we'll fail even at login.

As discussed, its 99.999% broken and we could provide a note with the Patch, that if someone configured it's base DN as referral, they should configure the referred LDAP uri instead as Base DN.

kleesc (approved):
lgtm

PR #1907, opened 2023-05-25
proxy: allowing expiring tags with timemachine set to 0 (PROJQUAY-5558)

bcaton85:
Adds the following changes:
- Allows `remove_tag_from_timemachine` to expire tags even if the time machine window is set to 0, immediately marking them for deletion. This allows the quota proxy pruner to expire tags with the same method call. This wasn't required for normal push/pulls as the user would just call the `DELETE /tag` endpoint.
- Remove `hidden = true` when expiring tags. For proxy Quay will attempt to lookup the tag referenced by the manifest in order to extend it's `lifetime_end_ms`. Hiding this tag prevents that logic from running correctly.

### DIFF
--- a/data/model/oci/tag.py
+++ b/data/model/oci/tag.py
@@ -698,6 +698,9 @@ def get_legacy_images_for_tags(tags):
 
 
 def get_tags_within_timemachine_window(repo_id, tag_name, manifest_id, timemaching_window_ms):
+    if timemaching_window_ms == 0:
+        return []
+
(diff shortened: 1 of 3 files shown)


flavianmissi (commented):
I'm having a hard time to picture the effect of these changes in Quay.
Can you elaborate more on this? Is there a chance expired tags will start showing up in the UI or start being listed in the API?
What problem are we solving here?
🤔

PR #1913, opened 2023-05-31
logs: ssl_ca_path should be optional (PROJQUAY-4993)

dmage:
When the Splunk instance uses a certificate from a well-known CA, ssl_ca_path is not needed.

### DIFF
--- a/data/logs_model/logs_producer/splunk_logs_producer.py
+++ b/data/logs_model/logs_producer/splunk_logs_producer.py
@@ -32,23 +32,21 @@ class SplunkLogsProducer(LogProducerInterface):
             "scheme": url_scheme,
             "verify": verify_ssl,
         }
+        # Create an SSLContext object for cert validation
+        context = ssl.create_default_context()
+        context.check_hostname = True
+        context.verify_mode = ssl.CERT_REQUIRED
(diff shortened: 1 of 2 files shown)


data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -32,23 +32,22 @@ def __init__(
             "scheme": url_scheme,
             "verify": verify_ssl,
         }
-        if ssl_ca_path:
-            # Create an SSLContext object for cert validation
-            context = ssl.create_default_context()
+        # Create an SSLContext object for cert validation
+        context = ssl.create_default_context()
+        if verify_ssl:
```
harishsurf, line 37: nit: we could simplify conditional for easier readability
```suggestion
        context.check_hostname = True
        context.verify_mode = ssl.CERT_REQUIRED
        if not verify_ssl:
            context.check_hostname = False
            context.verify_mode = ssl.CERT_NONE
```

dmage: I find it harder to read. But as that's the way it's written now, I don't mind keeping it as it is.

harishsurf (approved):

PR #1923, opened 2023-06-02
[redhat-3.7] superuser: paginate user's list (PROJQUAY-4297)

Sunandadadi:
Allow loading of superuser user list for larger registries by paginating API call.


bcaton85 (approved):
LGTM

PR #1925, opened 2023-06-02
[redhat-3.7] superuser: lowering page limit (PROJQUAY-5178)

Sunandadadi:
Need to merge [1923](https://github.com/quay/quay/pull/1923/files) before this PR.

Lowering the page limit allows for more operations to be done per requested user. Specifically in the scenario where for each user a LDAP lookup is done to check whether the user in the list is a superuser or not. This allows time for the request to complete and load the user list without error.


bcaton85 (approved):
LGTM

PR #1928, opened 2023-06-04
audit: ignore errors due to read-only mode in audit logs (PROJQUAY-5598)

dmesser:
This addresses https://issues.redhat.com/browse/PROJQUAY-5598

When the registry is in read-only mode, all writes to the database are suspended. Subsequently all log audit entry attempts will fail if the log model is based on the database. To avoid dealing with related exceptions from the database / model layer across the code base this PR adds handling at the relevant table logs model directly. This doesn't effect other log models where the log storage is an external system (ElasticSearch, Splunk, Document, etc)

### DIFF
--- a/data/logs_model/table_logs_model.py
+++ b/data/logs_model/table_logs_model.py
@@ -240,6 +240,7 @@ class TableLogsModel(SharedModel, ActionLogsDataInterface):
                 timestamp=timestamp,
             )
         except ReadOnlyModeException:
+            logger.debug("Skipping log action '%s' in read-only mode" % kind_name)
             pass
 
     def yield_logs_for_export(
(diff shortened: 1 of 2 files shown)


data/logs_model/table_logs_model.py
```diff
@@ -228,15 +229,18 @@ def log_action(
...
+            model.log.log_action(
+                kind_name,
+                namespace_name,
+                performer=performer,
+                repository=repository,
+                ip=ip,
+                metadata=metadata or {},
+                timestamp=timestamp,
+            )
+        except ReadOnlyModeException:
```
dmage, line 242: should we have a debug message that log_action failed?

dmage (approved):

PR #1929, opened 2023-06-04
db: change removed_tag_expiration_s  to bigint (PROJQUAY-2462)

dmesser:
This addresses https://issues.redhat.com/browse/PROJQUAY-2462.

In this case we see garbage collection stopping due to too large values in the removed_tag_expiration_s column of the user table. This column is currently set to integer which is fine for most use cases, but becomes problematic with garbage collection since we multiply this value by 1000. The maximum allowed number for a signed integer is [(1/2)*2^32]-1 or 2147483647, any number over this value will lead to errors. For example 4 weeks of time machine window will lead to this field overflowing.

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -675,7 +675,7 @@ class User(BaseModel):
     invoice_email = BooleanField(default=False)
     invalid_login_attempts = IntegerField(default=0)
     last_invalid_login = DateTimeField(default=datetime.utcnow)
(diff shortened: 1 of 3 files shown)


flavianmissi (approved):
LGTM

If we're using such big values in seconds it might be worth storing them in minutes.

dmage (approved):

PR #1935, opened 2023-06-05
[redhat-3.7] chore: Removing futures library dependency

Sunandadadi:


dmage:
It's better to backport https://github.com/quay/quay/pull/1743

Sunandadadi:
> It's better to backport #1743

Got it, Do we want to remove all the dependencies in 1743?

PR #1952, opened 2023-06-07
data: Fix error getting tag while calculating scan results SLO (PROJQUAY-5600)

Marcusk19:
Catch possible exception thrown when looking up tag `lifetime_start_ms` and set `created_at` property for a manifest to None.

If `created_at` is None, don't calculate indexing SLI for manifest.

### DIFF
--- a/data/registry_model/datatypes.py
+++ b/data/registry_model/datatypes.py
@@ -353,7 +353,10 @@ class Manifest(
         """
         When the manifest was created
         """
-        result = TagTable.get(TagTable.manifest_id == self.id).lifetime_start_ms
+        try:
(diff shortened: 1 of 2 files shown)


data/registry_model/datatypes.py
```diff
@@ -353,7 +353,10 @@ def created_at(self):
         """
         When the manifest was created
         """
-        result = TagTable.get(TagTable.manifest_id == self.id).lifetime_start_ms
+        try:
+            result = TagTable.get(TagTable.manifest_id == self.id).lifetime_start_ms
+        except:
```
bcaton85, line 358: You'll want to catch the specific exception `TagTable.DoesNotExist`. Any other exception is valid.

bcaton85 (approved):
LGTM

PR #1954, opened 2023-06-08
chore: Add nightly unit test for Power (PROJQUAY-5380)

yussufsh:
This PR will add a new nightly workflow for running unit tests on multiple architectures.
Initially, we are running only unit tests on amd64 and ppc64le. In the future we can add more tests eg: e2e.


yussufsh:
Same code was part of #1965

PR #1962, opened 2023-06-09
chore: Bump mypy

dmage:
Updating mypy to the latest version. It better recognizes the structure of Quay packages.

Two scripts are removed as they use non-existing packages.

```
tools/parsebuildpack.py:3: error: Cannot find implementation or library stub for module named "workers.dockerfilebuild"  [import]
tools/freeloaders.py:4: error: Cannot find implementation or library stub for module named "data.plans"  [import]
```


flavianmissi (approved):
LGTM

PR #1972, opened 2023-06-14
api: allow robot token creation with a pre-defined token (PROJQUAY-5414)

syed:
This is for usecases where we want to explicitly create a robot token with a pre-defined token in case of migration events when we migrate from one quay to another, we can re-use the same robot token to avoid resetting it in all places it is used


kleesc (dismissed):
LGTM

kleesc (approved):

PR #1973, opened 2023-06-14
ui: Add custom TOS support

ibazulic:
Enable adding of a custom TOS in the Quay footer for on-premise installations via the optional `TERMS_OF_SERVICE` config flag. If the flag is not defined, the TOS field will not be visible when the page is rendered. Also changes the behaviour of the discovery endpoint to include custom terms of service instead of pointing to Quay.io terms of service for all installations.

### DIFF
--- a/config.py
+++ b/config.py
@@ -48,6 +48,7 @@ CLIENT_WHITELIST = [
     "QUOTA_BACKFILL",
     "PERMANENTLY_DELETE_TAGS",
     "UI_V2_FEEDBACK_FORM",
+    "TERMS_OF_SERVICE",
 ]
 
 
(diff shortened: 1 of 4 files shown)


config.py
```diff
@@ -845,3 +846,6 @@ def create_transaction(db):
     ENTITLEMENT_RECONCILIATION_MARKETPLACE_ENDPOINT = ""
 
     FEATURE_RH_MARKETPLACE = False
+
+    # Set up custom TOS for on-premise installations
+    TERMS_OF_SERVICE = ""
```
syed, line 851: can we call it `TERMS_OF_SERVICE_URL`

ibazulic: Agreed. Changed in new commit.

syed (approved):

PR #1974, opened 2023-06-14
federationuser(ldap)!: avoid doing LDAP lookups for Robot accounts (PROJQUAY-5137)

michaelalang:
BREAKING CHANGE: LDAP lookup of robot accounts in the UI for granting
                     permission has been dropped in context of permissions granting.
                    This impacts Users from LDAP if they have not logged in to Quay 
                    already (pre provisioning) as a Federation User will only be available
                    after logging in or being part of a Team.
    
the functions:
   
* _ldap_single_user_search
* _ldap_user_search
    
have been updated to provide `context` for dropping the LDAP query. As
the functions are used during login as well as during setting
permissions it is mandatory to be able to distinguish the context.
    
the functions:
    
* get_nonrobot_user
* find_user_by_email
    
from the module `data.model.user` have been utilized to identify, if the
user should be looked up in LDAP or not.


michaelalang:
Guys,

this PR for avoiding LDAP queries on Robot accounts fails due to the unit-test `test_entity_search` does not federate an none DB user upfront. 

So its an ordering issue we need to address.
Compared to real-life implementations, the User method verify_and_link_user has not been called when executing, which happens at normally at login.

All other unit-tests called prior `test_entity_search` are not executing the method which explains why it's failing with the change to not do LDAP lookup for Robot accounts that are in the Database only. 

Manual verification in a LAB show that the PR change does work as expected as the method `verify_and_link_user` in users/federation calls module `externalldap.py` and the method `verify_credentials` which does the LDAP lookup accordingly.

Additional reason why the context for LDAP lookup in the `verify_credentials` method is sufficient, from Issue `PROJQUAY-5117` where it's stated, that the call is necessary because of permissions granting to the Robot, I want to challenge that for this being true, it would be necessary that Robot accounts are configured in LDAP too which is not the case.

Furthermore, in relation to `PROJQUAY-5657` I was able to verify that as long s the session has not expired, Users can be removed from LDAP completely and still get access granted, obsoleting the statement for necessity once more,.

Should I extend the patch to modify the Unit-test accordingly or do we want to drop the PR completely ?

michaelalang:
Dear all,

without a modification on the broken Unit test to federate the User prior utilizing the entity API to retrieve the user, the change cannot proceed and is blocked already for 3 weeks. 

Can someone please take a decision if updating the Unit test accordingly will be acceptable ? 
I can provide an update to the test if necessary but I do not want to propose such a change without acceptance of the community.

All the best and kind regards
Michi

PR #1975, opened 2023-06-14
federationuser(ldap): adding keepalive settings for LDAP connections (PROJQUAY-5137)

michaelalang:
Adding keepalive setting for LDAP connection in complex network scenarios.

introducing variables for 
* _DEFAULT_KEEPALIVE_IDLE
* _DEFAULT_KEEPALIVE_INTERVAL
* _DEFAULT_KEEPALIVE_PROBES

setting the corresponding LDAP options. 
The LDAP restart option ensures on issues that the connection will be re-established.

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -20,6 +20,7 @@ _DEFAULT_KEEPALIVE_IDLE = 10
 _DEFAULT_KEEPALIVE_INTERVAL = 5
 _DEFAULT_KEEPALIVE_PROBES = 3
 
+
 class LDAPConnectionBuilder(object):
     def __init__(
         self,


data/users/externalldap.py
```diff
@@ -16,7 +17,9 @@
 _DEFAULT_PAGE_SIZE = 1000
 # setting config LDAP_FOLLOW_REFERRALS: 0 to disable referral lookups
 _DEFAULT_REFERRALS = True
-
+_DEFAULT_KEEPALIVE_IDLE = 10
+_DEFAULT_KEEPALIVE_INTERVAL = 5
+_DEFAULT_KEEPALIVE_PROBES = 3
```
kleesc, line 22: Just curious as to how we figured out these were sane defaults. 
What are the values when not explicitly set?

data/users/externalldap.py
```diff
@@ -290,12 +302,17 @@ def _ldap_user_search(
             return (with_dns, None)
 
     def _ldap_single_user_search(
-        self, username_or_email, filter_superusers=False, filter_restricted_users=False
+        self, username_or_email, filter_superusers=False, filter_restricted_users=False, login=False
     ):
+        dbuser, dbmail = get_nonrobot_user(username_or_email), find_user_by_email(username_or_email)
```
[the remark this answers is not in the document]
michaelalang: @kleesc I based the second PR on the first (1974) that's why those lines are in. If you prefer, I can rebase them to the current main.

data/users/externalldap.py
michaelalang: the default values are 

KEEPALIVE_IDLE = 7200 # 2 hours
KEEPALIVE_INTERVAL = 75 # seconds
KEEPALIVE_PROBES = 9 # until connection get's closed.

with the `quay-operator` killing the POD within ~2 minutes if the health checks fail, it seems obvious to set them low as considered being internal network connectivity the size of 64b is worth mentioning and recovery should be possible.
The settings reflect a connection after 10 seconds of being IDLE will send very 5 seconds a keepalive and drop/close the connection after worst case 15 seconds no response received. If meanwhile data passes the connection, keepalive will reset for the 10 seconds IDLE.

data/users/externalldap.py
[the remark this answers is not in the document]
kleesc: Yea, I would split these if you think they're not related. Makes it easier to revert if needed. Otherwise LGTM

data/users/externalldap.py
michaelalang: do you want me to change/add config.yaml parameters for it ?

kleesc: I don't think that's necessary as you're currently just updating the default values. If you want to make it configurable, you would have to add these to https://github.com/quay/quay/blob/master/data/users/__init__.py#L56-L90.

michaelalang: well I guess we can save that for a different PR  :dancing_women: 
thanks for your input and support, will update the PR accordingly

kleesc (approved):

Sunandadadi:
@michaelalang Are you waiting on anything to merge this PR? If this is good to go, can you please merge?

michaelalang:
@Sunandadadi I am only contributing not deciding and or merging requests. Please give the guys the time to do so ...

PR #1978, opened 2023-06-16
chore: Run pre-commit checks on pull requests

dmage:
This change runs all [pre-commit checks](https://github.com/quay/quay/blob/master/.pre-commit-config.yaml) in a GitHub job. This job will be informing until we agree to make it required.


Sunandadadi (approved):
lgtm

PR #1989, opened 2023-06-22
(wip) billing: Assign SKU to org (PROJQUAY-5363)

Marcusk19:
* Create table for storing SKUs assigned to an org.
* Add endpoints `v1/organization/<orgname>/sku` for managing org SKUs and `v1/user/rhsubscriptions` for viewing all subscriptions a user has.


app.py
```diff
@@ -313,6 +304,9 @@ def _request_end(resp):
 
 tuf_metadata_api = TUFMetadataAPI(app, app.config)
 
+marketplace_users = MarketplaceUserApi(app)
+marketplace_subscriptions = MarketplaceSubscriptionApi(app)
```
kleesc, line 308: Are we deprecating `ENTITLEMENT_RECONCILIATION`?

data/database.py
```diff
@@ -2152,13 +2155,25 @@ def upstream_registry_hostname(self) -> str:
...
     account_number = IntegerField(unique=True)
 
 
+class OrganizationRhSkus(BaseModel):
+    """
+    Represents sku to org binding for
+    RH subscriptions
+    """
+
+    id = IntegerField(primary_key=True)
```
kleesc, line 2171: You shouldn't need to explicitly define the unique id of the table.

data/database.py
```diff
@@ -2152,13 +2155,25 @@ def upstream_registry_hostname(self) -> str:
...
+    subscription_id = IntegerField(unique=True)
+    user_id = IntegerField()
```
kleesc, line 2173: What's the user_id for if org_id already exists?

data/model/__init__.py
```diff
@@ -137,6 +137,10 @@ class UnsupportedQuotaSize(DataModelException):
     pass
 
 
+class OrgSubscriptionbindingExists(DataModelException):
```
kleesc, line 140: `OrgSubscriptionBindingAlreadyExists`

app.py
Marcusk19: No, the flag is still used for enabling the reconciler worker. This change was mostly for allowing the mocked user and subscription apis to be re-used in unit tests.

data/database.py
Marcusk19: The user_id is so we can track who assigned the sku to the org. That way when a user leaves an org we can remove all bindings in that table matching the user_id and org_id.

kleesc: In this case wouldn't we want that field to be a FK referencing the user table?

Marcusk19: Would it be ok to have multiple FK constraints on the user table since org_id already does that?

kleesc: I don't see why not. That way, you don't have to enforce that constraint in the application.

data/model/organization_skus.py
```diff
@@ -0,0 +1,73 @@
...
+from data import model
+from data.database import OrganizationRhSkus
+
+logger = logging.getLogger(__name__)
+
+
+def get_org_subscriptions(org_id):
+    subscriptions = []
+    try:
+        query = OrganizationRhSkus.select().where(OrganizationRhSkus.org_id == org_id).dicts()
```
kleesc, line 14: Could you return the query object itself here, and let the caller deal with that?
That way the results get lazily loaded only when necessary.

data/migrations/versions/b82361fba1cd_create_table_for_org_to_sku_relations.py
```diff
@@ -0,0 +1,35 @@
...
+        sa.Column("subscription_id", sa.Integer, nullable=False),
+        sa.Column("org_id", sa.Integer, nullable=False),
+        sa.Column("user_id", sa.Integer, nullable=False),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_organizationrhskus")),
+        sa.ForeignKeyConstraint(
+            ["user_id"], ["user.id"], name=op.f("fk_organizationrhskus_userid")
+        ),
+        sa.ForeignKeyConstraint(["org_id"], ["user.id"], name=op.f("fk_organizationrhskus_orgid")),
+    )
+    op.create_unique_constraint("uq_subscription_id", "organizationrhskus", ["subscription_id"])
```
kleesc, line 31: What does `create_unique_constraint` do? I don't think we've used it before.
You can achieve the same with an index:`create_index` with the `unique` parameter set.
That'll probably help with querying too (the order of the columns is relevant btw).

data/model/user.py
```diff
@@ -820,6 +820,13 @@ def get_user_by_id(user_db_id):
         return None
 
 
+def get_user_or_org_by_id(user_db_id):
+    try:
+        return User.get(User.id == user_db_id, User.robot == False)
+    except User.DoesNotExist:
+        return None
```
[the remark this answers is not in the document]
Marcusk19: The only thing that stops me from using `get_organization` is that it takes a username as the value to lookup by, and in this case I only have the org id, so I want to find the org name using the org id

data/database.py
```diff
@@ -2152,13 +2155,24 @@ def upstream_registry_hostname(self) -> str:
...
     account_number = IntegerField(unique=True)
 
 
+class OrganizationRhSkus(BaseModel):
+    """
+    Represents sku to org binding for
+    RH subscriptions
+    """
+
+    subscription_id = IntegerField(unique=True)
```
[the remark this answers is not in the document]
Marcusk19: No, this is supposed to be the unique id from the marketplace api that can be used to lookup the sku, etc.

data/model/user.py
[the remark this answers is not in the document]
Marcusk19: Nvm I realized the `org_id` I get from checking if it's bound is actually a user model itself so it turns out I can just get the org name from there

data/migrations/versions/b82361fba1cd_create_table_for_org_to_sku_relations.py
```diff
@@ -0,0 +1,38 @@
...
+        sa.Column("org_id", sa.Integer, nullable=False),
+        sa.Column("user_id", sa.Integer, nullable=False),
+        sa.PrimaryKeyConstraint("id", name=op.f("pk_organizationrhskus")),
+        sa.ForeignKeyConstraint(
+            ["user_id"], ["user.id"], name=op.f("fk_organizationrhskus_userid")
+        ),
+        sa.ForeignKeyConstraint(["org_id"], ["user.id"], name=op.f("fk_organizationrhskus_orgid")),
+    )
+    op.create_index(
+        "organizationrhskus_subscription_id", "organizationrhskus", ["subscription_id"], unique=True
```
kleesc, line 32: Shouldn't org id also be a foreign key?

I would add some composite index too: 
- (subscription_id, org_id)
- (subscription_id, org_id, user_id)

test/test_api_usage.py
```diff
@@ -5061,5 +5067,40 @@ def test_delete_message(self):
...
+            resource_name=OrganizationRhSku,
+            params=dict(orgname=ORGANIZATION),
+            data={"subscription_id": 12345},
+            expected_code=201,
+        )
+        json = self.getJsonResponse(
+            resource_name=OrganizationRhSku,
+            params=dict(orgname=ORGANIZATION),
+        )
+        self.assertEqual(len(json), 1)
```
kleesc, line 5083: Can we add a test with an SKU that's already bound to an org, where it wouldn't return a 2XX?

util/marketplace.py
```diff
@@ -165,15 +165,167 @@ def create_entitlement(self, customerId, sku):
...
+    def __init__(self):
+        self.subscription_extended = False
+        self.subscription_created = False
+
+    def lookup_subscription(self, customer_id, sku_id):
+        return None
+
+    def create_entitlement(self, customer_id, sku_id):
+        self.subscription_created = True
+        pass
```
kleesc, line 267: nit: don't really need this `pass`

util/marketplace.py
```diff
@@ -165,15 +165,167 @@ def create_entitlement(self, customerId, sku):
...
+
+    def extend_subscription(self, subscription_id, end_date):
+        self.subscription_extended = True
+
+    def get_subscription_sku(self, subscription_id):
+        pass
```
kleesc, line 273: Shouldn't the return some mocked value?

endpoints/api/organization.py
```diff
@@ -372,16 +366,27 @@ def get(self, orgname):
...
+                rh_subscriptions = list(query.dicts()) if query is not None else []
+                for subscription in rh_subscriptions:
+                    subscription_sku = marketplace_subscriptions.get_subscription_sku(
+                        subscription["subscription_id"]
+                    )
+                    equivalent_stripe_plan = get_plan_using_rh_sku(subscription_sku)
+                    if equivalent_stripe_plan:
+                        repos_allowed += equivalent_stripe_plan["privateRepos"]
+
+            data["privateAllowed"] = private_repos < repos_allowed
```
kleesc, line 389: Do we only return the number of private repos allowed as part of the response, or is the list of bound SKUs also included?

endpoints/api/billing.py
```diff
@@ -917,3 +930,124 @@ def delete(self, orgname, field_uuid):
...
+        permission = AdministerOrganizationPermission(orgname)
+        if permission.can():
+            organization = model.organization.get_organization(orgname)
+
+            model.organization_skus.remove_subscription_from_org(organization.id, subscription_id)
+            return ("Deleted", 204)
+        abort(401)
+
+
+@resource("/v1/user/marketplace")
```
kleesc, line 1022: Should this be `/v1/user/marketplace` to be consistent with the org endpoints?

data/migrations/versions/b82361fba1cd_create_table_for_org_to_sku_relations.py
Marcusk19: Is org id not a foreign key here? I thought it would just be another reference to the user table

endpoints/api/organization.py
Marcusk19: I think this actually just returns a boolean value for whether you're allowed to create a private repo and the number of existing private repos

util/marketplace.py
kleesc: ^

Marcusk19: It should be returning a value now

kleesc (approved):

kleesc (approved):

PR #1996, opened 2023-06-29
ui: adding create tag option and generic alerts (PROJQUAY-5290)

bcaton85:
Adds tag options dropdown and the create tag option. Also adds generic alerts through the `useAlerts` hook.

image

### DIFF
--- a/web/src/hooks/UseTags.ts
+++ b/web/src/hooks/UseTags.ts
@@ -1,0 +1,20 @@
+import { useMutation } from "@tanstack/react-query";
+import { createTag } from "src/resources/TagResource";
+
+export function useTags(org: string, repo: string){
+    const {
+      mutate: mutateCreateTag,
+      isSuccess: successCreateTag,
+      isError: errorCreateTag,
+    } = useMutation(
+      async ({tag, manifest}: {tag: string; manifest: string}) =>
+        createTag(org, repo, tag, manifest),
+      {},
+    );
+  
+    return {
+    createTag: mutateCreateTag,
+    successCreateTag: successCreateTag,
+    errorCreateTag: errorCreateTag,
+    };
+}
--- a/web/src/hooks/UseTags.ts
+++ b/web/src/hooks/UseTags.ts
@@ -9,7 +9,6 @@ export function useTags(org: string, repo: string){
     } = useMutation(
       async ({tag, manifest}: {tag: string; manifest: string}) =>
         createTag(org, repo, tag, manifest),
-      {},
     );
   
     return {
--- a/web/src/resources/TagResource.ts
+++ b/web/src/resources/TagResource.ts
@@ -264,3 +264,10 @@ export async function getSecurityDetails(
   assertHttpCode(response.status, 200);
   return response.data;
 }
+
+export async function createTag(org: string, repo: string, tag: string, manifest: string) {
+    await axios.put(
+      `/api/v1/repository/${org}/${repo}/tag/${tag}`,
+      {manifest_digest: manifest},
+    );
+}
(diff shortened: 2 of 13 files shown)


web/src/hooks/UseTags.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function useTags(org: string, repo: string){
+    const {
+      mutate: mutateCreateTag,
+      isSuccess: successCreateTag,
+      isError: errorCreateTag,
+    } = useMutation(
+      async ({tag, manifest}: {tag: string; manifest: string}) =>
+        createTag(org, repo, tag, manifest),
+      {},
```
harishsurf, line 12: why is this needed?

web/src/resources/TagResource.ts
```diff
@@ -264,3 +264,10 @@ export async function getSecurityDetails(
   assertHttpCode(response.status, 200);
   return response.data;
 }
+
+export async function createTag(org: string, repo: string, tag: string, manifest: string) {
+    await axios.put(
+      `/api/v1/repository/${org}/${repo}/tag/${tag}`,
+      {manifest_digest: manifest},
+    );
```
harishsurf, line 272: could we also add type checking, assert on status code and try catch in case of error - to keep it consistent with  existing axios requests?

web/src/routes/Alerts.tsx
```diff
@@ -0,0 +1,22 @@
+import { Alert, AlertActionCloseButton, AlertGroup } from "@patternfly/react-core";
+import { useRecoilState } from "recoil";
+import { alertState } from "src/atoms/AlertState";
+
+
+export default function Alerts(){
+    const [alerts, setAlerts] = useRecoilState(alertState);
+    return (
+    <AlertGroup isToast isLiveRegion>
+        {alerts.map(alert=><Alert
```
harishsurf, line 10: nit: doesn't seem to be formatted. It would be a good idea to add a format check for .tsx files in our CI

harishsurf, line 10: what do you think about adding a [timeout](https://www.patternfly.org/v4/components/alert/#alert-timeout) for alerts?

web/src/routes/RepositoryDetails/Tags/TagsActionsAddTagModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+            setValue('')
+            props.setIsOpen(false)
+        }
+    }, [errorCreateTag]);
+
+    return (
+        <>
+            <Modal
+                id="add-tag-modal"
+                title={`Add tag to manifest ${props.manifest}`}
```
harishsurf, line 34: could you make the font smaller so that the `sha` doesn't cut off?  The mocks seem to have smaller font size

harishsurf (commented):
some minor comments, otherwise lgtm!

web/src/hooks/UseTags.ts
bcaton85: It isn't really, just left over. It can be removed.

web/src/resources/TagResource.ts
bcaton85: This is now being handled by react query, we should consider removing that code from the other axios requests.

web/src/routes/Alerts.tsx
bcaton85: agreed

bcaton85: I was thinking we would want to keep an error alert so the user has a chance to debug. It makes sense for successful messages though.

web/src/routes/RepositoryDetails/Tags/TagsActionsAddTagModal.tsx
bcaton85: done

harishsurf (approved):
lgtm!

harishsurf (approved):

PR #1997, opened 2023-06-29
secsan: Add tags field to event data on new vulnerability notifications (PROJQUAY-5681)

Marcusk19:
Add tag field to event data for vulnerability notifications on new image pushes.

Missing tag field caused an exception for slack notifications, should allow slack notifications to fire for new images.

Screenshot 2023-06-29 at 11 49 55 AM

### DIFF
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -1,3 +1,4 @@
+import functools
 import itertools
 import logging
 import urllib
@@ -57,6 +58,7 @@ logger = logging.getLogger(__name__)
 
 
 DEFAULT_SECURITY_SCANNER_V4_REINDEX_THRESHOLD = 86400  # 1 day
+TAG_LIMIT = 100
 
 IndexReportState = namedtuple("IndexReportState", ["Index_Finished", "Index_Error"])(  # type: ignore[call-arg]
     "IndexFinished", "IndexError"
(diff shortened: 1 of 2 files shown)


data/secscan_model/secscan_v4_model.py
```diff
@@ -440,10 +442,15 @@ def should_skip_indexing(manifest_candidate):
                             import notifications
 
                             keys = list(found_vulnerabilities)
+                            tag_names = list(
+                                registry_model.tag_names_for_manifest(manifest, TAG_LIMIT)
+                            )
+                            print("TAG_NAMES: ", tag_names)
```
kleesc, line 448: Forgot to remove this print statement

data/secscan_model/secscan_v4_model.py
```diff
@@ -56,6 +57,7 @@
 
 
 DEFAULT_SECURITY_SCANNER_V4_REINDEX_THRESHOLD = 86400  # 1 day
+TAG_LIMIT = 100
```
hdonnay, line 60: Why the limit? What subset is chosen when it's exceeded?

Marcusk19: From what I understand the limit is for retrieving the tag names that point to that manifest up to that limit. So at max there will only be 100 tags attached to the event data for the notification.

hdonnay: Right, so how is that 100 chosen?
The limit will have to be documented somewhere, also.

Alternatively, is it possible to page through all the tags and emit multiple events if needed?

Marcusk19: The security notification worker already sets it as 100 which is why I chose that number. It's also possible to set it as None for no limit. 
Is it better to just go through all tag names and not limit it to 100?

hdonnay: Maybe? It could make the event object very large, which is why I suggested paging.

PR #2002, opened 2023-07-04
setup: Add models pkgs to setup.py (PROJQUAY-5414)

syed:
This is useful for places where we install quay using setup.py like service tool


kleesc (approved):
LGTM

PR #2005, opened 2023-07-06
storage: Do nothing when completing chunked upload if chunk list is empty (PROJQUAY-5489)

Marcusk19:
When completing a chunked upload, if the chunk list is empty do not attempt to assemble anything.

Using oras to copy an artifact from an outside registry to quay results in a 5XX error. This is because at some point the upload chunk list is empty and attempting to complete the chunked upload causes an exception. Not trying to write to storage if there are no chunks allows the copy operation to successfully complete.

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -1,31 +1,27 @@
-import os
-import logging
 import copy
+import logging
+import os
 import urllib
-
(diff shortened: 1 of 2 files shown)


kleesc (approved):
LGTM

PR #2006, opened 2023-07-07
Pagination: Fixing paginate for /api/v1/superuser/logs API

Sunandadadi:
Superuser logs are sorted on datetime field. The next page token contains the value of the start_index of the datetime field for the next page. datetime column on the LogEntry table is not unique, so in some cases,  multiple entries can have the same datetime field. 
For example, logs have a page limit of 20 entries. If 25 entries have the same datetime - d1, the start_index of the next page is again d1. With the current logic, the same query is run for the next page too. This PR addresses such issues and adds an offset in such cases where (the first 20 entries of page 1 and 1 or more entries of page 2 have the same datetime).

### DIFF
--- a/data/model/modelutil.py
+++ b/data/model/modelutil.py
@@ -26,35 +26,26 @@ def paginate(
     # to order by. The where clause, on the other hand, cannot use the alias because Postgres does
     # not allow aliases in where clauses.
     sort_field_name = sort_field_name or "id"
-    sort_id_field = getattr(model, "id")
     sort_field = getattr(model, sort_field_name)
 
     if sort_field_alias is not None:
         sort_field_name = sort_field_alias
         sort_field = SQL(sort_field_alias)
 
-    order_by_list = []
     if descending:
-        order_by_list.append(sort_field.desc())
+        query = query.order_by(sort_field.desc())
     else:
-        order_by_list.append(sort_field)
+        query = query.order_by(sort_field)
 
-    # Since datetime is not a unique key, multiple entries have the same datetime.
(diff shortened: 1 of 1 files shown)


data/model/modelutil.py
```diff
@@ -69,24 +77,33 @@ def pagination_start(page_token=None):
...
+                results[limit - 1], sort_field_name
+            )
+            offset_val = limit + offset_val if use_offset else 0
             is_datetime = True
 
         page_token = {
             "start_index": start_index,
             "page_number": page_number + 1 if page_number else 1,
             "is_datetime": is_datetime,
+            "use_offset": use_offset,
```
kleesc, line 105: Do you really need `use_offset`, or just to set `offset_val`?

Sunandadadi: nice, I can just use `offset_val `

data/model/modelutil.py
```diff
@@ -69,24 +77,32 @@ def pagination_start(page_token=None):
...
 
-def paginate_query(query, limit=50, sort_field_name=None, page_number=None):
+def paginate_query(query, limit=50, sort_field_name=None, page_number=None, offset_val=0):
     """
     Executes the given query and returns a page's worth of results, as well as the page token for
     the next page (if any).
     """
     results = list(query)
     page_token = None
+    offset_val = 0
```
kleesc, line 87: This is already defined as a default parameter

Sunandadadi: Ah, removed.

kleesc (approved):
LGTM

PR #2024, opened 2023-07-20
storage: Check for request context before getting host header (PROJQUAY-5860)

syed:
The download_url function is also used by the security worker which doesn't have an app context when it calls.


bcaton85 (approved):
LGTM

PR #2027, opened 2023-07-20
api: Adding ignore timezone flag when parsing datetime (PROJQUAY-5360)

Sunandadadi:
When the database time is altered or explicitly set, the dateutil library adds the offset - `±HH:MM` to the set datetime. This is not identified by the query language and hence gets ignored when the query is complied. `ignoretz` flag in the [parse function](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parse) when set to `True` does not add this offset to the datetime and thus the start_index from the page_token is identified and complied.

### DIFF
--- a/data/model/modelutil.py
+++ b/data/model/modelutil.py
@@ -71,8 +71,6 @@ def pagination_start(page_token=None):
     """
     if page_token is not None:
         start_index = page_token.get("start_index")
-        if page_token.get("is_datetime"):
-            start_index = dateutil.parser.parse(start_index)
         return start_index
     return None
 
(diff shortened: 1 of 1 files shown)


ibazulic:
Will this disable our ability to grab logs based on the time and date provided in the API log?

Sunandadadi:
> Will this disable our ability to filter logs based on the time and date provided in the API request?

No, this only removes the parsing of the `datetime` sent from the page_token.

dmage:
@Sunandadadi ignoretz is a code smell, should we fix generation of start_index instead? Perhaps it should be converted to UTC?

Sunandadadi:
@dmage Hmm, the start_index is generated correctly here. The problem is when a database is not in the UTC timezone, the dateutil.parse function adds an offset by default to the date time. This offset is not understood by the query language and ignores it when the query is executed. So, essentially the same query is run again instead as the new where clause is ignored.

kleesc (approved):

PR #2071, opened 2023-07-31
georep: Add error handling for remove location script (PROJQUAY-5878)

jonathankingfc:
- Gracefully handle errors that are thrown during remove location script
- This should handle scenarios where the location is not found or not provided, as well as other error cases

### DIFF
--- a/util/removelocation.py
+++ b/util/removelocation.py
@@ -39,21 +39,25 @@ def remove_location():
         print("Canceling operation...")
         return
 
-    # Get all image placements for the location
-    query = (
-        ImageStoragePlacement.select()
-        .join(ImageStorageLocation)
-        .where(ImageStorageLocation.name == location_name)
-    )
+    try:
+        # Get all image placements for the location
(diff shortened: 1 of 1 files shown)


util/removelocation.py
```diff
@@ -39,21 +39,25 @@ def remove_location():
...
+            print(f"Deleted placement {image_placement.id}")
 
-    # Delete storage location from database
-    ImageStorageLocation.delete().where(ImageStorageLocation.name == location_name).execute()
-    print(f"Deleted location {location_name}")
+        # Delete storage location from database
+        ImageStorageLocation.delete().where(ImageStorageLocation.name == location_name).execute()
+        print(f"Deleted location {location_name}")
+    except Exception as e:
+        print(f"Error: {e}")
```
harishsurf, line 59: do you want to add more context as to where the error message occurred, something like: `Error removing storage location` instead of `Error`

harishsurf (approved):

PR #2089, opened 2023-08-01
chore: drop deprecated tables and remove unused code (PROJQUAY-522)

kleesc:
Follow up to https://github.com/quay/quay/pull/1841


syed:
@kleesc  looks like a lot of tests are failing because of import errors like 

```
ImportError: cannot import name 'TagManifest' from 'data.database'
```

can you also update the tests?

data/model/_basequery.py
```diff
@@ -18,7 +18,6 @@
     Namespace,
     Visibility,
     ImageStorage,
-    Image,
     RepositoryKind,
     db_for_update,
     db_count_estimator,
```
syed, line 23: I see references to `Image` in this file for example https://github.com/quay/quay/blob/master/data/model/_basequery.py#L186

syed: looks like it's being used in the function `set_image_storage_metadata` but that function is never called so we could probably remove the whole flow

syed:
Ran it on a new instance on quay with no issues

```
15:54:49 INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
15:54:49 INFO  [alembic.runtime.migration] Running upgrade 46980ea2dde5 -> a0fc06d604a4, Unlink deprecated tables
15:54:50 INFO  [alembic.runtime.migration] Running upgrade a0fc06d604a4 -> 8a70b8777089, drop deprecated tables
```

syed:
Running on production copy gave a different result. The pod tries to drop the tables but since the health check is failing, kubernetes kill the pod mid-migration. When the next pod starts,  yrt errors out because some tables had been dropped and the drop table query will fail for them. Can you update it so it uses `DROP TABLE IF EXISTS`

syed (approved):
LGTM

PR #2098, opened 2023-08-07
chore: updated build-and-publish for s390x

ksdeekshith:
# Description
Updated Code Changes for IBM/Z 

## Changes that have been done

- Bump **PyYAML** version from `5.4` to `5.4.1`, since it is conflicting with cython_sources in Z Systems
- Added **libejpeg-turbo** and **libjpeg-turbo-devel** packages in the final container to resolve the issue for frontend crashing due to the following error in the logs,
```shell
gunicorn-web stdout | ImportError: libjpeg.so.62: cannot open shared object file: No such file or directory
``` 
- Updated Build-and-Publish workflow for building the image for s390x with **single ZVSI** and keep it running always.
- Removed the **wget in the Dockerfile** for `grpcio` and enabled `GRPC_PYTHON_BUILD_SYSTEM_OPENSSL 1` as ENV since it's required for Z Systems for both **grpcio** and **gevent** building.

## Tests that have been done

- Verified by [Building the image](https://github.com/ksdeekshith/quay/actions/runs/5782516664) through GitHub workflow.
- Tested the basic functionalities like accessing UI, creating a user, pushing images, scanning vulnerabilities, etc., in local environment.
- Verified the same using quay-operator

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,8 @@ RUN set -ex\
 		memcached \
 		nginx \
 		libpq-devel \
+		libjpeg-turbo \
+		libjpeg-turbo-devel \
 		openldap \
 		openssl \
 		python39 \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -17,6 +17,8 @@ RUN set -ex\
 		memcached \
 		nginx \
 		libpq-devel \
+		libjpeg-turbo \
+		libjpeg-turbo-devel \
```
dmage, line 21: Why do you need the devel package in the base image? Usually they are needed only in the build image.

ksdeekshith: sorry about that, I thought both might be needed so tested with both packages. Just now I tested by removing devel package, which is working fine. Will update accordingly

dmage (approved):

PR #2108, opened 2023-08-09
secscan: fix metrics to track all request types to secscan service (PROJQUAY-3839)

kleesc:
Also fixes typo in function call.

### DIFF
--- a/util/metrics/prometheus.py
+++ b/util/metrics/prometheus.py
@@ -4,16 +4,14 @@ import socket
 import sys
 import threading
 import time
-from collections import namedtuple
-import urllib.request
 import urllib.error
 import urllib.parse
+import urllib.request
+from collections import namedtuple
 
 from cachetools.func import lru_cache
-
 from flask import g, request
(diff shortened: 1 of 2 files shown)


util/metrics/prometheus.py
```diff
@@ -60,10 +60,10 @@
 )
 gc_iterations = Counter("quay_gc_iterations", "number of iterations by the GCWorker")
 
-secscan_index_request_duration = Histogram(
-    "quay_secscan_index_duration_seconds",
+secscan_request_duration = Histogram(
+    "quay_secscan_request_duration_seconds",
     "seconds taken to make an index request to the secscan service",
-    labelnames=["status"],
+    labelnames=["method", "route", "status"],
```
hdonnay, line 66: `Route` seems like a bad name, as it seems to be `action` at the call site and I initially thought it would be recording the path used in the request.

kleesc: Changed it from "route" to "action".

Marcusk19 (approved):
LGTM

PR #2109, opened 2023-08-10
chore: fix build of PyYAML on linux/ppc64le

dmage:
PyYAML 5.4.1 is the latest (at the moment) version that can be installed with Cython 3.0 (3.0b9).

As gevent requires Cython 3.0+, we cannot use PyYAML 6.0 that depends on Cython 0.x (<3.0).

As PyYAML 5.4.1 doesn't restrict the version of Cython that can be used to build it, pip selects Cython 3.0 when it creates an isolated build environment and fails to build PyYAML. To fix that, we preinstall build dependencies and disable the build isolation for PyYAML.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -88,7 +88,9 @@ RUN ARCH=$(uname -m) ; echo $ARCH; \
 	fi
 
 RUN set -ex\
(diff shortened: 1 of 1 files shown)


yussufsh (approved):
lgtm

PR #2111, opened 2023-08-10
api: Duplicate Robot accounts (PROJQUAY-5931)

Sunandadadi:
The tuples returned [here](https://github.com/quay/quay/blob/master/endpoints/api/robot_models_pre_oci.py#L37) were not distinct. So, in the for loop, if a robot_name already exists in the `robots` dictionary as [here](https://github.com/quay/quay/blob/master/endpoints/api/robot_models_pre_oci.py#L45) , the `robot_dict` is not constructed and previous items `robot_dict` is used for the current robots[robot_name]. So, the current robot_name is overwritten by previously constructed robot_dict.

By using distinct set of items, we won't run into the above issue.

### DIFF
--- a/data/model/user.py
+++ b/data/model/user.py
@@ -568,7 +568,7 @@ def list_entity_robot_permission_teams(entity_name, limit=None, include_permissi
         fields.append(Repository.name)
         fields.append(Team.name)
 
-    query = query.limit(limit).order_by(User.last_accessed.desc())
+    query = query.limit(limit).order_by(User.last_accessed.desc()).distinct()
     return TupleSelector(query, fields)
 
 
--- a/endpoints/api/robot_models_interface.py
+++ b/endpoints/api/robot_models_interface.py
@@ -66,12 +66,12 @@ class RobotWithPermissions(
     """
 
     def to_dict(self, include_token=False):
+        created = getattr(self, "created", None)
+        last_accessed = getattr(self, "last_accessed", None)
         data = {
             "name": self.name,
-            "created": format_date(self.created) if self.created is not None else None,
-            "last_accessed": format_date(self.last_accessed)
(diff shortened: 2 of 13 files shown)


ibazulic (approved):
Can't reproduce the issue. LGTM!

bcaton85:
Okay I see the issue. The [loop through the permissions](https://github.com/quay/quay/blob/master/endpoints/api/robot_models_pre_oci.py#L43) returned expects the list to be ordered by username but it is instead ordered by last accessed, and when that is null an entry with the same username can appear in different parts of the list. 

The issue is that [this loop](https://github.com/quay/quay/blob/master/endpoints/api/robot_models_pre_oci.py#L43) only updates [the robot_dict](https://github.com/quay/quay/blob/master/endpoints/api/robot_models_pre_oci.py#L45-L70) info when it encounters a robot name it hasn't seen before. So if the same robot name appears twice in different places in the list, `robot_dict` doesn't get updated and you assign robot permissions to the wrong robot. 

`distinct` works because it forces the list to be ordered by username. But i'm wary on using it since `distinct` as some performance costs. IMO having the code rely on this odd scoping mechanism should be refactored and fixed and that would prevent the need to modify any DB calls. Or at change the ordering to be by username. @kleesc what are your thoughts?

kleesc:
Was this intentional before, or just was never caught?
Like @bcaton85 mentioned, there's probably _some_ performance tradeoffs of using `distinct` (to what extent I'm not sure of, but it's probably not _that_ significant for this specific use case). You could also do the union of the tuples at the application layer instead of changing the db query.

Sunandadadi:
Was running into some issues with git so had to close this. Opened the fix here - https://github.com/quay/quay/pull/2192 instead

PR #2116, opened 2023-08-11
Use proper grpcio url for ppc64le

yussufsh:
New releases are named `grpcio-${GRPC_LATEST}-cp310-cp310-linux_ppc64le.whl`.
Always get the download URL from GitHub API and use it in wget.


yussufsh:
Not needed at present as cp310 indicate that it is built for python-3.10.

PR #2130, opened 2023-08-14
chore: remove husky

dmage:
husky, when installed via `npm install` in web/, replaces pre-commit, which is undesirable for this repository, as it essentially disables all pre-commit checks for Python code.

pre-commit already has all checks that are privded by husky.


bcaton85 (approved):
LGTM

PR #2155, opened 2023-08-25
user(robots): disallow robot login and create (PROJQUAY-5968)

michaelalang:
Add a config flag `ROBOTS_DISALLOW` to turn off Robot login and creation in Quay for all accounts. 
With the Flag set existing Robots cannot login anymore

```
$ podman login -u milang+test2 -p W7B...HQA quay.example.com
Error: logging into "quay.example.com": invalid username/password
```
    
This behavior will be reported in the logs accordingly
```
gunicorn-registry stdout | 2023-08-24 19:08:01,907 [253] [WARNING] [auth.credentials] Failed to validate credentials for robot milang+test2: Robot account has been disabled. Please contact your administrator.
```
    
as well as in the UI when creating a Robot Account which will display the `Robot account has been disabled. Please contact your  administrator.` message.


data/model/user.py
```diff
@@ -324,6 +324,10 @@ def update_enabled(user, set_enabled):
 
 def create_robot(robot_shortname, parent, description="", unstructured_metadata=None, token=None):
     (username_valid, username_issue) = validate_username(robot_shortname)
+    if config.app_config.get("ROBOTS_DISALLOW", False):
+        msg = "Robot account has been disabled. Please contact your administrator."
```
dmage, line 328: Should it be `Robot accounts have been disabled`?

dmage:
@michaelalang it's worth adding this config parameter to https://github.com/quay/quay/blob/master/util/config/schema.py

data/model/user.py
michaelalang: agreed ... but let's first get the Team triage if the change is something we want to consider as it disabled mirroring of repositories as well.

michaelalang:
will do accordingly ...
is there a generic rule on when config parameters should go in the schema and when not ?

michaelalang:
applied the language fix and added the schema definition for the option (otherwise I'll forget to add the suggested changes)

thanks @dmage

dmage (approved):

PR #2183, opened 2023-08-28
ui: adding robot account support for user namespace (PROJQUAY-5945)

Sunandadadi:


dmage (dismissed):

web/src/components/modals/CreateRobotAccountModal.tsx
```diff
@@ -116,81 +116,97 @@ export default function CreateRobotAccountModal(
...
+    ),
+  };
+
+  const AddToRepoStep = {
+    name: 'Add to repository (optional)',
+    component: (
+      <AddToRepository
+        namespace={props.namespace}
+        dropdownItems={props.RepoPermissionDropdownItems}
+        repos={repos}
```
harishsurf, line 163: nit: seeing a type warning for `repos` prop

harishsurf (commented):
Looks good! Just a fyi - when defining hooks instead of creating a separate file for every CRUD operation for a resource, I and Brandon thought it is better to have one file per resource and have all the operations defined inside it. (Eg: [UseTeams.ts](https://github.com/quay/quay/blob/master/web/src/hooks/UseTeams.ts)). This way it is easier to find methods and avoid duplication

Sunandadadi:
@harishsurf that makes sense. I will merge the different hooks for Robot accounts in a separate PR.

harishsurf (approved):

PR #2186, opened 2023-08-28
storage: adding maximum_chunk_size_gb storage option (PROJQUAY-2679)

bcaton85:
Users are currently using the `S3storage` option to use the IBM storage solution. The final copy of an image push from the `uploads` to the `sha256` storage directory currently fails for large image layers (10gb). This is due to the 10gb layer being chunked into 5gb ranges that causes delays in the IBM copy operation. This PR adds performance by removing an unnecessary copy and adds the `maximum_chunk_size_gb` option to reduce the size of the copy.

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -767,7 +767,9 @@ class S3Storage(_CloudStorage):
             access_key=s3_access_key or None,
             secret_key=s3_secret_key or None,
         )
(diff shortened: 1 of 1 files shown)


syed (approved):

syed (approved):
LGTM. We have to expose somewhere in the docs that `maximum_chunk_size_gb` is something that can be passed to the S3 storage provider

bcaton85:
Yup that's right, i'll bring in @stevsmit

kleesc (approved):

PR #2189, opened 2023-08-30
billing: marketplace UI for sku to org binding (PROJQUAY-5363)

Marcusk19:
Adds details in billing ui for managing org-bound subscriptions, also lets users view their subscriptions from Red Hat customer portal.

Screenshot 2023-08-30 at 9 44 08 AM

Screenshot 2023-08-30 at 9 44 33 AM


endpoints/api/billing.py
```diff
@@ -998,14 +1000,14 @@ def post(self, orgname):
...
 @path_param("subscription_id", "Marketplace subscription id")
-@related_user_resource(UserPlan)
 @show_if(features.BILLING)
 class OrganizationRhSkuSubscriptionField(ApiResource):
     """
     Resource for removing RH skus from an organization
     """
 
     @require_scope(scopes.ORG_ADMIN)
+    @nickname("removeSkuFromOrg")
```
kleesc, line 1010: Is there currently a way to swap available SKU from an org?

kleesc: Also, if I want to stack multiple SKUs, should I use the `POST` endpoint?

Marcusk19: There's no way to directly swap at the moment, you would need to remove the SKU from the org and assign it somewhere else. Multiple `POST` requests will stack the SKUs

endpoints/api/billing.py
```diff
@@ -1018,7 +1020,7 @@ def delete(self, orgname, subscription_id):
                 return ("Organization not valid", 400)
 
             model.organization_skus.remove_subscription_from_org(organization.id, subscription_id)
-            return ("Deleted", 204)
+            return ("Deleted", 200)
```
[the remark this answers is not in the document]
Marcusk19: I thought so too at first, but there were problems with determining if the SKU was actually removed because the 204 response can't return anything to verify that it was removed. There's an alert I want to show for a successful removal of a SKU

PR #2193, opened 2023-08-31
ui: Get public path form env variable (PROJQUAY-5938)

syed:
The console dot build 2 versions of the app one for beta and another for normal


bcaton85 (approved):
LGTM - I see that the path for `localChrome` is specific to your machine, should that be a general path?

syed:
> LGTM - I see that the path for `localChrome` is specific to your machine, should that be a general path?

Yes, that's for local dev

PR #2212, opened 2023-09-06
georeplication: Enqueue blobs for replication on manifest list pushes (PROJQUAY-5256)

ibazulic:
Previously, blob replication was only queued when normal manifests were pushed to the registry. This created problems with manifest lists who contain child manifests whose blobs were never queued. The end result was a huge discrepancy of storage usage between different defined storage engines. 

With this fix, blobs will always be enqueued for replication regardless of the image type.


bcaton85 (approved):
LGTM

PR #2217, opened 2023-09-07
quota: fixing registry total worker failing to start (PROJQUAY-6010)

bcaton85:
Worker fails to start because of wrong import + adding test.


kleesc (approved):
lgtm

PR #2228, opened 2023-09-11
ui: adding alerts for robot account actions (PROJQUAY-5946)

Sunandadadi:


web/src/routes/RepositoriesList/RobotAccountsList.tsx
```diff
@@ -478,6 +511,9 @@ export default function RobotAccountsList(props: RobotAccountsListProps) {
   }
   return (
     <>
+      <AlertGroup isToast isLiveRegion>
```
bcaton85, line 514: Do we want to use the `useAlerts` hook here instead of creating them directly in the component?

web/src/routes/OrganizationsList/Organization/Tabs/DefaultPermissions/createTeamWizard/AddTeamMember.tsx
```diff
@@ -94,9 +95,13 @@ export default function AddTeamMember(props: AddTeamMemberProps) {
   };
 
   const onCreateRobotAccount = async () => {
-    await createRobotAccntHook({
-      robotAccntName: newRobotAccntName,
+    await createNewRobot({
+      namespace: props.orgName,
+      robotname: newRobotAccntName,
       description: newRobotAccntDescription,
+      reposToUpdate: null,
```
bcaton85, line 102: Do we want to set these fields to nullable within the type declaration?

web/src/hooks/useRobotAccounts.ts
```diff
@@ -38,34 +38,91 @@ interface createNewRobotAccountForNamespaceParams {
   description: string;
 }
 
-export function useCreateRobotAccount(orgName, {onSuccess, onError}) {
+export function useCreateRobotAccount({namespace, onSuccess, onError}) {
   const queryClient = useQueryClient();
 
-  const createRobotAccntMutator = useMutation(
+  const updateRobotData = async (result) => {
```
bcaton85, line 44: These calls aren't handled with error handling. A success will still be reported to the user if the robot was created successfully even if these all fail.

Sunandadadi: Yes, I think this was done to make sure a robotAcct is created before adding repos/teams/etc to the robot acct.

PR #2233, opened 2023-09-13
chore: don't chmod on every file in Dockerfile

dmage:
This should speed up the build of the container on macOS a bit.


Sunandadadi (approved):
lgtm

PR #2234, opened 2023-09-14
documentation: Change tag template link location for build triggers (PROJQUAY-6044)

ibazulic:
The documentation link on the build trigger tag template screen is currently pointing to the location of the `schema` directly in the code. 

This is not really useful or user friendly. This will repoint that documentation link to the KCS article written for this purpose.


syed (approved):
LGTM. Thanks @ibazulic

PR #2235, opened 2023-09-14
secscan: add support for new enrichment type (PROJQUAY-5981)

crozzy:
This change adds the ability for Quay to reason with the rhcc-enrichment that Clair will return in the vulnerability reports. This information is meant to allow users or applications to determine which packages come from rhcc layers.


crozzy:
@hdonnay at the moment this ignores vulnerabilities that come from layers with an rhcc package *except* vulnerabilities directly related to the rhcc package. Do you think that is the desired behaviour? I'm not sure if we also want to include other vulnerabilities as well.

hdonnay:
I think that's right, but we should engage with stakeholders and drive a consensus.

PR #2247, opened 2023-09-21
quota: prevent tag creation on rejected manifest (PROJQUAY-3906)

bcaton85:
Currently if a new manifest is created that re-uses blobs already existing in the registry but not within the namespace it's possible for that manifest to be blocked if the newly referenced blobs exceed the quota limit. The issue being that the rejection only happens after the manifest has been created and tagged. This PR blocks the manifest upload before the creation of the tag and prevents that tag from appearing in the UI and being pullable. 

If the manifest upload get's rejected before the tag has been created for a new manifest, a temporary tag is created outside the time machine window so it is immediately available for GC.

### DIFF
--- a/data/model/oci/tag.py
+++ b/data/model/oci/tag.py
@@ -375,7 +375,7 @@ def create_temporary_tag_outside_timemachine(manifest):
         name=tag_name,
         repository=manifest.repository_id,
         lifetime_start_ms=now_ms,
-        lifetime_end_ms=0, # Start of unix epoch time
+        lifetime_end_ms=0,  # Start of unix epoch time
         reversion=False,
         hidden=True,
         manifest=manifest,
--- a/data/model/oci/test/test_oci_tag.py
+++ b/data/model/oci/test/test_oci_tag.py
@@ -1,7 +1,6 @@
 import json
 from calendar import timegm
 from datetime import datetime, timedelta
-from test.fixtures import *
 
 import pytest
 from mock import patch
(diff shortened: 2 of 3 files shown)


data/model/oci/tag.py
```diff
@@ -363,6 +363,41 @@ def create_temporary_tag_if_necessary(manifest, expiration_sec):
...
+            .get()
+        )
+    except User.DoesNotExist:
+        # Should never happen, but throw error anyway
+        raise DataModelException(
+            "Could not find namespace for repository: %s" % manifest.repository_id
+        )
+
+    time_machine_ms = namespace.removed_tag_expiration_s * 1000
+    end_ms = now_ms - time_machine_ms
```
dmage, line 387: Can we use just `0`?

bcaton85: Do you mean can we just set the end_ms to 0?

dmage: yes, can we? Or maybe `1`, or any other well-known timestamp in the past. I assume it'll save us 1 query to the database.

bcaton85: Yes we can actually. I typically do this because of the tag name, repository id, lifetime_end_ms unique index on the tag table. Since we can have expired tags that have the same name there is sometimes a clash if we try setting the lifetime_end_ms to the same value, in this case 0 or 1. Since we're using a 4 character hash for the name it shouldn't be an issue. I'll update.

bcaton85: It would only be an issue if the manifest was blocked by quota, also while during the push of each blob the quota was not exceeded, and the same 4 character hash was generated

dmage (approved):

PR #2250, opened 2023-09-22
feature: Expand rate limiting capabilities

ibazulic:
Previously, our rate limits were static in the sense that we could just turn them on or off and set specific namespaces where rate limiting is essentially turned off. With this PR we expand the rate limiting capability so that the user can define its own rates per defined bucket. The schema in Quay's `config.yaml` should look as follows:

~~~
RATE_LIMITS:
  http1:
     v2: number
     registry: number
     api_resources: number
  http2:
     v2: number
     registry: number
     api_resources: number
  namespaced:
     http1_v2: number
     http2_v2: number
     http1_registry: number
     http2_registry: number
     http1_api_resources: number
     http2_api_resources: number
~~~

We define three separate buckets for access:

* `http1`: handles HTTP/1 connections
* `http2`: handles HTTP/2 connections
* `namespaced`: handles access to particular namespaces when rate limiting is turned on

The values represent different endpoints that Quay exposes:

* `v2`: denotes the Docker v2 endpoint exposed on `/v2/`
* `registry`: deals with registry operations on manifests and blobs
* `api_resources`: deals with connections to Quay's `v1` API as well as Docker v2 `tags` and `_catalog` endpoint

Notable caveats:
* validation logic should at some point be moved to the config tool since it handles all other validation anyway
* if `FEATURE_RATE_LIMITS` is set to `true`, then all elements of the `RATE_LIMITS` setup must be present in the `config.yaml` file, else bootstrapping fails.


syed:
Thanks @ibazulic  this would be very useful to us. Will review this shortly

conf/init/rate_limiting_validate_config.py
```diff
@@ -0,0 +1,138 @@
...
+        exit(1)
+
+
+if __name__ == "__main__":
+    if os.path.exists(os.path.join(QUAYCONF_DIR, "stack/config.yaml")):
+        with open(os.path.join(QUAYCONF_DIR, "stack/config.yaml"), "r") as f:
+            config = yaml.safe_load(f)
+    else:
+        config = None
+    if config.get("FEATURE_RATE_LIMITS") == True:
```
syed, line 137: Will this give an error if `FEATURE_RATE_LIMITS == True` but there is no `RATE_LIMITS` configured?

conf/init/nginx_conf_create.py
```diff
@@ -121,10 +138,22 @@ def generate_rate_limiting_config(config):
     config = config or {}
     non_rate_limited_namespaces = config.get("NON_RATE_LIMITED_NAMESPACES") or set()
     enable_rate_limits = config.get("FEATURE_RATE_LIMITS", False)
+    if enable_rate_limits == True:
+        http1_bucket = [value for _, value in config["RATE_LIMITS"]["http1"].items()]
```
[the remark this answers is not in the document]
syed: Also, order isn't guarenteeed in the `items()` call. You might get the order as `['api_resources', 'registry', 'v2']`

conf/nginx/rate-limiting.conf.jnj
```diff
@@ -49,24 +49,27 @@ limit_req_zone $http_authorization zone=staticauth:10m rate=30r/s;
 limit_req_zone $request_id zone=staticauth:10m rate=300r/s;
 {% endif %}
 
-limit_req_zone $http1_bucket zone=dynamicauth_very_light_http1:10m rate=60r/s;
-limit_req_zone $http2_bucket zone=dynamicauth_very_light_http2:10m rate=600r/s;
-limit_req_zone $namespaced_http1_bucket zone=namespaced_dynamicauth_very_light_http1:10m rate=60r/s;
-limit_req_zone $namespaced_http2_bucket zone=namespaced_dynamicauth_very_light_http2:10m rate=600r/s;
+# Rate limits for the v2 API endpoint.
+limit_req_zone $http1_bucket zone=v2_endpoint_http1:10m rate={{ http1_bucket[0] }}r/s;
```
syed, line 53: consider using a dict like `http1_bucket['v2']`

conf/init/nginx_conf_create.py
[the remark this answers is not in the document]
ibazulic: I wondered if I can just send a dictionary, I wasn't sure. Fix is incoming because dictionaries are much better.

conf/nginx/rate-limiting.conf.jnj
ibazulic: Done in new commit.

PR #2261, opened 2023-09-26
chore: fix failing nightly and also use ssh action for ppc64le

yussufsh:


yussufsh:
Passing `GRPC_PYTHON_BUILD_SYSTEM_OPENSSL: 1` should build grpcio during tox deps for both Z/P.

dmage (approved):

PR #2266, opened 2023-09-27
ui: combining robot account hooks into a single file (PROJQUAY-5945)

Sunandadadi:
Combining different hooks associated with Robot Accounts tab into single file


web/src/hooks/useRobotAccounts.ts
```diff
@@ -182,3 +188,248 @@ interface createNewRobotForNamespaceParams {
...
+
+  const {usernames} = useOrganizations();
+  const isUser = usernames.includes(orgName);
+
+  const {
+    data: robotPermissions,
+    isLoading: loading,
+    error,
+  } = useQuery(
+    ['Namespace', namespace, 'robot', robotName, 'permissions'],
```
harishsurf, line 206: any reason why we need all these query keys?

web/src/hooks/useRobotAccounts.ts
```diff
@@ -182,3 +188,248 @@ interface createNewRobotForNamespaceParams {
   selectedTeams: IRobotTeam[];
   robotDefaultPerm: string;
 }
+
+export function useRobotPermissions({orgName, robName, onSuccess, onError}) {
```
harishsurf, line 192: nit: typo `robName`

web/src/hooks/useRobotAccounts.ts
```diff
@@ -182,3 +188,248 @@ interface createNewRobotForNamespaceParams {
...
+        onSuccess();
+      },
+      onError: (err) => {
+        onError(err);
+      },
+    },
+  );
+
+  return {
+    result: robotPermissions,
```
harishsurf, line 267: It seems as though `result` is not being used? I see it being referenced [here](https://github.com/quay/quay/blob/master/web/src/components/modals/RobotRepositoryPermissions.tsx#L29), but it doesn't seem to have any effect

web/src/hooks/useRobotAccounts.ts
```diff
@@ -182,3 +188,248 @@ interface createNewRobotForNamespaceParams {
...
+    loading: loading,
```
harishsurf, line 268: `loading: loading` seems redundant. You can just use `loading`

harishsurf (commented):
Thanks for making this change! left some comments. could you also fill in the PR description?

web/src/hooks/useRobotAccounts.ts
Sunandadadi: That's the query key for robot perms. I invalidate them [here](https://github.com/quay/quay/pull/2266/files#diff-ae5025f0fc8d98c584284426ef58c202d48f01937a65eff40694fc90ef01728aR229) and [here](https://github.com/quay/quay/pull/2266/files#diff-ae5025f0fc8d98c584284426ef58c202d48f01937a65eff40694fc90ef01728aR251) where repo perms are deleted/updated.

harishsurf: right, my thinking was that having too many query key strings seems confusing (eg:`'Namespace'` seems redundant when you already have `'robots'` along with other variables making it unique enough). From what I understand, query key uniquely identifies a given query and refetches cache only when one of its variable changes or if you use `queryClient.invalidateQueries([<query-key])`.  So we should be okay using fewer keys as long as it's unique. wdyt?

web/src/hooks/useRobotAccounts.ts
harishsurf: how about `robotAccntName` or `robotAcct` instead?

web/src/hooks/useRobotAccounts.ts
Sunandadadi: Different namespaces can have the same robot name. So, I am not sure if I can shorten it more. What do you have in mind?

harishsurf: (discussed offline)

harishsurf (approved):

PR #2267, opened 2023-09-28
quota: removing extra calls to get namespace quotas (PROJQUAY-6048)

bcaton85:
When quota is enabled and an API call is made to `GET /api/v1/repository` a query is made to fetch the namespace quota limit for each repository in the list. Most often, specifically when loading the repositories list page, the result of this call will be the same and can cause up to 99 duplicated calls with the default limit. This stores the result of the call in memory to prevent the duplicate queries.

### DIFF
--- a/endpoints/api/test/test_repository_models_pre_oci.py
+++ b/endpoints/api/test/test_repository_models_pre_oci.py
@@ -1,4 +1,3 @@
-
 import pytest
 from mock import ANY, MagicMock, patch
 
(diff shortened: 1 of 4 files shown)


dmage (dismissed):

endpoints/api/test/test_repository_models_pre_oci.py
```diff
@@ -0,0 +1,44 @@
+from endpoints.api.repository_models_pre_oci import pre_oci_model
+from test.fixtures import *
```
dmage, line 2: Please use `isort` on this file.

bcaton85: done

dmage (approved):

PR #2384, opened 2023-10-06
chore: add hack/update-requirements-build.sh

dmage:
Move build dependencies to their own file and use a cachito-provided script to generate these dependencies.


.github/workflows/CI-nightly.yaml
```diff
@@ -111,9 +111,7 @@ jobs:
           cd build && git checkout ${GH_REF}
           cat requirements-dev.txt | grep tox | xargs python3 -m pip -qq install
 
-          if [[ 'ppc64le' == $(uname -m) ]]; then
-            sed -i "s/PyYAML==5.4.1/PyYAML==6.0.1/" requirements.txt
-          else
+          if [[ 's390x' == $(uname -m) ]]; then
```
dmage, line 114: @modassarrana89 @yussufsh this PR updates PyYAML to 6.0.1 in requirements.txt; does this change look correct?

hack/update-requirements-build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -eux
+curl -fsSLO https://raw.githubusercontent.com/containerbuildsystem/cachito/master/bin/pip_find_builddeps.py
+python3 pip_find_builddeps.py requirements.txt -o requirements-build.in
```
flavianmissi, line 3: should `requirements-build.in` be checked in to this repo?

flavianmissi (commented):
does separating the build requirements from the requirements.txt file required any changes to how we install requirements on local dev envs? do we need to update the getting-started docs?
we might also need some docs about how these files are used. does the `update-requirements-build.sh` need to be run by a human? or is it run downstream by some build automation?
LGTM otherwise!

dmage:
> does separating the build requirements from the requirements.txt file requires any changes to how we install requirements on local dev envs? do we need to update the getting-started docs?

No, we don't.

> we might also need some docs about how these files are used. does the update-requirements-build.sh need to be run by a human? or is it run downstream by some build automation?

It needs to be run by a human. It generates the file that is used by our downstream cachito. Upstream builds don't use this file at all. Any suggestions where this info can be documented?

flavianmissi:
> Any suggestions where this info can be documented?

hmm I guess we don't need to document it in this repo then - maybe the downstream release docs on google drive?

PR #2393, opened 2023-10-09
autoprune: add auto-prune policy endpoints (PROJQUAY-6096)

bcaton85:
Add's the endpoints to create, update, delete, get, and list organization and user auto-prune policies.


dmage (dismissed):

endpoints/api/policy.py
```diff
@@ -0,0 +1,323 @@
...
+    def get(self, orgname):
+        permission = AdministerOrganizationPermission(orgname)
+        if not permission.can() and not allow_if_superuser():
+            raise Unauthorized()
+
+        policies = model.autoprune.get_namespace_autoprune_policies_by_orgname(orgname)
+
+        return {"policies": [policy.get_view() for policy in policies]}
+
+    @validate_json_request("AutoPrunePolicyConfig")
```
kleesc, line 59: `@require_scope(scopes.ORG_ADMIN)` decorator should be used here, as well as the other endpoints

bcaton85: added to the org endpoints. The user endpoints are already using `@require_user_admin()`

endpoints/api/policy.py
```diff
@@ -0,0 +1,323 @@
...
+        app_data = request.get_json()
+        method = app_data.get("method", None)
+        value = app_data.get("value", None)
+
+        if method is None or value is None:
+            request_error(message="Missing the following parameters: method, value")
+
+        policy_config = {
+            "method": method,
+            "value": value,
```
kleesc, line 150: I would check to make sure that this is an integer. Af far as i can tell your data model doesn't have that check, and just assumes an instance of `int`

bcaton85: the method and value are checked on creation and update [here](https://github.com/quay/quay/pull/2393/files#diff-0445600d80e07f49e4ac5d97b6853caebdeec7024679044c9c62c3b8eb31154bR132)

kleesc: Actually, never mind, I see that it's essentially a union of either str or int

kleesc: Do you see having other possible types in the future? If that's the case, it might be worthwhile to wrap this in it's own datatype.

bcaton85: How so?

kleesc: Similar to how we define datatypes in `data/registry_model/datatypes.py`. Say if at some point, we end up with some kind of dict as the value type.

But this should be fine as is for now.

kleesc (approved):
LGTM

endpoints/api/policy.py
bcaton85: yeah you're right, this will have to be apart of the future policy implementations.

PR #2399, opened 2023-10-10
autoprune: UI settings (PROJQUAY-6097)

bcaton85:
image


dmage (commented):
LGTM, but I'd like @Sunandadadi to review it too.

web/cypress/test/extra-config.yaml
```diff
@@ -0,0 +1 @@
+FEATURE_AUTO_PRUNE: true
```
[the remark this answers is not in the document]
bcaton85: We can enable it at some point, just leaving it disabled for now since it's still in development and I don't want to effect any other development efforts.

.github/workflows/CI.yaml
```diff
@@ -229,6 +229,12 @@ jobs:
         sleep 30
         make integration-test
 
+    - name: Apply extra config options
```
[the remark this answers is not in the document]
bcaton85: I had it there originally but the change kept getting reverted. I believe it's because of the checkout that happens right after the `Start Quay` step.

web/src/hooks/UseOrganization.ts
```diff
@@ -23,3 +30,122 @@ export function useOrganization(name: string) {
     organization,
   };
 }
+
+export function useNamespaceAutoPrunePolicies(
```
Sunandadadi, line 34: How about adding a new hook for all AutoPrune related functions? Might be better to separate it out.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/AutoPruning.tsx
```diff
@@ -0,0 +1,359 @@
...
+              <HelperTextItem>The method used to prune tags.</HelperTextItem>
+            </HelperText>
+          </FormHelperText>
+        </FormGroup>
+        <Conditional if={method === AutoPruneMethod.TAG_NUMBER}>
+          <FormGroup label="The number of tags to keep." fieldId="" isRequired>
+            <NumberInput
+              value={tagCount}
+              onMinus={() => {
+                tagCount > 0 ? setTagCount(tagCount - 1) : setTagCount(0);
```
Sunandadadi, line 261: If I save with number of tags as 0, I get a 400 error saying `Invalid Input`. I am guessing that [this](https://github.com/quay/quay/blob/master/data/model/autoprune.py#L46) line in the `valid_value` func is returning False. Mostly changing that line to `if value is None` should fix it.

bcaton85: hmm would we want users to specify a repo with 0 tags?

bcaton85: I instead added a check so a user can't input and empty value

Sunandadadi: Yes, the check works. I don't think we might need a repo with 0 tags.

Sunandadadi (approved):
lgtm

PR #2401, opened 2023-10-11
ui: Update all Patternfly deprecated components (PROJQUAY-6085)

jpuzz0:
### Summary
The following were updated to rid of legacy Patternfly components that will no longer exist in Patternfly v6:
- 25 `Dropdown`s (Many of which were Kebabs)
- 3 `Select`s (1 Typeahead - EntitySearch)
- 2 `Wizard`s (Both within modals)

(Table components were updated already as a part of the initial Patternfly v5 update PR)

As a result of the change, roughly half of the cypress tests had issues which were addressed mostly through updating selectors based on a newer DOM structure of the 3 types of components mentioned above.


jonathankingfc (approved):
LGTM

PR #2402, opened 2023-10-11
autoprune: background worker (PROJQUAY-6095)

bcaton85:
Implementation of the background worker for namespace autopruning.

Co-authored-by: @Sunandadadi


workers/autopruneworker.py
```diff
@@ -18,7 +24,31 @@ def __init__(self):
         self.add_operation(self.prune, POLL_PERIOD)
 
     def prune(self):
-        logger.info("Prune method not yet implemented...")
+        for _ in range(BATCH_SIZE):
```
dmage, line 27: Why do we need BATCH_SIZE, can we use `while True:`?

data/model/autoprune.py
```diff
@@ -218,3 +232,199 @@ def namespace_has_autoprune_task(namespace_id):
...
+
+        # When implementing repo policies, fetch repo policies and add it to the policies list here
+        list(
+            map(
+                lambda repo: execute_policies_for_repo(
+                    policies, repo, namespace_id, tag_page_limit
+                ),
+                repo_list,
+            )
+        )
```
dmage, line 427: Why do you use `list(map(...))`?

```suggestion
        for repo in repo_list:
            execute_policies_for_repo(
                policies, repo, namespace_id, tag_page_limit
            )
```

data/model/autoprune.py
```diff
@@ -218,3 +232,199 @@ def namespace_has_autoprune_task(namespace_id):
...
+        raise Exception(
+            f"Error deleting autoprune task for namespace id: {task.namespace_id} with error as: {str(err)}"
+        )
+
+
+def prune_repo_by_number_of_tags(repo, policy_config, namespace, tag_page_limit):
+    policy_method = policy_config.get("method", None)
+
+    if policy_method != AutoPruneMethod.NUMBER_OF_TAGS.value:
+        raise KeyError("Unsupported policy config provided", policy_config)
```
dmage, line 295: [KeyError](https://docs.python.org/3/library/exceptions.html#KeyError) is raised when a mapping (dictionary) key is not found in the set of existing keys.

This exception class does not apply here. Maybe ValueError?

data/model/autoprune.py
bcaton85: yeah we don't need that. I removed it.

data/model/autoprune.py
bcaton85: Agreed, added `InvalidNamespaceAutoPruneMethod` exception.

data/model/autoprune.py
```diff
@@ -4,18 +4,33 @@
...
+    repository,
+    user,
 )
 from data.model.user import get_active_namespace_user_by_username
 from util.timedeltastring import convert_to_timedelta
 
 logger = logging.getLogger(__name__)
+# Define a constant for the SKIP_LOCKED flag for testing purposes,
+# since we test with mysql 5.7 which does not support this flag.
+SKIP_LOCKED = True
```
kleesc, line 33: What does this statement do at the database level? What behavior on mysql can we expect when this is not used?

data/model/autoprune.py
```diff
@@ -218,3 +233,211 @@ def namespace_has_autoprune_task(namespace_id):
...
+    policy_method = policy_config.get("method", None)
+
+    if policy_method != AutoPruneMethod.NUMBER_OF_TAGS.value:
+        raise InvalidNamespaceAutoPruneMethod(
+            f"Expected prune method type {AutoPruneMethod.NUMBER_OF_TAGS.value} but got {policy_method}"
+        )
+
+    assert_valid_namespace_autoprune_policy(policy_config)
+
+    while True:
```
kleesc, line 306: Should this whole block be wrapped in a single transaction?

data/model/autoprune.py
```diff
@@ -218,3 +233,211 @@ def namespace_has_autoprune_task(namespace_id):
...
+
+    if policy_method != AutoPruneMethod.CREATION_DATE.value:
+        raise InvalidNamespaceAutoPruneMethod(
+            f"Expected prune method type {AutoPruneMethod.CREATION_DATE.value} but got {policy_method}"
+        )
+
+    assert_valid_namespace_autoprune_policy(policy_config)
+
+    time_ms = int(convert_to_timedelta(policy_config["value"]).total_seconds() * 1000)
+    while True:
```
kleesc, line 349: Likewise, should this be wrapped in a transaction, in order to avoid possible conflict with concurrent tags operations

data/model/autoprune.py
bcaton85: This appends `FOR UPDATE SKIP LOCKED` to the end of the select of the `autoprunetask`. It will basically prevent the autoprunetask row from being selected if a different worker has already selected it. If it's not used it's possible for multiple workers to pick up the same task.

data/model/autoprune.py
bcaton85: I'd prefer it wasn't actually. I'm thinking of instead just wrapping the select of the autoprune task in a transaction and updating the last_ran and status fields to running. And adding a check to not pick up any tasks that are running.

data/model/autoprune.py
bcaton85: yeah, we probably shouldn't do this, continuing the conversation in another thread

dmage (dismissed):

data/model/autoprune.py
kleesc: > If it's not used it's possible for multiple workers to pick up the same task

Would you need another locking mechanism in that case?

kleesc (approved):

data/model/autoprune.py
bcaton85: To prevent the scenario where two workers pick up and update the last_ran_ms at the same time.

PR #2413, opened 2023-10-17
autoprune: updating task status to running (PROJQUAY-6213)

bcaton85:
The task status should be updated to running when it is ran. Also dropping the level of the log for pruning a repository with a specific policy to debug to prevent flooding the logs.


Sunandadadi (approved):
lgtm

PR #2418, opened 2023-10-18
ui: Fix styling conflict with RedHatInsights (PROJQUAY-6085)

jpuzz0:
Using an older version of https://github.com/RedHatInsights/frontend-components, some styles in the patternfly v5 wizard were not being respected. Updating that package (and adding appropriate resolutions since it has a dependency on react v18), moving the import of patternfly css, and adding appropriate resolutions in the package.json seems to have remedied the styling conflicts.

Also, there was 1 page section that needed to remove padding still, which was leftover rework from the removal of deprecated patternfly components.

https://github.com/quay/quay/assets/96431149/16002d8a-2fe6-4727-adb6-01cecbfb0cc8

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -13,7 +13,7 @@
         "@patternfly/react-core": "^5.1.0",
         "@patternfly/react-icons": "^5.1.0",
         "@patternfly/react-table": "^5.1.0",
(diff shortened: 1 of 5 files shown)


harishsurf (approved):
lgtm!

PR #2425, opened 2023-10-20
autoprune: preventing prune of mirror or readonly repos (PROJQUAY-6235)

bcaton85:
Prevent prune of mirror or readonly repositories even if a auto-prune policy is configured.


Sunandadadi (approved):
lgtm

PR #2430, opened 2023-10-23
autoprune: validating input to autoprune policy (PROJQUAY-6230)

bcaton85:
Validates correct input on auto-prune policy creation.


Sunandadadi (approved):
lgtm

PR #2436, opened 2023-10-26
api: graceful error handling for robot acct already exists (PROJQUAY-6261)

Sunandadadi:
This PR takes care of graceful error handling instead of raising an exception for create robot account endpoint. Added the same error to propagate on the new UI too.

### DIFF
--- a/endpoints/api/robot.py
+++ b/endpoints/api/robot.py
@@ -121,12 +121,16 @@ class UserRobot(ApiResource):
         """
         parent = get_authenticated_user()
         create_data = request.get_json(silent=True) or {}
-        robot = model.create_user_robot(
-            robot_shortname,
-            parent,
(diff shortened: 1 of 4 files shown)


endpoints/api/robot.py
```diff
@@ -121,6 +122,12 @@ def put(self, robot_shortname):
         """
         parent = get_authenticated_user()
         create_data = request.get_json(silent=True) or {}
+
+        if model.robot_account_exists(robot_shortname, parent, user=True) is not None:
```
bcaton85, line 126: Is it better to catch the duplicate name error returned by `create_user_robot` and re-raise the request_error? That way we don't add any additional API requests. Same for the org endpoint.

bcaton85 (approved):
LGTM

PR #2437, opened 2023-10-26
marketplace: return None if subscription api times out (PROJQUAY-5363)

Marcusk19:
For all methods that make a request to the subscription watch API, if the request times out then there should be `None` returned.


kleesc (approved):
lgtm

kleesc (approved):

PR #2449, opened 2023-11-01
ui: Add form for repository state (PROJQUAY-5715)

jonathankingfc:
- Allow user to switch repository between normal, mirror, and read-only states


Sunandadadi (approved):
lgtm

PR #2459, opened 2023-11-06
storage: pass S3 region to S3Storage init (PROJQUAY-6375)

syed:
This is needed for correctly signing URLs for CloudFlare


Marcusk19 (approved):
LGTM

PR #2464, opened 2023-11-07
ui: fix search for bulk delete default permission modal (PROJQUAY-6369)

harishsurf:
Fix search for bulk delete default permission modal


bcaton85 (approved):
LGTM

PR #2469, opened 2023-11-10
ui: display manifest list size range (PROJQUAY-6393)

dmesser:
Instead of displaying `N/A` or `Unknown` in the size column, a tag pointing to a manifest list should display size information. Since clients by default select only a specific child manifest of a manifest list, I opted here for a display of the size of the smallest and the largest child manifest.

https://github.com/quay/quay/assets/12664117/b42c3e78-8ae1-4592-9262-41c4c5a9d768

### DIFF
--- a/web/src/components/Table/ManifestListSize.tsx
+++ b/web/src/components/Table/ManifestListSize.tsx
@@ -1,17 +1,17 @@
 import {Skeleton} from '@patternfly/react-core';
 import prettyBytes from 'pretty-bytes';
 import {Manifest} from 'src/resources/TagResource';
-import {useRecoilValueLoadable} from 'recoil';
+import {useRecoilValue} from 'recoil';
 import {childManifestSizeState} from 'src/atoms/TagListState';
 
 export default function ManifestListSize(props: ManifestListSizeProps) {
   const childManifestSizes = props.manifests.map((manifest) =>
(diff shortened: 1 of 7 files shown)


JustinXHale:
I like this

web/src/components/Table/ManifestListSize.tsx
```diff
@@ -0,0 +1,38 @@
+import {Skeleton} from '@patternfly/react-core';
+import prettyBytes from 'pretty-bytes';
+import {Manifest} from 'src/resources/TagResource';
+import {useRecoilValueLoadable} from 'recoil';
+import {childManifestSizeState} from 'src/atoms/TagListState';
+
+export default function ManifestListSize(props: ManifestListSizeProps) {
+  const childManifestSizes = props.manifests.map((manifest) =>
+    useRecoilValueLoadable(childManifestSizeState(manifest.digest)),
```
bcaton85, line 9: do we have to use `useRecoilValueLoadable` instead of `useRecoilValue` here? IIUC `useRecoilValueLoadable` is for asynchronous operations happening within the atom.

dmesser: I think I misunderstood this then, you are right, there is no async operation inside the atom, it just gets mutated asynchronously. Will refactor.

PR #2505, opened 2023-11-27
federationuser(ldap)!: avoid doing LDAP lookups for Robot accounts (PROJQUAY-5137)

michaelalang:
FYI, this is a `re-submit` of the original PROJQUAY-5137 since the code for dropping LDAP-robot lookups has not been merged so far.

BREAKING CHANGE: LDAP lookup of robot accounts in the UI for granting
permission has been dropped in context of permissions granting. This impacts Users from LDAP if they have not logged in to Quay already (pre provisioning) as a Federation User will only be available after logging in or being part of a Team.

the functions:

* _ldap_single_user_search
* _ldap_user_search

have been updated to provide context for dropping the LDAP query. As the functions are used during login as well as during setting permissions it is mandatory to be able to distinguish the context.

the functions:

* get_nonrobot_user
* find_user_by_email

from the module data.model.user have been utilized to identify, if the user should be looked up in LDAP or not.

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -8,6 +8,7 @@ from ldap.filter import escape_filter_chars, filter_format
 
 from data.users.federated import FederatedUsers, UserInformation
 from util.itertoolrecipes import take
+from data.model.user import get_nonrobot_user, find_user_by_email
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


michaelalang:
https://github.com/quay/quay/pull/2635 needs to be incorprated into this PR to pass unittests.

data/users/externalldap.py
```diff
@@ -276,6 +283,13 @@ def _ldap_user_search(
         if not username_or_email:
             return (None, "Empty username/email")
 
+        try:
+            lookup_robot(username_or_email)
+            return get_username(username_or_email)
```
bcaton85: It doesn't look like the calls to `_ldap_user_search` or `_ldap_single_user_search` expect a user(robot) table row to be returned here. It looks like this contexts that call this method will break if a robot is returned here?

michaelalang: ack true .... it didn't show because it fails the process anyway but you are right ... will fix it ASAP

bcaton85:
3.14 cherrypick https://github.com/quay/quay/pull/4147

PR #2511, opened 2023-11-27
database: use psycopg2 instead of psycopg2-binary (PROJQUAY-6453)

syed:
psycopg2-binary is for dev only in prod we should use psycopg2


Sunandadadi (approved):
lgtm

PR #2515, opened 2023-11-29
api: add caching for get_repository method (PROJQUAY-6472)

Sunandadadi:
Adds caching to `model.repository.get_repository` func called from api/v2/manifest endpoints.


syed (approved):
LGTM

PR #2519, opened 2023-12-01
federatedusers(externalldap): report Exceptions with AD integration (PROJQUAY-6481)

michaelalang:
During debugging AD integration which do look correct from configuration POV, we see logs within Quay that do not report the reason/explanation of an Exception which is raised during authenticating users.

```
[DEBUG] [data.users.externalldap] LDAP search exception
```

The fix is trivial and reporting in the logs the Exception code and reason.

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -248,8 +248,9 @@ class LDAPUsers(FederatedUsers):
                 logger.debug("LDAP referral search exception")
                 return (None, "Username not found")
 
-        except ldap.LDAPError:
+        except ldap.LDAPError as ldaperr:
             logger.debug("LDAP search exception")
+            logger.debug(str(ldaperr))
             return (None, "Username not found")
 
     def _ldap_user_search(


ibazulic (approved):
LGTM!

data/users/externalldap.py
```diff
@@ -248,8 +248,9 @@ def _ldap_user_search_with_rdn(
                 logger.debug("LDAP referral search exception")
                 return (None, "Username not found")
 
-        except ldap.LDAPError:
+        except ldap.LDAPError as ldaperr:
             logger.debug("LDAP search exception")
+            logger.debug(str(ldaperr))
```
jbpratt, line 253: Should be able to do something like which is a bit cleaner

```
        except ldap.LDAPError:
            logger.debug("LDAP search exception", exc_info=True)
```

https://docs.python.org/3/library/logging.html#logging.Logger.debug

> If exc_info does not evaluate as false, it causes exception information to be added to the logging message. If an exception tuple (in the format returned by [sys.exc_info()](https://docs.python.org/3/library/sys.html#sys.exc_info)) or an exception instance is provided, it is used; otherwise, [sys.exc_info()](https://docs.python.org/3/library/sys.html#sys.exc_info) is called to get the exception information.

jbpratt:
@michaelalang could you rebase this against master and we can get the tests running?

data/users/externalldap.py
michaelalang: @jbpratt yeap agreed and in https://github.com/quay/quay/pull/2761 we'll fix it to that syntax anyway ...

michaelalang:
In favor of adding LDAP Exception handling and logging with PR https://github.com/quay/quay/pull/2761 we'll drop this single liner.

PR #2521, opened 2023-12-01
cache: adding caching on look_up repository (PROJQUAY-6472)

Sunandadadi:
This PR caches the response of `model.repository.get_repository` as a json object instead of picking as done in https://github.com/quay/quay/pull/2515.


syed (approved):
lgtm

PR #2524, opened 2023-12-01
api: adding nickname decorator to autoprune policy endpoints (PROJQUAY-6483)

bcaton85:
Allows for the HTTP methods for the auto-prune policy endpoints to be returned by the  `/api/v1/discovery` endpoint.


Sunandadadi (approved):
lgtm

PR #2540, opened 2023-12-04
sso: adding error log for export compliance (PROJQUAY-6486)

Sunandadadi:
Adding error log instead of raising exception when we get unknown response/error from export compliance service. `exc_info` should send the error log to sentry.

### DIFF
--- a/oauth/services/rhsso.py
+++ b/oauth/services/rhsso.py
@@ -61,8 +61,12 @@ class RHSSOOAuthService(OIDCLoginService):
                 # also any issues with reaching the export
                 # compliance API should trigger this
                 # Adding error log to capture the error on sentry without causing quay.io logins to fail
+                msg = ""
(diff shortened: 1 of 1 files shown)


oauth/services/rhsso.py
```diff
@@ -60,6 +60,13 @@ def exchange_code_for_login(self, app_config, http_client, code, redirect_suffix
                 # This generates a generic OAUTH error page
                 # also any issues with reaching the export
                 # compliance API should trigger this
-                raise OAuthLoginException(str(e))
+                # Adding error log to capture the error on sentry without causing quay.io logins to fail
+                logger.error(
+                    f"Got unknown response from export compliance service: {result.text}, with status_code: {result.status_code}"
```
syed, line 65: `result` can be `None` this can cause another exception

PR #2553, opened 2023-12-11
ui: show UI toogle to all users in quay.io (PROJQUAY-6505)

syed:
Removes the check that limits the UI toggle to RedHat users


bcaton85 (approved):
LGTM

PR #2557, opened 2023-12-14
WIP Reimplement rehash to support OpenSSL 3

dmage:


rehash-openssl/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        let md_data = state
+            .get("md_data")
+            .ok_or_else(|| {
+                PyErr::new::<pyo3::exceptions::PyValueError, _>("state does not contain md_data")
+            })?
+            .extract::<&[u8]>(py)?;
+        let len = md_data.len();
+        const SHA256_SIZE: usize = std::mem::size_of::<openssl::sha::Sha256>();
+        const POINTER_SIZE: usize = 8;
```
hdonnay, line 69: This probably wants to be a `std::mem::size_of::<*openssl::sha::Sha256>();`? Unless the API has it hard-coded to `8` instead of being the machine's pointer size.

rehash-openssl/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_sha256_size() {
+        // As we want future versions of rehash-openssl to load states from older versions, we want
+        // to detect when its internal implementation changes. The size of the object isn't the
+        // best way to do this, but it's better than nothing.
```
hdonnay, line 97: This could also check that a memcmp of the raw bytes looks as expected to get the test to fail.

rehash-openssl/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+            )
+        };
+        // return a dict that is compatible with rehash
+        let mut dict: HashMap<String, PyObject> = HashMap::new();
+        dict.insert("name".to_string(), "sha256".into_py(py));
+        dict.insert("md_data".to_string(), state.into_py(py));
+        Ok(dict)
+    }
+
+    fn __setstate__(&mut self, py: Python, state: HashMap<String, PyObject>) -> PyResult<()> {
```
hdonnay, line 47: After talking with some rustaceans, it sounds like this could eventually use a pure-rust implementation that officially supports a resumption API. With that in mind, it might make sense to start adding a key that indicates the format of the serialized state to enable a seamless upgrade in the future.

syed:
fixed in #3780

PR #2570, opened 2023-12-19
ui: adding build trigger table (PROJQUAY-6295)

bcaton85:
Adds table to view build triggers. Includes the the following:
- Ability to enable/disable trigger through in-row options
- Ability to delete trigger
- Ability to view the trigger credentials
- Display's and gives option to delete partially intialized triggers, similar to the angular UI
- Display's triggers that have been automatically disabled and gives a prompt allowing the user to re-enable the trigger


Sunandadadi (approved):
lgtm

PR #2574, opened 2023-12-27
secscan: CVE suppression (PROJQUAY-6567)

dmesser:
This PR adds UI and API capabilities to selectively suppress vulnerabilities in secscan reports.

- CVE suppression occurs entirely in Quay and are gated by a feature toggle `SECURITY_VULNERABILITY_SUPPRESSION` which is `true` by default
- an organization admin can define a list of CVE identifiers, which are then suppressed in any CVE report of any manifest in any repository in that organization
- a repository admin can define a list of CVE identifiers at the repository level, which are then suppressed in any CVE report of any manifest in that repository
- a repository admin can define a list of CVE identifiers in the context of a manifest, which are then suppressed  in the CVE report of the manifest
- the security API endpoint in Quay will, by default, omit suppressed CVEs but can be called in a way where suppressed CVEs are part of the report (`suppressions` parameter)
- the manifest API got a new endpoint that allows to set/unset the a list of CVE identifiers for suppression (`/suppressed_vulnerabilities)
- the organization API now accepts a property called `suppressed_vulnerabilities` as a way to specify org-level CVE suppression (`UpdateOrg`) and also returns a property called `suppressed_vulnerabilities` containing them (`getOrganization`)
- the repository API (`updateRepo`) now accepts a property called `suppressed_vulnerabilities` as a way to specify repo-level CVE suppression and also returns a property called `suppressed_vulnerabilities` containing them (`getRepo`)
- CVE reports can optionally show suppressed CVEs in the UI but don't do that by default (needs to be possible in API and UI)
- CVE reports in the UI by default show the number of suppressed CVEs in summary sections
- if different CVE identifiers are configured to be suppressed at the organization and/or repository and/or manifest level, they will combined into a larger set that will be used to filter the vulnerability report of a manifest
- suppressed CVEs never cause or are included in Repository notifications
- modifications of CVE suppression settings are logged in the action logs
- CVE identifiers are entered as alphanumeric strings without spaces
- when retrieving a vulnerability report, filtering for suppressed CVEs is done by matching the configured CVE identifiers as substrings against the CVE identifiers in the original report from Clair
- CVE suppression is only implemented in the new UI, the old UI will only show reported filtered for configured CVE suppressions

Screenshot 2023-12-27 at 12 46 54
Screenshot 2023-12-27 at 12 46 19
Screenshot 2023-12-27 at 12 44 37
Screenshot 2023-12-27 at 12 43 56
Screenshot 2023-12-27 at 12 43 06


ibazulic:
I am not sure that the point of the registry is to manipulate with image metadata. It is my strong belief that we should provide complete and concize information about the image, particularly related to security reports.  If some filtering needs to be done, it can be done after a security report has already been fetched via the secscan API (the returned object is a JSON anyway so filtering should be easy to do). Another point I'm not certain of is the ability to define CVEs to ignore on both org and repo levels. What happens if on the org level an admin of the org defines a CVE that should be filtered out but the repo owner needs that info to fix a bug in its own code? Finally, what happens if a CVE changes level? We have seen in the past that certain trivial CVEs were upgraded to a more severe level, who will track those changes and how will it/they track them if the CVE is ignored?

dmesser:
@ibazulic Thanks for your comments. This is indeed a feature requested by users. The ability to suppress reported items is also very common in security tooling.
The logic for Quay is that some CVEs are disputed, and this feature gives users a persistent way to exclude them from the reports. The security API is the source of these reports and, therefore, the best place to do so because it impacts both the UI and repository notifications. I guess your suggestion is to have everyone do it outside of Quay? I would say that this would miss out on user expectations who want to rely on Quay as the source of reports.

While I can see your point that we'd want to report the most concise information at all times, ultimately, this puts the user in control. The feature can be disabled, preventing users from being able to create exclusions. I don't think the use case here is to exclude trivial CVEs but rather those CVEs which are false positives because of mismatches in the scanner or simply because they don't apply. We would ideally fix the former, but depending on the root cause, this can take significant effort and time. To prevent security teams from having to investigate every time a false positive pops up (especially when they are in base images), the CVE suppression puts an effective measure in place to declare this particular CVE to be safe to be ignored. The latter case can happen because the scanner leverages upstream vulnerability databases, which match against components / packages / imports found in an image, but the author of the image decided that there is no code path to exploit the vulnerability.

Regarding the repo owner wanting to overrule the org owner, I would say that in our permission hierarchy the org owner is more powerful than the repo owner, if they are indeed two different individuals. The concept is also employed with auto-pruning and our general access permissions. In a conflict situation, the org owner could be asked to remove the exclusion at the org level, but a repo owner should never be able to overrule an org setting. I think the usage scenario for org-level suppressions is more about CVEs reported by multiple advisories or suppressing an entire class of advisories, most likely caused by something in a commonly used base image.

Hope this makes sense.

ibazulic:
@dmesser 

> The logic for Quay is that some CVEs are disputed, and this feature gives users a persistent way to exclude them from the reports.

Question is for whom. One CVE that is disputed for you might not be disputed for me. In fact, it could show a problem in my own code. And now I cannot see that CVE because the admin of the org masked that CVE out. So I effectively lose information.

> I guess your suggestion is to have everyone do it outside of Quay?

Correct. In my opinion, the purpose of a registry is to serve images and to provide the most correct and accurate metadata about an image that the registry at some point has. This includes unmodified security reports (what the scanner returns), but is not limited to just that (everything that can be considered as image metadata from date created to build artifacts should be reported by the registry, if that data exists).

> While I can see your point that we'd want to report the most concise information at all times, ultimately, this puts the user in control. The feature can be disabled, preventing users from being able to create exclusions.

Well, no, it actually limits and removes control. I don't see a way to disable it, apart as a feature flag, most users of Quay do not have access to feature flags because they don't have access to Quay's `config.yaml` file. If it *is* enabled and an org admin *does* define a list of CVEs, as it stands now (and correct me if I'm wrong here) there is no way for the user in Quay to know which CVEs are suppressed on his own repo because supporession of CVEs on org level is allowed only to org admins. If global CVE suppressions are allowed, then they must be printed out on the page where repo suppressions should be, in a separate box so that a user knows what was defined.

> I don't think the use case here is to exclude trivial CVEs but rather those CVEs which are false positives because of mismatches in the scanner or simply because they don't apply.

If you allow org wide CVE suppressions, you actually decide in advance what false positives might be. And you cannot make that assumption. 

> The latter case can happen because the scanner leverages upstream vulnerability databases, which match against components / packages / imports found in an image, but the author of the image decided that there is no code path to exploit the vulnerability.

This to me sounds like better heuristics are needed on the detection side. What exactly is in the image. The author of the image *might* decide at this moment that a specific CVE is irrelevant to him bht that might change in the future, as I have said in my comment. If a CVE level suddenly changes, what will inform the user about that change if they are suppressing that particular CVE?

> Regarding the repo owner wanting to overrule the org owner, I would say that in our permission hierarchy the org owner is more powerful than the repo owner, if they are indeed two different individuals. The concept is also employed with auto-pruning and our general access permissions. In a conflict situation, the org owner could be asked to remove the exclusion at the org level, but a repo owner should never be able to overrule an org setting. I think the usage scenario for org-level suppressions is more about CVEs reported by multiple advisories or suppressing an entire class of advisories, most likely caused by something in a commonly used base image.

Org owners oould be managers, people who are non-technical. How would they have the power to decide what CVE is important and what isn't? CVEs are not an org issue. CVEs, if they do exist, are image issues, which means the only person who should decide what to suppress are repo owners. The comparison with quotas and auto pruning is, in my opinion, not valid here, since both quotas and auto pruning (which can to some extent be viewed as an extension of the quota system) are properties of the organization or an individual user. Org admins already have access to all repositories inside the organization, so they can change repo quotas as they see fit.

dmesser:
The feature can be disabled via a flag, or people can decide not to use it. It seems like your main points are that it could be used wrong and specifically, because the org owner could be an entirely different persona, even non-technical, and be somewhat removed from the actual content stored in the repos.
In my opinion, this is not the case for most scenarios. Members of the same team that owns the content inside the org typically also manage it, and they have some relation to the content. In case they do not, I would agree with you. They have no business in setting CVE suppressions or pruning rules, as they all depend on the content type stored. But that doesn't mean that org-level control features like this are not useful in other scenarios. For instance where you want to disable an entire class of advisories or a specific CVE that you know is always going to be a false positive because it comes from a base image and all your images leverage the same base image. But yes, you would need to know what you are doing.

Your scenario where a CVE level changes does not track for me. If I am publishing my software on quay.io, and it gets flagged for CVE in a library that is not used at runtime in my app or only in a way where that CVE is not exploitable at all, then the CVE rating update does not change anything about that. If it turns out that the vulnerability can be exploited in another way, that would typically be a new CVE? For example, like in the history of log4j: https://news.apache.org/foundation/entry/apache-log4j-cves

PR #2580, opened 2024-01-03
ui: adding build avatar image size (PROJQUAY-6575)

bcaton85:
Sets the image size for avatars shown in the build history table. 

How to test:
1. Run the `Displays build status` test in `web/cypress/e2e/builds.cy.ts`
2. Verify the row with build id `build008` displays the avatar with the correct size


harishsurf (approved):
lgtm!

PR #2582, opened 2024-01-03
marketplace: update reconciliationworker to use webCustomerId instead of ebsAccountNumber (PROJQUAY-233)

Marcusk19:
* fix reconciler where it was incorrectly using the ebsAccountNumber to create subscriptions
* add job to reconciler so that it reconciles different ids between the database and the user api
* separate skus to be used by billing and skus to be used by reconciler

### DIFF
--- a/data/billing.py
+++ b/data/billing.py
@@ -348,7 +348,8 @@ PLANS = [
     },
 ]
 
-RH_SKUS = [plan["rh_sku"] for plan in PLANS if plan.get("rh_sku") is not None]
+RH_SKUS = ["MW02701"]
+RECONCILER_SKUS = [plan["rh_sku"] for plan in PLANS if plan.get("rh_sku") is not None]
 
 
 def get_plan(plan_id):
--- a/data/billing.py
+++ b/data/billing.py
@@ -223,6 +223,7 @@ PLANS = [
         "privateRepos": 5,
         "stripeId": "personal-2018",
         "rh_sku": "MW00584MO",
+        "sku_billing": False,
         "audience": "Individuals",
         "bus_features": False,
         "deprecated": False,
@@ -235,6 +236,7 @@ PLANS = [
         "price": 3000,
         "privateRepos": 10,
         "rh_sku": "MW00585MO",
+        "sku_billing": False,
         "stripeId": "bus-micro-2018",
         "audience": "For startups",
         "bus_features": True,
@@ -248,6 +250,7 @@ PLANS = [
         "price": 6000,
         "privateRepos": 20,
         "rh_sku": "MW00586MO",
+        "sku_billing": False,
         "stripeId": "bus-small-2018",
         "audience": "For small businesses",
         "bus_features": True,
@@ -261,6 +264,7 @@ PLANS = [
         "price": 12500,
         "privateRepos": 50,
         "rh_sku": "MW00587MO",
+        "sku_billing": False,
         "stripeId": "bus-medium-2018",
         "audience": "For normal businesses",
         "bus_features": True,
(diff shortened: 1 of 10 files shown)


data/billing.py
```diff
@@ -348,7 +348,8 @@
     },
 ]
 
-RH_SKUS = [plan["rh_sku"] for plan in PLANS if plan.get("rh_sku") is not None]
+RH_SKUS = ["MW02701"]
+RECONCILER_SKUS = [plan["rh_sku"] for plan in PLANS if plan.get("rh_sku") is not None]
```
kleesc, line 352: Do these need to be separate?

util/marketplace.py
```diff
@@ -285,6 +313,8 @@ def __init__(self):
         self.subscription_created = False
 
     def lookup_subscription(self, customer_id, sku_id):
+        if customer_id == TEST_USER["account_number"] and sku_id == "MW02701":
```
kleesc, line 316: MW02701 is hardcoded to `RH_SKUS`

util/marketplace.py
```diff
@@ -294,24 +324,12 @@ def extend_subscription(self, subscription_id, end_date):
         self.subscription_extended = True
 
     def get_subscription_sku(self, subscription_id):
-        if id == 12345:
-            return "FakeSku"
+        valid_ids = [subscription["id"] for subscription in TEST_USER["subscriptions"]]
+        if subscription_id in valid_ids:
+            return "MW02701"
```
kleesc, line 329: MW02701 is hardcoded to `RH_SKUS`

data/billing.py
Marcusk19: Yeah, because the old skus will be used to track stripe customers in the RH portal, and the new skus will be what is used for binding to orgs, etc.

util/marketplace.py
Marcusk19: I can add to the `PLANS` list so that it doesn't have to be hardcoded, but was thinking I would do that in a separate PR so it doesn't clutter this one

data/model/entitlements.py
```diff
@@ -6,16 +6,37 @@
...
+            {RedHatSubscriptions.account_number: web_customer_id}
+        ).where(RedHatSubscriptions.user_id == user.id)
+        query.execute()
+    except model.DataModelException as ex:
+        logger.error("Problem updating customer id for %s: %s", user.username, ex)
+
+
+def remove_web_customer_id(user, web_customer_id):
+    try:
+        id = RedHatSubscriptions.get(
```
kleesc, line 36: nit: Can you rename this? `id` is also a built-in function

workers/reconciliationworker.py
```diff
@@ -48,21 +47,39 @@ def _perform_reconciliation(self, user_api, marketplace_api):
...
 
-            # go to user api if no ebsAccountNumber is found
-            if ebsAccountNumber is None:
-                logger.debug("Looking up ebsAccountNumber for email %s", email)
-                ebsAccountNumber = user_api.lookup_customer_id(email)
-                logger.debug("Found %s number for %s", str(ebsAccountNumber), user.username)
-                if ebsAccountNumber:
-                    entitlements.save_ebs_account_number(user, ebsAccountNumber)
+            # check against user api
+            logger.debug("Looking up webCustomerId for email %s", email)
```
kleesc, line 56: nit: A little redundant. You can just add the email to the following log statement

util/marketplace.py
kleesc: I meant `MW02701` is already defined in `data/billing.py`. You could just import that variable instead of setting `MW02701` in multiple places

kleesc (approved):

PR #2590, opened 2024-01-08
marketplace: make sure customer id from api is returned as an int (PROJQUAY-233)

Marcusk19:
Make sure when parsing data from user api that the customer id is returned as an integer type. Fixes problem with reconciler where it thinks two of the same id are different.


kleesc (dismissed):
LGTM

kleesc (approved):

PR #2595, opened 2024-01-09
billing: marketplace UI (PROJQUAY-6551)

Marcusk19:
adds UI in billing section for managing user and org-bound skus
add new skus for bulk purchase orders
change api endpoints to allow bulk attaching skus to orgs


endpoints/api/billing.py
```diff
@@ -1019,7 +1058,7 @@ def delete(self, orgname, subscription_id):
                 return ("Organization not valid", 400)
 
             model.organization_skus.remove_subscription_from_org(organization.id, subscription_id)
-            return ("Deleted", 204)
+            return ("Deleted", 200)
```
kleesc, line 1061: Any reason to change from 204 to 200?

endpoints/api/billing.py
```diff
@@ -1040,7 +1079,7 @@ def get(self):
         user = get_authenticated_user()
         account_number = marketplace_users.get_account_number(user)
         if not account_number:
-            raise NotFound()
+            return []
```
kleesc, line 1082: Likewise, why go from 404 to 200? The client should be able to handle this.
An empty list should be returned in the case where the account number exists, but no subscriptions is attached to that user

endpoints/api/billing.py
Marcusk19: I was having trouble updating the frontend component when removing a sku because 204 returned an empty response

endpoints/api/billing.py
Marcusk19: Same reason as above, the billing ui would look like it was indefinitely loading when the api returned the 404. I found a way to properly handle the frontend though and updated the response codes accordingly

kleesc (approved):

PR #2601, opened 2024-01-13
storage: Fix big layer uploads for Ceph/RADOS driver (PROJQUAY-6586)

ibazulic:
Current uploads of large images usually fail on Ceph/RADOS compatible implementations (including Noobaa) because during the last assembly, copy is done all at once. For large layers, this takes a long while and Boto times out. With this patch, we limit the size of the used chunk to 32 MB so the final copy is done in parts of up to 32 MB each. The size can be overridden by specifying the parameter `maximum_chunk_size_mb` in the driver settings, for example:

~~~
DISTRIBUTED_STORAGE_CONFIG:
    default:
        - RadosGWStorage
        - ...
           maximum_chunk_size_mb: 100
~~~

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -990,8 +990,8 @@ class RadosGWStorage(_CloudStorage):
         connect_kwargs = {
             "endpoint_url": _build_endpoint_url(hostname, port=port, is_secure=is_secure),
             "config": Config(
-                connect_timeout=600 if server_side_assembly == False else 60,
-                read_timeout=600 if server_side_assembly == False else 60,
+                connect_timeout=600 if not server_side_assembly else 60,
+                read_timeout=600 if not server_side_assembly else 60,
             ),
         }
 
@@ -1032,8 +1032,8 @@ class RadosGWStorage(_CloudStorage):
         return super(RadosGWStorage, self).get_direct_upload_url(path, mime_type, requires_cors)
 
     def complete_chunked_upload(self, uuid, final_path, storage_metadata):
-        logger.debug("Multipart upload is set to {}.".format(self.server_side_assembly))
-        if self.server_side_assembly == True:
+        logger.debug("Server side assembly is set to {}.".format(self.server_side_assembly))
+        if self.server_side_assembly:
             logger.debug("Initiating multipart upload and server side assembly for final push.")
             return super(RadosGWStorage, self).complete_chunked_upload(
                 uuid, final_path, storage_metadata
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -984,10 +984,15 @@ class RadosGWStorage(_CloudStorage):
         bucket_name,
         port=None,
         maximum_chunk_size_mb=None,
+        multipart_upload=True,
     ):
         upload_params = {}
         connect_kwargs = {
             "endpoint_url": _build_endpoint_url(hostname, port=port, is_secure=is_secure),
+            "config": Config(
+                connect_timeout=600 if multipart_upload == False else 60,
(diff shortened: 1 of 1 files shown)


storage/cloud.py
```diff
@@ -1018,15 +1024,6 @@ def get_direct_upload_url(self, path, mime_type, requires_cors=True):
 
         return super(RadosGWStorage, self).get_direct_upload_url(path, mime_type, requires_cors)
 
-    def complete_chunked_upload(self, uuid, final_path, storage_metadata):
```
bcaton85, line 1021: To my knowledge the RADOS driver is used as a catch all for other storage engines that aren't explicitly supported, won't this break for users using this driver for storage engines that don't support multipart uploads? FWIW this probably isn't a very common scenario.

ibazulic: Added a compatiblity switch to the driver: if `multipart_upload: true` then we force server side assembly and push of the end file in chunks of 32 MB (or whatever the user tries). If `multipart_upload: false`, we fall back to client side assembly, but in order to accomodate big layer pushes, we increase `boto` timeout to 600 seconds (instead of default 60).

ibazulic:
For backward compatibility, a new parameter has been added to the `RadosGW` config:

a) if `server_side_assembly = True` then we force the end layer to be pushed to the blob tree in chunks of at most `maximum_chunk_size_mb` (or at most 32 MB if the parameter is missing).
b) if `server_side_assembly = False` we force client side assembly of pushed layer. To support pushes of large layers, we also change Boto timeout to 600 seconds instead of 60.

Example configuration:

~~~
DISTRIBUTED_STORAGE_CONFIG:
    rados:
        - RadosGWStorage
        - access_key: 593N1F05GJWJO1AFT1WG
          bucket_name: quay
          hostname: 10.0.43.173
          is_secure: false
          port: "9000"
          secret_key: nlTLb5kSVxxs826AaAKs7PEkJ3PJT6YsFCd7PUei
          storage_path: /datastorage/registry
          maximum_chunk_size_mb: 100
          server_side_assembly: true
~~~

If `server_side_assembly` is missing, we presume it to be `true`.

storage/cloud.py
```diff
@@ -1019,13 +1032,21 @@ def get_direct_upload_url(self, path, mime_type, requires_cors=True):
...
     def complete_chunked_upload(self, uuid, final_path, storage_metadata):
-        self._initialize_cloud_conn()
-
-        # RadosGW does not support multipart copying from keys, so we are forced to join
-        # it all locally and then reupload.
-        # See https://github.com/ceph/ceph/pull/5139
-        chunk_list = self._chunk_list_from_metadata(storage_metadata)
-        self._client_side_chunk_join(final_path, chunk_list)
+        logger.debug("Multipart upload is set to {}.".format(self.server_side_assembly))
+        if self.server_side_assembly == True:
```
bcaton85, line 1036: nit: can leave as `if self.server_side_assembly:`

storage/cloud.py
```diff
@@ -983,10 +983,16 @@ def __init__(
...
         bucket_name,
         port=None,
+        maximum_chunk_size_mb=None,
+        server_side_assembly=True,
     ):
         upload_params = {}
         connect_kwargs = {
             "endpoint_url": _build_endpoint_url(hostname, port=port, is_secure=is_secure),
+            "config": Config(
+                connect_timeout=600 if server_side_assembly == False else 60,
```
bcaton85, line 993: nit: can be `connect_timeout=600 if not server_side_assembly else 60,`

storage/cloud.py
ibazulic: Fixed.

PR #2613, opened 2024-01-17
marketplace: return array of valid subscriptions when looking up subscription (PROJQUAY-6551)

Marcusk19:
Currently `lookup_subscription` only returns the most recent subscription found in subscription watch. For the new skus it should return an array of valid subscriptions.

### DIFF
--- a/util/marketplace.py
+++ b/util/marketplace.py
@@ -111,19 +111,20 @@ class RedHatSubscriptionApi(object):
             return None
 
         try:
-            subscription = max(
-                json.loads(r.content), key=lambda i: (i["effectiveEndDate"]), default=None
-            )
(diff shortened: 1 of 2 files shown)


util/marketplace.py
```diff
@@ -111,19 +111,20 @@ def lookup_subscription(self, webCustomerId, skuId):
             return None
 
         try:
-            subscription = max(
-                json.loads(r.content), key=lambda i: (i["effectiveEndDate"]), default=None
-            )
+            subscriptions = json.loads(r.content)
```
syed, line 114: is this a list or a single dict?

Marcusk19: this is a list from the api

syed (approved):
lgtm

PR #2616, opened 2024-01-18
cdn: add namespace and username to CDN redirect for usage calculation (PROJQUAY-5939)

syed:
We use the referer header to infer the namespace but that is not always reliable as some clients don't add that header when sending the request to the CDN


kleesc (approved):
LGTM

PR #2627, opened 2024-01-22
api: Return layer compressed size with manifest request

ibazulic:
Previously, we weren't returning the total compressed layer size (in bytes) which is part of the manifest table. This makes it cumbersome to identify the real size of an image via `api/v1/repository/{LOCATION}/manifest/{manifestref}` endpoint since only individual layer sizes were returned and the total would then need to be manually summed over each individual layer. With this change, the API endpoint will return the compressed size of all layers if we're referencing a real image, if a reference points to an OCI index or a manifest list, the returned size will be 0.


syed (approved):
lgtm

PR #2633, opened 2024-01-24
marketplace: add support for quantity from subscriptions api (PROJQUAY-6551)

Marcusk19:
Adds handling for when a subscription returned from the subscription watch api has a `quantity` greater than 1. Number of private repos should be correctly calculated using the quantity. 

Updates ui so that subscriptions can only be added to an org as a group, i.e. a subscription with `quantity` = 2 cannot be split across organizations.

### DIFF
--- a/data/migrations/versions/41d15c93c299_add_quantity_field_to_orgrhskus.py
+++ b/data/migrations/versions/41d15c93c299_add_quantity_field_to_orgrhskus.py
@@ -15,9 +15,7 @@ import sqlalchemy as sa
 
 def upgrade(op, tables, tester):
     op.add_column("organizationrhskus", sa.Column("quantity", sa.Integer(), nullable=True))
-    pass
 
 
 def downgrade(op, tables, tester):
     op.drop_column("organizationrhskus", "quantity")
-    pass
--- a/data/migrations/versions/41d15c93c299_add_quantity_field_to_orgrhskus.py
+++ b/data/migrations/versions/41d15c93c299_add_quantity_field_to_orgrhskus.py
@@ -1,0 +1,23 @@
+"""add quantity field to orgRhSkus
+
+Revision ID: 41d15c93c299
+Revises: 3f8e3657bb67
+Create Date: 2024-01-24 11:19:19.095256
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = "41d15c93c299"
+down_revision = "3f8e3657bb67"
+
+import sqlalchemy as sa
+
+
+def upgrade(op, tables, tester):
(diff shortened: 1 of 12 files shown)


data/migrations/versions/41d15c93c299_add_quantity_field_to_orgrhskus.py
```diff
@@ -0,0 +1,23 @@
...
+# revision identifiers, used by Alembic.
+revision = "41d15c93c299"
+down_revision = "3f8e3657bb67"
+
+import sqlalchemy as sa
+
+
+def upgrade(op, tables, tester):
+    op.add_column("organizationrhskus", sa.Column("quantity", sa.Integer(), nullable=True))
+    pass
```
kleesc, line 18: Don't need 'pass`

data/migrations/versions/41d15c93c299_add_quantity_field_to_orgrhskus.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def downgrade(op, tables, tester):
+    op.drop_column("organizationrhskus", "quantity")
+    pass
```
kleesc, line 23: Likewise

endpoints/api/billing.py
```diff
@@ -980,6 +984,9 @@ def post(self, orgname):
                 subscription_id = subscription.get("subscription_id")
                 if subscription_id is None:
                     break
+                # quantity = subscription.get("quantity")
+                # if quantity is None:
+                #     quantity = 1
```
kleesc, line 989: Can you remove this if this is not needed

endpoints/api/billing.py
```diff
@@ -989,9 +996,16 @@ def post(self, orgname):
 
                 user_subscription_ids = [int(subscription["id"]) for subscription in subscriptions]
                 if int(subscription_id) in user_subscription_ids:
+                    quantity = 1
```
kleesc, line 999: I'm assuming this means that if a subscription is returned from the marketplace, there's at least one of it.
Is there a case where this could be 0?

endpoints/api/billing.py
```diff
@@ -989,9 +996,16 @@ def post(self, orgname):
...
+                    for subscription in subscriptions:
+                        if subscription["id"] == subscription_id:
+                            quantity = subscription["quantity"]
```
kleesc, line 1002: You probably want to add a `break` if this matches correct?

endpoints/api/billing.py
Marcusk19: I don't think so, a quantity of 0 should just mean the subscription is terminated anyways I would imagine

kleesc (approved):

PR #2647, opened 2024-02-02
manifest: support empty config for artifacts (PROJQUAY-6658)

syed:
Artifacts can have empty config defined
skip parsing if the config is empty


arborite-rh (dismissed):
seems reasonable to me

arborite-rh (approved):

PR #2657, opened 2024-02-05
oci: remove platform requirement for image index (PROJQUAY-6658)

syed:
The image index spec https://github.com/opencontainers/image-spec/blob/main/image-index.md
    specifics `platform` as an optional field.


ibazulic (approved):
LGTM.

PR #2658, opened 2024-02-06
proxy: Add a separate configuration option for repo visibility in case of proxy registry (PROJQUAY-4850)

verdel:
It is possible that we might want to create public repositories by default, but at the same time, repositories created when using Quay as a proxy should be created as private. It seems to me that it would be worth providing the ability to regulate the type of repository created in this case separately.

### DIFF
--- a/config.py
+++ b/config.py
@@ -787,6 +787,9 @@ class DefaultConfig(ImmutableConfig):
     # Repos created by push default to private visibility
     CREATE_PRIVATE_REPO_ON_PUSH = True
 
+    # Repos created by pull through proxy default to private visibility
+    CREATE_PRIVATE_REPO_ON_PULL_THROUGH = True
+
     # Create organization on push if it does not exist
     CREATE_NAMESPACE_ON_PUSH = False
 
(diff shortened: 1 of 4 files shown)


ibazulic:
Thanks for the contribution @verdel. I'm not sure I understand exactly the use case, why would you want proxied repositories to be private? You can change visiblity of repos now via API. Also, the downside to this approach is it would make all repos across all organizations private. Wouldn't it be better if the setting is per org, rather than global, if needed?

verdel:
@ibazulic

Currently, there is a configuration option `CREATE_PRIVATE_REPO_ON_PUSH`, which sets the default visibility at the Quay instance level for all repositories created through push.

In our case, we use Quay to store helm charts in OCI image format and to proxy requests to Docker Hub.

We want repositories for helm charts to be created with public visibility type, and repositories created in the process of using Quay as a proxy to have a private visibility type.

Currently, this is not achievable because one variable in the configuration is responsible for the default behavior.

If it were possible to configure the behavior differently for all regular repositories and separately for repositories in proxy mode, it would immediately solve our problem.

Unfortunately, I cannot find in the documentation or UI how to configure the default visibility of repositories at the organization level.

verdel:
@ibazulic
Is there anything I can do to assist with the review of this Pull Request? Perhaps some additional information is required from me?

PR #2659, opened 2024-02-06
registry: allow pushing manifests with artifactType and subject fields (PROJQUAY-6673)

kleesc:
Allow manifests to be pushed with a artifactType and subject fields.

### DIFF
--- a/image/oci/manifest.py
+++ b/image/oci/manifest.py
@@ -74,9 +74,13 @@ OCI_MANIFEST_DIGEST_KEY = "digest"
 OCI_MANIFEST_LAYERS_KEY = "layers"
 OCI_MANIFEST_URLS_KEY = "urls"
 OCI_MANIFEST_ANNOTATIONS_KEY = "annotations"
+OCI_MANIFEST_SUBJECT_KEY = "subject"
+OCI_MANIFEST_ARTIFACT_TYPE_KEY = "artifactType"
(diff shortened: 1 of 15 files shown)


image/oci/manifest.py
```diff
@@ -129,6 +136,15 @@ def get_meta_schema(self, ignore_unknown_mediatypes=False):
...
                 },
+                OCI_MANIFEST_SUBJECT_KEY: get_descriptor_schema(
+                    [],
+                    ignore_unknown_mediatypes=True,
+                ),
+                OCI_MANIFEST_ANNOTATIONS_KEY: {
+                    "type": "object",
+                    "description": "The annotations, if any, on this manifest",
+                    "additionalProperties": True,
+                },
```
ibazulic, line 147: According to [the spec](https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules), there are rules how annotations can be added. We should probably add those validations as well at some point.

kleesc: It would be nice to have at some point. But for this, I don't think the all the constraints in the spec can be enforced using only jsonschema.

syed (approved):
lgtm

LiZhang19817:
@sudhaponnaganti @kleesc @ibazulic want to know **any test image** that we can use to verify the fix in Quay 3.11

PR #2663, opened 2024-02-08
chore: extend market place subscription enddate (chore)

Sunandadadi:


arborite-rh (approved):
will revisit this in 2027 :)

PR #2665, opened 2024-02-12
chore: use postgres 16.2 for development

dmage:


arborite-rh (approved):
no more EOL warnings :)

PR #2675, opened 2024-02-12
PROJQUAY-6696: feat(api): raw clair vulnerability reports

arborite-rh:
Allow returning the full unprocessed Clair report


hdonnay:
Is there any appetite to land this? I can bring it up to date if need be.

PR #2685, opened 2024-02-18
script: Adding python script to automate removal of config editor (PROJQUAY-6707)

sayalibhavsar:
The removal of the config editor component from Quay version 3.10 onwards requires manual intervention by users. Customers are obligated to take additional steps. To streamline this process, consider adding the above script to the upgrade process that automatically removes associated objects(such as pod deployments, secrets, routes, and services) of config editor. This script works when quay-operator is deployed in single namespace as well as in all namespaces.

### DIFF
--- a/remove_config_editor_script.py
+++ b/remove_config_editor_script.py
@@ -1,0 +1,42 @@
+
+import subprocess
+import sys
+def check_oc_command():
+    try:
+        subprocess.run(["oc", "version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    except subprocess.CalledProcessError:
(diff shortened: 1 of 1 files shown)


dmage:
That should be automatically done by github.com/quay/quay-operator
@jonathankingfc does quay-operator delete these objects?

sayalibhavsar:
As of now operator doesn't clear out the objects and one has to manually run steps from 2.3. Removing config editor objects on Red Hat Quay Operator https://access.redhat.com/documentation/en-us/red_hat_quay/3.10/html/upgrade_red_hat_quay/operator-upgrade#config-editor-removal 
I would like to know if something similar is in the pipelines

Varsha-Shri:
Having this script can reduce manual interventions and increase experience for quay users during operator deployment

PR #2693, opened 2024-02-20
backend: Syncing teams with OIDC group (PROJQUAY-6290)

Sunandadadi:


data/migrations/versions/2744c27ad007_.py
```diff
@@ -0,0 +1,21 @@
+"""empty message
```
syed, line 1: can we remove this?

syed (approved):
lgtm! Few notes for future

1. We need to look at using `jsonb` 
2. Add index to the config 
3. update the query to use json query instead of `contains`
4. protections around self-deletion/removal from org

PR #2697, opened 2024-02-23
logging: reduce noise in quay logs and add missing worker names (PROJQUAY-6562)

sayalibhavsar:
1) Changing logger.info to logger.debug to stop too many INFO log lines from displaying in Quay logs 
2) Adding the missing worker name that print those log lines 

The idea of the log is to tell you which block is being procesed by worker that is processing it. These logs are missing the worker names
```
2023-11-27 15:24:05,825 [80] [INFO] [util.migrate.allocator] No candidates, marking entire block completed 210690-213109
2023-11-27 15:24:05,832 [80] [INFO] [util.migrate.allocator] No candidates, marking entire block completed 131657-134076
[sbhavsar-mac:~ sayalibhavsar]$ grep "No candidates, marking entire block completed" quay | wc -l
5300
```
For huge registries above logs  could be a bit noisy. We could hide it as a debug message rather than as an info message.
@sayalibhavsar

### DIFF
--- a/util/migrate/allocator.py
+++ b/util/migrate/allocator.py
@@ -5,7 +5,7 @@ from threading import Event
 from bintrees import RBTree
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
+logger.setLevel(logging.DEBUG)  # Set the default log level to DEBUG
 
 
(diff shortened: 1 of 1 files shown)


Marcusk19 (approved):
lgtm

PR #2704, opened 2024-02-26
quota: removing repo size from quota verification (PROJQUAY-6637)

bcaton85:
During quota verification, Quay fetches the size of the image currently being uploaded. That method incorrectly added the repository size to the size of the image being uploaded. That total returned was then added to the namespace size resulting in the following check: `repo_size + namespace_size + uploaded blob > quota_size_limit`. This causes the repo_size to be counted twice and images to be blocked early.

### DIFF
--- a/data/model/repository.py
+++ b/data/model/repository.py
@@ -708,7 +708,4 @@ def get_size_during_upload(repo_id: int):
         )
     ).get()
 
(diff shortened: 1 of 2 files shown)


kleesc (approved):
LGTM

PR #2709, opened 2024-02-28
quota: feature flagging quota edit/view/enforce (PROJQUAY-6734)

bcaton85:
Adds feature flags to enable/disable functionality related to the view/edit/enforcement of quota's.


arborite-rh (commented):
makes sense to me. should we consider test coverage for this behavior change?

bcaton85:
@arborite-rh I'll have the tests added in a follow up PR that refactors this a bit. I split it up to make it easier to review.

arborite-rh (approved):

PR #2712, opened 2024-02-29
quota: refactor, tests, and optimizations (PROJQUAY-6735)

bcaton85:
Adds the following changes:
- Additional logging
- Moving checks before expensive DB calls
- Additional tests


config.py
```diff
@@ -806,6 +806,9 @@ def create_transaction(db):
 
     # Add quota management configuration, caching, and validation
     FEATURE_QUOTA_MANAGEMENT = False
+
+    FEATURE_QUOTA_SUPPRESS_FAILURES = False
```
syed, line 810: can you add a comment on top of what this feature flag does

data/model/gc.py
```diff
@@ -425,12 +425,8 @@ def _garbage_collect_manifest(manifest_id, context):
...
         )
 
-        # Subtract blobs from namespace/repo total after they've been deleted successfully.
-        # If the quota feature is disabled mark the total as stale
-        if features.QUOTA_MANAGEMENT:
-            subtract_blob_size(manifest.repository_id, manifest_id, blob_sizes)
-        else:
-            reset_backfill(manifest.repository_id)
+        # Subtract blob sizes if quota management is enabled
+        update_quota(manifest.repository_id, manifest_id, blob_sizes, "subtract")
```
syed, line 429: can you create a enum for the possible operations to `update_quota`

syed:
Not in the current PR but the function which does the DB query to check for duplicate blobs
```
def blob_exists_in_repository(repository_id: int, manifest_id: int, blob_id: int):
    # Return true if there exists some other manifest within the repository that
    # references this blob
    return (
        ManifestBlob.select(1)
        .where(
            ManifestBlob.repository == repository_id,
            ManifestBlob.blob == blob_id,
            ManifestBlob.manifest != manifest_id,
        )
        .exists()
    )
```

Can be run faster if you use the `LIMIT` this way, the query will stop at the first result which is what you seem to be doing here. I'm not sure if the `exists()` does the same thing or it waits for the full query result. If it returns at the first row then we should be good

bcaton85:
`exists` should add a `LIMIT 1` but let me confirm

data/model/gc.py
bcaton85: yes

util/config/schema.py
```diff
@@ -1285,6 +1285,11 @@
             "description": "Enables configuration, caching, and validation for quota management feature",
             "x-example": False,
         },
+        "FEATURE_QUOTA_SUPPRESS_FAILURES": {
+            "type": "boolean",
+            "description": "Enables configuration, caching, and validation for quota management feature",
```
syed, line 1290: can you update the description here

data/model/quota.py
```diff
@@ -25,29 +26,140 @@
...
 
 
-def add_blob_size(repository_id: int, manifest_id: int, blobs: dict):
-    update_sizes(repository_id, manifest_id, blobs, "add")
-
+def update_quota(repository_id: int, manifest_id: int, blobs: dict, operation: str):
+    # If quota management is disabled, mark the total as stale and return
+    if not features.QUOTA_MANAGEMENT:
+        if config.app_config.get("QUOTA_INVALIDATE_TOTALS", True):
+            reset_backfill(repository_id)
```
syed, line 33: if I am never enabling quota management, this looks like it will always write to the DB for every push. Ideally if there's no quota management, we should not write to the quota tables

bcaton85: This is to handle the scenario where quota was disabled and a push/GC happened to the namespace invalidating the current total. Here we mark it stale to indicate that it has to be re-backfilled. Ideally we wouldn't have to do this but i couldn't find another way. The other option would be to re-backfill the entire table when quota is re-enabled but i saw this as an optimization.

syed: instead of an `update` can it be modified to be a check and update. check would ideal `select` against an indexed field. Another option is to delete the row and you can check for non existence on an indexed field

bcaton85: Doesn't that still write to the quota tables?

syed: it will only write once per repo/namespace. With the current logic, `update` gets called every time even if the data is the same. Updates cause a DB write even if the data is the same

bcaton85: Couldn't I add those as conditions to the `where` statement? It shouldn't write if those conditions fail correct?

data/model/quota.py
```diff
@@ -177,6 +223,13 @@ def write_repository_total(
...
     if repository_size_exists:
+        logger.info(
+            "updating repository size for manifest %s in repository %s, %s %s",
+            manifest_id,
+            repository_id,
+            operation,
+            repository_total,
+        )
         params = {}
         if operation == "add":
```
syed, line 234: replace with enum

data/model/quota.py
```diff
@@ -132,6 +166,13 @@ def write_namespace_total(
     # manifest in the repository, it needs to be handled by the backfill worker.
     # If it does exist we can add/subtract the total
     if namespace_size_exists:
+        logger.info(
+            "updating namespace size for manifest %s in namespace %s, %s %s",
+            manifest_id,
+            namespace_id,
+            operation,
```
syed, line 173: replace with enum

data/model/quota.py
syed: yes that should be good

syed (approved):
lgtm

PR #2717, opened 2024-03-04
chore: Add auto-prune tests + fix namespace autoprune policy UI issue (PROJQUAY-6773)

harishsurf:
- Adds more repository auto-prune tests 
- Fixes UI issue: when user is not an admin, the namespace autoprune policy under repository auto-prune policies tab shouldbe hidden


bcaton85 (approved):
LGTM

PR #2746, opened 2024-03-11
storage: Disable pushes on registry (PROJQUAY-6870)

ibazulic:
The current read-only option for Quay is not sometimes feasible, since it requires an insert of the service key and other manual config changes. For instance, if you want to just recalculate quota on the registry, but would like to allow all registry operations (including UI) without the possibility of pushes until recalculation is done, setting the whole registry `read-only` cannot be done since it makes the database read only as well.

This PR introduces a new flag called `DISABLE_PUSHES` which allows all registry operations to continue (changing tags, repo editing, robot account creation/deletion, user creation etc.) but will disable pushes of new images to the registry (i.e. backend storage will not change). If a registry already contains the image and a new tag is simply being added, that operation should succeed.

The following message would appear in the logs:

~~~
gunicorn-registry stdout | 2024-03-11 14:41:58,363 [285] [DEBUG] [endpoints.v2.blob] Cannot push images, pushes disabled on registry.
gunicorn-registry stdout | 2024-03-11 14:41:58,363 [285] [DEBUG] [endpoints.v2] sending response: b'{"errors":[{"code":"METHOD NOT ALLOWED","detail":{},"message":"Pushes to the registry are currently disabled. Please contact the administrator for more information."}]}\n'
gunicorn-registry stdout | 2024-03-11 14:41:58,364 [285] [DEBUG] [app] Ending request: urn:request:86c4b59b-3ca2-4038-a7ef-fff18e1c30f2 (/v2/ibazulic/big-layer-test/blobs/uploads/) {'endpoint': 'v2.start_blob_upload', 'request_id': 'urn:request:86c4b59b-3ca2-4038-a7ef-fff18e1c30f2', 'remote_addr': '172.24.20.50', 'http_method': 'POST', 'original_url': 'https://ubuntu.skynet/v2/ibazulic/big-layer-test/blobs/uploads/', 'path': '/v2/ibazulic/big-layer-test/blobs/uploads/', 'parameters': {}, 'json_body': None, 'confsha': '5c365a98', 'user-agent': 'docker/25.0.4 go/go1.21.8 git-commit/061aa95 kernel/6.5.0-25-generic os/linux arch/amd64 UpstreamClient(Docker-Client/25.0.4 \\(linux\\))'}
gunicorn-registry stdout | 2024-03-11 14:41:58,364 [285] [INFO] [gunicorn.access] 172.24.20.50 - - [11/Mar/2024:14:41:58 +0000] "POST /v2/ibazulic/big-layer-test/blobs/uploads/ HTTP/1.1" 405 169 "-" "docker/25.0.4 go/go1.21.8 git-commit/061aa95 kernel/6.5.0-25-generic os/linux arch/amd64 UpstreamClient(Docker-Client/25.0.4 \(linux\))"
nginx stdout | 172.24.20.50 (-) - - [11/Mar/2024:14:41:58 +0000] "POST /v2/ibazulic/big-layer-test/blobs/uploads/ HTTP/1.1" 405 169 "-" "docker/25.0.4 go/go1.21.8 git-commit/061aa95 kernel/6.5.0-25-generic os/linux arch/amd64 UpstreamClient(Docker-Client/25.0.4 \x5C(linux\x5C))" (0.003 1333 0.003)
~~~

The flag defaults to `False` (pushes enabled), unless set otherwise.

### DIFF
--- a/endpoints/v2/blob.py
+++ b/endpoints/v2/blob.py
@@ -44,6 +44,7 @@ from endpoints.v2.errors import (
     NameUnknown,
     QuotaExceeded,
     Unsupported,
+    PushesDisabled,
 )
 from util.cache import cache_control
 from util.names import parse_namespace_repository
@@ -149,6 +150,7 @@ def _try_to_mount_blob(repository_ref, mount_blob_digest):
     """
     Attempts to mount a blob requested by the user from another repository.
     """
+
     logger.debug("Got mount request for blob `%s` into `%s`", mount_blob_digest, repository_ref)
     from_repo = request.args.get("from", None)
     if from_repo is None:
@@ -246,6 +248,10 @@ def _try_to_mount_blob(repository_ref, mount_blob_digest):
 @anon_protect
 @check_readonly
 def start_blob_upload(namespace_name, repo_name):
+
+    if app.config.get("DISABLE_PUSHES", False):
+        logger.debug("Cannot push images, pushes disabled on registry.")
+        raise PushesDisabled
 
     repository_ref = registry_model.lookup_repository(namespace_name, repo_name)
     if repository_ref is None:
@@ -347,6 +353,10 @@ def fetch_existing_upload(namespace_name, repo_name, upload_uuid):
 @anon_protect
 @check_readonly
 def upload_chunk(namespace_name, repo_name, upload_uuid):
+    if app.config.get("DISABLE_PUSHES", False):
+        logger.debug("Cannot push images, pushes disabled on registry.")
+        raise PushesDisabled
+
     repository_ref = registry_model.lookup_repository(namespace_name, repo_name)
     if repository_ref is None:
         raise NameUnknown("repository not found")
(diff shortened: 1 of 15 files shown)


endpoints/v2/blob.py
```diff
@@ -389,6 +399,11 @@ def upload_chunk(namespace_name, repo_name, upload_uuid):
 @anon_protect
 @check_readonly
 def monolithic_upload_or_last_chunk(namespace_name, repo_name, upload_uuid):
+    # Check if pushes are disabled on the registry
+    if app.config.get("DISABLE_PUSHES", False):
+        logger.debug("Cannot push images, pushes disabled on registry.")
+        raise PushesDisabled
```
kleesc, line 405: You'll need to also check the other non-GET endpoints no?
Couldn't you add this check to the existing `check_readonly` decorator?

ibazulic: The check is added to all other non-GET endpoints: it's added to the `POST` and `PATCH`. Direct blob deletions are not possible anyway, and for blob requests that have an upload id it makes no sense to add the check since it's obvious that pushes are enabled there. I didn't however add the check to the manifest `DELETE` request which would delete the manifest and subsequent blobs. I think I need to add it there as well.

I was thinking about the decorator too, I'll revise that part.

PR #2748, opened 2024-03-12
oidc: allows users to login via cli (PROJQUAY-6873)

Sunandadadi:


data/users/externaloidc.py
```diff
@@ -37,10 +39,37 @@ def verify_credentials(self, username_or_email, password):
...
+                    if response is None:
+                        return (None, "External OIDC Group Sync: Got no user info")
+
+                    user_info = service.get_user_info(
+                        service._http_client, response["access_token"]
+                    )
+
+                    return (
+                        UserInformation(
+                            username=user_info.get("preferred_username"),
```
syed, line 62: `preferred_username` is not always present. Check the logic in `exchange_code_for_login` to see how it deals with this. If possible you could reuse part of that logic

data/users/externaloidc.py
```diff
@@ -37,10 +39,37 @@ def verify_credentials(self, username_or_email, password):
...
-            None,
-            f"Unsupported login option. Please sign in with the configured {self._federated_service} provider",
-        )
+        if not password:
+            return (None, "Anonymous binding not allowed.")
+
+        if not username_or_email:
+            return (None, "Missing username or email.")
+
+        for service in app.oauth_login.services:
```
syed, line 48: can there be multiple instances in this case? Can you identify the correct OIDC service and send the creds there? otherwise you'll be sending the full password to a service that's not supposed to receive it

syed (approved):
lgtm

syed (approved):
lgtm

PR #2750, opened 2024-03-12
logs: add action log for team sync enable/disable (PROJQUAY-6763)

Marcusk19:
Add logging for enabling and disabling team sync.


Sunandadadi (approved):
lgtm

PR #2754, opened 2024-03-13
logs: adjust graph padding dynamically (PROJQUAY-6857)

Marcusk19:
Adjusts the padding of the usage log domain based on the length of log data available. 

Should prevent overlap between the chart bars and the legend.


Sunandadadi (approved):
lgtm

PR #2755, opened 2024-03-13
storage: Disable pushes on registry (PROJQUAY-6870)

ibazulic:
The current read-only option for Quay is not sometimes feasible, since it requires an insert of the service key and other manual config changes. For instance, if you want to just recalculate quota on the registry, but would like to allow all registry operations (including UI) without the possibility of pushes until recalculation is done, setting the whole registry `read-only` cannot be done since it makes the database read only as well.

This PR introduces a new flag called `DISABLE_PUSHES` which allows all registry operations to continue (changing tags, repo editing, robot account creation/deletion, user creation etc.) but will disable pushes of new images to the registry (i.e. backend storage will not change). If a registry already contains the image and a new tag is simply being added, that operation should succeed.

The following message would appear in the logs:

~~~
gunicorn-registry stdout | 2024-03-13 20:19:49,414 [369] [DEBUG] [endpoints.v2] sending response: b'{"errors":[{"code":"METHOD NOT ALLOWED","detail":{},"message":"Pushes to the registry are currently disabled. Please contact the administrator for more information."}]}\n' 
gunicorn-registry stdout | 2024-03-13 20:19:49,414 [369] [INFO] [gunicorn.access] 172.17.0.1 - - [13/Mar/2024:20:19:49 +0000] "PUT /v2/ibazulic/mariadb/manifests/sha256:c4694ba424e0259694a5117bbb510d67340051f0bdb7f9fa8033941a2d66e53e HTTP/1.1" 405 169 "-" "skopeo/1.9.3" 
nginx stdout | 172.17.0.1 (-) - - [13/Mar/2024:20:19:49 +0000] "PUT /v2/ibazulic/mariadb/manifests/sha256:c4694ba424e0259694a5117bbb510d67340051f0bdb7f9fa8033941a2d66e53e HTTP/1.1" 405 169 "-" "skopeo/1.9.3" (0.002 3813 0.002) 
~~~

The flag defaults to `False` (pushes enabled), unless set otherwise.


ibazulic:
@kleesc I did something wrong in my previous PR so I closed it. But I did take your advice and created a new decorator that would do what I did previously with the exception. I couldn't incorporate this into the already existing decorator because API requests are also checking it, so it seemed a better solution to add a new one.

endpoints/decorators.py
```diff
@@ -197,6 +197,31 @@ def wrapper(*args, **kwargs):
...
+def check_pushes_disabled(func):
+    """
+    Validates that a non-GET method is not invoked when the registry has pushes disabled, unless
+    explicitly marked as being allowed.
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        # Skip if a GET method.
+        if request.method == "GET":
```
syed, line 209: could you add an exception `HEAD`  as well?

workers/storagereplication.py
```diff
@@ -220,6 +220,11 @@ def create_gunicorn_worker():
                 has_local_storage = True
                 break
 
+    if app.config.get("DISABLE_PUSHES", False):
+        logger.debug("Pushes to the registry are disabled; skipping startup")
+        while True:
+            time.sleep(100000)
+
```
kleesc, line 227: This isn't technically pushing an image, just replicating the storage from one location to another one. 
Is there a specific reason we would want to prevent these types of operations, while allowing things like quota calculations.

workers/repositorygcworker.py
```diff
@@ -79,6 +79,11 @@ def create_gunicorn_worker():
         while True:
             time.sleep(100000)
 
+    if app.config.get("DISABLE_PUSHES", False):
+        logger.debug("Pushes to the registry are disabled; skipping startup")
+        while True:
+            time.sleep(100000)
+
```
kleesc, line 86: Wondering if these workers should be disabled in this case. If we're allowing _some_ operations, like adding or deleting tags, the GC workers should work normally no?

Couldn't we only disable the v2 endpoints concerned?

workers/storagereplication.py
ibazulic: Removed.

workers/repositorygcworker.py
ibazulic: A possible scenario where `gc` could potentially cause problems is as follows: 

1. Disable pushes on registry for maintenance purposes (for instance, one wants to upgrade Quay).
2. Create a snapshot of the database. 
3. Delete something from the database after a snapshot has been taken. GC process would start up to 30 seconds after deletion.
4. Create a snapshot of storage blobs in the backend storage while GC is taking place.
5. Restore the backed up database from your snapshot to revert changes.

The outcome here is that now we cannot guarantee consistency between what's stored in the db vs. what's stored in the storage engine. There's a possibility that the restored snapshot contains references to blobs that were deleted during the GC process which would break pulls of images. So we need to keep them in sync. The disabled pushes would be a temporary thing anyway so even if GC is turned off for that duration, it should not cause problems (the items should be GCed after read only is enabled).

kleesc (approved):
LGTM

PR #2761, opened 2024-03-15
federatedusers(externalldap): fix UI/cli 500 errors by catching all LDAP exceptions (PROJQUAY-6900)

michaelalang:
The LDAP code we have in Quay does not care about all possible situations which causes various unnecessary 500 UI/cli errors when encountering an exception.

The exception handling ensures, we catch all ldap errors and log classified administrative interaction necessary ones in `warning` mode and all LDAP internal and none Quay related operations (write,...) in `debug` log level.

 

The classification for warning covers:

- AUTH_UNKNOWN,
- CONFIDENTIALITY_REQUIRED,
- CONSTRAINT_VIOLATION,
- INVALID_DN_SYNTAX,
- INVALID_SYNTAX,
- SERVER_DOWN,
- STRONG_AUTH_NOT_SUPPORTED,
- STRONG_AUTH_REQUIRED,
- UNAVAILABLE,
- UNWILLING_TO_PERFORM,
- CONTROL_NOT_FOUND,
- INAPPROPRIATE_AUTH,
- INSUFFICIENT_ACCESS,

A more detailed description of the result codes listed can be seen

 

Wrapping or extending all `self._ldap.get_connection` methods should as well make the UI/cli workers more robust on temporary errors in LDAP backend and the calling methods of `self._ldap.get_connection` will return the most common error response.

*NOTE* the change is only meant to be logged not returned to any client as this could expose sensitive information

This Bug PR goes in line with https://issues.redhat.com/browse/PROJQUAY-6481

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -247,7 +247,26 @@ class LDAPUsers(FederatedUsers):
             except ldap.LDAPError:
                 logger.debug("LDAP referral search exception")
                 return (None, "Username not found")
-
+        except (
+            ldap.AUTH_UNKNOWN,
+            ldap.CONFIDENTIALITY_REQUIRED,
+            ldap.CONSTRAINT_VIOLATION,
+            ldap.INVALID_DN_SYNTAX,
+            ldap.INVALID_SYNTAX,
+            ldap.SERVER_DOWN,
+            ldap.STRONG_AUTH_NOT_SUPPORTED,
+            ldap.STRONG_AUTH_REQUIRED,
+            ldap.UNAVAILABLE,
+            ldap.UNWILLING_TO_PERFORM,
+            ldap.CONTROL_NOT_FOUND,
+            ldap.INAPPROPRIATE_AUTH,
+            ldap.INSUFFICIENT_ACCESS,
+        ) as ldaperr:
+            for args in ldaperr.args:
+                logger.warning(
+                    f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
+                )
+            return (None, "Invalid username or password.")
         except ldap.LDAPError:
             logger.debug("LDAP search exception")
             return (None, "Username not found")
@@ -270,33 +289,92 @@ class LDAPUsers(FederatedUsers):
                 pass
         except ldap.INVALID_CREDENTIALS:
             return (None, "LDAP Admin dn or password is invalid")
-
-        with self._ldap.get_connection() as conn:
-            logger.debug("Incoming username or email param: %s", username_or_email.__repr__())
-
-            for user_search_dn in self._user_dns:
-                (pairs, err_msg) = self._ldap_user_search_with_rdn(
-                    conn,
-                    username_or_email,
-                    user_search_dn,
-                    suffix=suffix,
-                    filter_superusers=filter_superusers,
-                    filter_restricted_users=filter_restricted_users,
+        except (
+            ldap.AUTH_UNKNOWN,
+            ldap.CONFIDENTIALITY_REQUIRED,
+            ldap.CONSTRAINT_VIOLATION,
+            ldap.INVALID_DN_SYNTAX,
+            ldap.INVALID_SYNTAX,
+            ldap.SERVER_DOWN,
+            ldap.STRONG_AUTH_NOT_SUPPORTED,
+            ldap.STRONG_AUTH_REQUIRED,
+            ldap.UNAVAILABLE,
+            ldap.UNWILLING_TO_PERFORM,
+            ldap.CONTROL_NOT_FOUND,
+            ldap.INAPPROPRIATE_AUTH,
+            ldap.INSUFFICIENT_ACCESS,
+        ) as ldaperr:
+            for args in ldaperr.args:
+                logger.warning(
+                    f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
                 )
-                if pairs is not None and len(pairs) > 0:
-                    break
+            return (None, "Invalid username or password.")
+        except Exception as ldaperr:
+            try:  # just in case we are not an LDAP exception
+                for args in ldaperr.args:
+                    logger.debug(
+                        f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
+                    )
+            except:
+                logger.debug(str(ldaperr))
+            return (None, "Invalid username or password.")
 
-            if err_msg is not None:
-                return (None, err_msg)
+        try:
+            with self._ldap.get_connection() as conn:
+                logger.debug("Incoming username or email param: %s", username_or_email.__repr__())
 
-            dn_lst = [pair[0] for pair in pairs]
-            logger.debug("Found matching DNs: %s" % dn_lst)
+                for user_search_dn in self._user_dns:
+                    (pairs, err_msg) = self._ldap_user_search_with_rdn(
+                        conn,
+                        username_or_email,
+                        user_search_dn,
+                        suffix=suffix,
+                        filter_superusers=filter_superusers,
+                        filter_restricted_users=filter_restricted_users,
+                    )
+                    if pairs is not None and len(pairs) > 0:
+                        break
 
-            results = [LDAPUsers._LDAPResult(*pair) for pair in take(limit, pairs)]
+                if err_msg is not None:
+                    return (None, err_msg)
 
-            # Filter out pairs without DNs. Some LDAP impls will return such pairs.
-            with_dns = [result for result in results if result.dn]
-            return (with_dns, None)
+                dn_lst = [pair[0] for pair in pairs]
+                logger.debug("Found matching DNs: %s" % dn_lst)
+
+                results = [LDAPUsers._LDAPResult(*pair) for pair in take(limit, pairs)]
+
+                # Filter out pairs without DNs. Some LDAP impls will return such pairs.
+                with_dns = [result for result in results if result.dn]
+                return (with_dns, None)
+        except (
+            ldap.AUTH_UNKNOWN,
+            ldap.CONFIDENTIALITY_REQUIRED,
+            ldap.CONSTRAINT_VIOLATION,
+            ldap.INVALID_DN_SYNTAX,
+            ldap.INVALID_SYNTAX,
+            ldap.SERVER_DOWN,
+            ldap.STRONG_AUTH_NOT_SUPPORTED,
+            ldap.STRONG_AUTH_REQUIRED,
+            ldap.UNAVAILABLE,
+            ldap.UNWILLING_TO_PERFORM,
+            ldap.CONTROL_NOT_FOUND,
+            ldap.INAPPROPRIATE_AUTH,
+            ldap.INSUFFICIENT_ACCESS,
+        ) as ldaperr:
+            for args in ldaperr.args:
+                logger.warning(
+                    f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
+                )
+            return (None, "Invalid username or password.")
+        except Exception as ldaperr:
+            try:  # just in case we are not an LDAP exception
+                for args in ldaperr.args:
+                    logger.debug(
+                        f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
+                    )
+            except:
+                logger.debug(str(ldaperr))
+            return (None, "Invalid username or password.")
 
     def _ldap_single_user_search(
         self, username_or_email, filter_superusers=False, filter_restricted_users=False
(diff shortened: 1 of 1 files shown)


michaelalang:
added unittests for exception handling checks introduced in this PR

michaelalang:
Codecov doesn't recognize that the unittests do cover the newly added code. 
I do not understand how I should fix the codecov/patch check to be satisfied.

For a better understanding of the unittests:

The newly added unittest call the LDAPUser methods which have been changed to not raise any exception but instead log them and return an "invalid user" response to Quay instead.
The tests cover the exceptions that we want to log in WARNING level and catch the logs generated to understand if the exception was handled. The DEBUG level exceptions  are represented by IOError and could be any except the WARNING level ones.

jbpratt:
@coderabbit full review

data/users/externalldap.py
```diff
@@ -283,34 +302,82 @@
...
+            return (None, "Invalid username or password.")
+        except Exception as ldaperr:
+            try:  # just in case we are not an LDAP exception
+                for args in ldaperr.args:
+                    logger.debug(
+                        f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
+                    )
+            except:
+                logger.debug(str(ldaperr))
+            return (None, "Invalid username or password.")
```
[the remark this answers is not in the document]
michaelalang: the bare except is for catching any in logger print statement only (formatting something) since it's about printing prior raise exception, dumping it as str instead of details another more detailed handling is not necessary

data/users/externalldap.py
```diff
@@ -365,6 +432,26 @@
...
+            ldap.UNWILLING_TO_PERFORM,
+            ldap.CONTROL_NOT_FOUND,
+            ldap.INAPPROPRIATE_AUTH,
+            ldap.INSUFFICIENT_ACCESS,
+        ) as ldaperr:
+            for args in ldaperr.args:
+                logger.warning(
+                    f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
+                )
+            return (False, args.get("info", args.get("desc")))
```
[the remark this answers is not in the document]
michaelalang: that is true but you miss that the complete exception handling on that section is wrapped into another exception handling which will return the ldaperror raise as string if there's an exception raised like you mention.

data/users/externalldap.py
```diff
@@ -378,43 +465,102 @@
...
+                    if filter_restricted_users:
+                        if self._ldap_restricted_user_filter:
+                            search_flt = self._add_restricted_user_filter(search_flt)
+                        else:
+                            return (False, "Superuser filter not set")
+                    elif filter_superusers:
+                        if self._ldap_superuser_filter:
+                            search_flt = self._add_superuser_filter(search_flt)
+                        else:
+                            return (False, "Restricted user filter not set")
```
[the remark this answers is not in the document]
michaelalang: this is the original code from commit c84067a4d and expected to return as such

data/users/externalldap.py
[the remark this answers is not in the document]
michaelalang: we don't want to capture the formatting error to be returned or logged but the LDAP error. So if any of the formating rules fail, it's still returning the LDAP error and not a formatting error.

data/users/externalldap.py
```diff
@@ -259,7 +259,26 @@ def _ldap_user_search_with_rdn(
...
+            ldap.INVALID_DN_SYNTAX,
+            ldap.INVALID_SYNTAX,
+            ldap.SERVER_DOWN,
+            ldap.STRONG_AUTH_NOT_SUPPORTED,
+            ldap.STRONG_AUTH_REQUIRED,
+            ldap.UNAVAILABLE,
+            ldap.UNWILLING_TO_PERFORM,
+            ldap.CONTROL_NOT_FOUND,
+            ldap.INAPPROPRIATE_AUTH,
+            ldap.INSUFFICIENT_ACCESS,
```
jbpratt: @michaelalang :eyes: we should be able to reduce some of the exception handling duplication by doing something like this

```
  LDAP_CONNECTION_ERRORS = (
      ldap.AUTH_UNKNOWN,
      ldap.CONFIDENTIALITY_REQUIRED,
      ldap.CONSTRAINT_VIOLATION,
      ldap.INVALID_DN_SYNTAX,
      ldap.INVALID_SYNTAX,
      ldap.SERVER_DOWN,
      ldap.STRONG_AUTH_NOT_SUPPORTED,
      ldap.STRONG_AUTH_REQUIRED,
      ldap.UNAVAILABLE,
      ldap.UNWILLING_TO_PERFORM,
      ldap.CONTROL_NOT_FOUND,
      ldap.INAPPROPRIATE_AUTH,
      ldap.INSUFFICIENT_ACCESS,
  )

  def _log_ldap_error(exc, level=logging.WARNING):
      """Log LDAP exception details."""
      for arg in exc.args:
          if isinstance(arg, dict):
              msg = arg.get('info', arg.get('desc', str(arg)))
          else:
              msg = str(arg)
          logger.log(level, f"{exc.__class__.__name__} {getattr(exc, 'errnum', 'N/A')} {msg}")

...
  except LDAP_CONNECTION_ERRORS as ldaperr:
      _log_ldap_error(ldaperr)
      return (None, "Invalid username or password.")
```

data/users/externalldap.py
```diff
@@ -259,7 +259,26 @@ def _ldap_user_search_with_rdn(
             except ldap.LDAPError:
                 logger.debug("LDAP referral search exception")
                 return (None, "Username not found")
-
+        except (
+            ldap.AUTH_UNKNOWN,
```
jbpratt: Here are a few other exceptions, do we want to handle any of these as well?

```
ldap.BUSY
ldap.TIMEOUT
ldap.OPERATIONS_ERROR
ldap.CONNECT_ERROR
ldap.PROTOCOL_ERROR
```

data/users/externalldap.py
```diff
@@ -283,34 +302,82 @@ def _ldap_user_search(
...
+            ldap.INAPPROPRIATE_AUTH,
+            ldap.INSUFFICIENT_ACCESS,
+        ) as ldaperr:
+            for args in ldaperr.args:
+                logger.warning(
+                    f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
+                )
+            return (None, "Invalid username or password.")
+        except Exception as ldaperr:
+            try:  # just in case we are not an LDAP exception
```
jbpratt: Not sure I understand this branch, if we aren't an ldap exception we won't have `.errnum` and info, right?

data/users/externalldap.py
```diff
@@ -283,34 +302,82 @@ def _ldap_user_search(
...
+                for args in ldaperr.args:
+                    logger.debug(
+                        f"{ldaperr.__class__.__name__} {ldaperr.errnum} {args.get('info', args.get('desc'))}"
+                    )
+            except:
```
jbpratt: bare excepts are a python anti-pattern, and aren't we masking the original `ldaperr` exception in this nested exception block?

data/users/externalldap.py
michaelalang: @jbpratt agreed but than we would need to add it to a submodule like ldaperrors which would be a more reliable to avoid duplication. If you agree I can update the PR to do so.

data/users/externalldap.py
michaelalang: @jbpratt I bumped through all exception the module can raise and came up with the list of what makes sense.
Having those particular listed ones.
on Busy or Timeout, we wouldn't gain more insights than in a generice catch all exception.
Connect_error cannot happen at that stage in reality
Operations_error and Protocol_error also would happen at the very first object create (and Operations_error should only be raised when writing LDAP which we don't do) 

this is why I excluded them from being reported explicitly ...

data/users/externalldap.py
michaelalang: @jbpratt that's the same the Rabbit asked ... it the handling of some formatting error during the exception handling. it wouldn't drop the original ldaperr but instead just use string representation as something in the debug log statement wasn't resolving from the python-ldap library as it should ... (we could drop the handling but I wanted to ensure no matter what, the code will not fail the application in total because of formatting errors)

data/users/externalldap.py
jbpratt: @michaelalang I don't mind putting it in this file, its only ~25 lines for the setup and would only be used here so I don't think a submodule is necessary - WDYT?

data/users/externalldap.py
jbpratt: ACK, thanks - just trying to understand the reasoning behind the exceptions chosen (i.e. ldap returning busy while quay is trying to function seems to me the same as server down, granted I don't know specifically _why_ busy would be returned)

data/users/externalldap.py
michaelalang: oh sorry I mis read the weird formatting from the snipped ... you mean somwhere on top of externalldap.py the static Exception errors and the function that wraps the exception printing for all of them ... 
yeap makes sense too ...

jbpratt: sorry for the confusing snippet, yes exactly! Thanks :grin:

michaelalang: Ack done ... I also added a unittest for the "new" function accordingly ... thanks for your support :heart:

data/users/test/test_exceptions.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def emit(self, record):
+        try:
+            msg = self.format(record)
+            self.content.append(msg)
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except:
+            self.handleError(record)
+
```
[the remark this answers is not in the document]
michaelalang: The reason KeyboardInterrupt and SystemExit are are excluded and bare except handling the others is that the except logic adds the record to the exception handler where we check on Logs being propagated correctly.. Since it's a unittest it's not necessary to change the logic

michaelalang:
the E2E MySQL failure is not related to the code change

data/users/externalldap.py
```diff
@@ -283,48 +309,50 @@ def _ldap_user_search(
...
-            # Filter out pairs without DNs. Some LDAP impls will return such pairs.
-            with_dns = [result for result in results if result.dn]
-            return (with_dns, None)
+                # Filter out pairs without DNs. Some LDAP impls will return such pairs.
+                with_dns = [result for result in results if result.dn]
+                return (with_dns, None)
+        except LDAP_CONNECTION_ERRORS as ldaperr:
+            _log_ldap_error(ldaperr)
+            return (None, "Invalid username or password.")
+        except Exception as ldaperr:
```
jbpratt, line 352: This could not be an `ldaperr`, right? Should we add another block capturing `ldap.LDAPError` and not call the log ldap error when it's just a general exception?

data/users/externalldap.py
```diff
@@ -635,123 +680,125 @@ def has_restricted_users(self) -> bool:
...
+                                break
+                        else:
+                            # Pagination is not supported.
+                            logger.debug("Pagination is not supported for this LDAP server")
+                            break
+        except LDAP_CONNECTION_ERRORS as ldaperr:
+            _log_ldap_error(ldaperr)
+        except Exception as ldaperr:
+            # capture any other error to ensure we can track and fix problems
+            _log_ldap_error(ldaperr)
```
jbpratt, line 804:
```suggestion
        except Exception as ldaperr:
            # capture any other error to ensure we can track and fix problems
            _log_ldap_error(ldaperr)
```

Can you help me understand the difference in the addition and my suggestion? I don't think we want to be catching general exceptions and handing it the same way as the LDAP connection errors, otherwise we don't have to specifically handle the LDAP errors - this just handles everything

jbpratt (changes requested):

data/users/test/test_exceptions.py
```diff
@@ -0,0 +1,273 @@
...
+        (ldap.INVALID_SYNTAX, 21),
+        (ldap.SERVER_DOWN, -1),
+        (ldap.STRONG_AUTH_NOT_SUPPORTED, 7),
+        (ldap.STRONG_AUTH_REQUIRED, 8),
+        (ldap.UNAVAILABLE, 52),
+        (ldap.UNWILLING_TO_PERFORM, 53),
+        (ldap.CONTROL_NOT_FOUND, -13),
+        (ldap.INAPPROPRIATE_AUTH, 48),
+        (ldap.INSUFFICIENT_ACCESS, 50),
+    ):
```
jbpratt, line 199: 😀 we are duplicating this in all the tests, if we have to add or remove an exception it would require updating in every place. Best if we have one source of truth for the exceptions

data/users/externalldap.py
```diff
@@ -27,6 +20,37 @@
...
+    message = ""
+    for arg in exc.args:
+        if isinstance(arg, dict):
+            msg = arg.get("info", arg.get("desc", str(arg)))
+        else:
+            msg = str(arg)
+        message = f"{exc.__class__.__name__} {getattr(exc, 'errnum', 'N/A')} {msg}"
+        logger.log(level, additional) if additional != "" else None
+        logger.log(level, message)
+    return message if details else None
```
jbpratt, line 52: Why do we return something here? We don't handle the return in any of the calls

data/users/externalldap.py
```diff
@@ -386,9 +407,17 @@ def ping(self):
                 pass
         except ldap.INVALID_CREDENTIALS:
             return (False, "LDAP Admin dn or password is invalid")
+        except LDAP_CONNECTION_ERRORS as ldaperr:
```
jbpratt, line 410: Hmm, I'm struggling to understand the addition of this block which matches the exception block on line 415 - they are handled the same, why not always fall through to LDAPError? As far as I can tell, a TIMEOUT would be handled the exact same as any other LDAPError exception which would defeat the purpose of the `LDAP_CONNECTION_ERRORS` tuple 🤔

PR #2770, opened 2024-03-18
api: reducing db calls in repo list endpoints with quota enabled (PROJQUAY-6895)

bcaton85:
Reducing the number of DB calls in the repo list endpoint with quota enabled by:
- Adding the id to `RepositoryBaseElement` when the repositories are initially fetched, removing the need to fetch the repository ID's again
- Fetching the repository sizes with a single DB call using the `IN` operator

### DIFF
--- a/data/model/repository.py
+++ b/data/model/repository.py
@@ -715,7 +715,11 @@ def get_repository_sizes(repo_ids: list):
         )
 
     tuples = (
-        QuotaRepositorySize.select(QuotaRepositorySize.repository, QuotaRepositorySize.size_bytes)
+        QuotaRepositorySize.select(
+            QuotaRepositorySize.repository,
+            QuotaRepositorySize.size_bytes,
+            can_use_read_replica=True,
+        )
         .where(QuotaRepositorySize.repository << repo_ids)
         .tuples()
     )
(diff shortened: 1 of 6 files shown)


data/model/repository.py
```diff
@@ -701,6 +701,37 @@ def get_repository_size(repo_id: int):
...
+    Returns a map from repository ID to the size of the repository in bytes.
+    List of repo_ids should be kept below 1000 to avoid performance issues.
+    """
+    if not repo_ids:
+        return {}
+
+    if len(repo_ids) > 1000:
+        logger.warning(
+            "Fetching more than 1000 repository sizes at once, you may experience performance issues."
+        )
```
kleesc, line 715: What is this function used for, wouldn't we want a way to paginate this instead?
Are the repo_ids restricted to a single namespace or is this a full table lookup?

bcaton85: I could add pagination but for this particular use case it uses what ever the page size is for listing repositories. The repo_ids will belong to a single namespace the majority of the time but it's possible that they belong to separate namespaces. 

I could also replace this with individual requests for fetching each repository size but it looked like this gave better performance.

kleesc (approved):
lgtm

PR #2772, opened 2024-03-18
Changed wording from 'Vulnerabilities' to 'Advisories' on Security tab

BillDett:
This is needed to get the Clair Certification process moving with ProdSec


bcaton85 (approved):
LGTM

PR #2783, opened 2024-03-21
db: add config variable for DB pooling (PROJQUAY-6397)

syed:
adding `DB_CONNECTION_POOLING` param to enable/disable connection pooling for primary and replica DBs


bcaton85 (approved):
LGTM

PR #2788, opened 2024-03-22
teamsync: run team sync worker for teams synced via ldap or keystone (PROJQUAY-6762)

Sunandadadi:


data/users/teamsync.py
```diff
@@ -82,7 +83,7 @@ def sync_team(authentication, stale_team_sync):
     # Collect all the members currently found in the group, adding them to the team as we go
     # along.
     group_membership = set()
-    for (member_info, err) in member_iterator:
+    for member_info, err in member_iterator:
```
bcaton85, line 86: Running a quick test locally I don't think this changes anything but doesn't necessarily need to be removed.

bcaton85 (dismissed):
LGTM - we may want to consider adding a test to cover this

bcaton85 (approved):
LGTM

PR #2814, opened 2024-04-10
ui: raise accurate errors to usage logs view (PROJQUAY-6934)

Marcusk19:
Makes it so that failed requests to logs endpoints should raise accurate errors for the new UI. Allows for errors like "Method not implemented" to show when viewing logs
Screenshot 2024-04-10 at 12 21 00 PM


endpoints/decorated.py
```diff
@@ -65,6 +65,7 @@ def handle_readonly(ex):
 
 
 @app.errorhandler(NotImplementedError)
+@crossorigin()
```
bcaton85, line 68: Do we need this?

Marcusk19: Actually I see that the above issue was because I was running the patternfly ui in a dev environment outside the container - removing the decorator since it's not necessary outside of that

PR #2822, opened 2024-04-15
codecov: passing -Z flag (PROJQUAY-7009)

Sunandadadi:


.github/workflows/CI.yaml
```diff
@@ -100,10 +100,10 @@ jobs:
       run: tox -e py39-unit -- --cov=./ --cov-report=xml
 
     - name: Upload coverage reports to Codecov
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
-        flags: unit
+        flags: unit, -Z
```
dmage, line 106: `flags` will eventually be [converted into `-F unit -F -Z`](https://github.com/codecov/codecov-action/blob/8deb5f497d6af48d764f4983aea375e70bd09747/src/buildExec.ts#L280-L282) which isn't what we want.

According to [their documentation](https://github.com/codecov/codecov-action/), we need to add `fail_ci_if_error: true`.

Sunandadadi: > which isn't what we want
Why? If we don't pass `-Z` as a flag, we are not able to read the error message and we see [this error msg](https://github.com/quay/quay/actions/runs/8637183179/job/23892515611?pr=2805#step:6:39) in logs.

dmage: We need `-F unit -Z`.

`-F -Z` means that we want to tag results with the tag `-Z`.

PR #2824, opened 2024-04-16
ui: show 'unlimited' if private repo count is over max number (PROJQUAY-7016)

Marcusk19:
Since the unlimited sku returns a private repo count equal to a max integer, the ui should show 'unlimited' instead of the actual number.
unlimited-repos


Sunandadadi (approved):
lgtm

PR #2827, opened 2024-04-18
ui: batching requests in repository list page (PROJQUAY-6972)

bcaton85:
This change batches requests when trying to fetch all repositories the current user has access too. Otherwise the UI will make simultaneous requests equal to the number of organizations, where if the number is great enough it will exhaust the browser resources and cause the page to crash.


Sunandadadi (approved):
lgtm

jwojnarowicz:
FWIW, Did you consider displaying the first page immediately? 

Right now, if there are lots of repos, the X in 1-25 of X keeps increasing by 100 until it reaches the end and after that the first page is displayed. It can take up to several seconds for lots of repos (over 30s with few thousand repos).

It would be a nice improvement.

bcaton85:
@jwojnarowicz I agree, I created a ticket to track this.

PR #2834, opened 2024-04-22
marketplace: check for terminated subscriptions when fetching details (PROJQUAY-7035)

Marcusk19:
If a subscription is deleted in the subscription watch API it will set the field `terminatedDate` for the subscription object. This should be taken into consideration when fetching a subscription's details and the subscription should be removed from an org if it is terminated like how we already do for expired subscriptions.


kleesc (approved):
LGTM

PR #2839, opened 2024-04-24
chore:  Update alembic migration script to support sqlite db (PROJQUAY-7097)

harishsurf:
This PR updates alembic migration scripts in order for quay to support sqlite db


Makefile
```diff
@@ -186,7 +186,7 @@ node_modules/.npm-install-stamp: package.json package-lock.json | build-image-lo
 
 .PHONY: local-dev-clean
 local-dev-clean:
-	rm -f ./conf/jwtproxy_conf.yaml ./conf/mitm.cert ./conf/mitm.key ./conf/quay.kid ./conf/quay.pem ./conf/supervisord.conf
```
dmage, line 189: This change is unrelated to this PR

data/migrations/versions/3e8cc74a1e7b_add_severity_and_media_type_to_global_.py
```diff
@@ -23,24 +23,24 @@ def upgrade(op, tables, tester):
...
-    )
-    # ### end Alembic commands ###
+    with op.batch_alter_table("messages") as batch_op:
+        batch_op.create_foreign_key(
+            op.f("fk_messages_media_type_id_mediatype"),
+            "mediatype",
+            ["media_type_id"],
+            ["id"],
+        )
+    ### end Alembic commands ###
```
dmage, line 43: Other comments in this file are `# ### ... ###`, so I doubt it's a good change

data/migrations/versions/3383aad1e992_add_uploadedblob_table.py
```diff
@@ -23,7 +23,11 @@ def upgrade(op, tables, tester):
     if "uploadedblob" not in table_names:
         op.create_table(
             "uploadedblob",
-            sa.Column("id", sa.BigInteger(), nullable=False),
+            sa.Column(
+                "id",
+                sa.Integer() if op.get_context().dialect.name == "sqlite" else sa.BigInteger(),
```
dmage, line 28:
```
sa.Column(
    "id",
    sa.BigInteger().with_variant(sa.Integer, "sqlite"),
    nullable=False,
)
...
```
might be a bit cleaner option.

It is described at https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#allowing-autoincrement-behavior-sqlalchemy-types-other-than-integer-integer

data/migrations/versions/34c8ef052ec9_repo_mirror_columns.py
```diff
@@ -89,6 +89,8 @@ def upgrade(op, tables, tester):
 
     logger.info("Migrating to external_reference from existing columns")
     op.add_column("repomirrorconfig", sa.Column("external_reference", sa.Text(), nullable=True))
+    if op.get_context().dialect.name != "sqlite":
+        op.get_bind().execute("COMMIT")
```
dmage, line 93: It is worth adding a comment on why COMMIT is required for all DBs except SQLite

util/migrate/table_ops.py
```diff
@@ -11,7 +11,10 @@ def copy_table_contents(source_table, destination_table, conn):
         conn.execute(
             "INSERT INTO `%s` SELECT * FROM `%s` WHERE 1" % (destination_table, source_table)
         )
-        result = list(conn.execute("Select Max(id) from `%s` WHERE 1" % destination_table))[0]
-        if result[0] is not None:
-            new_start_id = result[0] + 1
-            conn.execute("ALTER TABLE `%s` AUTO_INCREMENT = %s" % (destination_table, new_start_id))
+        if conn.engine.name != "sqlite":
```
dmage, line 14: Shouldn't we do some magic around `sqlite_sequence` for sqlite?

harishsurf: Reading through the [sqlite docs](https://www.sqlite.org/autoinc.html#:~:text=SQLite%20keeps%20track%20of%20the,an%20AUTOINCREMENT%20column%20is%20created.), I found that the `sqlite_sequence` is automatically updated during the `INSERT INTO ...` stmts while copying contents from one table to another, so the next insert wouldn't need a special handling.

tox.ini
```diff
@@ -37,6 +37,20 @@ environment =
     MYSQL_ALLOW_EMPTY_PASSWORD=1
     MYSQL_USER=quay
 
+[testenv:py39-unit]
```
dmage, line 40: why do we need this new block?

harishsurf: this was after our discussion that we want to add `SKIP_DB_SCHEMA=true` for `py39-unit` and essentially have 3 unit test jobs with different DBs. Did you have a different suggestion?

tox.ini
```diff
@@ -37,6 +37,21 @@ environment =
...
+setenv =
+    PYTHONDONTWRITEBYTECODE = 1
+    PYTHONPATH={toxinidir}{:}{toxinidir}
+    TEST=true
+    SKIP_DB_SCHEMA=true
+    MARKERS="not e2e"
+    TEST_DATABASE_URI=sqlite:///data/quay_sqlite_test.db
```
dmage, line 47: data is a directory with python code, we shouldn't create temporary files there

tox.ini
```diff
@@ -37,6 +37,21 @@ environment =
...
+    registry: FILE=test/registry/registry_tests.py
+    e2e: MARKERS="e2e"
```
dmage, line 49: These markers make sense only in the generic [testenv], i.e. these envvars will be set in py39-registry and py39-e2e respectively. They shouldn't be in `[testenv:py39-unit]`, as they don't change anything here.

tox.ini
dmage: I would recommend to remove test/data/test.db from git - https://github.com/quay/quay/pull/2945 - and use test/data/test.db here.

dmage (approved):

bcaton85 (approved):
LGTM

PR #2840, opened 2024-04-24
ui: fix locale and layout issues with the time picker in tag expiration modal (PROJQUAY-7069)

dmesser:
This addresses a broader issue with hard-coded locale in the UI in some places of the UI vs. locale-dependent formatting of date and time output in other areas. Now, all parts of the UI leverage the browser's locale. This also means that the Cypress tests now expect locale-specific date and time formatting.

This addresses a more specific issue where the time picker in the tag expiration modal is causing error messages because it does not switch to the 24-hour mode when the locale demands it. This PR also addresses UI layout glitches in this modal and cleans up the error state visualization and enhances the guidance to valid inputs for expiration time by using the `minTime` property of the `TimePicker` component if the expiration time is on the same day as the current day.

Screenshot 2024-04-24 at 11 37 02
Screenshot 2024-04-24 at 11 36 49

Lastly, because I ran into this in CI, I am addressing a Cypress flake where the clipboard access in test runs sometimes fails. This is implemented by stubbing the `writeText` calls to the clipboard.

### DIFF
--- a/web/cypress/e2e/builds.cy.ts
+++ b/web/cypress/e2e/builds.cy.ts
@@ -1922,7 +1922,13 @@ describe('Repository Builds - View build logs', () => {
               );
             });
           });
-        });
+
+        cy.contains('Copy')
+          .click()
+          .then(() => {
+            expect(clipboardStub).to.be.called;
+          });
+      });
 
       // Download
       cy.contains('a', 'Download').should(
--- a/web/cypress/e2e/builds.cy.ts
+++ b/web/cypress/e2e/builds.cy.ts
@@ -1922,13 +1922,7 @@ describe('Repository Builds - View build logs', () => {
               );
             });
           });
-
-        cy.contains('Copy')
-          .click()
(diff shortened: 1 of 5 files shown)


web/cypress/e2e/builds.cy.ts
```diff
@@ -1031,6 +1031,7 @@ describe('Repository Builds', () => {
 describe('Repository Builds - Create Custom Git Build Triggers', () => {
   beforeEach(() => {
     cy.exec('npm run quay:seed');
+    cy.intercept('GET', '/config', {fixture: 'config.json'}).as('getConfig');
```
Sunandadadi, line 1034: do we need this?

dmesser: Without this for me the test fails, which makes sense because in the default config builds are disabled and so the builds UI is missing. I have no idea why it worked so far.

Sunandadadi (approved):
lgtm

PR #2846, opened 2024-04-27
logs: allow without strict logging (PROJQUAY-7116)

dmesser:
This introduce the config option `ALLOW_WITHOUT_STRICT_LOGGING` which when set to `True` will make any error encountered during writing logs to remote systems non-fatal and logged to stdout instead.

This is useful, when an external log storage system like Splunk or ElasticSearch is intermittently down and the user favors the uptime of the registry more than audit log coverage.

Defaults to false.

### DIFF
--- a/data/logs_model/document_logs_model.py
+++ b/data/logs_model/document_logs_model.py
@@ -511,9 +511,15 @@ class DocumentLogsModel(SharedModel, ActionLogsDataInterface, ElasticsearchLogsM
         try:
             self._logs_producer.send(log)
         except LogSendException as lse:
-            strict_logging_disabled = config.app_config.get("ALLOW_PULLS_WITHOUT_STRICT_LOGGING")
-            logger.exception("log_action failed", extra=({"exception": lse}).update(log.to_dict()))
-            if not (strict_logging_disabled and kind_name in ACTIONS_ALLOWED_WITHOUT_AUDIT_LOGGING):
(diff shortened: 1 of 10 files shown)


data/logs_model/document_logs_model.py
```diff
@@ -511,9 +511,15 @@ def log_action(
         try:
             self._logs_producer.send(log)
         except LogSendException as lse:
-            strict_logging_disabled = config.app_config.get("ALLOW_PULLS_WITHOUT_STRICT_LOGGING")
-            logger.exception("log_action failed", extra=({"exception": lse}).update(log.to_dict()))
-            if not (strict_logging_disabled and kind_name in ACTIONS_ALLOWED_WITHOUT_AUDIT_LOGGING):
+            strict_logging_disabled = config.app_config.get("ALLOW_WITHOUT_STRICT_LOGGING") or (
+                config.app_config.get("ALLOW_PULLS_WITHOUT_STRICT_LOGGING")
```
syed, line 515: do we need two config options? Can we just use `ALLOW_WITHOUT_STRICT_LOGGING` ?

dmesser: That's an existing config option that controls the behavior of the registry when it is in read-only mode. That's a different aspect than tolerating logging failures for everything to happen. I am undecided if we should remove that existing option, it might violate user expectations.

syed (approved):
LGTM

Marcusk19 (approved):

PR #2847, opened 2024-04-29
api: human-readable limit requests for quotas (PROJQUAY-7122)

dmesser:
This adds an optional new field in the namespace and superuser quota management APIs called `limit` which accepts a human-readable storage capacity string. The field is mutually exclusive to the existing `limit_bytes` field and adds convenience when scripting against Quay.


endpoints/api/namespacequota.py
```diff
@@ -116,7 +126,14 @@ def post(self, orgname):
             raise Unauthorized()
 
         quota_data = request.get_json()
-        limit_bytes = quota_data["limit_bytes"]
+
+        if "limit" in quota_data:
+            try:
+                limit_bytes = bitmath.parse_string_unsafe(quota_data["limit"]).to_Byte().value
+            except ValueError:
+                raise request_error(message="Invalid limit format")
```
syed, line 134: can you put info about what the expected format is?

endpoints/api/namespacequota.py
```diff
@@ -143,12 +160,36 @@ class OrganizationQuota(ApiResource):
...
+                        "limit": {
+                            "type": "string",
+                            "description": "Human readable storage capacity of the organization",
+                            "pattern": r"^(\d+\s?(B|KiB|MiB|GiB|TiB|PiB|EiB|ZiB|YiB|Ki|Mi|Gi|Ti|Pi|Ei|Zi|Yi|KB|MB|GB|TB|PB|EB|ZB|YB|K|M|G|T|P|E|Z|Y)?)$",
+                        },
+                    },
+                    "required": ["limit"],
+                    "additionalProperties": False,
+                },
+                {
```
syed, line 185: what does this do?

dmesser: Currently, the API, for whatever reason, allows a valid request to be submitted that does not contain `limit_bytes`. This schema makes it so that either `limit` or `limit_bytes` or neither of them can be supplied, essentially allowing an empty request. I don't necessarily understand why the parameter `limit` can be omitted since the API call doesn't seem to do anything useful in this case, but I needed to retain this behavior since this API is released fora while and clients may rely on it.

syed (approved):
LGTM

PR #2869, opened 2024-05-08
oauth: allowing to assign token to user (PROJQUAY-7074)

bcaton85:
image


data/database.py
```diff
@@ -2035,6 +2036,19 @@ class RepositoryAutoPrunePolicy(BaseModel):
     policy = JSONField(null=False, default={})
 
 
+class OauthAssignedToken(BaseModel):
+    uuid = CharField(default=uuid_generator, max_length=36, index=True, unique=True, null=False)
+    assigned_user = QuayUserField(index=True, unique=False, null=False)
+    application = ForeignKeyField(OAuthApplication, index=True, null=False, unique=False)
+    redirect_uri = CharField(max_length=255, null=True)
```
syed, line 2043: does the redirect URI in this case contain the access token? if so consider encrypting this field

data/model/oauth.py
```diff
@@ -378,12 +399,42 @@
     return query
 
 
+def get_assigned_authorization_for_user(user_obj, uuid):
+    try:
+        assigned_token = (
+            OauthAssignedToken.select()
+            .join(OAuthApplication)
+            .where(OauthAssignedToken.assigned_user == user_obj, OauthAssignedToken.uuid == uuid)
```
syed, line 407: isn't the uuid unique? does it need to be combined with the user to get the assigned_token?

data/model/oauth.py
```diff
@@ -406,3 +457,59 @@
...
+    if uuid is None:
+        return None
+
+    try:
+        return (
+            OauthAssignedToken.select()
+            .join(OAuthApplication)
+            .where(
+                OauthAssignedToken.uuid == uuid,
+                OauthAssignedToken.assigned_user == db_user,
```
syed, line 493: same: if uuid is unique, the user in the join is not required

data/database.py
bcaton85: The redirect URI should never contain any access token. The UI will set this to the config value `LOCAL_OAUTH_HANDLER` when it is being called [here](https://github.com/quay/quay/pull/2869/files#diff-2613f878141b14e270c68d1698010d584081398652c15f757e0dff579bf58617R104). If we feel that users may accidentally send tokens if they want to call this API themselves then i can add a check if need be

data/model/oauth.py
bcaton85: It's used to express `token that exists for this user with this ID`. I could use just the uuid but then i would need an additional check in python for the `assigned_user` field. This basically moves the condition check to the DB. Same for the other lookup.

data/model/oauth.py
```diff
@@ -189,15 +191,24 @@ def get_auth_denied_response(self, response_type, client_id, redirect_uri, **par
...
 
         # Check for a valid client ID.
-        is_valid_client_id = self.validate_client_id(client_id)
-        if not is_valid_client_id:
+        oauth_application = self.get_application_for_client_id(client_id)
+
+        if oauth_application is None or (
+            not is_org_admin(self.get_authorized_user(), oauth_application.organization)
+            and get_token_assignment(
+                assignment_uuid, self.get_authorized_user(), oauth_application.organization
```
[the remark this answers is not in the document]
bcaton85: moved to `is_org_admin_or_has_token_assignment`

syed (approved):
LGTM

PR #2875, opened 2024-05-14
ui: show repo count in marketplace UI (PROJQUAY-7185)

Marcusk19:
Change angular marketplace UI to show private repo count instead of SKU. Remove "Monthly" from section heading.

### DIFF
--- a/static/directives/org-binding.html
+++ b/static/directives/org-binding.html
@@ -1,9 +1,9 @@
 <div class="org-binding-settings-element" >
-  <span><h3>Monthly Subscriptions From Red Hat Customer Portal</h3></span>
+  <span><h3>Subscriptions From Red Hat Customer Portal</h3></span>
   <div class="cor-loader-inline" ng-show="marketplaceLoading"></div>
   <span ng-show="!organization && !marketplaceLoading">
     <div ng-repeat="subscription in userMarketplaceSubscriptions">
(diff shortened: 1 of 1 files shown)


static/directives/org-binding.html
```diff
@@ -1,32 +1,32 @@
 <div class="org-binding-settings-element" >
-  <span><h3>Monthly Subscriptions From Red Hat Customer Portal</h3></span>
+  <span><h3>Subscriptions From Red Hat Customer Portal</h3></span>
   <div class="cor-loader-inline" ng-show="marketplaceLoading"></div>
   <span ng-show="!organization && !marketplaceLoading">
     <div ng-repeat="subscription in userMarketplaceSubscriptions">
-      {{subscription.quantity}}x {{ subscription.sku }}
+      {{subscription.quantity}}x {{ subscription.metadata.privateRepos }} private repos
```
Sunandadadi, line 6: do we always get `metadata` and `privateRepos`?

Marcusk19: `metadata` should be always be there for user and org skus. As for `privateRepos` it should always be part of metadata since `metadata` is the equivalent [plan](https://github.com/quay/quay/blob/master/data/billing.py#L12) that we have hardcoded in the billing section

Sunandadadi (approved):
lgtm

PR #2894, opened 2024-05-20
db: updates for image expiry notification (PROJQUAY-7075)

Sunandadadi:
database migrations for image expiry notification ([PROJQUAY-7075](https://issues.redhat.com//browse/PROJQUAY-7075))

seeding db changes for cypress

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -752,6 +752,7 @@ class User(BaseModel):
                     AutoPruneTaskStatus,
                     RepositoryAutoPrunePolicy,
                     OauthAssignedToken,
+                    TagNotificationSuccess,
                 }
                 | appr_classes
                 | v22_classes
@@ -973,6 +974,7 @@ class Repository(BaseModel):
                 QuotaNamespaceSize,
                 QuotaRepositorySize,
                 RepositoryAutoPrunePolicy,
+                TagNotificationSuccess,
             }
             | appr_classes
             | v22_classes
@@ -1427,6 +1429,7 @@ class RepositoryNotification(BaseModel):
     config_json = TextField()
     event_config_json = TextField(default="{}")
     number_of_failures = IntegerField(default=0)
+    last_ran_ms = BigIntegerField(null=True)
 
 
 class RepositoryAuthorizedEmail(BaseModel):
@@ -2047,6 +2050,12 @@ class OauthAssignedToken(BaseModel):
     class Meta:
         database = db
         read_only_config = read_only_config
+
+
+class TagNotificationSuccess(BaseModel):
+    notification = ForeignKeyField(RepositoryNotification, index=True, null=False)
+    tag = ForeignKeyField(Tag, index=True, null=False)
+    method = ForeignKeyField(ExternalNotificationMethod, null=False)
 
 
 # Defines a map from full-length index names to the legacy names used in our code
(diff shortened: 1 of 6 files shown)


data/database.py
```diff
@@ -1426,6 +1428,7 @@ class RepositoryNotification(BaseModel):
     config_json = TextField()
     event_config_json = TextField(default="{}")
     number_of_failures = IntegerField(default=0)
+    last_ran_ms = BigIntegerField(null=True)
```
bcaton85, line 1431: You'll want an index on this field if you're going to be looking up by `last_ran_ms`

data/database.py
```diff
@@ -2035,6 +2038,12 @@ class RepositoryAutoPrunePolicy(BaseModel):
     policy = JSONField(null=False, default={})
 
 
+class TagNotificationSuccess(BaseModel):
+    notification = ForeignKeyField(RepositoryNotification, index=True, null=False)
+    tag = ForeignKeyField(Tag, index=True, null=False)
+    method = ForeignKeyField(ExternalNotificationMethod, null=False)
```
bcaton85, line 2044: You'll have to test when ever the foreign notification or tag rows are deleted that this row get's deleted as well without causing any issues.

data/migrations/versions/2664723e1b4b_add_last_ran_ms_to_repository_.py
```diff
@@ -0,0 +1,23 @@
...
+# revision identifiers, used by Alembic.
+revision = "2664723e1b4b"
+down_revision = "f67fe4871771"
+
+import sqlalchemy as sa
+
+
+def upgrade(op, tables, tester):
+    op.add_column(
+        "repositorynotification", sa.Column("last_ran_ms", sa.BigInteger(), nullable=True)
```
bcaton85, line 18: as mentioned before, this will most likely need an index

data/database.py
Sunandadadi: yes, I added both in the second PR, [tags clean up](https://github.com/quay/quay/pull/2895/files#diff-d54089ecfb2c9ed6312fb03ca1ea1cd1be9555c8dedf9138a2feac65ea530ff5R496) and [notification clean up](https://github.com/quay/quay/pull/2895/files#diff-8f5c31fa9c15f136ba9dd9d240d5c8b265556d6a2246f7af9f30df806543f3caR246)

data/migrations/versions/2664723e1b4b_add_last_ran_ms_to_repository_.py
Sunandadadi: good point, I added an index here

data/migrations/versions/2664723e1b4b_add_last_ran_ms_to_repository_.py
```diff
@@ -0,0 +1,35 @@
...
+            "last_ran_repositorynotification",
+            "repositorynotification",
+            ["last_ran_ms"],
+            unique=False,
+        )
+
+
+def downgrade(op, tables, tester):
+    op.drop_column("repositorynotification", "last_ran_ms")
+    op.drop_index("last_ran_repositorynotification", table_name="repositorynotification")
```
bcaton85, line 35: I would imagine that the index needs to be dropped before the column, could be wrong though

Sunandadadi: good point. I reversed the operations

bcaton85 (approved):
LGTM

bcaton85 (approved):
LGTM

PR #2900, opened 2024-05-22
ui: add caching to vulnerability reports (PROJQUAY-7250)

dmesser:
This moves the API calls to the `secscan` endpoint of Quay to react-query and also adds appropriate caching. This improves load time and resource utilization on the client and server side, since reports can be quite large (>1MB JSON) and there are many of them when loading a tag list view page with lots of manifest lists / OCI indexes, where a secscan API call is issued for every manifest child. The latter are now only issued when someone actually expands the manifest list row in the tag list table to view the child manifests.

Using react-query allowed to clean up the code quite a bit and stop leveraging atoms for caching. It even introduced transparent and staggering retry (up to 3 times) which is useful in case the `secscan` endpoint was overloaded.

This also changes the empty state view of a manifest detail page on the _Packages_ and _Security_ screen to simply have a message that this manifest didn't contain any content our scanner could recognize instead of still rendering the donut chart and package tables.

PS: There are still too many API calls going to render a tag list view due to a naive approach I took with #2469 - something I will fix in the next PR


ibazulic:
Putting the code aside for a moment, how good is, from a design standpoint, to move what is essentially a backend process to the frontend and then asking the frontend to keep data? We already cache (or do have the ability to cache) security reports on the API level, so the UI should just render that data. The securityworker doesn't need to do anything particular, since data is pulled from the memcached instance.

dmesser:
@ibazulic The response of the security manifest endpoint can be very large in size. As described in the initial comment, when loading a tag list with sufficiently old image content that carries a lot of vulnerabilities, the individual security vulnerability reports can be very large in size, and they would be fetched whenever you navigate back and forth in the UI, even though the data is exactly the same. Even if the vulnerability report is cached on the server side, it is still always transferred to the client, needlessly.  Users have reported slowness and total response traffic north of 40MB per tag list page, each report being 1~2MB worth of JSON.
While this PR does not address that in itself, it at least avoids loading that many bytes again and also reloading the individual vuln report for a manifest from the server when navigating from the tag list to the tag detail page. The caching is also rather a by-product of using react-query instead of directly embedding Axios calls into the component codes, helping with decluttering the code and avoiding props drilling.

PR #2913, opened 2024-05-29
storage: Add S3 optimization for CloudFlare provider (PROJQUAY-7267)

syed:
This adds the optimization in CloudFlare where if a request is from the primary region then instead of redirecting to the CDN, we return the S3 URL to save egress cost


jonathankingfc (approved):
LGTM

PR #2916, opened 2024-05-30
ui: manifest list size calculation on server side (PROJQUAY-7275)

dmesser:
This reverts most of #2469, which led to an excessive number of API calls when listing all tags in the repo view.
It moves this calculation to the server side, allowing clients scripting against the Quay API to know the size of manifest lists. It also better informs Quay users about the amount of storage a manifest list takes up when they are about to pull or delete it. Later, this will also allow for an easier implementation of sortable table columns.

But primarily, this is the basis for saving all these manifest API calls for each child manifest of a manifest list that are issued right on the first render.  Another PR is required to actually achieve this by refactoring the implementation to use `react-query` and deferring the call to the point where the manifest list is actually expanded in the UI. 

For now, this speeds up the UI rendering of the initial tag list view and provides more clarity via a popover.

Screenshot 2024-05-30 at 22 52 56


dmesser:
Currently, to calculate the aggregate size of a manifest list, _n_ API calls need to made to get the size of each sub manifest. When listing 20 image tags per page in the tag list view of a repo, this can easily add up to 100s of API calls just for the size calculation. This PR gets the aggregated size directly from the Quay API in a single API call.

Next, the API call to get the individual sub-manifests must be delayed until the user clicks "expand" on the manifest list row. This is the next PR I am eluding to. You are correct that this will yield additional calls. Still, at least they only occur when the user actually wants to see this information by expanding the manifest list and not right away for all manifest lists in the result page upon rendering the table.

bcaton85:
ah okay now i understand. By having the calls happen only on render this means we can't get the manifest list size until the row is expanded. Makes sense

data/model/oci/tag.py
```diff
@@ -116,6 +116,19 @@ def get_child_manifests(repo_id: int, manifest_id: int):
...
+def get_child_manifests_aggregate_size(repo_id: int, manifest_id: int):
+    return (
+        Manifest.select(fn.Sum(Manifest.layers_compressed_size))
+        .join(ManifestChild, on=(Manifest.id == ManifestChild.child_manifest))
+        .where(
+            (ManifestChild.repository == repo_id)
+            & (ManifestChild.manifest == manifest_id)
+            & (Manifest.layers_compressed_size.is_null(False))
+        )
+        .scalar()
```
bcaton85, line 128: The reason I haven't approved this yet is concerns over this query. The list tags endpoint is our most requested endpoint so this will be ran many, many times. I'm thinking it shouldn't be an issue because the look ups will happen via indexed manifest ID columns but the `layers_compressed_size` is not indexed. Can we see the execution plan of this query via `EXPLAIN` or another method? We need to make sure this won't cause significant performance issues if deployed.

kleesc: This query will get the aggregate if the immediate child manifests, but what about other manifest lists. Wouldn't you need to recursively go through manifest lists to get an accurate size?

PR #2917, opened 2024-05-31
ldap: allow global readonly superuser to be filtered (PROJQUAY-7044)

bcaton85:
Allows the global readonly superuser to be specified via a ldap filter `LDAP_GLOBAL_READONLY_SUPERUSER_FILTER`.

### DIFF
--- a/data/users/__init__.py
+++ b/data/users/__init__.py
@@ -69,6 +69,9 @@ def get_users_handler(config, _, override_config_dir, oauth_login):
         network_timeout = config.get("LDAP_NETWORK_TIMEOUT")
         ldap_user_filter = config.get("LDAP_USER_FILTER", None)
         ldap_superuser_filter = config.get("LDAP_SUPERUSER_FILTER", None)
+        ldap_global_readonly_superuser_filter = config.get(
+            "LDAP_GLOBAL_READONLY_SUPERUSER_FILTER", None
+        )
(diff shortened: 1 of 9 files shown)


data/users/__init__.py
```diff
@@ -69,6 +69,9 @@ def get_users_handler(config, _, override_config_dir, oauth_login):
         network_timeout = config.get("LDAP_NETWORK_TIMEOUT")
         ldap_user_filter = config.get("LDAP_USER_FILTER", None)
         ldap_superuser_filter = config.get("LDAP_SUPERUSER_FILTER", None)
+        ldap_global_readonly_superuser_filter = config.get(
+            "LDAP_GLOBAL_READONLY_SUPERUSER_FILTER", None
```
kleesc, line 73: Which one will take precedence if both `LDAP_SUPERUSER_FILTER` and `LDAP_GLOBAL_READONLY_SUPERUSER_FILTER` are defined?

kleesc: So that means that the same user will always have the superset of permissions? In the example above, if defined in both group, the user would still have write permissions, not just readonly, correct?

bcaton85: It would be a superset, that same user would have superuser permissions and also global readonly permissions

kleesc (approved):
LGTM

PR #2923, opened 2024-06-04
secscan: Fix Slack notification creation on initial index (PROJQUAY-7037)

ibazulic:
This fixes the Slack and e-mail notifications when images are indexed on initial push, where `tags` information was missing. If an image is a manifest child, instead of the tag, we provide a full SHA digest to the client.

Also adds the ability to filter security vulnerabilites depending on their severity. If the `NOTIFICATION_MIN_SEVERITY_ON_NEW_INDEX` is not set in the `config.yaml` file, we will automatically create notifications only for vulnerabilities marked "high" or "critical". This variable can take values defined [here](https://github.com/quay/quay/blob/3248a72da68ee70f4bd2b00dd03adaf76719269c/util/secscan/__init__.py#L1).

Example:

```
NOTIFICATION_MIN_SEVERITY_ON_NEW_INDEX: Medium
```

### DIFF
--- a/config.py
+++ b/config.py
@@ -846,7 +846,7 @@ class DefaultConfig(ImmutableConfig):
 
     # Set minimal security level for new notifications on detected vulnerabilities. Avoids
     # creation of large number of notifications after first index. If not defined, defaults to "High".
-    NOTIFICATION_MIN_SEVERITY_ON_NEW_INDEX = ""
+    NOTIFICATION_MIN_SEVERITY_ON_NEW_INDEX = "High"
 
     FEATURE_SUPERUSERS_FULL_ACCESS = False
     FEATURE_SUPERUSERS_ORG_CREATION_ONLY = False
--- a/config.py
+++ b/config.py
@@ -844,6 +844,10 @@ class DefaultConfig(ImmutableConfig):
     # Feature Flag: Enables notifications about vulnerabilities to be sent for new pushes
     FEATURE_SECURITY_SCANNER_NOTIFY_ON_NEW_INDEX = True
 
+    # Set minimal security level for new notifications on detected vulnerabilities. Avoids
+    # creation of large number of notifications after first index. If not defined, defaults to "High".
+    NOTIFICATION_MIN_SEVERITY_ON_NEW_INDEX = ""
+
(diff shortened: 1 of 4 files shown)


config.py
```diff
@@ -844,6 +844,10 @@ def create_transaction(db):
     # Feature Flag: Enables notifications about vulnerabilities to be sent for new pushes
     FEATURE_SECURITY_SCANNER_NOTIFY_ON_NEW_INDEX = True
 
+    # Set minimal security level for new notifications on detected vulnerabilities. Avoids
+    # creation of large number of notifications after first index. If not defined, defaults to "High".
+    NOTIFICATION_MIN_SEVERITY_ON_NEW_INDEX = ""
```
kleesc, line 849: Can you set it to "High" if that should be the default behavior?

ibazulic: @kleesc I set it in `secscan_v4_model.py`. Do I need to put it in the `config.py` as well?

kleesc: Just so it's easier to figure out what the default value is

data/secscan_model/secscan_v4_model.py
```diff
@@ -450,28 +452,70 @@ def should_skip_indexing(manifest_candidate):
...
+                                    vuln["normalized_severity"], PRIORITY_LEVELS["Unknown"]
                                 )
 
+                                if found_severity["score"] >= lowest_severity["score"]:
+                                    tag_names = list(
+                                        registry_model.tag_names_for_manifest(manifest, TAG_LIMIT)
+                                    )
+                                    if tag_names:
+                                        event_data = {
+                                            "tags": list(tag_names),
```
Marcusk19, line 486: nit: could we do something like
```
"tags": list(tag_names) if tag_names else [manifest.digest]
```
to avoid having to write out `event_data` twice?

ibazulic: @Marcusk19 I actually tried that as my first thought. And then what I got is a bunch of notifications for `s`, `n`, `a`... i.e. the whole SHA digest (each letter and number) was interpreted as a tag name. This does not happen with the two different events.

Marcusk19 (approved):
lgtm

Marcusk19 (approved):

PR #2949, opened 2024-06-14
storage: Add min chunk size for S3 storage for multipart upload (PROJQUAY-7328)

syed:
Adds an option for min_chunk_size to the S3 storage provider


Sunandadadi (approved):
lgtm

PR #2950, opened 2024-06-14
autoprune: Registry level autopruning (PROJQUAY-7392)

bcaton85:
Allows for adding a default autopruning policy that will be applied to all namespaces.


dmage (approved):

util/config/schema.py
```diff
@@ -1536,5 +1536,22 @@
...
+            "description": "Default auto-prune policy applied to newly created organizations",
+            "properties": {
+                "method": {
+                    "type": "string",
+                    "description": "",
+                    "enum": ["number_of_tags", "creation_date"],
+                    "x-example": "number_of_tags",
+                },
+                "value": {
+                    "type": "string",
```
Sunandadadi, line 1550: config tool checks that if method type is `number_of_tags`, value must be `int`. And I see that is what we check in [validation](https://github.com/quay/quay/blob/master/data/model/autoprune.py#L80) too

workers/autopruneworker.py
```diff
@@ -95,5 +161,6 @@ def create_gunicorn_worker():
             time.sleep(100000)
 
     logging.config.fileConfig(logfile_path(debug=False), disable_existing_loggers=False)
+    GlobalLock.configure(app.config)
```
Sunandadadi, line 164: How about configuring the lock if default auto prune policy exists (around [here](https://github.com/quay/quay/pull/2950/files#diff-7db0e9f1060956a8ddbfed7876897f57385740a3a58742640f2761fa902dde35R51))?

util/config/schema.py
bcaton85: good point, updated

workers/autopruneworker.py
bcaton85: add a conditional for creating the lock

Sunandadadi (approved):
lgtm

PR #2954, opened 2024-06-17
api: OCI-compliant tag listing behavior with limit = 0 (PROJQUAY-7327)

dmesser:
This corrects the behavior when requesting a list of tags with a an `n` parameter set to 0. Whereas before at least one tag was returned, now, an empty list is returned.


dmage (approved):

LiZhang19817:
@dmage Pls cherry-pick this fix into Quay 3.12 branch.

PR #2959, opened 2024-06-20
storage: use managed copy for single chunk uploads (PROJQUAY-7328)

syed:
We do a multi-part copy from the staging location to the final blob location in 5GB chunks sequentially. For large layers this is extremely slow. Use managed `copy` to move the blob to the final location faster


bcaton85 (approved):
LGTM

PR #2961, opened 2024-06-21
tag: look at child labels when creating and retargeting tag (PROJQUAY-7245)

Marcusk19:
When detecting expiration date for a manifest using the `quay.expires-after` label, if the manifest has any children the expiration should be set based on their labels. Fixes an issue where manifest lists only had the expiration set on the first manifest.

### DIFF
--- a/data/registry_model/registry_oci_model.py
+++ b/data/registry_model/registry_oci_model.py
@@ -16,7 +16,9 @@ from data.database import (
     db_transaction,
 )
 from data.model import DataModelException, QuotaExceededException, namespacequota, oci
+from data.model.oci.label import list_manifest_labels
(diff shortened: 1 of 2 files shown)


ibazulic:
What happens if you have multiple child manifests, each having their own expiry date? What date will be applied to the manifest list? Can we even have expiration of images set for images which are part of a manifest list?

Marcusk19:
> What happens if you have multiple child manifests, each having their own expiry date? What date will be applied to the manifest list? Can we even have expiration of images set for images which are part of a manifest list?

The expiration date from the latest child manifest should be applied. Not very familiar with manifest lists - would it be a problem to have expiration for images part of a manifest list?

PR #2962, opened 2024-06-21
storage: add debug log for time taken to upload blob (PROJQUAY-7328)

syed:
Adds logging on uploads to object storage
during the initial upload and the final copy


bcaton85 (approved):
LGTM

PR #2965, opened 2024-06-22
metrics: add metrics for registry, namespace and repo sizes plus quota-related information (PROJQUAY-6560)

dmesser:
This adds the following metrics related to the quota system:

- `namespace_stats_used_bytes` - Storage consumption in bytes per namespace (labels for namespace and entity type (whether it's an "organization" or a "user"))
- `namespace_quota_stats_capacity_bytes` - Storage quota in bytes per namespace, if defined (labels for namespace and entity type)
- `namespace_quota_stats_available_bytes` - Remaining storage capacity per quota in bytes per namespace, if quota exists (labels for namespace and entity type)
- `repository_stats_used_bytes` - Storage consumption in bytes per repository (labels for repository name and parent namespace)
- `registry_total_used_bytes` - Storage consumption in bytes for the entire registry

It also introduces a new config tunable called `QUOTA_METRICS`, which is `False` by default. Only if it's true will the above metrics be emitted. `GLOBAL_PROMETHEUS_STATS_FREQUENCY` is used to control how often the metrics are refreshed; the default is 1 hour. 
The default values for both settings are due to the potential performance impact on the database for very large installs with thousands of orgs and tens of thousands of repositories. These environments may lead to high cardinality on the Prometheus side too.

### DIFF
--- a/data/model/namespacequota.py
+++ b/data/model/namespacequota.py
@@ -137,6 +137,16 @@ def get_namespace_quota_limit(quota, limit_id):
         return None
 
 
+def get_namespaces_with_quotas():
+    namespaces_with_quotas = (
+        UserOrganizationQuota.select(User.id, User.username, UserOrganizationQuota.limit_bytes)
+        .join(User)
+        .dicts()
+    )
+
+    return list(namespaces_with_quotas)
+
+
 def create_namespace_quota_limit(quota, quota_type, percent_of_limit):
     if not percent_of_limit > 0 or not percent_of_limit <= 100:
         raise InvalidNamespaceQuotaLimit("Quota limit threshold must be between 1 and 100")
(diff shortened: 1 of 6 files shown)


data/model/namespacequota.py
```diff
@@ -137,6 +137,16 @@ def get_namespace_quota_limit(quota, limit_id):
         return None
 
 
+def get_namespaces_with_quotas():
+    namespaces_with_quotas = (
+        UserOrganizationQuota.select(User.id, User.username, UserOrganizationQuota.limit_bytes)
+        .join(User)
+        .dicts()
```
bcaton85, line 144: These and the other queries are going to be doing full table scans. This is going to put a lot of pressure on the database. Can we [paginate these calls to improve performance?](https://github.com/quay/quay/blob/master/data/model/modelutil.py#L7)

bcaton85 (changes requested):
Everything looks good except the queries will be doing full table scans. If we can get those paginated we should be good to go.

PR #2993, opened 2024-07-03
api: adding permissions for global readonly superuser (PROJQUAY-7177)

bcaton85:
The global readonly superuser is missing read only permissions on certain GET api's. This adds change adds those permissions.


kleesc (approved):
LGTM

PR #2996, opened 2024-07-04
mirror: increase upstreamregistry's password length (PROJQUAY-7430)

sayalibhavsar:
Changing the data type in RepoMirrorConfig for external_registry_password from EncryptedCharField to EncryptedTextField(null=True) and creating an alembic migration script that describe changes to be applied to the database

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -752,6 +752,7 @@ class User(BaseModel):
                     AutoPruneTaskStatus,
                     RepositoryAutoPrunePolicy,
                     OauthAssignedToken,
+                    TagNotificationSuccess,
                 }
                 | appr_classes
                 | v22_classes
@@ -973,6 +974,7 @@ class Repository(BaseModel):
                 QuotaNamespaceSize,
                 QuotaRepositorySize,
                 RepositoryAutoPrunePolicy,
+                TagNotificationSuccess,
             }
             | appr_classes
             | v22_classes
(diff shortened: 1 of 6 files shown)


data/migrations/versions/506f29126e2b_increase_repomirror_s_upstreamregistry_.py
```diff
@@ -0,0 +1,36 @@
...
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision: str = '506f29126e2b'
+down_revision: Union[str, None] = None
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
```
Marcusk19, line 18: You can get rid of these lines and just specify a down revision here, it should be the latest version in alembic's history (946f0e90f9c9)

data/migrations/versions/506f29126e2b_increase_repomirror_s_upstreamregistry_.py
```diff
@@ -0,0 +1,36 @@
+"""increase repomirror's upstreamregistry password length
+
+Revision ID: 506f29126e2b
+Revises:
```
Marcusk19, line 4: nit: could you also include the version this comes after (`946f0e90f9c9`)?

PR #3007, opened 2024-07-08
featureflag: add feature flag for image expiry trigger (PROJQUAY-7460)

Sunandadadi:


config.py
```diff
@@ -891,3 +891,6 @@ def create_transaction(db):
 
     FEATURE_ASSIGN_OAUTH_TOKEN = True
     DEFAULT_NAMESPACE_AUTOPRUNE_POLICY: Optional[Dict[str, str]] = None
+
+    # Allows users to set up notifications on image expiry, can remove flag once feature is tested
+    FEATURE_IMAGE_EXPIRY_TRIGGER = False
```
bcaton85, line 896: you can default this to True, we can set it to false manually

web/src/hooks/UseEvents.tsx
```diff
@@ -89,7 +89,7 @@ export function useEvents() {
       type: NotificationEventType.imageExpiry,
       title: 'Image expiry trigger',
       icon: <HistoryIcon />,
-      enabled: true,
+      enabled: config?.features.IMAGE_EXPIRY_TRIGGER,
```
bcaton85, line 92: you'll have to add `FEATURE_IMAGE_EXPIRY_TRIGGER` to `CLIENT_WHITELIST` in config.py

config.py
Sunandadadi: Since, this is a temporary flag, I thought why add it in app-interface and clean it afterwards if that makes sense. I plan to remove this flag after QE is done testing.

web/src/hooks/UseEvents.tsx
Sunandadadi: yes, missed that.

config.py
bcaton85: yeah that's fair, i usually like to have the control to enable/disable certain functionality it some cases. We can leave it

features/__init__.pyi
```diff
@@ -219,3 +219,6 @@ REFERRERS_API: FeatureNameValue
 
 SUPERUSERS_FULL_ACCESS: FeatureNameValue
 ASSIGN_OAUTH_TOKEN: FeatureNameValue
+
+# Feature Flag: If set to true, supports setting notifications on image expiry
+FEATURE_IMAGE_EXPIRY_TRIGGER: FeatureNameValue
```
bcaton85, line 224: you'll need to remove the `FEATURE_` prefix

Sunandadadi: good catch, fixed it

bcaton85 (approved):
LGTM

bcaton85 (approved):
LGTM

bcaton85 (approved):

PR #3015, opened 2024-07-10
ui: add helper text for input (PROJQUAY-7442)

Sunandadadi:
Screenshot 2024-07-10 at 1 43 36 PM

### DIFF
--- a/web/src/routes/RepositoryDetails/Settings/RepoEventExpiry.tsx
+++ b/web/src/routes/RepositoryDetails/Settings/RepoEventExpiry.tsx
@@ -1,18 +1,34 @@
-import {FormGroup, TextInput, ValidatedOptions} from '@patternfly/react-core';
+import {
+  FormGroup,
+  TextInput,
+  ValidatedOptions,
+  HelperText,
+  HelperTextItem,
+} from '@patternfly/react-core';
 import {NotificationEventConfig} from 'src/hooks/UseEvents';
-import {useState} from 'react';
+import {useEffect, useState} from 'react';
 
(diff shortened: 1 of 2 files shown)


web/src/routes/RepositoryDetails/Settings/RepoEventExpiry.tsx
```diff
@@ -28,6 +44,13 @@ export default function RepoEventExpiry(props: RepoEventExpiryProps) {
         required
         validated={valid}
       />
+      {valid == ValidatedOptions.error ? (
```
harishsurf, line 47: nit: You can use <[Conditional](https://github.com/quay/quay/blob/master/web/src/components/empty/Conditional.tsx#L2)> component instead of ternary operator

web/src/routes/RepositoryDetails/Settings/RepoEventExpiry.tsx
```diff
@@ -1,19 +1,35 @@
...
       setValid(ValidatedOptions.success);
     }
   };
 
+  const onChange = (_event, value) => {
+    props.setEventConfig({days: Number(value)});
+    validateNumber(value);
+  };
+
+  useEffect(() => {
```
harishsurf, line 29: shouldn't the `useEffect` be run only when the text input for days change and not on every render?

harishsurf (commented):
minor comment, otherwise lgtm!

harishsurf (dismissed):
lgtm!

harishsurf (approved):

PR #3020, opened 2024-07-11
webhook: Fix failure in sending webhook POST requests (PROJQUAY-7468)

ibazulic:
`urllib` requires that certs be parsed as tuple, not as a list. This small fix addresses that issue.


kleesc (approved):
lgtm

PR #3021, opened 2024-07-12
storage(RadosGWStorage): add Region to RadosGWStorage for multi-zone setups (PROJQUAY-7300)

michaelalang:
The config-tool validation does not accept Region as parameter and does not pass Region at anytime. To avoid setting `IGNORE_VALIDATION=true` to get Quay starting up, the PR add's the possibility *without changing behaviour*.

Behavior: 

* Setting default parameter for `region` to `us-east-1` 
    * this ensures that any configuration ommiting the region parameter will still work
* Passing either the args.Region or the default to the `minio.New` API  in the S3 call validation function

### DIFF
--- a/config-tool/pkg/lib/shared/storage_validators.go
+++ b/config-tool/pkg/lib/shared/storage_validators.go
@@ -30,6 +30,7 @@ func ValidateStorage(opts Options, storageName string, storageType string, args 
 	var secretKey string
 	var isSecure bool
 	var bucketName string
+	var region string = "us-east-1"
 	var token string = ""
 
 	switch storageType {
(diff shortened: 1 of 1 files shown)


michaelalang:
Guys,
I do not see where the code changes I added would fail the cypress tests on  `repository-permissions.cy.ts `
```
  │ ✖  repository-permissions.cy.ts             00:48        6        5        1        -        - │
```

I assume that has nothing do do with the change to the validator tool at all ..
Can someone please consider merging the PR as the Quay python code has already been merged ?

bcaton85:
That's a known flake in our tests. Can you rebase when you have a moment? Then I'll merge

PR #3026, opened 2024-07-15
configtool: default namespace autoprune policy validation (PROJQUAY-7407)

Sunandadadi:
If DEFAULT_NAMESPACE_AUTOPRUNE_POLICY.method is number_of_tags, the integer should be greater than 0.


bcaton85 (approved):
LGTM

PR #3033, opened 2024-07-16
oauth: logging assign oauth action to audit logs (PROJQUAY-7431)

bcaton85:
Adds audit log when an OAuth token is assigned.
image


jonathankingfc (approved):
LGTM

PR #3041, opened 2024-07-17
storage(StorageClasses): support changing signature version for pre-signed URL (PROJQUAY-7491)

michaelalang:
Pre-signed URL's are only on the `S3Storage` Class configured vor `s3v4` (hard coded).

This PR adds the attribute `signature_version` to all StorageClass definitions to be configured individually.
The behavior when not set defaults back to `v2` for all StorageClasses except `S3Storage` which defaults to `s3v4` 

Manual verification has been documented in https://issues.redhat.com/browse/PROJQUAY-7491 as  we do not have any validation tests as presigning is `client-local only` and therefor only would validate the string returned by the underlying function of botocore.

### DIFF
--- a/config-tool/pkg/lib/shared/structs.go
+++ b/config-tool/pkg/lib/shared/structs.go
@@ -27,6 +27,8 @@ type DistributedStorageArgs struct {
 	AccessKey   string `default:"" validate:"" json:"access_key,omitempty" yaml:"access_key,omitempty"`
 	SecretKey   string `default:"" validate:"" json:"secret_key,omitempty" yaml:"secret_key,omitempty"`
 	BucketName  string `default:"" validate:"" json:"bucket_name,omitempty" yaml:"bucket_name,omitempty"`
+
+	Signature string `default:"s3v2" validate: "" json:"signature_version,omitempty" yaml:"signature_version,omitempty"`
 	// Args for S3Storage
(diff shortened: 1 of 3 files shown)


config-tool/pkg/lib/shared/structs.go
```diff
@@ -27,6 +27,8 @@ type DistributedStorageArgs struct {
 	AccessKey   string `default:"" validate:"" json:"access_key,omitempty" yaml:"access_key,omitempty"`
 	SecretKey   string `default:"" validate:"" json:"secret_key,omitempty" yaml:"secret_key,omitempty"`
 	BucketName  string `default:"" validate:"" json:"bucket_name,omitempty" yaml:"bucket_name,omitempty"`
+
+	Signature string `default:"s3v2" validate: "" json:"signature_version,omitempty" yaml:"signature_version,omitempty"`
```
ibazulic, line 31: Do we need this since we're not doing any other validations related to the sig version?

michaelalang: @ibazulic  without the definition the config-validator will reject if it's defined (had the same with `region_name`)

ibazulic: Gotcha!

ibazulic (approved):
LGTM.

PR #3042, opened 2024-07-17
ui: adding message that registry wide autoprune has been enabled (PROJQUAY-7452)

bcaton85:
image
image


harishsurf (approved):
tested it out, lgtm!

PR #3043, opened 2024-07-17
ui: assign OAuth token to user form checks (PROJQUAY-7433)

bcaton85:
Fixes 2 bugs:
- Disables assign token button if user has selected a user to assign token to but hasn't selected and scopes (https://issues.redhat.com/browse/PROJQUAY-7461)
- Resolves issue where if user cancels assignment the button will still attempt to assign the token to the user

### DIFF
--- a/static/js/pages/manage-application.js
+++ b/static/js/pages/manage-application.js
@@ -161,6 +161,15 @@
       });
     };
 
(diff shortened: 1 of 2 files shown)


Sunandadadi (approved):
lgtm

PR #3051, opened 2024-07-18
registry: hidden tag should be included when checking a manifest's referrers (PROJQUAY-7519)

kleesc:
Referrers with only a temp tag (hidden) still need to be listed to validate that they exist when listing referrers.


bcaton85 (approved):
LGTM

PR #3053, opened 2024-07-18
bug: Fix UI x-axis display (PROJQUAY-7467)

ibazulic:
Several fixes were made:

- We weren't properly displaying the `x` axis. This change increases the border padding to properly display the axis at the bottom.
- We weren't properly setting the legend font colors which made it impossible to see it when dark mode was turned on. We now properly apply them.
- If loaded logs were empty, the graph would not render properly. We check if any data should be rendered, if not, we display a message instead of rendering the graph.


Marcusk19 (approved):
lgtm

web/src/routes/UsageLogs/UsageLogsGraph.tsx
```diff
@@ -76,48 +80,63 @@ export default function UsageLogsGraph(props: UsageLogsGraphProps) {
...
+            }
+            name="usage-logs-graph"
+            padding={{
+              bottom: 20 * getLegendData().length,
+              left: 80,
+              right: 500, // Adjusted to accommodate legend
+              top: 50,
+            }}
+            domainPadding={{x: 5 * Object.keys(logData).length}}
+            height={400}
```
bcaton85, line 127: Ther x-axis was still covered up for me, I had to change this to 450 to make it visible

PR #3054, opened 2024-07-19
events: fix notification methods for image expiry events (PROJQUAY-7487)

Sunandadadi:
Passing notification uuid to fetch matching notification event.

### DIFF
--- a/data/model/notification.py
+++ b/data/model/notification.py
@@ -266,9 +266,7 @@ def list_repo_notifications(
         )
 
     if notification_uuid:
-        query = query.switch(RepositoryNotification).where(
-            RepositoryNotification.uuid == notification_uuid
-        )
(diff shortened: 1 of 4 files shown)


data/model/notification.py
```diff
@@ -263,6 +265,11 @@ def list_repo_notifications(namespace_name, repository_name, event_name=None):
             .where(ExternalNotificationEvent.name == event_name)
         )
 
+    if notification_uuid:
+        query = query.switch(RepositoryNotification).where(
```
bcaton85, line 269: i don't think you need this switch since `RepositoryNotification` is the table being selected from

Sunandadadi: yes, fixed in the latest commit.

bcaton85 (approved):
LGTM

kleesc (approved):

PR #3056, opened 2024-07-22
audit: fix auto-prune audit logging (PROJQUAY-7423)

dmesser:
This fixes an issue where the autoprune logic would not use the logs data model proxy but directly reach into the table-based audit logs model, leading to log coverage gaps when using another backend for audit logs.


data/model/autoprune.py
```diff
@@ -509,11 +508,14 @@ def delete_autoprune_task(task):
 
 
 def prune_tags(tags, repo, namespace):
+    # imported here to avoid cyclic imports
+    from data.logs_model import logs_model
```
bcaton85, line 512: Not a huge fan of this fix, this is a patch that increases complexity and may lead to maintainability issues later on

PR #3066, opened 2024-07-22
bug: Fix UI x-axis display (PROJQUAY-7467)

ibazulic:
Change bottom padding from dynamically allocating it to a fixed value. The dynamic allocation caused an issue with axis not being displayed properly when only one action on the namespace is logged.


bcaton85 (approved):
LGTM

PR #3071, opened 2024-07-23
registry: return 400 instead of 404 for manifest errors on referrers api (PROJQUAY-7523)

kleesc:
Given that a repository is found, the request must not return a 404.


ibazulic (approved):
LGTM!

PR #3089, opened 2024-07-26
nginx: dropping x-forwarded-host (PROJQUAY-7563)

michaelalang:
Description: dropping x-forwared-host from the request ensures S3 signatures are valid

With Hitachi Vantara S3 Buckets integration, the `x-forwarded-host` invalidated the Signature in v4 mode.
Removing it accordinlgy in the nginx `_storage_proxy` location match enables us to use HCP Buckets.

### DIFF
--- a/conf/nginx/server-base.conf.jnj
+++ b/conf/nginx/server-base.conf.jnj
@@ -90,6 +90,7 @@ location ~ ^/_storage_proxy/([^/]+)/([^/]+)/([^/]+)/(.+) {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header Host $3;
     proxy_set_header Authorization "";
+    proxy_set_header x-forwarded-host "";
     proxy_ssl_name $3;
     proxy_ssl_server_name on;
 


conf/nginx/server-base.conf.jnj
```diff
@@ -90,6 +90,7 @@ location ~ ^/_storage_proxy/([^/]+)/([^/]+)/([^/]+)/(.+) {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header Host $3;
     proxy_set_header Authorization "";
+    proxy_set_header x-forwarded-host "";
```
ibazulic, line 93: Does this change break any other storage engine? We don't have a specific nginx storage proxy config for different storage engines so this has to work with everything we support.

michaelalang: per definition 

```
The X-Forwarded-Host (XFH) header is a de-facto standard header for identifying the original host requested by the client in the Host HTTP request header.
``` 

Which means only if someone specified in their S3 storage system to not provide content for a specific host/network that could break (similar to statistics from parsed logs). From the operational POV it should not break anything as it's a Hint to the backend system who the originator of the request was.

As mentioned in Slack, that header is only injected in OCP deployments where the OCP ingress add's that header. Quay does not have code for adding the header, and even verifying with a copy of the instantiated nginx configuration in a simple podman container verified, the configuration is OCP specific.

michaelalang: https://docs.openshift.com/container-platform/4.16/networking/ingress-operator.html#:~:text=httpHeaders%20defines%20the%20policy%20for%20HTTP%20headers.

ibazulic (approved):

LiZhang19817:
@ibazulic @michaelalang @HammerMeetNail @bcaton85 Not seeing this fix was cherry pick to Quay 3.12, suppose coming Quay 3.1.2 Z-Stream should have this fix, pls review.

PR #3097, opened 2024-07-29
ui: Fix UI x-axis display (PROJQUAY-7467)

ibazulic:
Several fixes were made:

- We weren't properly displaying the x axis. This change increases the border padding to properly display the axis at the bottom.
- We weren't properly setting the legend font colors which made it impossible to see it when dark mode was turned on. We now properly apply them.
- If loaded logs were empty, the graph would not render properly. We check if any data should be rendered, if not, we display a message instead of rendering the graph.

Manual cherrypick of https://github.com/quay/quay/pull/3053 and https://github.com/quay/quay/pull/3066 .


Marcusk19 (approved):
lgtm

PR #3119, opened 2024-08-07
ui: add refetch to usage logs search (PROJQUAY-7109)

Marcusk19:
Usage logs are not updating when the user changes the filter value in the search bar.

Adding a callback to refetch the query when filter value changes to have the log data update with filter value state.


ibazulic:
Can we adda a test for this particular case?

Marcusk19:
> Can we adda a test for this particular case?

Sure, updated the test case for filtering

PR #3126, opened 2024-08-08
bug: too strict validateTeamName regex (PROJQUAY-7606)

eb4x:
If you PUT against /api/v1/organization/mirrors/team/CAPITAL-LETTERS you'll get a different regex back.

  Invalid team name CAPITAL-LETTERS: Namespace must match expression
  ^([a-z0-9]+(?:[._-][a-z0-9]+)*)$

Let's use this more permissive regex in the web-ui instead.

### DIFF
--- a/web/src/libs/utils.ts
+++ b/web/src/libs/utils.ts
@@ -52,7 +52,7 @@ export function isValidEmail(email: string): boolean {
 }
 
 export const validateTeamName = (name: string) => {
-  return /^[a-z][a-z0-9]+$/.test(name);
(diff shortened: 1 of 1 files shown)


ibazulic:
Hey @eb4x ,

Yeah, that seems like it's too restrictive. The valid team name regex is described here:

https://github.com/quay/quay/blob/dde4dde03dd47b57d17b93b95ca84c7b8519f724/data/model/team.py#L35

So that's what we should use for the UI as well. Good catch, thank you for that!

ibazulic:
@harishsurf any objections on this fix?

harishsurf (approved):
lgtm!

PR #3129, opened 2024-08-09
nginx: revert #3098 (PROJQUAY-7573)

syed:
reverting this change because we won't use keepalive with gunicorn workers


Sunandadadi (approved):
lgtm

PR #3146, opened 2024-08-19
logs: Audit export logs requests (PROJQUAY-7679)

ibazulic:
We add the ability to audit export logs requests that were previously not tracked.


ibazulic:
Sample logs:

image

static/js/directives/ui/logs-view.js
```diff
@@ -892,3 +916,8 @@ angular.module('quay').directive('logsView', function () {
 
   return directiveDefinitionObject;
 });
+
+function obfuscate_email(email) {
```
Marcusk19, line 920: Should we obfuscate the email? If we want the ability to audit log exports shouldn't the user be able to see which email it was sent to?

web/src/routes/UsageLogs/UsageLogsTable.tsx
```diff
@@ -92,7 +92,7 @@ export function UsageLogsTable(props: UsageLogsTableProps) {
           <Table variant="compact" borders={false} style={{margin: '20px'}}>
             <Thead>
               <Tr>
-                <Th>Date & Time</Th>
+                <Th width={15}>Date & Time</Th>
```
Marcusk19, line 95: Just curious, why does a specific width need to be set for this header?

Marcusk19 (dismissed):
Overall looks good, just left a couple of comments. One thing you may need to add though is an entry for these new log kinds in `initdb.py` - like [this](https://github.com/quay/quay/blob/master/initdb.py#L478-L479) for example

static/js/directives/ui/logs-view.js
ibazulic: I'm generally not in favour of printing out e-mails in raw text format in browsers, due to potential malware being present on computers which can read screens. The e-mail address is stored as part of the log in the `logentry3` table and would be exported in logs as well. It just wouldn't be shown as clear text in the UI. The full information is still available via API:

```
    {
      "kind": "export_logs_success",
      "metadata": {
        "export_id": "0f38180f-0e11-48c5-97f1-48cfdbfa179e",
        "url": null,
        "email": "some@email.com"
      },
      "ip": "172.17.0.1",
      "datetime": "Mon, 19 Aug 2024 22:11:35 -0000",
      "performer": {
        "kind": "user",
        "name": "ibazulic",
        "is_robot": false,
        "avatar": {
          "name": "ibazulic",
          "hash": "9b6277f06ccaf5ee50f00e85ade1ab0a7210c13c7e7c5739d1f318d4feca183c",
          "color": "#a1d99b",
          "kind": "user"
        }
      }
    },
```

ibazulic:
> Overall looks good, just left a couple of comments. One thing you may need to add though is an entry for these new log kinds in `initdb.py` - like [this](https://github.com/quay/quay/blob/master/initdb.py#L478-L479) for example

Didn't occur to me. I'll add it there as well.

web/src/routes/UsageLogs/UsageLogsTable.tsx
ibazulic: The idea was to stop splitting the date and time column across multiple lines. Because this split occurs on random places (for instance, sometimes letters `AM` and `PM` are moved to the 2nd line, while the rest of the date is left on the first line). This doesn't really look good and is occurring consistently, especially when larger descriptions are in place and the browser window is zoomed. I thought that maybe setting the width to a fixed size will stop that but I guess something else needs to be added to achieve it.

ibazulic: @Marcusk19 I added a truncate option on the text field, that should keep the `date/time` always on one line and will truncate it if zoom levels are too high so that text cannot be displayed in the defined column width.

Marcusk19 (commented):
Looks like there's a cypress test failing for the usage logs. I think this error has something to do with the pop-up alert that should show after an export happens on the patternfly ui:
```
     AssertionError: Timed out retrying after 25000ms: Expected to find content: 'Logs exported with id' but never did.
```

ibazulic:
> Looks like there's a cypress test failing for the usage logs. I think this error has something to do with the pop-up alert that should show after an export happens on the patternfly ui:
> 
> ```
>      AssertionError: Timed out retrying after 25000ms: Expected to find content: 'Logs exported with id' but never did.
> ```

Yeah, I saw that. But I cannot replicate it. Whenever I export the logs through the UI, the id always appears. And the changes I did do not change the return logic. So I don't understand why this test is failing.

endpoints/api/logs.py
```diff
@@ -440,8 +477,28 @@ def post(self, orgname, parsed_args):
...
 
+        log_action(
+            "export_logs_failure",
+            orgname,
+            {
+                "error": "unauthorized",
+                "url": request.get_json().get("callback_url") or None,
+                "email": request.get_json().get("callback_email") or None,
+            },
+        )
```
kleesc, line 503: I think instead of logging failures at the api call level, we should log them if the actual background job fails.

ibazulic: I don't think that's possible. I did add some additional logic that would grab some exceptions that the API may have. But what you're saying is out of the API scope. Exporting of logs is an async operation, the worker does the job and it picks it up from `queueitem` table. The API is only here to tell the worker that some work needs to be done. If the worker generates an exception and cannot continue, the worker should send a message to the initial e-mail or URL and inform the client that failure happened. I think it does that today. I'll change the wording in the audit logs to reflect that: not `successful` but `queued for execution`.

PR #3152, opened 2024-08-20
logging: reduce noise in quay logs and add missing worker names

sayalibhavsar:
Changing logger.info to logger.debug to stop too many INFO log lines from displaying in Quay logs

### DIFF
--- a/util/migrate/allocator.py
+++ b/util/migrate/allocator.py
@@ -1,4 +1,4 @@
-# change3
+# change4
 import logging
 import random
 from threading import Event
--- a/util/migrate/allocator.py
+++ b/util/migrate/allocator.py
@@ -4,7 +4,6 @@ from threading import Event
 
 from bintrees import RBTree
 
-# Assuming the logger level is set globally or in the application's configuration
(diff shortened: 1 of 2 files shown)


util/migrate/allocator.py
```diff
@@ -141,19 +116,9 @@ def get_block_start_index(self, block_size_estimate):
...
 ):
-    """
-    This method will yield items from random blocks in the database.
-
-    We will track metadata about which keys are available for work, and we will complete the
-    backfill when there is no more work to be done. The method yields tuples of (candidate, Event),
-    and if the work was already done by another worker, the caller should set the event. Batch
-    candidates must have an "id" field which can be inspected.
-    """
-
```
ibazulic, line 152: I don't think we should remove method descriptions.

ibazulic:
Not sure I understand what is the goal here. Can you please explain what are you trying to achieve?

sayalibhavsar:
HI Ivan 
Those are undesired changes that got pulled in as a byproduct of another work. Let me close it

PR #3179, opened 2024-08-29
deps: Bump werkzeug to 3.0.3 (PROJQUAY-7142)

harishsurf:


ibazulic:
It seems that this will require much more work to get it to function. As it currently stands, it breaks half of Quay, tried to upgrade it on my rig a couple of months ago and found that half of things weren't operational. Oleg was working on this last year but didn't finish it: #2195 .

bcaton85 (approved):

PR #3188, opened 2024-09-03
api: adding tag_pattern to autoprune API (PROJQUAY-7668)

bcaton85:
Allows users to specify a regex tag pattern when creating namespace/repository autoprune policies via the new UI. Users will have the option to prune tags that only match the tag pattern or exclude tags that match the tag pattern.

### DIFF
--- a/data/model/oci/tag.py
+++ b/data/model/oci/tag.py
@@ -15,6 +15,7 @@ from data.database import (
     Tag,
     User,
     db_random_func,
+    db_regex_search,
     db_transaction,
     get_epoch_timestamp_ms,
 )
@@ -780,7 +781,7 @@ def reset_child_manifest_expiration(repository_id, manifest, expiration=None):
 
 
 def fetch_paginated_autoprune_repo_tags_by_number(
-    repo_id, max_tags_allowed: int, items_per_page, page
+    repo_id, max_tags_allowed: int, items_per_page, page, tag_pattern=None, tag_pattern_matches=True
 ):
     """
     Fetch repository's active tags sorted by creation date & are more than max_tags_allowed
@@ -801,6 +802,13 @@ def fetch_paginated_autoprune_repo_tags_by_number(
             .offset(tags_offset)
             .limit(items_per_page)
         )
+        if tag_pattern is not None:
+            query = db_regex_search(
+                Tag.select(query.c.name).from_(query),
+                query.c.name,
+                tag_pattern,
+                matches=tag_pattern_matches,
+            )
         return list(query)
     except Exception as err:
         raise Exception(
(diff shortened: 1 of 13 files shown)


data/text.py
```diff
@@ -56,3 +59,16 @@ def match_like(field, search_query):
...
+    return query.where(field.regexp(pattern)) if matches else query.where(~field.regexp(pattern))
+
+
+def regex_sqlite(query, field, pattern, matches=True):
+    rows = query.execute()
+    return (
+        [row for row in rows if re.search(pattern, getattr(row, field.name))]
+        if matches
+        else [row for row in rows if not re.search(pattern, getattr(row, field.name))]
+    )
```
Sunandadadi, line 74: Instead of iterating over all rows, will `like 'pattern'` work here?

data/model/oci/tag.py
```diff
@@ -801,6 +802,13 @@ def fetch_paginated_autoprune_repo_tags_by_number(
             .offset(tags_offset)
             .limit(items_per_page)
         )
+        if tag_pattern is not None:
+            query = db_regex_search(
+                Tag.select(query.c.name).from_(query),
+                query.c.name,
```
Sunandadadi, line 808: what does `query.c.name` return? why are we querying for `Tag.select(query.c.name).from_(query)` here while we are just passing the query [here](https://github.com/quay/quay/pull/3188/files#diff-d54089ecfb2c9ed6312fb03ca1ea1cd1be9555c8dedf9138a2feac65ea530ff5R845)

data/text.py
bcaton85: `like` uses a SQL specific way of pattern matching while we want to match via regex patterns. Currently sqlite doesn't support regex pattern matching

data/model/oci/tag.py
bcaton85: `query.c.name` is the same as `Tag.name` from the original query. It's being used as a subquery via the `.from_(query)` because we only want to match on tags that **exceed** the limit. Any tags that match the regex will still be counted towards the tag count limit.

Sunandadadi: I am just trying to understand why you chose to use `query.c.name` here but `Tag.name` [here](https://github.com/quay/quay/pull/3188/files#diff-d54089ecfb2c9ed6312fb03ca1ea1cd1be9555c8dedf9138a2feac65ea530ff5R845)

data/text.py
Sunandadadi: yes, that makes sense

data/model/oci/tag.py
bcaton85: It's because of how peewee alias' the tables. The subquery is aliased as `T2` but the query references `Tag.name` as `T1.name` but since we're using a subquery in the `from` there is no `T1`. So that results in an error similar to `T1 does not exist`.

Sunandadadi: ok, that makes sense.

Sunandadadi (approved):
lgtm

PR #3160, opened 2024-08-23
logging: hide entries when debuglog is set to false and add missing worker names (PROJQUAY-6562)

sayalibhavsar:
1. Stop log entries from the util.migrate.allocator.py module from being displayed in Quay logs when the debug log is false.
2. Only show them when the debug log is set to true with the name of the worker printing them

PR #3193, opened 2024-09-05
logging: hide entries when debuglog is set to false and add missing worker names (PROJQUAY-6562)

sayalibhavsar:
fixing log discrepancies as per reviewer's suggestion from https://github.com/quay/quay/pull/3160

### DIFF
--- a/util/migrate/allocator.py
+++ b/util/migrate/allocator.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import random
+import time
 from threading import Event
 
 from bintrees import RBTree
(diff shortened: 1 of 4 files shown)


util/migrate/allocator.py
```diff
@@ -10,11 +11,18 @@
...
-else:
-    logging.basicConfig(level=logging.INFO)
+
+# Define the logging format to include worker_name and stream_type
+log_format = "%(asctime)s [%(process)d] [%(levelname)s] [%(name)s] %(message)s"
+
+# Set the logging level and format based on DEBUGLOG
+
+# Set up logging
+logging.basicConfig(
```
Marcusk19, line 21: Should we be using [basicConfig](https://docs.python.org/3/library/logging.html#logging.basicConfig) here? It seems like this configures the root logger which might not be what we want

sayalibhavsar: You are right. logging.basicConfig affects all logging calls and we want granular control for module util.migrate.allocator. Doing it with [setLevel](https://docs.python.org/3/library/logging.html#logging.Logger.setLevel) with handlers and formatters

Marcusk19 (approved):
lgtm

PR #3206, opened 2024-09-09
[redhat-3.12] logging: hide entries when debuglog is set to false and add missing worker names (PROJQUAY-6562)

sayalibhavsar:
This is a manually cherry pick of #3193


Marcusk19 (approved):
lgtm

PR #3209, opened 2024-09-10
autoprune: support multiple organization and repository policies (PROJQUAY-7695)

Sunandadadi:


bcaton85 (changes requested):
Overall looks great. We should change the autoprune worker to execute the repository policies once [here](https://github.com/quay/quay/blob/PROJQUAY-7695/data/model/autoprune.py#L710-L719). There could also be some UI improvements for deleting and listing policies. The pruning itself works great.

bcaton85 (approved):

PR #3224, opened 2024-09-13
bug: fixing NaN value error for quota displayed on member org page (PROJQUAY-6465)

deshpandevlab:
https://issues.redhat.com/browse/PROJQUAY-6465

Current member view:
orig_member_org_quota_view

Fixed member view:
member_org_quota_view

Unauth view of the organization:
unauth_org_view

### DIFF
--- a/endpoints/api/organization.py
+++ b/endpoints/api/organization.py
@@ -106,10 +106,10 @@ def org_view(o, teams):
         view["tag_expiration_s"] = o.removed_tag_expiration_s
         view["is_free_account"] = o.stripe_id is None
 
-        if features.QUOTA_MANAGEMENT and features.EDIT_QUOTA:
-            quotas = model.namespacequota.get_namespace_quota_list(o.username)
-            view["quotas"] = [quota_view(quota) for quota in quotas] if quotas else []
-            view["quota_report"] = model.namespacequota.get_quota_for_view(o.username)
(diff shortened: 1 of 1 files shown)


endpoints/api/organization.py
```diff
@@ -106,10 +106,10 @@ def org_view(o, teams):
         view["tag_expiration_s"] = o.removed_tag_expiration_s
         view["is_free_account"] = o.stripe_id is None
 
-        if features.QUOTA_MANAGEMENT and features.EDIT_QUOTA:
-            quotas = model.namespacequota.get_namespace_quota_list(o.username)
-            view["quotas"] = [quota_view(quota) for quota in quotas] if quotas else []
-            view["quota_report"] = model.namespacequota.get_quota_for_view(o.username)
+    if features.QUOTA_MANAGEMENT and features.EDIT_QUOTA:
```
Sunandadadi, line 109: do we want `EDIT_QUOTA` to be enabled to view quota?

bcaton85, line 109: How will this effect permissions? Since we're removing the check if the user is admin we'll most likely want a check if the user is a team member.

deshpandevlab: > How will this effect permissions? Since we're removing the check if the user is admin we'll most likely want a check if the user is a team member.

will update in order for quota to be accessible only users with members or admin permissions.

Sunandadadi (approved):
lgtm

PR #3299, opened 2024-10-03
api: put /superuser/users/<username> will now show up in swagger (PROJQUAY-7579)

sayalibhavsar:
remove internal-only decorator from PUT request


Marcusk19 (approved):
lgtm

PR #3302, opened 2024-10-03
marketplace: remove database lookup when retrieving customer ids (PROJQUAY-8017)

Marcusk19:
Removes the database lookup from occurring when getting account number for a user in quay.io.

Have a user that is unable to view their subscription in quay.io. When looking at the database, only one of their accounts is present for the user, but the subscription is attached to another account they are part of. Quay is unable to retrieve the subscription for the account because if it finds an account number in the database it uses that and doesn't query the user api.

Skip querying the database so the user api is used as the source of truth when looking for customer ids.

### DIFF
--- a/util/marketplace.py
+++ b/util/marketplace.py
@@ -22,13 +22,7 @@ class RedHatUserApi(object):
 
     def get_account_number(self, user):
         email = user.email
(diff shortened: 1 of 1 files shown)


kleesc (approved):
lgtm

PR #3310, opened 2024-10-07
api: allowing superuser to read catalog endpoint (PROJQUAY-8023)

bcaton85:
Allows for the full access and global readonly superuser to read all repositories from the `/v2/_catalog` endpoint.

### DIFF
--- a/endpoints/api/__init__.py
+++ b/endpoints/api/__init__.py
@@ -490,7 +490,7 @@ log_unauthorized_delete = log_unauthorized("delete_tag_failed")
 
 
 def allow_if_superuser():
-    return features.SUPERUSERS_FULL_ACCESS and SuperUserPermission().can()
+    return bool(features.SUPERUSERS_FULL_ACCESS and SuperUserPermission().can())
 
 
 def allow_if_global_readonly_superuser():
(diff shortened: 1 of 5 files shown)


ibazulic:
@bcaton85 is this influenced by the following featore flag?

https://github.com/quay/quay/blob/9fa48ea376260a2562d5e33a187f5ea68268a86a/config.py#L396-L398

Or does this flag still need to be turned on for `_catalog` to return everything?

bcaton85:
That flag shouldn't make a difference if the user calling is a superuser.

endpoints/api/__init__.py
```diff
@@ -490,7 +490,7 @@ def wrapper(*args, **kwargs):
 
 
 def allow_if_superuser():
-    return features.SUPERUSERS_FULL_ACCESS and SuperUserPermission().can()
+    return bool(features.SUPERUSERS_FULL_ACCESS and SuperUserPermission().can())
```
kleesc, line 493: You don't really need this do you?

bcaton85: the type returned ends up being `FeatureNameValue` because of the way the conditional runs. I'm setting it so the linter doesn't complain and also to be explicit

kleesc (approved):
LGTM

PR #3314, opened 2024-10-10
autoprune: fix tags returned by autoprune worker (PROJQUAY-8070)

Sunandadadi:
We have the offset being applied before querying for tag_pattern. This is returning wrong results. So, moved offset to be applied on `db_regex_search` if tag_pattern exists.

### DIFF
--- a/data/text.py
+++ b/data/text.py
@@ -61,12 +61,16 @@ def match_like(field, search_query):
     return Field.__pow__(field, clause)
 
 
-def regex_search(query, field, pattern, matches=True):
-    return query.where(field.regexp(pattern)) if matches else query.where(~field.regexp(pattern))
+def regex_search(query, field, pattern, offset, matches=True):
+    return (
+        query.where(field.regexp(pattern)).offset(offset)
+        if matches
+        else query.where(~field.regexp(pattern)).offset(offset)
+    )
 
 
-def regex_sqlite(query, field, pattern, matches=True):
-    rows = query.execute()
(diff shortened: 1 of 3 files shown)


data/text.py
```diff
@@ -61,14 +61,19 @@ def match_like(field, search_query):
...
+    return (
+        query.where(field.regexp(pattern)).offset(offset).limit(limit)
+        if matches
+        else query.where(~field.regexp(pattern)).offset(offset).limit(limit)
+    )
 
 
-def regex_sqlite(query, field, pattern, matches=True):
+def regex_sqlite(query, field, pattern, offset, limit, matches=True):
     rows = query.execute()
```
bcaton85, line 73: Can we get a warning and/or a comment saying that the query is going to fetch all rows regardless of what offset and limit is set to?

workers/test/test_autopruneworker.py
```diff
@@ -711,24 +711,29 @@ def test_registry_prune_invalid_policy(initialized_db):
...
         (
             ["match1", "match2", "match3", "match4", "match5"],
             ["match1", "match2", "match3", "match4", "match5"],
             False,
         ),
+        (
+            ["test1", "test2", "test3", "test4", "test5"],
+            ["test1", "test2", "test3"],
+            False,
+        ),
```
bcaton85, line 736: I have some additional test cases that i'll send over

data/text.py
Sunandadadi: good call, added.

bcaton85 (approved):

PR #3322, opened 2024-10-12
workers(quotaregistrysize): added to run quotasizeworker when SUPERUSERS are retrieved from LDAP (PROJQUAY-8086)

michaelalang:
worker(quotaregistrysize) lacks the check to run calculation if no **superusers** are defined statically but retrieved from **LDAP** instead.

Reproducer:
```
$ grep 'SUPERUSER' conf/stack/config.yaml
FEATURE_NONSUPERUSER_TEAM_SYNCING_SETUP: false
LDAP_SUPERUSER_FILTER: (|(memberOf=cn=quay-superusers,ou=Groups,dc=example,dc=com)(memberOf=cn=quay-readonly-superuser,ou=Groups,dc=example,dc=com))
FEATURE_SUPERUSERS_FULL_ACCESS: true
```
```
$  python workers/quotaregistrysizeworker.py
Super users disabled or none specified; skipping quotaregistrysizeworker
```

### DIFF
--- a/workers/quotaregistrysizeworker.py
+++ b/workers/quotaregistrysizeworker.py
@@ -39,10 +39,14 @@ if __name__ == "__main__":
             time.sleep(100000)
 
     # Registry size is only viewable by superusers, don't calculate if not used
-    if not any(
+    if not all(
         [
-            features.SUPER_USERS,
-            len(app.config.get("SUPER_USERS", [])) == 0,
+            any(
+                [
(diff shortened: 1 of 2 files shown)


michaelalang:
Codecov doesn't understand the logic in the tests covering the three use-cases of running the worker or not.
Who ever approves, please double check that I did not miss any condition.

PR #3341, opened 2024-10-18
ui: conditionally support custom tag expiration durations (PROJQUAY-6700)

dmesser:
This adds support for user-supplied tag expiration values from `TAG_EXPIRATION_OPTIONS` via config.yaml in the new UI. In the current code, these values weren't properly processed and led to the bug described in https://issues.redhat.com/browse/PROJQUAY-6700

This implementation also adds a number of additional improvements:

- conditional display of the time machine configuration depending on the `CHANGE_TAG_EXPIRATION` feature in the account settings
- actualy support for setting the time machines' default tag expiration in the account settings UI
- improved logic to set the currently selected expiration value in the dropdown list (based on the value in seconds instead of the human readable form)
- improved human-readable output of custom durations that prevent excessive rounding, e.g. `80d` is now displayed as 80 days instead of 3 months
- cypress unit test coverage
- the UI form will now also ignore values from `TAG_EXPIRATION_OPTIONS` that neither moment.js nor Quay supports

Attached a screenshot how the user-supplied tag expiration options are now rendered in human-friendly form in the new UI with the following setting from the bug report:

```
TAG_EXPIRATION_OPTIONS:
- 0s
- 30m
- 1800s    
- 1h
- 2w
- 4w
- 10w
- 6048000s
- 80d 
```

Screenshot 2024-10-18 at 09 27 38

### DIFF
--- a/web/cypress/e2e/account-settings.cy.ts
+++ b/web/cypress/e2e/account-settings.cy.ts
@@ -1,5 +1,7 @@
 /// <reference types="cypress" />
 
+import {humanizeTimeForExpiry, parseTimeDuration} from 'src/libs/utils';
+
 describe('Account Settings Page', () => {
   beforeEach(() => {
     cy.exec('npm run quay:seed');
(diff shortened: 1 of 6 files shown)


Sunandadadi (approved):
lgtm

PR #3370, opened 2024-10-29
ui: tag expiration settings support year values (PROJQUAY-8171)

dmesser:
Adds support for yearly values as tag expiration options with the `y` suffix.


Sunandadadi (approved):
lgtm

PR #3378, opened 2024-11-05
endpoints(quota): update Quota Human readable parsing/handling

michaelalang:
As described in https://issues.redhat.com/browse/RFE-6540, the Quota API `/v1/superuser/users/<namespace>/quota` and `/v1/superuser/organization/<namespace>/quota` on the `PUT` and `POST` methods do not validate the submitted values correctly in relation to the supported Unit values of the underlying module. Furthermore, it is not possible to utilize floats that are anyway converted to bytes which are stored as `integer`.

An explicit statement in the Documentation on expected values while converting between values multiplied by `1000` (KB, MB, GB,..) and `1024` (KiB, MiB, GiB) should simplify expectation of customers.

Examples:

```
$ curl -H "Authorization: Bearer $(cat .token)" \
    -H "Content-Type: application/json"  \
    -d '{"limit": "100.5 GiB", "path": "milang"}' \
     https://quay.apps.example.com/api/v1/superuser/users/milang/quota -s | jq -r 
Created
```

without the PR the value of `100.5 GiB` isn't accepted due to the matching in validation of `\d+`. That means, if User want to set `100.5 GiB` they would first need to convert it into `bytes` to be accepted.

On any of `creating` / `update` of the Quota the additional check to not exceed the max value of `BigInteger` from Postgres was added and a error response with that limit was added ...

```
$ curl -X PUT -H "Authorization: Bearer $(cat .token)" \
    -H "Content-Type: application/json" \
    -d '{"limit_bytes": 9223372036854775808}' \
    https://quay.apps.example.com/api/v1/superuser/users/milang/quota/4 -s | /
    jq -r 
{
  "detail": "Maximum supported Quota is less than 8 EiB",
  "error_message": "Maximum supported Quota is less than 8 EiB",
  "error_type": "invalid_request",
  "title": "invalid_request",
  "type": "https://quay.apps.example.com/api/v1/error/invalid_request",
  "status": 400
}

$ curl -H "Authorization: Bearer $(cat .token)"  \
    -H "Content-Type: application/json" \
    -d '{"limit": "100.5 YB", "path": "milang"}' \
    https://quay.apps.example.com/api/v1/superuser/users/milang/quota -s | \
    jq -r 
{
  "detail": "Maximum supported Quota is less than 8 EiB",
  "error_message": "Maximum supported Quota is less than 8 EiB",
  "error_type": "invalid_request",
  "title": "invalid_request",
  "type": "https://quay.apps.example.com/api/v1/error/invalid_request",
  "status": 400
}
``` 

Any unsupported `Unit` will be returned as error from the module utilised for conversion.
```
$ curl -H "Authorization: Bearer $(cat .token)" \
    -H "Content-Type: application/json" \
    -d '{"limit": "100.5 YiB", "path": "milang"}' \
    https://quay.apps.example.com/api/v1/superuser/users/milang/quota -s | \
    jq -r 
{
  "detail": "'100.5 YiB' does not match '^(\\\\d+(|\\\\.\\\\d+)\\\\s?(Bit|Byte|kb|kB|Mb|MB|Gb|GB|Tb|TB|Pb|PB|Eb|EB|Zb|ZB|Yb|YB|Kib|KiB|Mib|MiB|Gib|GiB|Tib|TiB|Pib|PiB|Eib|EiB)?)$'\n\nFailed validating 'pattern' in schema[1]['properties']['limit']:\n    {'description': 'Human readable storage capacity of the '\n                    'organizationMaximum supported limit is less than 8 '\n                    'EiB',\n     'pattern': '^(\\\\d+(|\\\\.\\\\d+)\\\\s?(Bit|Byte|kb|kB|Mb|MB|Gb|GB|Tb|TB|Pb|PB|Eb|EB|Zb|ZB|Yb|YB|Kib|KiB|Mib|MiB|Gib|GiB|Tib|TiB|Pib|PiB|Eib|EiB)?)$',\n     'type': 'string'}\n\nOn instance['limit']:\n    '100.5 YiB'",
  "error_message": "'100.5 YiB' does not match '^(\\\\d+(|\\\\.\\\\d+)\\\\s?(Bit|Byte|kb|kB|Mb|MB|Gb|GB|Tb|TB|Pb|PB|Eb|EB|Zb|ZB|Yb|YB|Kib|KiB|Mib|MiB|Gib|GiB|Tib|TiB|Pib|PiB|Eib|EiB)?)$'\n\nFailed validating 'pattern' in schema[1]['properties']['limit']:\n    {'description': 'Human readable storage capacity of the '\n                    'organizationMaximum supported limit is less than 8 '\n                    'EiB',\n     'pattern': '^(\\\\d+(|\\\\.\\\\d+)\\\\s?(Bit|Byte|kb|kB|Mb|MB|Gb|GB|Tb|TB|Pb|PB|Eb|EB|Zb|ZB|Yb|YB|Kib|KiB|Mib|MiB|Gib|GiB|Tib|TiB|Pib|PiB|Eib|EiB)?)$',\n     'type': 'string'}\n\nOn instance['limit']:\n    '100.5 YiB'",
  "error_type": "invalid_request",
  "title": "invalid_request",
  "type": "https://quay.apps.example.com/api/v1/error/invalid_request",
  "status": 400
}
```

### DIFF
--- a/endpoints/api/namespacequota.py
+++ b/endpoints/api/namespacequota.py
@@ -87,8 +87,11 @@ class OrganizationQuotaList(ApiResource):
                     "properties": {
                         "limit": {
                             "type": "string",
-                            "description": "Human readable storage capacity of the organization",
-                            "pattern": r"^(\d+\s?(B|KiB|MiB|GiB|TiB|PiB|EiB|ZiB|YiB|Ki|Mi|Gi|Ti|Pi|Ei|Zi|Yi|KB|MB|GB|TB|PB|EB|ZB|YB|K|M|G|T|P|E|Z|Y)?)$",
+                            "description": "Human readable storage capacity of the organization"
+                            + "Maximum supported limit is less than 8 EiB",
+                            "pattern": r"^(\d+(|\.\d+)\s?("
+                            + "|".join(bitmath.ALL_UNIT_TYPES)
+                            + ")?)$",
                         },
                     },
                 },
@@ -153,6 +156,10 @@ class OrganizationQuotaList(ApiResource):
         if quotas:
             raise request_error(message="Organization quota for '%s' already exists" % orgname)
 
+        if not limit_bytes <= int(bitmath.parse_string_unsafe("8 EiB").to_Byte().value) - 1:
+            # the Postgres maximum of an BigInteger is 9223372036854775807
+            raise request_error(message="Maximum supported Quota is less than 8 EiB")
+
         try:
             model.namespacequota.create_namespace_quota(org, limit_bytes)
             return "Created", 201
@@ -183,8 +190,11 @@ class OrganizationQuota(ApiResource):
                     "properties": {
                         "limit": {
                             "type": "string",
-                            "description": "Human readable storage capacity of the organization",
-                            "pattern": r"^(\d+\s?(B|KiB|MiB|GiB|TiB|PiB|EiB|ZiB|YiB|Ki|Mi|Gi|Ti|Pi|Ei|Zi|Yi|KB|MB|GB|TB|PB|EB|ZB|YB|K|M|G|T|P|E|Z|Y)?)$",
+                            "description": "Human readable storage capacity of the organization"
+                            + "Maximum supported limit is less than 8 EiB",
+                            "pattern": r"^(\d+(|\.\d+)\s?("
+                            + "|".join(bitmath.ALL_UNIT_TYPES)
+                            + ")?)$",
                         },
                     },
                     "required": ["limit"],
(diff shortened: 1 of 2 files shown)


ibazulic:
Can you please fix the tests as well? If this goes in? Also, it seems that the `error_message` and `detail` return the same thing at the moment. We should probably put only the `error` in the error message and the regex pattern (or some other explanation) in the detail.

michaelalang:
@ibazulic sure ... I didn't check on any of the Unittests so far as I wasn't sure how it would return. Of Course I will update the Test as well now that I have some response on what the test does. Thanks for the input <3

PR #3380, opened 2024-11-05
registry: support for extending quay with pulp/rpm support

kleesc:


pulp/rpm.py
```diff
@@ -0,0 +1,66 @@
...
+
+@pulp_bp.route(BASE_PULP_PLUGIN_RPM_ROUTE, methods=["GET"])
+@parse_repository_name()
+@anon_protect
+def pulp_plugin_support(namespace_name, repo_name):
+    raise NotImplementedError()
+
+
+@pulp_bp.route(BASE_PULP_PLUGIN_RPM_ROUTE, methods=["POST"])
+@route_show_if(features.PULP_PLUGIN)
```
bcaton85, line 38: `PULP_PLUGIN` needs to be added to features

data/model/gc.py
```diff
@@ -201,6 +189,8 @@ def _purge_repository_contents(repo):
         if not found:
             break
 
+    # TODO(kleesc): Check for orphaned manifests that woudln't be caught be the above
```
bcaton85, line 192: Can this be a JIRA ticket instead? We never really come back to take care of TODO's

ibazulic:
Can anything here be used for other artifact types? Or will we need to write a specific implementation for every artifact that we want to support?

pulp/rpm.py
```diff
@@ -0,0 +1,66 @@
...
+@check_readonly
+@check_pushes_disabled
+def pulp_publish_rpm(namespace_name, repo_name):
+    ref_name = "/".join([namespace_name, repo_name])
+
+    pulp_repo = pulp_client.get_repository(ref_name)
+    if pulp_repo.json().get("count") == 0:
+        assert pulp_client.create_repository(ref_name).status_code // 100 == 2
+
+    repository_prn = pulp_repo.json()["results"][0]["prn"]
```
bcaton85, line 53: If the repository count is 0 (`pulp_repo.json().get("count") == 0`) won't this result in an index error? Do we need another variable for the created repo on line 52 and reference that in the scenario we're creating it on demand?

pulp/rpm.py
```diff
@@ -0,0 +1,66 @@
...
+    pulp_dist = pulp_client.get_distribution(ref_name)
+    if pulp_dist.json().get("count") == 0:
+        assert pulp_client.create_distribution(ref_name, repository_prn, basepath=ref_name).status_code // 100 == 2
+
+    fname, file_obj = next(request.files.items())
+    filename = secure_filename(fname)
+    # file_obj.save(os.path.join(app.config['UPLOAD_FOLDER'], filename))
+
+    # content_upload = pulp_client.create_content(repository_prn, os.path.join(app.config['UPLOAD_FOLDER'], filename))
+    content_upload = pulp_client.create_content(repository_prn, f=file_obj)
```
bcaton85, line 64: We'll need error handling here for uploads that fail

bcaton85, line 64: Will streaming this content support large rpm's? I would imagine it's possible to have some rpm's reach GB's in size right?

pulp/client.py
```diff
@@ -0,0 +1,241 @@
...
+    #     used by the `pulp` CLI tool
+    #     """
+    #     path = os.path.expanduser(path or "pulp/cli.toml")
+    #     with open(path, "rb") as fp:
+    #         config = toml.load(fp)
+    #     return cls(config["cli"])
+
+    def __init__(self, config):
+        self.config = config
+        self.timeout = 60
```
bcaton85, line 46: Is this timeout going to cut off larger pushes?

pulp/rpm.py
ibazulic: I'm not sure why it would cause an index error. If count is already 0, then we call the repo creation function. If it fails, the assertion will fail as well and an exception will be raised and execution will stop. If everything passes, then `repository_prn` should contain something after execution, `["results"][0]["prn"]` should not be `Null` but should have some value.

bcaton85: won't `pulp_repo.json()` will still contain the results from the first call when the repo didn't exist? So `len(pulp_repo.json()["results"]) == 0`

pulp/rpm.py
ibazulic: RPMs are usually up to several hundred MB in size. Looking at CentOS Stream for instance, the biggest package I found in `AppStrean` repo is `java-21-openjdk-jmods-21.0.4.0.7-1.el10.x86_64.rpm` which is 311 MB in size. Biggest source pacakge is for `TeXLive` which is almost 590 MB.

pulp/rpm.py
ibazulic: Oh, now I get what you're saying. Yeah, we're not saving the result from the `create_repo` function.

ibazulic:
Another global question, this might be for a later discussion: 

From Quay's point of view, do we track distributions or individual RPMs? These create/delete functions call on Pulp to create RPM repo or delete it, I assume this is stored in Pulp's database. What happens if a deletion is done in Pulp? Does Pulp call Quay then? I assume both functionalities will have to be implemented on our side, right?

pulp/rpm.py
bcaton85: okay not bad

pulp/rpm.py
syed: can we add pulp tu the generic plugin framework? We have another plugin for model registry that needs to be added as well

pulp/rpm.py
ipanova: this is not a freqent thing, but we've encountered rpms being almost a GB

PR #3385, opened 2024-11-07
fix(ui): Show the Apple icon when Darwin is used

gastaldi:
- Because there isn't a `fa-darwin` entry in FontAwesome, it should use `fa-apple`  instead
- It also fixes the icon when the os is unknown


jbpratt:
Thanks for the PR! Closing this as the old UI is deprecated.

PR #3389, opened 2024-11-07
marketplace: splittable sku for MW02702 (PROJQUAY-8151)

Marcusk19:
* Alembic migration to drop unique constraint on the orgrhsubscriptions table
* Can split sub quantities of MW02702 across multiple orgs
* Can specify quantity for the MW02702 SKU across orgs on react UI
* Update angular UI to allow user to specify quantities for MW02702
* Write unit tests for splittable skus

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -2022,7 +2022,7 @@ class OrganizationRhSkus(BaseModel):
     RH subscriptions
     """
 
-    subscription_id = IntegerField(index=True, unique=True)
+    subscription_id = IntegerField(index=True)
     user_id = ForeignKeyField(User, backref="org_bound_subscription")
     org_id = ForeignKeyField(User, backref="subscription")
     quantity = IntegerField(index=True, null=True)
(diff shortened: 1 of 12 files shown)


data/database.py
```diff
@@ -2022,7 +2022,7 @@ class OrganizationRhSkus(BaseModel):
     RH subscriptions
     """
 
-    subscription_id = IntegerField(index=True, unique=True)
+    subscription_id = IntegerField(index=True)
     user_id = ForeignKeyField(User, backref="org_bound_subscription")
     org_id = ForeignKeyField(User, backref="subscription")
```
kleesc, line 2027: Note: when defining peewee models, you don't need `_id` at the end of field names. Peewee allows you to access the id by `<field_name>_id`. In this case, if you access `user_id`, will that return a reference to the other model, or the actual db id? e.g. `org_sku.user_id`

kleesc (approved):
lgtm

PR #3400, opened 2024-11-12
auth: Implement is_restricted_user for federated auth systems (PROJQUAY-8208)

ibazulic:
Currently, if any federated auth system is set as an authentication mechanism and restricted users is set, Quay will return a `501 Not Implemented` on invocation. Now, Quay will properly check the restricted user whitelist for federated users. 

Additionally, if user restriction is in place and super user's username was **not** explicitly whitelisted, super users would not be able to create new content inside the registry. Now, the username is explicitly checked in the UI to allow super users to create both organizations and repos regardless of restricted users whitelist.

### DIFF
--- a/data/users/__init__.py
+++ b/data/users/__init__.py
@@ -405,6 +405,12 @@ class UserManager(object):
 
         return self.state.is_restricted_user(username)
 
+    def is_superuser(self, username):
+        if not features.SUPER_USERS:
+            return False
+
+        return self.state.is_superuser(username)
+
 
 class FederatedUserManager(ConfigUserManager):
     """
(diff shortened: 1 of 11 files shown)


data/users/__init__.py
```diff
@@ -405,6 +405,12 @@ def is_restricted_user(self, username):
 
         return self.state.is_restricted_user(username)
 
+    def is_superuser(self, username):
+        if not features.SUPER_USERS:
+            return False
```
Sunandadadi, line 410: Can you add this check [here](https://github.com/quay/quay/blob/7c6a909130c1f8b65e4957567bdd16c5f4f096e9/data/users/__init__.py#L366) too?

endpoints/api/organization.py
```diff
@@ -158,8 +158,10 @@ def post(self):
         org_data = request.get_json()
         existing = None
 
-        if features.RESTRICTED_USERS and usermanager.is_restricted_user(user.username):
-            raise Unauthorized()
+        # Super users should be able to create new orgs regardless of user restriction
+        if user.username not in app.config.get("SUPER_USERS", None):
```
Sunandadadi, line 162: Can you confirm with Daniel on which should take precedence?

data/users/__init__.py
ibazulic: @Sunandadadi I didn't add it there because I don't see it added for other functions that also have the same logic.

Sunandadadi (dismissed):

Sunandadadi (approved):
lgtm

PR #3420, opened 2024-11-22
deps: Upgrade cryptography package to 43.0.3 (PROJQUAY-7032)

kleesc:


bcaton85:
@kleesc does this need to be cherrypicked into redhat-3.13 too?

kleesc:
@bcaton85 yes

PR #3427, opened 2024-11-26
oci: fix nested index parsing (PROJQUAY-8272)

syed:
image index can refer to another index when parsing. We currently assume only a manifest can be referenced from an index. This fixes the parse logic


Sunandadadi (approved):
lgtm

deshpandevlab (approved):

PR #3451, opened 2024-12-04
registry: handle empty blob downloads (PROJQUAY-8290)

kleesc:


endpoints/v2/blob.py
```diff
@@ -105,6 +105,9 @@ def download_blob(namespace_name, repo_name, digest, registry_model):
     # Build the response headers.
     headers = {"Docker-Content-Digest": digest}
 
+    if blob.compressed_size == 0:
+        return Response(headers=headers)
+
```
ibazulic, line 110: Wouldn't the client expect a download link here? Because during download, when a blob is requested, a 302 is returned with the location of the blob. What does it mean "empty"? Is that a file with 0 bytes? Or an "empty layer"?

kleesc: Either. You could have an entry in the imagestorage table with size 0, but storage side, Quay's not uploading an empty file. Client-side, the initial HEAD request would return 200, but the following GET requests would try to look for an non-existent key in S3, for example.

ibazulic: Ah, wait, I see it now, we return a redirect after we conjured a download link. So this response would actually be a `HTTP 200` because status is not defined?

PR #3526, opened 2024-12-20
storage(adjustment): ensure we default to empty for none CDN configuration

michaelalang:
CDN isn't a everyone's configuration Item. We might encounter that configurations without `CDN_SPECIFIC_NAMESPACES`  configured fail downloading blobs accordingly with an error Traceback of

```
gunicorn-registry stdout |   File "/quay-registry/endpoints/v2/blob.py", line 224, in _is_cdn_specific
gunicorn-registry stdout |     return namespace in namespaces or []
gunicorn-registry stdout | TypeError: argument of type 'NoneType' is not iterable
```


Marcusk19 (approved):
lgtm

PR #3550, opened 2025-01-06
teamsync: peewee integrity error check on teamsync (PROJQUAY-7747)

Sunandadadi:
This issue should fix [PROJQUAY-7747](https://issues.redhat.com/browse/PROJQUAY-7747) and [PROJQUAY-8130](https://issues.redhat.com/browse/PROJQUAY-8130). Adding `IntegrityError` to the try except block to skip adding existing user to team.

### DIFF
--- a/data/model/team.py
+++ b/data/model/team.py
@@ -65,12 +65,10 @@ def create_team(name, org_obj, team_role_name, description=""):
 
 
 def add_user_to_team(user_obj, team):
-    try:
-        return TeamMember.create(user=user_obj, team=team)
-    except Exception:
-        raise UserAlreadyInTeam(
-            "User %s is already a member of team %s" % (user_obj.username, team.name)
-        )
+    if user_exists_in_team(user_obj, team):
+        return
+
+    return TeamMember.create(user=user_obj, team=team)
 
 
 def remove_user_from_team(org_name, team_name, username, removed_by_username):
--- a/data/model/team.py
+++ b/data/model/team.py
@@ -170,7 +170,10 @@ def add_or_invite_to_team(inviter, team, user_obj=None, email=None, requires_inv
 
     # If we have a valid user and no invite is required, simply add the user to the team.
     if user_obj and not requires_invite:
-        add_user_to_team(user_obj, team)
+        if add_user_to_team(user_obj, team) is None:
+            raise UserAlreadyInTeam(
+                "User %s is already a member of team %s" % (user_obj.username, team.name)
+            )
         return None
 
     email_address = email if not user_obj else None
--- a/data/model/team.py
+++ b/data/model/team.py
@@ -619,3 +619,11 @@ def get_oidc_team_from_groupname(group_name, login_service_name):
             response.append(row)
 
     return response
+
+
+def user_exists_in_team(user_obj, team):
+    try:
+        TeamMember.get(TeamMember.user == user_obj, TeamMember.team == team)
+        return True
+    except TeamMember.DoesNotExist:
+        return False
--- a/data/model/team.py
+++ b/data/model/team.py
@@ -66,7 +66,9 @@ def create_team(name, org_obj, team_role_name, description=""):
 
 def add_user_to_team(user_obj, team):
     if user_exists_in_team(user_obj, team):
-        return
+        raise UserAlreadyInTeam(
+            "User %s is already a member of team %s" % (user_obj.username, team.name)
+        )
 
     return TeamMember.create(user=user_obj, team=team)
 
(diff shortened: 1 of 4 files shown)


data/users/externaloidc.py
```diff
@@ -150,7 +152,7 @@ def sync_oidc_groups(self, user_groups, user_obj):
                     logger.exception(
                         f"External OIDC Group Sync: Exception occurred when adding user: {user_obj.username} to quay team: {team_synced.team} as {err}"
                     )
-                except UserAlreadyInTeam:
+                except (IntegrityError, UserAlreadyInTeam):
```
bcaton85, line 155: Instead of catching the integrity error do we want to check if this entry exists instead? That would remove unnecessary errors from logs and would be easier to understand for the future

bcaton85, line 155: Can we also add a test case for this?

bcaton85 (changes requested):

data/model/team.py
```diff
@@ -608,14 +608,14 @@ def get_oidc_team_from_groupname(group_name, login_service_name):
...
-            .join(LoginService)
-            .where(TeamSync.config.contains(group_name), LoginService.name == login_service_name)
-        )
-        for row in query_result:
-            if json.loads(row.config).get("group_name", None) == group_name:
-                response.append(row)
+    query_result = (
+        TeamSync.select()
+        .join(LoginService)
+        .where(TeamSync.config.contains(group_name), LoginService.name == login_service_name)
```
bcaton85, line 614: is contains supported across postgres, mysql, and sqlite?

data/users/externaloidc.py
```diff
@@ -143,16 +145,16 @@ def sync_oidc_groups(self, user_groups, user_obj):
...
                     )
 
+                if team.user_exists_in_team(user_obj, team_synced.team):
+                    continue
+
                 # add user to team
                 try:
                     team.add_user_to_team(user_obj, team_synced.team)
-                except InvalidTeamException as err:
+                except Exception as err:
```
bcaton85, line 154: We don't want to catch general exceptions, as this won't surface real errors. We can keep the except block as it was before

bcaton85 (changes requested):

data/model/test/test_team.py
```diff
@@ -166,3 +170,38 @@ def test_get_federated_user_teams(login_service_name, initialized_db):
...
+
+    # add user to team already part of
+    with pytest.raises(UserAlreadyInTeam):
+        add_user_to_team(dev_user, team_1)
+
+    team_2 = create_team("team_2", new_org, "member")
+    assert user_exists_in_team(dev_user, team_2) is False
+
+
+def test_get_get_oidc_team_from_groupname(initialized_db):
```
bcaton85, line 191: typo `get_get`

data/model/team.py
```diff
@@ -65,12 +65,10 @@ def create_team(name, org_obj, team_role_name, description=""):
 
 
 def add_user_to_team(user_obj, team):
-    try:
-        return TeamMember.create(user=user_obj, team=team)
-    except Exception:
-        raise UserAlreadyInTeam(
```
bcaton85, line 71: Is there code that expects to catch this `UserAlreadyInTeam` error?

data/model/team.py
```diff
@@ -608,14 +609,22 @@ def get_oidc_team_from_groupname(group_name, login_service_name):
...
 
     return response
+
+
+def user_exists_in_team(user_obj, team):
+    try:
+        TeamMember.get(TeamMember.user == user_obj, TeamMember.team == team)
+        return True
+    except TeamMember.DoesNotExist:
+        return False
```
bcaton85, line 630: nit: doesn't need to be fixed now but this can just be `return TeamMember.select().where(TeamMember.user == user_obj, TeamMember.team == team).exists()`

data/users/externaloidc.py
```diff
@@ -2,6 +2,8 @@
 import logging
 from urllib.parse import urlparse
 
+from peewee import IntegrityError
```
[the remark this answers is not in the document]
Sunandadadi: no, will remove it

data/model/team.py
Sunandadadi: nice! will update

data/model/team.py
Sunandadadi: yes, [this](https://github.com/quay/quay/blob/master/endpoints/api/team.py#L441) method is expected to throw 400 when user is already a member of the team. I moved up this exception in [this](https://github.com/quay/quay/pull/3550/commits/7e458e8eeb45682a25c88ea82c87d397b35ac65c) commit.

data/model/team.py
```diff
@@ -608,14 +609,23 @@ def get_oidc_team_from_groupname(group_name, login_service_name):
...
     return response
+
+
+def user_exists_in_team(user_obj, team):
+    try:
+        return (
+            TeamMember.select().where(TeamMember.user == user_obj, TeamMember.team == team).exists()
+        )
+    except TeamMember.DoesNotExist:
+        return False
```
bcaton85, line 631: you don't need the `try ... except` here

bcaton85: if it doesn't exist `.exists()` will return false

data/model/team.py
Sunandadadi: If I am not wrong, peewee should convert `contains` based on the db.

bcaton85 (approved):

PR #3560, opened 2025-01-10
worker: include orgs in reconciler runs (PROJQUAY-8431)

Marcusk19:
Adds an `include_orgs` param to the active users query used by the reconciler and sets it to true for reconciler runs

Reconciler is not including orgs as a candidate for creating corresponding RH entitlements. As a result it misses users with stripe billing that are considered orgs.


ibazulic (approved):
LGTM.

PR #3566, opened 2025-01-15
util: parse forwarded headers for ip (PROJQUAY-8444)

Marcusk19:
Add `real_ip_recursive` directive to nginx config to properly get client IP addresses


syed (approved):
lgtm

PR #3592, opened 2025-01-29
grafana: update the slo dashboard (PROJQUAY-8455)

Marcusk19:
Adds adjustments to the SLO dashboard:
- error budget calculations are now done based on an SLO of 99.9 instead of 99.999
- ALB SLI is calculated based on a ratio of 5XX/2XX instead of 5XX/total # reqs

### DIFF
--- a/deploy/dashboards/grafana-dashboard-quay-slo.configmap.yaml
+++ b/deploy/dashboards/grafana-dashboard-quay-slo.configmap.yaml
@@ -96,7 +96,7 @@ data:
                   },
                   {
                     "color": "green",
-                    "value": 0.9999
+                    "value": 0.999
                   }
                 ]
               },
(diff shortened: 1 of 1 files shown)


deploy/dashboards/grafana-dashboard-quay-slo.configmap.yaml
```diff
@@ -398,11 +399,11 @@ data:
                   },
                   {
                     "color": "#EAB839",
-                    "value": 0.9999
+                    "value": 0.998
```
deshpandevlab, line 402: minor: should the display value be 0.999?

deshpandevlab: correcting myself: it's a warning line. LGTM

deshpandevlab (approved):

PR #3593, opened 2025-01-29
endpoints(user): added filter to avoid NoneType being return to function that expects objects (PROJQUAY-8497)

michaelalang:
Setting up a Quay with a config bundle containing:
```
PUBLIC_NAMESPACES:

some
someother
``´
will fail login in any User into the UI with
```
gunicorn-web stdout |   File "/quay-registry/endpoints/api/user.py", line 118, in org_view
gunicorn-web stdout |     admin_org = AdministerOrganizationPermission(o.username)
gunicorn-web stdout | AttributeError: 'NoneType' object has no attribute 'username'
``` 

reason for that is that the missing organization entries in the `user` table exlode to `None` in the function `list(organizations.values())`from `endpoints/api/user.py`

### DIFF
--- a/endpoints/api/test/test_user.py
+++ b/endpoints/api/test/test_user.py
@@ -1,5 +1,3 @@
-from test.fixtures import *
-
 import pytest
 from flask import url_for
 from mock import patch
@@ -8,6 +6,7 @@ from endpoints.api.test.shared import conduct_api_call
 from endpoints.api.user import User
 from endpoints.test.shared import client_with_identity, conduct_call
 from features import FeatureNameValue
+from test.fixtures import *
 
 
 def test_user_metadata_update(app):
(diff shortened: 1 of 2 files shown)


bcaton85:
So in this scenario the `some` organization has not been created yet?

michaelalang:
> So in this scenario the `some` organization has not been created yet?

ACK .. most likely will not happen in a running and used setup .. I was looking into a out-of-the-box mirror/proxy solution when I did hit the issue ...

bcaton85:
Okay looks good, can we add a simple test for this?

michaelalang:
> Okay looks good, can we add a simple test for this?

hmm need to check if it's easy possible to `screw` the config for one test ...

michaelalang:
> Okay looks good, can we add a simple test for this?

I don't see how to check this in the mock area to populate a Quay instance but no content so I added a method in the unittest emulating the situation and marked it also in the modified method to keep it linked.

I doubt that we can draft this differently in the unittest without rewriting ton's and that code is in since 2022-10-14 I doubt it's hitting a lot of people.

PR #3603, opened 2025-01-30
marketplace: handle exception thrown by stripe when customer has no subscription (PROJQUAY-8431)

Marcusk19:
If there is no subscription in the customer object found by stripe an `AttributeError` will be thrown. This should be caught and handled as if there is no subscription present so that the reconciler can continue its work for other users/orgs


kleesc (approved):
LGTM

PR #3632, opened 2025-02-04
chore: update thresholds for SLO error budgets (PROJQUAY-0000)

Marcusk19:
Update the thresholds for the error budget display so the colors are more accurate. Changing so that anything below 0.005 (0.5%) for the error budget is yellow and anything below 0% is red.


Sunandadadi (approved):
lgtm

PR #3643, opened 2025-02-06
storage: encoding signature and extending timeout (PROJQUAY-8511)

bcaton85:


storage/akamaistorage.py
```diff
@@ -69,6 +80,26 @@ def get_direct_download_url(
 
         # add akamai signed token
         try:
+            if self.amz_encryption_token is not None and self.amz_encryption_iv is not None:
+                query_params = urllib.parse.parse_qs(akamai_url_parsed.query)
+                amz_signature_params = query_params.get("X-Amz-Signature", [])
+                amz_signature = amz_signature_params[0] if len(amz_signature_params) == 1 else None
```
syed, line 86: can you check `encryption.py` to see if you can re-use those functions?

storage/akamaistorage.py
```diff
@@ -69,6 +80,26 @@ def get_direct_download_url(
 
         # add akamai signed token
         try:
+            if self.amz_encryption_token is not None and self.amz_encryption_iv is not None:
+                query_params = urllib.parse.parse_qs(akamai_url_parsed.query)
```
syed, line 84: you want `iv` to be random and added to the message, same thing with the `nonce`

bcaton85: there's a field for `iv` on the akamai side too, i assumed they had to match

syed: yes, in Akamai, you'll have to extract the nonce and iv from the encrypted message and store them in a variable there. Every request will have a different nonce and an IV

storage/akamaistorage.py
bcaton85: hm kind of. It uses AEM-CCM which we need to see if it's supported by akamai. The methods are also specific for the db but i can modify them to make something work if need be

storage/akamaistorage.py
bcaton85: so those fields in the akamai dashboard are for the key's to pick up on the iv and nonce? How does akamai find those values?

syed: you have to send those values in the message either as another query param

bcaton85: how will akamai consume those values though?

bcaton85: updated to return unique iv each time

storage/akamaistorage.py
bcaton85: akamai doesn't have an option for CCM encryption mode

storage/akamaistorage.py
```diff
@@ -69,6 +79,28 @@ def get_direct_download_url(
 
         # add akamai signed token
         try:
+            if self.amz_encryption_token is not None:
+                query_params = urllib.parse.parse_qs(akamai_url_parsed.query)
+                amz_signature_params = query_params.get("X-Amz-Signature", [])
+                amz_signature = amz_signature_params[0] if len(amz_signature_params) == 1 else None
+
+                if amz_signature is not None:
+                    padder = padding.PKCS7(algorithms.AES256.block_size).padder()
```
syed, line 88: can you wrap this encryption logic in a function in the `encryption.py` file

storage/test/test_akamai.py
```diff
@@ -46,7 +51,7 @@ def test_direct_download_cdn_specific(ipranges_populated, test_ip_range_cache, a
             "some/path",
             _TEST_BUCKET,
             _TEST_REGION,
-            None,
+            amz_encryption_token=token,
```
syed, line 54: The encryption token is not for Amazon. can you call it something else like `akamai_sig_encryption_key`

PR #3660, opened 2025-02-10
api: Disallow push of manifests with negative layer size (PROJQUAY-8560)

ibazulic:
Under certain conditions, clients may create a manifest (OCI or Docker v2) that contains negative layer sizes. Our current validation schema does not take that corner case into account, it only checks if the manifest is properly formatted or not. With this change, Quay will reject manifests that have negative layer sizes and raise a `400` with a proper exception. An example can be seen here:

~~~
gunicorn-registry stdout | 2025-02-10 22:34:54,930 [377] [ERROR] [endpoints.v2.manifest] failed to parse manifest when writing by tagname gunicorn-registry stdout | Traceback (most recent call last):
gunicorn-registry stdout |   File "/quay-registry/endpoints/v2/manifest.py", line 362, in _parse_manifest
gunicorn-registry stdout |     return parse_manifest_from_bytes(
gunicorn-registry stdout |   File "/quay-registry/image/shared/schemas.py", line 40, in parse_manifest_from_bytes
gunicorn-registry stdout |     return DockerSchema2Manifest(manifest_bytes)
gunicorn-registry stdout |   File "/quay-registry/image/docker/schema2/manifest.py", line 172, in __init__
gunicorn-registry stdout |     raise MalformedSchema2Manifest("layer size is negative")
gunicorn-registry stdout | image.docker.schema2.manifest.MalformedSchema2Manifest: negative layer size
gunicorn-registry stdout | 2025-02-10 22:34:54,931 [377] [DEBUG] [endpoints.v2] sending response: b'{"errors":[{"code":"MANIFEST_INVALID","detail":{"message":"failed to parse manifest: negative layer size"},"message":"manifest invalid"}]}\n
~~~

### DIFF
--- a/image/oci/manifest.py
+++ b/image/oci/manifest.py
@@ -169,7 +169,7 @@ class OCIManifest(ManifestInterface):
 
         for layer in self._parsed["layers"]:
             if not layer["size"] > 0:
-                raise MalformedOCIManifest("negative layer size")
+                raise MalformedOCIManifest("invalid layer size")
 
         for layer in self.filesystem_layers:
             if layer.is_remote and not layer.urls:
--- a/image/oci/manifest.py
+++ b/image/oci/manifest.py
@@ -167,6 +167,10 @@ class OCIManifest(ManifestInterface):
         except ValidationError as ve:
             raise MalformedOCIManifest("manifest data does not match schema: %s" % ve)
 
+        for layer in self._parsed["layers"]:
+            if not layer["size"] > 0:
+                raise MalformedOCIManifest("negative layer size")
+
         for layer in self.filesystem_layers:
             if layer.is_remote and not layer.urls:
                 raise MalformedOCIManifest("missing `urls` for remote layer")
--- a/endpoints/v2/test/test_manifest.py
+++ b/endpoints/v2/test/test_manifest.py
@@ -1,5 +1,5 @@
+import json
 import time
-from test.fixtures import *  # noqa: F401, F403
 from unittest.mock import patch
 
 from flask import url_for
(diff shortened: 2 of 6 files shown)


bcaton85:
How did they create a manifest with a negative layer size?

ibazulic:
> How did they create a manifest with a negative layer size?

That's a good question. @robbmanes is trying to replicate it with `podman/skopeo`, from what I understood, it looks like a racing condition during manifest assembly. Never the less, we still should not allow these manifests, these images are not pullable and they also appear in the UI with size `NaN` which might cause problems for calculating quotas and disk usage.

image/oci/manifest.py
```diff
@@ -167,6 +167,10 @@ def __init__(self, manifest_bytes, validate=False):
         except ValidationError as ve:
             raise MalformedOCIManifest("manifest data does not match schema: %s" % ve)
 
+        for layer in self._parsed["layers"]:
+            if not layer["size"] > 0:
```
kleesc, line 171: What about a layer with size `0`, should that be valid?
Also, can this be defined as part of the jsonschema?

ibazulic: @kleesc Discussion is a bit older:

https://github.com/opencontainers/image-spec/issues/350#issuecomment-250567107

@wking mentions in that discussion that layers must be positive. And that makes sense since the spec also says:

https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties

> `size` _int_ This REQUIRED property specifies the size in bytes of the blob. This property exists so that a client will have an expected size for the content before validating. If the length of the retrieved content does not match the specified length, the content SHOULD NOT be trusted.

In other words, if I get a layer and the layer size says `-1` or `0` and real layer size is `> 0`, then clients might reject such a download since they get more data then expected. For this reason, having these two layer sizes makes no sense.

In terms of where the check is done, I haven't explored the possibility that the check is done within JSON schema. Is there a problem with current implementation? It's a really simple `for` loop, should not take too long, even for huge manifests.

ibazulic:
Closing in favor of #3683 which has additional fixes.

PR #3666, opened 2025-02-12
storage(cloudfront): fixed presign uri for multi-region (PROJQUAY-8532)

michaelalang:
Fixed super initialize to include region_name in CloudFrontedS3Storage.
Added unittest for it.

### DIFF
--- a/storage/test/test_cloudfront.py
+++ b/storage/test/test_cloudfront.py
@@ -272,7 +272,7 @@ def test_direct_download_regions(
     ipresolver = IPResolver(app)
     context = StorageContext("nyc", None, config_provider, ipresolver)
     # Create a test bucket and put some test content.
-    boto3.client("s3").create_bucket(Bucket=bucket)
+    boto3.client("s3").create_bucket(Bucket="bucket")
     for region_name in boto3.Session("s3").get_available_regions("s3"):
         engine = CloudFlareS3Storage(
             context,


storage/cloud.py
```diff
@@ -1151,7 +1151,7 @@ def __init__(
         **kwargs,
     ):
         super(CloudFrontedS3Storage, self).__init__(
-            context, storage_path, s3_bucket, *args, **kwargs
+            context, storage_path, s3_bucket, s3_region=s3_region, *args, **kwargs
         )
 
         self.s3_region = s3_region
```
ibazulic, line 1157: Should we add a default region here? Like, `us-east-1`? Just for the sake of Quay.io, if nothing else.

bcaton85: @syed is it possible this will break anything on Quay.io? I think since we use [this optimization](https://github.com/quay/quay/blob/master/storage/cloud.py#L1175-L1178) it would return the s3 endpoint with the region (`s3.us-east-1.amazonaws.com`) in the host name, not sure if that would break any existing users?

syed: it should be fine. We should be using the region specific endpoint anyways

bcaton85 (approved):

storage/cloud.py
bcaton85: okay i ran a curl in a pod running in aws and it returned with the region in the hostname, so we should be good

PR #3670, opened 2025-02-13
[Feature] storage: Modify the STS S3 implementation of the storage backend to use Web Identity Tokens when available (PROJQUAY-8576)

bouchardmathieu-qc:
When deploying Quay in a Secure AWS environment, we can't use IAM Access Keys or Secrets since these credentials are often blocked for multiple reasons (credentials are long-lived, can be shared / stolen, etc.). So the preferred deployment method is to use an alternative method, like the Web Identity Token files that are automatically created in a Kubernetes cluster that has a federation link with IAM using the OIDC provider federation.

The current code of Quay force the use of an IAM account that is then used to assume another role that has S3 access to store the image files. The current pull request removes the need to use that IAM account and allows to directly assume the correct role using Web Identity Tokens while retaining compatibility with the old method of using IAM credentials.

The code relies on the automatic detection of the correct configurations using environment variables where possible. The code has been tested on an OpenShift cluster deployed using manual mode with AWS STS.

The Quay storage configuration used is:

```
DISTRIBUTED_STORAGE_CONFIG:
    default:
      - STSS3Storage
      - s3_bucket: test-quay-bucket
        storage_path: /datastorage/registry
```

The Quay pod is using a Service Account that has the following annotation:

```
kind: ServiceAccount
apiVersion: v1
metadata:
  name: quay
  namespace: quay
  annotations:
    eks.amazonaws.com/role-arn: 'arn:aws:iam::111111111111:role/test-quay-bucket-role'
```

That role has the following trust relationship:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Quay",
            "Effect": "Allow",
            "Principal": {
                "Federated": "arn:aws:iam::111111111111:oidc-provider/my-provider-url.com/my-cluster-name"
            },
            "Action": "sts:AssumeRoleWithWebIdentity",
            "Condition": {
                "StringEquals": {
                    "my-provider-url.com/my-cluster-name:sub": "system:serviceaccount:quay:quay"
                }
            }
        }
    ]
}
```

Basic tests that have been done:

- Start Quay in OpenShift with database / redis integration and a default route
- Created a test repository
- Pushed a test image in the repository

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -1245,28 +1245,39 @@ class STSS3Storage(S3Storage):
         maximum_chunk_size_gb=None,
         signature_version="s3v4",
     ):
-        sts_client = boto3.client(
-            "sts", aws_access_key_id=sts_user_access_key, aws_secret_access_key=sts_user_secret_key
-        )
-        assumed_role = sts_client.assume_role(RoleArn=sts_role_arn, RoleSessionName="quay")
-        credentials = assumed_role["Credentials"]
-        deferred_refreshable_credentials = DeferredRefreshableCredentials(
-            refresh_using=create_assume_role_refresher(
-                sts_client, {"RoleArn": sts_role_arn, "RoleSessionName": "quay"}
-            ),
-            method="sts-assume-role",
-        )
+        if sts_user_access_key == "" or sts_user_secret_key == "":
+            sts_client = boto3.client("sts")
+        else:
+            sts_client = boto3.client(
+                "sts",
+                aws_access_key_id=sts_user_access_key,
+                aws_secret_access_key=sts_user_secret_key,
+            )
 
         # !! NOTE !! connect_kwargs here initializes the S3Storage Class not the s3 connection (mis leading re-use of the name)
         connect_kwargs = {
-            "s3_access_key": credentials["AccessKeyId"],
-            "s3_secret_key": credentials["SecretAccessKey"],
-            "aws_session_token": credentials["SessionToken"],
             "s3_region": s3_region,
             "endpoint_url": endpoint_url,
             "maximum_chunk_size_gb": maximum_chunk_size_gb,
-            "deferred_refreshable_credentials": deferred_refreshable_credentials,
             "signature_version": signature_version,
         }
+        if sts_role_arn is not None:
+            assumed_role = sts_client.assume_role(RoleArn=sts_role_arn, RoleSessionName="quay")
+            credentials = assumed_role["Credentials"]
+            deferred_refreshable_credentials = DeferredRefreshableCredentials(
+                refresh_using=create_assume_role_refresher(
+                    sts_client, {"RoleArn": sts_role_arn, "RoleSessionName": "quay"}
+                ),
+                method="sts-assume-role",
+            )
+
+            connect_kwargs.update(
+                {
+                    "s3_access_key": credentials["AccessKeyId"],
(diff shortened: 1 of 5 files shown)


ibazulic:
How would this work for standalone Quay deployments on standalone VMs? That may or may not be deployed in AWS?

bouchardmathieu-qc:
Hi @ibazulic,

When deploying Quay locally, you can still use the older configuration with the same parameters (sts_role_arn, sts_user_access_key, sts_user_secret_key, etc.)

marbindrakon:
I've created PROJQUAY-8576 for this enhancement request

bouchardmathieu-qc:
Also note that the code is a little more complex that what i would have wanted because the golang implementation of Quay is using the AWS SDK for Go v1, which doesn't load Web Identity Tokens in the default credentials chain. So, I have to manually set the Web Identity Token configuration in the Go code.

With the AWS SDK for Go v2, the AWS_WEB_IDENTITY_TOKEN_FILE environment variable would have been used automatically and the code could be simplified to only creating a default session if the Access Key and Secret String are not provided (like the Python implementation).

storage/cloud.py
```diff
@@ -1245,28 +1245,39 @@ def __init__(
...
-        assumed_role = sts_client.assume_role(RoleArn=sts_role_arn, RoleSessionName="quay")
-        credentials = assumed_role["Credentials"]
-        deferred_refreshable_credentials = DeferredRefreshableCredentials(
-            refresh_using=create_assume_role_refresher(
-                sts_client, {"RoleArn": sts_role_arn, "RoleSessionName": "quay"}
-            ),
-            method="sts-assume-role",
-        )
+        if sts_user_access_key == "" or sts_user_secret_key == "":
+            sts_client = boto3.client("sts")
```
bcaton85, line 1249: i'm assuming boto3 is able to pick up on the location of the web identity file through environment variables?

bouchardmathieu-qc: That is the same documentation as the Go SDK for AWS v2

jonathankingfc (approved):
LGTM. I've verified the new changes, we should be good to pass this to QE.

bcaton85:
@bouchardmathieu-qc can you retarget this branch to master instead? All changes go into master first

bouchardmathieu-qc:
> @bouchardmathieu-qc can you retarget this branch to master instead? All changes go into master first

I rebased the branch on master and tested it in our OpenShift cluster.

bcaton85:
@bouchardmathieu-qc for some reason our CI is failing to start, can you push a smaller commit to force the tests to restart?

bouchardmathieu-qc:
> @bouchardmathieu-qc for some reason our CI is failing to start, can you push a smaller commit to force the tests to restart?

I updated my branch

bcaton85:
@bouchardmathieu-qc there still seems to be linting issues. If you run `make install-pre-commit-hook` it will run the checks on commit and make the changes for you. The full guide is [here](https://github.com/quay/quay/blob/master/docs/getting-started.md#requirements).

bouchardmathieu-qc:
> @bouchardmathieu-qc there still seems to be linting issues. If you run `make install-pre-commit-hook` it will run the checks on commit and make the changes for you. The full guide is [here](https://github.com/quay/quay/blob/master/docs/getting-started.md#requirements).

I fixed it. The problem was that I modified a file that already had a trailing space in it

bcaton85 (approved):

PR #3672, opened 2025-02-13
chore: updated target group names in stage dashboard (PROJQUAY-8506)

deshpandevlab:
updated target group names in stage dashboard


Marcusk19 (approved):
lgtm

PR #3681, opened 2025-02-17
repomirror(skopeo): changed skopeo calls to use temporary authfile and not expose creds in process list (PROJQUAY-7340)

michaelalang:
changing the repomirror skopeo calls from list of arguments containing the credentials
```
args = ["/usr/bin/skopeo", 
             ...,
             "--dest-creds",
             "user:password",
             "--src-creds",
             "user:password",
             ...
             ]
```
which ends up with the command being printing the credentials in the process list.
The PR changes that to create and use a `auth.json` file which is used in context (meaning being removed once the context closes.

The PR also add's unittests for the copy and tag skopeo class.

### DIFF
--- a/util/repomirror/skopeomirror.py
+++ b/util/repomirror/skopeomirror.py
@@ -1,19 +1,48 @@
+import base64
 import json
 import logging
 import os
 import subprocess
+import urllib.parse
 from collections import namedtuple
 from pipes import quote
-from tempfile import SpooledTemporaryFile
+from tempfile import NamedTemporaryFile, SpooledTemporaryFile
 from typing import Optional
 
 logger = logging.getLogger(__name__)
+
+SKOPEO_TIMEOUT_SECONDS = 300
 
 # success: True or False whether call was successful
 # tags: list of tags or empty list
 # stdout: stdout from skopeo subprocess
 # stderr: stderr from skopeo subprocess
 SkopeoResults = namedtuple("SkopeoResults", "success tags stdout stderr")
+AuthContent = namedtuple("AuthContent", ["location", "username", "password"])
+
+
+def wrap_anonymous(user: str = "", passwd: str = "") -> str:
+    if user in ("", None):
+        return ""
+    if passwd in ("", None):
+        return ""
+    return f"{user}:{passwd}"
+
+
+def create_authfile_content(content: list = [namedtuple]) -> dict:
+    return dict(
+        auths=dict(
+            map(
+                lambda x: (
+                    f"{x.location}",
+                    {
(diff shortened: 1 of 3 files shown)


ibazulic:
@michaelalang I think this would be great to import to my PR #3723 . What do you think?

util/repomirror/skopeomirror.py
```diff
@@ -18,6 +20,20 @@
...
+                lambda x: (
+                    f"{x[0]}",
+                    {"auth": f"{base64.b64encode(x[1].encode('utf8')).decode('utf8')}"},
+                ),
+                filter(lambda y: y[1] != "", content),
+            )
+        )
+    )
+
+
```
ibazulic, line 36: Do we need a lambda function here? We will only have one set of credentials at any given time for a specific job. We could just build the JSON ourselves.

util/repomirror/skopeomirror.py
```diff
@@ -50,13 +72,25 @@ def copy(
...
+            [
+                (
+                    urllib.parse.urlparse(src_image).netloc,
+                    f"{wrap_anonymous(src_username, src_password)}",
+                ),
+                (
+                    urllib.parse.urlparse(dest_image).netloc,
+                    f"{wrap_anonymous(dest_username, dest_password)}",
+                ),
+            ]
```
ibazulic, line 85: For the destination registry we use robot account credentials anyway, do we need to mask that?

util/repomirror/skopeomirror.py
michaelalang: @ibazulic agreed that the current use-case it single creds for one destination but I cannot ack that we might not need further creds in the future (like a composed manifest pulling in from multiple registries). Since the authfile can have multiple auths inside I would vote for keeping the syntax. 
I picked the lambda as it's a conversation that isn't needed anywhere else. If you want we could make it a method of `create_authfile_content` instead and map to that instead ?

util/repomirror/skopeomirror.py
michaelalang: @ibazulic it's necessary to wrap anonymous as mirroring might be without creds (like docker unauthenticated or any registry where no credentials are provided) and since json cannot deal with `None` the value needs to be mapped to `""` instead which is the purpose of the `wrap_anonymous` (the name might be mis leading)...

michaelalang: @ibazulic in regards to incorporating sure. Do you want this PR to be merging yours or the other way around ? (might be the easier one as this PR verify specific on that particular function only including the unittest of course)

ibazulic:
@michaelalang I think this needs rebasing as I merged my changes for skopeo timeout. Once rebased, I'm more than happy to re-review and approve. Can you please do that? Thanks!

PR #3683, opened 2025-02-18
api: Disallow push of manifests with negative layer size (PROJQUAY-8560)

ibazulic:
Under certain conditions, clients may create a manifest (OCI or Docker v2) that contains negative layer sizes. Our current validation schema does not take that corner case into account, it only checks if the manifest is properly formatted or not. With this change, Quay will reject manifests that have negative layer sizes and raise a `400` with a proper exception. An example can be seen here:

~~~
gunicorn-registry stdout | 2025-02-10 22:34:54,930 [377] [ERROR] [endpoints.v2.manifest] failed to parse manifest when writing by tagname gunicorn-registry stdout | Traceback (most recent call last):
gunicorn-registry stdout |   File "/quay-registry/endpoints/v2/manifest.py", line 362, in _parse_manifest
gunicorn-registry stdout |     return parse_manifest_from_bytes(
gunicorn-registry stdout |   File "/quay-registry/image/shared/schemas.py", line 40, in parse_manifest_from_bytes
gunicorn-registry stdout |     return DockerSchema2Manifest(manifest_bytes)
gunicorn-registry stdout |   File "/quay-registry/image/docker/schema2/manifest.py", line 172, in __init__
gunicorn-registry stdout |     raise MalformedSchema2Manifest("layer size is negative")
gunicorn-registry stdout | image.docker.schema2.manifest.MalformedSchema2Manifest: negative layer size
gunicorn-registry stdout | 2025-02-10 22:34:54,931 [377] [DEBUG] [endpoints.v2] sending response: b'{"errors":[{"code":"MANIFEST_INVALID","detail":{"message":"failed to parse manifest: negative layer size"},"message":"manifest invalid"}]}\n
~~~

### DIFF
--- a/endpoints/v2/test/test_manifest.py
+++ b/endpoints/v2/test/test_manifest.py
@@ -1,3 +1,4 @@
+import hashlib
 import json
 import time
 from unittest.mock import patch
@@ -83,7 +84,7 @@ def test_e2e_query_count_manifest_norewrite(client, app):
         assert counter.count <= 27
 
 
-MANIFEST_BYTES = json.dumps(
+INVALID_DOCKER_V2_MANIFEST = json.dumps(
     {
         "schemaVersion": 2,
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
@@ -123,7 +124,7 @@ def test_push_malformed_manifest_docker_v2s2(client, app):
 
     params = {
         "repository": "devtable/simple",
-        "manifest_ref": "sha256:d2fc93351b0929db9127713600adc3867fea4b665e5e9724f8cc9279c4a1ccde",
+        "manifest_ref": "sha256:" + hashlib.sha256(INVALID_DOCKER_V2_MANIFEST).hexdigest(),
     }
 
     user = model.user.get_user("devtable")
@@ -153,11 +154,11 @@ def test_push_malformed_manifest_docker_v2s2(client, app):
         params,
         expected_code=400,
         headers=headers,
-        raw_body=MANIFEST_BYTES,
+        raw_body=INVALID_DOCKER_V2_MANIFEST,
     )
 
 
-SAMPLE_MANIFEST = json.dumps(
+INVALID_OCI_MANIFEST = json.dumps(
     {
         "schemaVersion": 2,
         "config": {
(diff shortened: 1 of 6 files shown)


mtrmac (commented):
(A drive-by comment by someone unfamiliar with the codebase:) I don’t see any obvious reason to reject zero-size layers; now that OCI artifacts tend to be (mis) used to store collections of arbitrary files (with ~no schema or semantics to speak of), I think users will expect that to work.

ibazulic:
@mtrmac 

> (A drive-by comment by someone unfamiliar with the codebase:) I don’t see any obvious reason to reject zero-size layers; now that OCI artifacts tend to be (mis) used to store collections of arbitrary files (with ~no schema or semantics to speak of), I think users will expect that to work.

The 0 layer size would mean an empty file. Does that actually carry any kind of information? Of any sort? It's not hard to change the condition to allow 0 layer size manifests, I'm just wondering what would be the point. @robbmanes and I have had several discussions on this topic and our conclusion was that the spec didn't necessarily prohibit 0 layer sizes, but that eventually the concensus was that such a layer is not valid. That's why I opted for this solution.

bcaton85 (approved):

mtrmac:
I think it’s a clearly not valid _image layer_, but I don’t see why it wouldn’t be a valid _artifact blob_. I dunno, it might be an “error log” where an empty file = no errors, but a missing file = assertion error (that’s just something I invented at the moment, my only point is that it’s not _too hard_ to imagine).

mtrmac:
Well… it makes _no sense_. (And, in my experience as a maintainer of a client, Quay has actually historically been the registry with the strictest validation, in other respects.)

Sure, there, there are some image writers and image readers that don’t care at all.

But some readers do use the value, and fail after receiving a manifest that declares a negative layer size. And those readers are rather far removed from the original uploader of the image, making it potentially difficult to diagnose the reason why the uploaded manifest contains a negative layer size. Failing immediately can surface the condition much closer to the cause, and potentially make it much easier to diagnose/fix.

In the motivating customer case, the root cause was
> Dynatrace on the Quay registry cluster removing "Content-Length" headers from HEAD replies
(affecting image pushes) and the effect was
> a cascade of failures, the largest of which being that the OpenShift internal registry isn't accepting layers with negative sizes so they can't use these images in their environments.

mtrmac:
FWIW, https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties documents `size` as required.

PR #3691, opened 2025-02-24
chore: allow OIDC tokens for fresh login required API endpoints (PROJQUAY-7804)

matheusfm:
## Description
These changes allow API endpoints that require a fresh login (`@require_fresh_login`) to be accessed with an OIDC token.

## Linked Issues
This appears to be related to the following issue: https://issues.redhat.com/browse/PROJQUAY-7804

## Steps to reproduce

1. Run Quay with OIDC authentication type:
    ```
    AUTHENTICATION_TYPE: OIDC
    AUTH0_LOGIN_CONFIG:
      CLIENT_ID: *****
      CLIENT_SECRET: *****
      OIDC_SERVER: https://*****.auth0.com/
      SERVICE_NAME: Auth0
    ```
2. Generate a token directly from your auth server.
3. Call an API endpoint that **does not** require a fresh login
    ```
    curl -X GET 'http://localhost:8080/api/v1/user/' -H 'Authorization: Bearer <YOUR_ACCESS_TOKEN>'
    ```
    This returns a successful response.
4. Call an API endpoint that **requires** a fresh login (annotated with `@require_fresh_login`)
    ```
    curl -XPOST 'http://localhost:8080/api/v1/user/apptoken' \
    -H 'Authorization: Bearer <YOUR_ACCESS_TOKEN>' \
    -H 'Content-Type: application/json' \
    -d '{"title": "my token"}'
    ```
    This returns 401.

## Expected behavior
A success response is expected in this scenario, similar to when using an [OAuth token](https://docs.projectquay.io/api_quay.html#creating-oauth-access-token).

### DIFF
--- a/endpoints/api/__init__.py
+++ b/endpoints/api/__init__.py
@@ -19,6 +19,7 @@ from auth.auth_context import (
     get_authenticated_context,
     get_authenticated_user,
     get_validated_oauth_token,
+    get_sso_token,
 )
 from auth.decorators import process_oauth
 from auth.permissions import (
@@ -539,7 +540,7 @@ def require_fresh_login(func):
         if not user or user.robot:
             raise Unauthorized()
 
-        if get_validated_oauth_token():
+        if get_validated_oauth_token() or get_sso_token():
             return func(*args, **kwargs)
 
         last_login = session.get("login_time", datetime.datetime.min)


ibazulic:
Hey!

I'm not sure how your change works exactly given our model. Oauth tokens are generated through Oauth applications created inside an org account. These tokens have scopes which are attached to them via the database. Scopes define the level of access and each API endpoint has a specific decorator added to it that checks scopes. For instance, for app specific tokens:

https://github.com/quay/quay/blob/76008625f4309b4c885800a11fb417173429c00a/endpoints/api/appspecifictokens.py#L81

How would you provide a set of scopes for a specific token? How would you limit access for a specific user when you're not providing a legitimate bearer token that Quay created, one that contains the set of scopes?

syed:
Looks good. Thank you for the patch

PR #3692, opened 2025-02-25
ui: implement change to render modelcard stored in layers (PROJQUAY-8642)

kleesc:
When a manifest has a certain annotation and modelcard stored as its last layer, renders it in a new tags detail tab.
Screenshot 2025-02-25 at 8 41 04 AM


web/src/routes/TagDetails/TagDetailsTabs.tsx
```diff
@@ -59,6 +61,12 @@ export default function TagTabs(props: TagTabsProps) {
       >
         <Packages />
       </Tab>
+      <Tab
+        eventKey={TabIndex.ModelCard}
+        title={<TabTitleText>ModelCard</TabTitleText>}
+      >
+        <ModelCard />
+      </Tab>
```
bcaton85, line 69: Should we show this tab if there's no model card data?

web/src/routes/TagDetails/ModelCard/ModelCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const repo = parseRepoNameFromUrl(location.pathname);
+  const tag = parseTagNameFromUrl(location.pathname);
+
+  const [modelcardMarkdown, setModelcardMarkdown] = useState<string>('');
+
+  useEffect(() => {
+    (async () => {
+
+      try {
+	const resp: TagsResponse = await getTags(org, repo, 1, 100, tag);
```
bcaton85, line 36: We use react query to fetch data. This should be in a hook similar to [this](https://github.com/quay/quay/blob/master/web/src/hooks/UseTags.ts#L11-L33)

image/oci/manifest.py
```diff
@@ -354,11 +354,11 @@ def _manifest_image_layers(self, content_retriever):
...
-        if len(history) < len(self.filesystem_layers):
-            raise MalformedOCIManifest(
-                "Found less history (%s) than layer blobs (%s)"
-                % (len(history), len(self.filesystem_layers))
-            )
+        # if len(history) < len(self.filesystem_layers):
+        #     raise MalformedOCIManifest(
+        #         "Found less history (%s) than layer blobs (%s)"
+        #         % (len(history), len(self.filesystem_layers))
+        #     )
```
bcaton85, line 361: Are removing these checks required for the feature to work? Can we add these back in for now?

web/src/routes/TagDetails/TagDetailsTabs.tsx
```diff
@@ -59,6 +61,12 @@ export default function TagTabs(props: TagTabsProps) {
       >
         <Packages />
       </Tab>
+      <Tab
+        eventKey={TabIndex.ModelCard}
+        title={<TabTitleText>ModelCard</TabTitleText>}
```
bcaton85, line 66: Can we also feature flag this? This will require additional work to get running on quay.io

bcaton85 (changes requested):
Looks like a good foundation. Just need some clarification on some questions. Can we also add tests?

endpoints/api/manifest.py
```diff
@@ -120,7 +130,23 @@ def get(self, namespace_name, repository_name, manifestref):
         if manifest is None or manifest.internal_manifest_bytes.as_unicode() == "":
             raise NotFound()
 
-        return _manifest_dict(manifest)
+        manifest_dict = _manifest_dict(manifest)
+
+        if parsed_args["include_modelcard"]:
+            parsed = manifest.get_parsed_manifest()
+            layer_digest = str(parsed.filesystem_layers[-1].digest)
```
bcaton85, line 137: How do we differentiate manifests that contain model cards to those that don't? We certainly don't want to try to read in the blob of images that don't contain a model card

endpoints/api/manifest.py
```diff
@@ -107,7 +110,14 @@ class RepositoryManifest(RepositoryParamResource):
     @require_repo_read(allow_for_superuser=True)
     @nickname("getRepoManifest")
     @disallow_for_app_repositories
-    def get(self, namespace_name, repository_name, manifestref):
+    @parse_args()
```
syed, line 113: could we have getting the model card info a seperate API call which the frontend can determine based on the artifactType?

endpoints/api/manifest.py
```diff
@@ -120,7 +130,23 @@ def get(self, namespace_name, repository_name, manifestref):
...
             raise NotFound()
 
-        return _manifest_dict(manifest)
+        manifest_dict = _manifest_dict(manifest)
+
+        if parsed_args["include_modelcard"]:
+            parsed = manifest.get_parsed_manifest()
+            layer_digest = str(parsed.filesystem_layers[-1].digest)
+            retriever = RepositoryContentRetriever(repo_ref._db_id, storage)
+            content = retriever.get_blob_bytes_with_digest(layer_digest)
```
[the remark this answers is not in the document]
syed: the RHEL AI folks are using `  "artifactType": "application/x-mlmodel",` in their manifests. We should filter on that

endpoints/api/manifest.py
syed, line 137: are we positive that the modelcard is always going to be in the last layer? I was looking at some of the AI images and it seems like it is the first layer. They also have annotated layer with the filename for each layer. 
```
      "annotations": {
        "org.opencontainers.image.title": "README.md"
      }
```

check the full manifest 

```
oras manifest fetch registry.redhat.io/rhelai1/granite-8b-code-instruct@sha256:3b04dfe5fcbd4bcdcd1d4185fb2645974104fb173b437efe4cea1f37518aefe5 | jq
```

syed: we could use annotations instead of a fixed position in the layers IMO

image/oci/manifest.py
kleesc: I added these back. I think the correct way if for the images to have an empty descriptor instead

endpoints/api/manifest.py
```diff
@@ -120,7 +131,43 @@ def get(self, namespace_name, repository_name, manifestref):
         if manifest is None or manifest.internal_manifest_bytes.as_unicode() == "":
             raise NotFound()
 
-        return _manifest_dict(manifest)
+        manifest_dict = _manifest_dict(manifest)
+
+        if parsed_args["include_modelcard"]:
```
bcaton85, line 136: the feature flag will need to be checked here as well

web/src/routes/TagDetails/TagDetailsTabs.tsx
```diff
@@ -59,6 +64,13 @@ export default function TagTabs(props: TagTabsProps) {
...
+        isHidden={!quayConfig.features.UI_MODELCARD || !props.tag.modelcard}
```
bcaton85, line 70: this may not load right away, it will need to be `!quayConfig?.features?.UI_MODELCARD`

bcaton85 (approved):

PR #3721, opened 2025-03-18
oci: skip validation of labels for nested manifest indexes (PROJQUAY-8272)

syed:
When parsing a nested manifest index, the nested index will not have labels. Skip that check if the nested manifest is an index

### DIFF
--- a/data/model/oci/manifest.py
+++ b/data/model/oci/manifest.py
@@ -327,7 +327,7 @@ def _create_manifest(
 
             # Retrieve its labels.
             labels = child_manifest.get_manifest_labels(retriever)
-            if labels is None and isinstance(child_manifest, ManifestInterface):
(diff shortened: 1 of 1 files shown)


stivassc:
@syed Is this something that is planned to be merged still? We're making use of nested indexes but it is broken without this PR, so I would like to see this being merged if possible, thanks.

hellt:
I am interested in this fix as well. Would be fair to assume oci images with nested images would be supported by quay

hellt:
@jbpratt can someone from RH take a look at that?

jbpratt:
hey @hellt , I can get this in our queue - thanks for the ping!

jbpratt:
Closed in favor of https://github.com/quay/quay/pull/6892

PR #3723, opened 2025-03-18
mirror: Add job timeout to mirror configurations (PROJQUAY-7249)

ibazulic:
Previous global job timeout of 5 minutes was inadequate for big images. The timeout should now be configurable in much the same way as sync is. Minimum job length is 300 seconds/5 minutes.

The PR is still work in progress.


util/repomirror/skopeomirror.py
```diff
@@ -113,10 +113,13 @@
 
         return env
 
-    def run_skopeo(self, args, proxy):
+    def run_skopeo(self, args, proxy, timeout):
         # Using a tempfile makes sure that if --debug is set, the stdout and stderr output
         # doesn't get truncated by the system's pipe limit.
+
         with SpooledTemporaryFile() as stdoutpipe, SpooledTemporaryFile() as stderrpipe:
+            logger.debug("Creating mirroring job, command to execute: %s", args)
```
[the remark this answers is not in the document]
ibazulic: Reference removed.

endpoints/api/mirror.py
```diff
@@ -291,6 +299,12 @@ def post(self, namespace_name, repository_name):
 
         data = request.get_json()
 
+        if data["skopeo_timeout_interval"] < 300:
```
bcaton85, line 302: `skopeo_timeout_interval` is not marked as required, if the parameter is sent with the request will this fail? should we check the existence of `skopeo_timeout_interval` first?

storage/cloud.py
```diff
@@ -1079,8 +1081,18 @@ def __init__(
...
+
+        # For very big layers over 100 GiB, chunk size of 5 MiB is not enough.
+        # S3 supports up to 10000 parts in a multipart upload, with 5 MiB chunk
+        # we can only upload up to 50 GiB files. If a layer is bigger than that,
+        # storage will return a 400 and upload will fail.
+        # This makes minimum chunk size configurable for both RADOS and OCS storage.
+        self.minimum_chunk_size = (
+            (minimum_chunk_size_mb if minimum_chunk_size_mb is not None else 5) * 1024 * 1024
+        )
+
```
bcaton85, line 1095: Is this change related to the skopeo timeouts? should this be in a separate PR?

bcaton85 (commented):
looks good, i'd like to test it out tomorrow morning before giving an approval

storage/cloud.py
ibazulic: @bcaton85 I needed it to test my changes for skopeo timeout. It was a part of a new PR (https://github.com/quay/quay/pull/3958) and I asked the contributor whether he would want us to merge it, but because the change in driver behaviour would not be added to a z-stream, but rather to a new y-stream, we decided to leave it here.

endpoints/api/mirror.py
ibazulic: @bcaton85 I don't think so because of the `validate_json_request` wrapper that is being called. If the parameter is missing, it should cancel the request. I'll double check to be sure.

https://github.com/quay/quay/blob/140ee4ab800f379d5cfe9f06c3266600ed092386/endpoints/api/__init__.py#L589

storage/cloud.py
bcaton85: ah gotcha, ideally this would be a separate PR so we don't conflate changes but this should be fine for now

endpoints/api/mirror.py
ibazulic: @bcaton85 yeah, I forgot to add the property to the list of required properties, you were correct. Now it should work as intended.

PR #3746, opened 2025-03-27
healthcheck: Make gunicorn health check timeout configurable (PROJQUAY-8757)

Sunandadadi:


health/services.py
```diff
@@ -40,13 +40,20 @@ def fn(app):
         client = app.config["HTTPCLIENT"]
         registry_url = _compute_internal_endpoint(app, endpoint)
         try:
-            status_code = client.get(registry_url, verify=False, timeout=2).status_code
+            status_code = client.get(
+                registry_url, verify=False, timeout=app.config["GUNICORN_HEALTH_CHECK_TIMEOUT"]
```
bcaton85, line 44: you'll want to do `app.config.get("GUNICORN_HEALTH_CHECK_TIMEOUT", 2)` instead

Sunandadadi: good point, updated, thanks!

health/services.py
```diff
@@ -40,13 +40,22 @@ def fn(app):
...
             return (okay, message)
         except Exception as ex:
-            logger.exception("Exception when checking worker health: %s", registry_url)
-            return (False, "Exception when checking worker health: %s" % registry_url)
+            logger.exception(
+                f"Exception when checking worker health on endpoint {registry_url} as {str(ex)}"
+            )
+            return (
+                False,
+                f"Exception when checking worker health on endpoint {registry_url} as {str(ex)}",
```
bcaton85, line 57: any possibility this string conversion fails?

Sunandadadi: I'll revert this for now.

bcaton85: when fixing prod issues we want to make the least amount of changes as possible

Sunandadadi: Agreed, one change at a time

syed: this should probably be fine, anything subclassed from `Exception` will have a default `__str__` or `__repr__`

bcaton85 (approved):

PR #3747, opened 2025-03-28
operations: added ELB calculations to ALB based SLO timeseries (PROJQUAY-8508)

deshpandevlab:
- Current ALB based time series only describes ALB 2xx. This change add ELB errors to the metric of availability to display true state of the service.


Marcusk19 (approved):
lgtm

PR #3749, opened 2025-03-28
operations: corrected metric expression to span over all targets (PROJQUAY-8508)

deshpandevlab:
tested in stage and corrected metric to span over entire load balancer


Marcusk19 (approved):
lgtm

PR #3757, opened 2025-04-03
fix: do not fail on splunk errors when flag set (PROJQUAY-8595)

jonathankingfc:


data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -62,6 +65,15 @@ def __init__(
             raise Exception(
                 "Connection to Splunk refused, check if Splunk instance is available: %s" % e
             )
+        except Exception as e:
+            strict_logging_disabled = config.app_config.get("ALLOW_WITHOUT_STRICT_LOGGING") or (
+                config.app_config.get("ALLOW_PULLS_WITHOUT_STRICT_LOGGING")
```
ibazulic, line 70: Do we want to set a default value here?

bcaton85: it's default value is set in config.py but it's fine to set it again here

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -62,6 +65,15 @@ def __init__(
...
+                and kind_name in ACTIONS_ALLOWED_WITHOUT_AUDIT_LOGGING
```
bcaton85, line 71: where is `kind_name` defined?

bcaton85 (changes requested):
undefined var, check comment

data/logs_model/logs_producer/splunk_logs_producer.py
jonathankingfc: We shouldn't redefine defaults if we already have them defined in config.py

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -62,6 +65,15 @@ def __init__(
...
+            )
+            if strict_logging_disabled:
+                logger.exception("log_action failed", extra=({"exception": lse}).update(log_data))
```
bcaton85: let's print out the exception here, we have some unreferenced variables

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -62,6 +65,12 @@ def __init__(
             raise Exception(
```
bcaton85, line 65: if splunk is unable to connect here we don't want to re-raise this exception

data/logs_model/logs_producer/splunk_logs_producer.py
```diff
@@ -62,6 +65,12 @@ def __init__(
...
                 "Connection to Splunk refused, check if Splunk instance is available: %s" % e
             )
+        except Exception as e:
+            strict_logging_disabled = config.app_config.get("ALLOW_WITHOUT_STRICT_LOGGING")
+            if strict_logging_disabled:
+                logger.exception("log_action failed: %s", e)
```
bcaton85: technically this isn't a `log_action`, we should output something like `unable to create index`

bcaton85: or `unable to configure splunk instance`

bcaton85 (approved):

bcaton85 (approved):

PR #3778, opened 2025-04-14
notifications: Support slash in repository names (PROJQUAY-7538)

asgeirn:
Fix for PROJQUAY-7538 discussed in #3069 by only considering the first slash when separating namespace and repository.

I did a search for other examples where `.split("/")` was used and expecting two values, and also found one in gitlabtrigger.

### DIFF
--- a/endpoints/gitlabtrigger.py
+++ b/endpoints/gitlabtrigger.py
@@ -25,7 +25,7 @@ def attach_gitlab_build_trigger():
         abort(400)
     state = state[len("repo:") :]
     try:
-        [namespace, repository] = state.split("/")
+        [namespace, repository] = state.split("/", maxsplit=1)
     except ValueError:
         abort(400)
 
--- a/util/jinjautil.py
+++ b/util/jinjautil.py
@@ -48,7 +48,7 @@ def user_reference(username):
 
 def repository_tag_reference(repository_path_and_tag):
     (repository_path, tag) = repository_path_and_tag
-    (namespace, repository) = repository_path.split("/")
+    (namespace, repository) = repository_path.split("/", maxsplit=1)
     owner = model.user.get_namespace_user(namespace)
     if not owner:
         return tag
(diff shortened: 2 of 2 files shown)


asgeirn:
@Sunandadadi - could you have a look?

Sunandadadi:
@asgeirn thank you for submitting the PR. Can you add tests for this?

asgeirn:
> @asgeirn thank you for submitting the PR. Can you add tests for this?

@Sunandadadi I'm looking at adding tests, but there aren't any existing tests for these parts, do you have any guidance on how to achieve this?

For the notifications (where I discovered the bug) I think this would test it: https://github.com/quay/quay/blob/0c1405f7293e5dc86d99ac3cdc4c1ad90ccb5992/endpoints/api/test/test_repositorynotification.py#L89-L100

However, for the Gitlab trigger, I only find a Cypress test, but I'm not sure how to add a fixture with a repository with slashes in the name? https://github.com/quay/quay/blob/0c1405f7293e5dc86d99ac3cdc4c1ad90ccb5992/web/cypress/e2e/builds.cy.ts#L1543-L1549

.devcontainer/devcontainer.json
Sunandadadi, line 1: Did you add this by mistake? I don't see this file being used anywhere.

Sunandadadi:
@asgeirn sorry for the late reply. Thank you for adding the tests. We should be good with the added python tests.

.devcontainer/devcontainer.json
Sunandadadi: Are there any other changes you want to push here?

Sunandadadi: There seems to be some tests failing.

asgeirn: I've removed devcontainer.json now.

Which tests would that be, are they reported here?

asgeirn:
I attempted to fix the failing unit test by adding the repository in initdb.py, but this caused other tests to start failing, so I'm not sure which way to proceed.

asgeirn:
@Sunandadadi Rebased with latest master commit and removed changes to tests.

bcaton85 (approved):

PR #3797, opened 2025-04-17
gc: garbage collect manifests not targetted by any tags when deleting repository (PROJQUAY-8136)

kleesc:


bcaton85 (changes requested):
Can we have tests added for this?

data/model/gc.py
```diff
@@ -201,6 +201,30 @@ def _purge_repository_contents(repo):
         if not found:
             break
 
+    # Purge manifests that not referenced by tags
+    while True:
+        found = False
+        for manifests in _chunk_iterate_for_deletion(
+            Manifest.select().where(Manifest.repository == repo)
```
Sunandadadi, line 208: can we send this query to the read replica?

kleesc: For GC I would say that we want to make sure that we have the up-to-date data

PR #3799, opened 2025-04-18
reconciler: Remove database calls for storing/changing web customer ids (PROJQUAY-0000)

Marcusk19:
Remove adding/updating web customer ids in database during the reconciler run. Using the database to reference web customer ids was removed in https://github.com/quay/quay/pull/3302


Sunandadadi (approved):
lgtm

PR #3814, opened 2025-04-24
proxy: moving manifest check to after upstream manifest fetch (PROJQUAY-8536)

bcaton85:
The check for the existing manifest was occurring before the request to the upstream registry to fetch the manifest. This caused race conditions where two requests for the same manifest would try to create the manifest at the same time, leading to registry 500's.

### DIFF
--- a/data/registry_model/registry_proxy_model.py
+++ b/data/registry_model/registry_proxy_model.py
@@ -254,7 +254,10 @@ class ProxyModel(OCIModel):
         if wrapped_manifest is None:
             try:
                 wrapped_manifest, _ = self._create_and_tag_manifest(
-                    repository_ref, manifest_digest, self._create_manifest_with_temp_tag
+                    repository_ref,
+                    manifest_digest,
+                    self._create_manifest_with_temp_tag,
+                    verify_manifest_exists=True,
                 )
             except (UpstreamRegistryError, ManifestDoesNotExist) as e:
                 raise ManifestDoesNotExist(str(e))
@@ -319,7 +322,6 @@ class ProxyModel(OCIModel):
             repository_ref,
             manifest_digest,
             allow_dead=True,
-            allow_hidden=True,
             require_available=False,
             raise_on_error=True,
         )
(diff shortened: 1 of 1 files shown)


data/registry_model/registry_proxy_model.py
```diff
@@ -254,7 +254,10 @@ def lookup_manifest_by_digest(
         if wrapped_manifest is None:
             try:
                 wrapped_manifest, _ = self._create_and_tag_manifest(
-                    repository_ref, manifest_digest, self._create_manifest_with_temp_tag
+                    repository_ref,
```
deshpandevlab, line 257: I think this approach while being proactive for checking manifest before attempting to create might suffer from [Time-of-check to time-of-use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) race condition. This is a good first step

data/registry_model/registry_proxy_model.py
```diff
@@ -399,6 +402,12 @@ def _create_and_tag_manifest(
         """
         self._proxy.manifest_exists(manifest_ref, ACCEPTED_MEDIA_TYPES)
         upstream_manifest = self._pull_upstream_manifest(repo_ref.name, manifest_ref)
+        if verify_manifest_exists:
```
deshpandevlab, line 405: should the check happen outside of create function? the functionality and naming of the function seems off

bcaton85: the check has to happen after `upstream_manifest = self._pull_upstream_manifest(repo_ref.name, manifest_ref)`. There's a better way to refactor this so the parameter isn't so single use case specific. I just went the least intrusive option

data/registry_model/registry_proxy_model.py
deshpandevlab: discussed this offline. Posting here for future reference:

My concern was that there could be a race condition where two concurrent requests verify_manifest_exists and don't find the manifest, both try to create it and we again run into the same issue. 

Brandon explained the root cause of this race condition is the [_pull_upstream_manifest](https://github.com/quay/quay/blob/master/data/registry_model/registry_proxy_model.py#L401) method. This method is the long running one as it is making a call to upstream repo. It sits between the original [lookup_manifest_by_digest](https://github.com/quay/quay/blob/master/data/registry_model/registry_proxy_model.py#L251) and actual _create_manifest method. Giving the request enough time to create the race condition.

The verify_manifest_exists however sits right before create.

PR #3840, opened 2025-04-30
storage: encrypting aws signature for akamai (PROJQUAY-8511)

Sunandadadi:


data/test/test_encryption.py
```diff
@@ -78,3 +86,46 @@ def test_encryption_value(secret_key, encrypted_value, expected_decrypted_value)
     decrypted = encrypter.decrypt_value(encrypted_value)
 
     assert decrypted == expected_decrypted_value
+
```
syed, line 89: nice 👍

storage/akamaistorage.py
```diff
@@ -66,9 +78,24 @@ def get_direct_download_url(
...
+                amz_signature = amz_signature_params[0] if len(amz_signature_params) == 1 else None
+
+                if amz_signature is not None:
+                    # Generate a random 16-byte Initialization Vector
+                    iv = os.urandom(16)
+                    encoded_string = aes256_encrypt(
+                        amz_signature, self.akamai_sig_encryption_key, iv
+                    )
+
+                    query_params["X-Amz-Signature"] = [base64.b64encode(encoded_string).decode()]
```
syed, line 95: do you need to wrap this in a list if you have only one value?

Sunandadadi: [urlencode](https://github.com/quay/quay/pull/3840/files#diff-2d91dc6d8204ceb32233b24732caad8aa6354817d14c55dbd56adc5e568dde6aR97) function here converts `query_params` into url encoded query string. list will let us add more values in future if we want. Either way in the current case, list with a single item and directly assigning the value both result in the same query string.

PR #3865, opened 2025-05-06
storage: Increase proxy JWT expiry to 10 minutes (PROJQUAY-8894)

ibazulic:
The current JWT expiry time of 30 seconds is sometimes inadequate, especially for scanning purposes when download is not as quick as one expects. This will increase the expiry time to 10 minutes allowing downloading of layers even in slower environments.

### DIFF
--- a/storage/downloadproxy.py
+++ b/storage/downloadproxy.py
@@ -14,7 +14,7 @@ from util.security.registry_jwt import (
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


ibazulic:
New token after making these changes:

~~~
{
  "iss": "quay",
  "aud": "QUAY_HOSTNAME",
  "nbf": 1746549378,
  "iat": 1746549378,
  "exp": 1746549978,
  "sub": "storageproxy",
...
~~~

bcaton85 (approved):

PR #3885, opened 2025-05-13
migration: alter id column type for table manifestblob

Sunandadadi:


data/migrations/versions/e8ed3fb547da_alter_manifestblob_id.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+        op.execute(
+            """
+            ALTER SEQUENCE manifestblob_id_seq MAXVALUE 9223372036854775807;
+        """
+        )
+
+
+def downgrade(op, tables, tester):
+    pass
```
syed, line 37: can you add the downgrade steps. It's basically the reverse BIGINT -> INT

kuklyy: You can’t safely migrate from BIGINT to INT without checking the data first. Any existing values outside the INT range will cause truncation or errors.

PR #3950, opened 2025-05-22
build: updating hack script for gathering python dependencies on konflux (PROJQUAY-8804)

Marcusk19:
- Rewrites the `hack/update-requirements-build.sh` script for prefetching pip dependencies according to the[ konflux docs](https://konflux-ci.dev/docs/building/prefetching-dependencies/#pip). Enterprise contract requires that all dependencies are built from source so we need to generate a definitive list of all build dependencies. 
- Removes `data.queue` from list of packages in `setup.py` due to causing issues with `pip-compile` during script execution
- Adds a `constraints.txt` file used to workaround a [bootstrapping issue](https://github.com/python/importlib_metadata/issues/516) with importlib-metadata during script execution
- Updates PyYAML to fix versioning conflict with Cython during script execution


setup.py
```diff
@@ -79,7 +79,6 @@
         "data.logs_model.logs_producer",
         "data.model",
         "data.model.oci",
-        "data.queue",
```
bcaton85, line 82: this might break the service tool, any reason we're removing this?

bcaton85: if it's too much effort we can just merge this and deal with the service tool later

bcaton85: oh odd you're right, i wonder why the sub packages are listed as well. nvm we should be fine then

bcaton85 (approved):

PR #3958, opened 2025-05-27
storage: Adds minimum_chunk_size_mb to RadosGWStorage storage provider

mlowery:
Same as #2949 but for `RadosGWStorage` driver.

Adds optional parameter `minimum_chunk_size_mb`. If not specified, previously hard-coded value (5 MiB) is used.

### DIFF
--- a/storage/cloud.py
+++ b/storage/cloud.py
@@ -1014,7 +1014,9 @@ class RadosGWStorage(_CloudStorage):
         region_name=None,
         signature_version=None,
         port=None,
+        # Chunk size options for multipart upload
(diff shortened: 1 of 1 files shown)


ibazulic:
Hey,

Thanks for the commit. This change is already part of some larger changes related to mirror workers: #3723.

The PR will most likely be included in the next Quay release. We can definitely test and merge this before that PR but we wouldn't be able to cherry pick it for current Quay versions as driver changes, even small ones, change the workflow so it would most likely be part of the same release as the mentioned PR.

PR #4007, opened 2025-06-05
proxycache: Download blob not cached when pulling manifests with blobs available locally (PROJQUAY-6708)

kleesc:


data/registry_model/registry_proxy_model.py
```diff
@@ -707,6 +709,18 @@ def _create_placeholder_blobs(
...
+                json.dumps(
+                    {
+                        "digest": str(layer.digest),
+                        "repo_id": repo_id,
+                        "username": self._user.username,
+                        "namespace": self._user.username,
+                    }
+                ),
+            )
+            # self._download_blob(RepositoryReference.for_id(repo_id), layer.digest)
```
bcaton85: won't this have the worker pull blobs for all manifest that are pulled through the proxy org?

bcaton85: if so that will cause a very large amount of redundant pulls, we should only pull the missing blobs

workers/proxycacheblobworker.py
```diff
@@ -0,0 +1,121 @@
...
+            user_ref,
+        )
+
+        logger.debug(
+            "Starting proxy cache blob download for digest %s for repo id %s",
+            digest,
+            repo_id,
+        )
+
+        blob = registry_proxy_model._get_shared_storage(digest)
```
bcaton85, line 66: i'm assuming this is just to check for the global empty layer? otherwise this check looks redundant to line 53

data/registry_model/registry_proxy_model.py
```diff
@@ -707,6 +709,19 @@ def _create_placeholder_blobs(
...
+                    {
+                        "digest": str(layer.digest),
+                        "repo_id": repo_id,
+                        "username": self._user.username,
+                        "namespace": self._user.username,
+                    }
+                ),
+                available_after=5,
+            )
+            # self._download_blob(RepositoryReference.for_id(repo_id), layer.digest)
```
bcaton85: can we also delete this

workers/proxycacheblobworker.py
```diff
@@ -0,0 +1,126 @@
...
+                    .join(ManifestBlob)
+                    .where(
+                        ManifestBlob.repository_id == repo_id,
+                        ImageStorage.content_checksum == digest,
+                    )
+                    .get()
+                )
+            except ImageStorage.DoesNotExist:
+                # There should be placeholder blobs after manifest requests
+                return False
```
bcaton85, line 80: this will result in the blobs being pulled by the worker, do we want to do that if the placeholder blobs haven't been created?

bcaton85: i think we want to pull only if the placeholder blobs have been created but the imagestorage placement has not

bcaton85 (changes requested):

bcaton85 (approved):

bcaton85 (approved):

PR #4008, opened 2025-06-05
build: remove aws-ip-ranges from downstream deps (PROJQUAY-8804)

Marcusk19:
If aws-ip-ranges.json fails to download during build continue without it. 
Removes aws-ip-ranges and other unused dependencies from our `artifacts.lock.yaml`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -159,7 +159,7 @@ COPY --chown=0:0 . .
 RUN set -ex\
 	; chmod -R g=u ./conf\
     ; curl -fsSL https://ip-ranges.amazonaws.com/ip-ranges.json -o util/ipresolver/aws-ip-ranges.json 2>/dev/null \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -159,7 +159,7 @@ COPY --chown=0:0 . .
 RUN set -ex\
 	; chmod -R g=u ./conf\
     ; curl -fsSL https://ip-ranges.amazonaws.com/ip-ranges.json -o util/ipresolver/aws-ip-ranges.json 2>/dev/null \
-    || cp /cachi2/output/deps/generic/aws-ip-ranges.json util/ipresolver/aws-ip-ranges.json \
```
bcaton85, line 162: we're removing the copy here, i believe we still need that for quay.io builds

Marcusk19:
Closing, since we're going with the approach of using a [run-script task](https://github.com/konflux-ci/build-definitions/tree/main/task/run-script-oci-ta/0.1) in our pipeline to remove this line before build

PR #4076, opened 2025-06-16
proxycache: fix queueitem payload for proxycachblobworker (PROJQUAY-9018)

kleesc:


bcaton85 (changes requested):
good catch, can we get tests for this? also can we get a config option for enabling/disabling the worker? just in case it gives customers performance issues

workers/proxycacheblobworker.py
```diff
@@ -113,7 +113,7 @@ def create_gunicorn_worker():
         while True:
             time.sleep(100000)
 
-    if not features.PROXY_CACHE:
+    if not features.PROXY_CACHE or not features.PROXY_CACHE_BLOB_DOWNLOAD:
```
bcaton85, line 116: this needs to be added to the feature's module

bcaton85 (changes requested):
small bug

bcaton85 (approved):

PR #4087, opened 2025-06-25
Update request.py to get Real IP address on Quay UI/Database

wenxintan:


util/request.py
```diff
@@ -20,8 +20,15 @@ def get_request_ip():
     remote_addr = request.remote_addr or None
     if os.getenv("TEST", "false").lower() == "true":
         remote_addr = request.headers.get("X-Override-Remote-Addr-For-Testing", remote_addr)
+        
+    """
+    Returns the Real IP address of the client making the current Flask request or None if none.
+    """
+    x_forwarded_for = request.headers.get("X-Forwarded-For", None)
+    real_ip = x_forwarded_for.split(",")[-2] or request.remote_addr
```
bcaton85, line 28: can we get a null default here in case that index is empty?

bcaton85:
Hi @wenxintan thank you for your contribution! can we get a test added for this?

util/request.py
michaelalang: I see a more problematic issue here as `[-2]` assumes only two values are populated in x-forwarded-for.
According to the definition it's
```
If a request goes through multiple proxies, the IP addresses of each successive proxy are listed. 
This means that the rightmost IP address is the IP address of the most recent proxy and the leftmost IP address is the address of the originating client.
```
so we could avoid running into an index error by going like
```
x_forwarded_for = request.headers.get("X-Forwarded-For", request.remote_addr)
real_ip = x_forwarded_for.split(",")[0]
```
```
>>> request.remote = "127.0.0.1"
>>> request.headers = {"no-x-forwarded": True}
# no header present
>>> request.headers.get("X-Forwarded-For", request.remote)
'127.0.0.1'
>>> request.headers.get("X-Forwarded-For", request.remote).split(",")[0]
'127.0.0.1'
# one item in the list
>>> request.headers.get("X-Forwarded-For", request.remote).split(",")[0]
'203.0.113.195'
# two items in the list
>>> request.headers = {"X-Forwarded-For": "203.0.113.195, 2001:db8:85a3:8d3:1319:8a2e:370:7348"}
>>> request.headers.get("X-Forwarded-For", request.remote).split(",")[0]
'203.0.113.195'
# more items in the list
>>> request.headers = {"X-Forwarded-For": "203.0.113.195,2001:db8:85a3:8d3:1319:8a2e:370:7348,10.1.1.1,10.2.2.2,10.3.3.3"}
>>> request.headers.get("X-Forwarded-For", request.remote).split(",")[0]
'203.0.113.195'
```

and in case we would be interested in the last proxy the request passed it would be also always `[-1]` 

additional we should consider alternative as recommended by RFC7239
```
Forwarded: for=192.0.2.60;proto=http;by=203.0.113.43
Forwarded: for="[2001:db8::1234]"
```

michaelalang: what about following idea instead of just that use-case in a one liner ?
```
# utils/something...
def get_client_from_headers() -> str:
    origin = request.headers.get(
        "X-Forwarded-For",
        request.headers.get("Forwarded", f"{request.remote},{request.remote}"),
    )
    if ";" in origin:
        # do RFC7239 parsing
        # according to the RFC we split paramters by, for, host, proto on ";"
        # and multi occurances of each by ","
        def get_for(orig):
            # extract the for record
            for ips in list(
                map(
                    lambda y: y.split(","),
                    filter(lambda x: "for=" in x, orig.split(";")),
                )
            ):
                for ip in ips:
                    yield ip.split("=")[-1]
            yield request.remote

        def get_by(orig):
            # extract the by record
            for ips in list(
                map(
                    lambda y: y.split(","),
                    filter(lambda x: "by=" in x, orig.split(";")),
                )
            ):
                for ip in ips:
                    yield ip.split("=")[-1]
            yield request.remote

        ip = list(get_for(origin))
        proxy = list(get_by(origin))
        return (ip[0].strip(), proxy[0].strip())
    else:
        # parse X-forwarded-for
        origin = origin.split(",")
        ip, proxy = origin[:: len(origin) - 1]
        return (ip.strip(), proxy.strip())
```

and you can than select if you add origin ip and proxy or just ip 
and testing that theory looks like
```
Origin: 1.2.3.4,8.8.8.8
('1.2.3.4', '8.8.8.8')
Origin: 127.0.0.2,127.0.0.2
('127.0.0.2', '127.0.0.2')
Origin: for=1.2.3.4;proto=http;by=8.8.8.8,by=9.9.9.9
('1.2.3.4', '8.8.8.8')
Origin: for=1.2.3.4, for=2.3.4.5, for=3.4.5.6;proto=http;by=8.8.8.8,by=9.9.9.9
('1.2.3.4', '8.8.8.8')
```
```

PR #4092, opened 2025-06-27
quota: only create notification if one doesn't already exist (PROJQUAY-8857)

bcaton85:


data/model/notification.py
```diff
@@ -146,6 +146,37 @@ def delete_matching_notifications(target, kind_name, **kwargs):
...
+    # For each, match the metadata to the specified values.
+    for notification in notifications:
+        matches = True
+        try:
+            metadata = json.loads(notification.metadata_json)
+        except:
+            continue
+
+        for key, value in kwargs.items():
+            if not key in metadata or metadata[key] != value:
```
kleesc, line 170: I see in service_keys.py, one of the metadata field is an expiration date. Could this cause issue, where a notification already exists with different date value?

bcaton85: For this scenario the only metadata would be namespace name, so for this fix it won't make a difference

kleesc (approved):

PR #4132, opened 2025-07-18
api: add missing read permissions for readonly superuser

kleesc:


endpoints/api/billing.py
```diff
@@ -20,6 +20,7 @@
 from endpoints.api import (
     ApiResource,
     abort,
+    allow_if_global_readonly_superuser,
```
bcaton85: this has been added to a lot of endpoints that are only used on quay.io, for example getting the credit card or subscription. The permissions add can be removed on those endpoints

kleesc: Reverted the changes to the billing endpoints

bcaton85: i'm still seeing the changes in this PR

bcaton85 (approved):

PR #4152, opened 2025-07-28
oauth: Fallback to using depricated_sub since rhsso now has a new value for sub field  (PROJQUAY-9124)

harishsurf:
There was a recent change made to RHSSO oauth flow which generates a new value for the `sub` claim field which breaks quay on RH console dot UI. 
This change adds fallback to use `deprecated_sub` field which has the old value which quay relies on to map to it's usernames.

### DIFF
--- a/oauth/test/test_oidc.py
+++ b/oauth/test/test_oidc.py
@@ -12,6 +12,8 @@ from cryptography.hazmat.primitives import serialization
 from httmock import HTTMock, urlmatch
 from six.moves.urllib.parse import quote
 
+from oauth.base import OAuthUserIdException
+from oauth.login_utils import get_sub_username_email_from_token
 from oauth.oidc import OAuthLoginException, OIDCLoginService, PasswordGrantException
 from util.config import URLSchemeAndHostname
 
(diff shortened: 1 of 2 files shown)


oauth/login_utils.py
```diff
@@ -93,7 +96,17 @@ def get_sub_username_email_from_token(
                 config.get("PREFERRED_GROUP_CLAIM_NAME", None)
             )
 
-    return decoded_id_token["sub"], lusername, email_address, additional_info
+    # Call the login service's get_user_id() if it exists (e.g., RHSSO), otherwise use sub directly
+    if (
```
syed: instead of this logic, can you make `get_user_id` a required function in the base metaclass and have the subclass use the generic function of returning the `sub` in the token

oauth/services/rhsso.py
```diff
@@ -74,3 +74,18 @@ def exchange_code_for_login(self, app_config, http_client, code, redirect_suffix
...
+        Returns `deprecated_sub` if `sub` is a digit, otherwise returns `sub`.
+        """
+        sub = decoded_id_token.get("sub")
+        if not sub:
+            raise OAuthLoginException("Token missing 'sub' field")
+
+        if sub.isdigit():
+            # For older tokens, use deprecated_sub if available, fallback to sub
+            return decoded_id_token.get("deprecated_sub", sub)
+        else:
```
syed: nit: else is not needed

syed: just the `return` statement is enough

syed: we can say that if `deprecated_sub` exists, always return that. else return the `sub`

harishsurf: ack, removed the `sub.isdigit()` check and instead use `deprecated_sub` if exists, else the `sub`

oauth/login_utils.py
harishsurf: moved `get_user_id()` to `OAuthService` base meta class and added the override for rhsso

syed (approved):

PR #4153, opened 2025-07-28
ui: add retry logic for user api to prevent infinite api call (PROJQUAY-9124)

harishsurf:
This PR prevents infinite api calls to `/user` endpoint on console dot env when user is not successfully authenticated

### DIFF
--- a/web/src/hooks/UseCurrentUser.ts
+++ b/web/src/hooks/UseCurrentUser.ts
@@ -11,6 +11,14 @@ export function useCurrentUser() {
     error,
   } = useQuery(['user'], fetchUser, {
     staleTime: Infinity,
+    retry: (failureCount, error: any) => {
+      // Retry up to 5 times for 401 errors, then stop to prevent infinite loops
+      if (error?.response?.status === 401) {
+        return failureCount < 5;
+      }
+      // Default retry behavior for other errors
+      return failureCount < 3;
+    },
   });
 
   const isSuperUser =


web/src/hooks/UseCurrentUser.ts
```diff
@@ -11,6 +11,14 @@ export function useCurrentUser() {
     error,
   } = useQuery(['user'], fetchUser, {
     staleTime: Infinity,
+    retry: (failureCount, error: any) => {
```
syed: what happens after the 5 retries? is there an error page that pops up?

harishsurf: If API call fails, the `user` obj is undefined which should _hopefully_ show up [this banner](https://github.com/quay/quay/pull/4136/files#diff-3faab70fa4a9eb7ed0d484beb55d3b9d4df711af1b3afe5b74aab02a455b905dR99)

web/src/routes/PluginMain.tsx
```diff
@@ -67,14 +68,22 @@ function PluginMain() {
...
   const setIsPluginState = useSetRecoilState(IsPluginState);
   const [isConfirmUserModalOpen, setConfirmUserModalOpen] = useState(false);
+  const [tokenReady, setTokenReady] = useState(false);
 
-  chrome?.auth?.getToken().then((token) => {
-    GlobalAuthState.bearerToken = token;
-  });
+  useEffect(() => {
+    if (chrome?.auth) {
+      chrome.auth.getToken().then((token) => {
```
bcaton85: you still need the auth check on `chrome.auth?.`

web/src/routes/PluginMain.tsx
```diff
@@ -67,14 +68,22 @@ function PluginMain() {
...
+        GlobalAuthState.bearerToken = token;
+        setTokenReady(true);
+      });
+    } else {
+      setTokenReady(true);
+    }
+  }, [chrome]);
```
bcaton85: IIRC the input here has to be a primitive (number, string) otherwise if it's an object it does incorrect renders

web/src/routes/PluginMain.tsx
```diff
@@ -67,14 +68,22 @@ function PluginMain() {
...
-    GlobalAuthState.bearerToken = token;
-  });
+  useEffect(() => {
+    if (chrome?.auth) {
+      chrome.auth.getToken().then((token) => {
+        GlobalAuthState.bearerToken = token;
+        setTokenReady(true);
+      });
+    } else {
+      setTokenReady(true);
```
bcaton85, line 78: it seems like this will set tokenReady to true even if chrome?.auth isn't instantiated?

harishsurf: true, removed the `else` part

web/src/routes/PluginMain.tsx
harishsurf: having `useEffect` with no dependency array would trigger it on every component re-render. I have kept it empty `[]` so that token is only fetched once when the component initially loads. wdyt?

bcaton85 (approved):

PR #4159, opened 2025-07-29
nginx: rate limit on namespace (PROJQUAY-9203)

Marcusk19:
Should extract the namespace from auth scope param and implement rate limits based on that. Should fall back to auth header if namespace cannot be extracted for some reason

### DIFF
--- a/conf/nginx/rate-limiting.conf.jnj
+++ b/conf/nginx/rate-limiting.conf.jnj
@@ -20,14 +20,12 @@ map $arg_scope $scope_namespace {
 }
 
 # Create namespace-based rate limiting bucket for auth endpoint
-# Falls back to $request_id if no namespace is found (effectively disabling rate limiting)
 map $scope_namespace $auth_namespace_bucket {
   {% for namespace in non_rate_limited_namespaces %}
-  "{{ namespace }}" $request_id;
+  "{{ namespace }}"  $request_id;
   {% endfor %}
   {% if enable_rate_limits %}
-  ~^(.+)$ $1;  # Use namespace if found
-  default $http_authorization;  # Fallback to auth header if no namespace
+  default $http_authorization;  # Fallback to auth header
   {% else %}
   default $request_id;  # Disable rate limiting when not enabled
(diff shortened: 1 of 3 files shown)


conf/nginx/rate-limiting.conf.jnj
```diff
@@ -44,9 +65,14 @@ map $namespace $namespaced_http2_bucket {
 }
 
 {% if enable_rate_limits %}
+# Original auth rate limiting by authorization header
 limit_req_zone $http_authorization zone=staticauth:10m rate=120r/s;
```
syed: you can remove this since it's not being used anymore

conf/nginx/rate-limiting.conf.jnj
```diff
@@ -44,9 +65,14 @@ map $namespace $namespaced_http2_bucket {
...
+
+# New namespace-based auth rate limiting
+limit_req_zone $auth_namespace_bucket zone=namespace_auth:10m rate=120r/s;
 {% else %}
 limit_req_zone $request_id zone=staticauth:10m rate=300r/s;
```
syed: same here, this is not needed

conf/nginx/rate-limiting.conf.jnj
```diff
@@ -12,6 +12,27 @@ map $http2 $http2_bucket {
...
+}
+
+# Create namespace-based rate limiting bucket for auth endpoint
+# Falls back to $request_id if no namespace is found (effectively disabling rate limiting)
+map $scope_namespace $auth_namespace_bucket {
+  {% for namespace in non_rate_limited_namespaces %}
+  "{{ namespace }}" $request_id;
+  {% endfor %}
+  {% if enable_rate_limits %}
+  ~^(.+)$ $1;  # Use namespace if found
```
syed: you don't want this. The idea is if the `namespace` is in the non rate-limited list then basically limit by `request_id` which is essentially no limit. But, if it is any other namespace, limit by the old method which is the auth header.

PR #4162, opened 2025-07-29
Further increase ngnix rate limit on the auth header

arewm:
#4133 increased the rate limit but we are still seeing some rate limiting when trying to check images in parallel. I have been making changes on the [client side](https://github.com/conforma/cli/pull/2660), but the auth 429s are still occurring.


arewm:
Closing in favor of  #4161.

PR #4164, opened 2025-07-31
nginx: html encoded matching for regex pattern (PROJQUAY-9203)

Marcusk19:
Nginx rate limits are not properly picking up the namespace. It appears that the scope is html encoded so adjusting the regex to properly pick up the namespace. Tested locally to validate that rate limit key is the namespace.

### DIFF
--- a/conf/nginx/rate-limiting.conf.jnj
+++ b/conf/nginx/rate-limiting.conf.jnj
@@ -15,7 +15,7 @@ map $http2 $http2_bucket {
 # Extract namespace from scope query parameter
 # Docker scope format: "repository:namespace/repo:actions"
 map $arg_scope $scope_namespace {
-  ~^repository:([^/]+)/.* $1;
+  ~^repository%3A(.+?)(%2F|%3A).* $1;
   default "";
 }
 


conf/nginx/rate-limiting.conf.jnj
```diff
@@ -15,7 +15,7 @@ map $http2 $http2_bucket {
 # Extract namespace from scope query parameter
 # Docker scope format: "repository:namespace/repo:actions"
 map $arg_scope $scope_namespace {
-  ~^repository:([^/]+)/.* $1;
+  ~^repository(?:|%3A)([^(?\/|%2F)]+)[(?\/|%2F)].+ $1;
```
syed: looking at this regex, if all you want is to match something like `repository%3Aredhat-user-workloads%2Fsomerepo%3Apull%2Cpush` or `repository%3Asome-repo%3Apull%2Cpush`  you can simplify this to `^repository%3A(.+?)(%2F|%3A).*` what do you think?

Marcusk19: sounds good to me, I was thinking maybe it should try to be flexible with either `/` or `%2F` but if we can simplify it that would be better

syed (approved):

PR #4166, opened 2025-08-01
nginx: fix rate limiting bucket (PROJQUAY-9203)

Marcusk19:
Bucket was improperly using the auth header as a key even though it should be using the namespace with the new changes.


Marcusk19:
Closing, as discussed using namespace for rate limit key is not desired since the unintended result would be namespaces with heavy use getting severely rate limited

PR #4169, opened 2025-08-01
Updated raven dependency to be compatible with python 3.12 

deshpandevlab:
Updated raven dependency to be compatible with python 3.12

### DIFF
--- a/buildman/builder.py
+++ b/buildman/builder.py
@@ -4,8 +4,7 @@ import os
 import socket
 import time
 
-from raven.conf import setup_logging
-from raven.handlers.logging import SentryHandler
+import sentry_sdk
 
 import features
 from app import (
@@ -98,9 +97,16 @@ if __name__ == "__main__":
     logging.getLogger("boto").setLevel(logging.WARN)
 
     if app.config.get("EXCEPTION_LOG_TYPE", "FakeSentry") == "Sentry":
-        buildman_name = "%s:buildman" % socket.gethostname()
-        setup_logging(
-            SentryHandler(app.config.get("SENTRY_DSN", ""), name=buildman_name, level=logging.ERROR)
-        )
+        sentry_dsn = app.config.get("SENTRY_DSN", "")
+        if sentry_dsn:
+            buildman_name = "%s:buildman" % socket.gethostname()
+            sentry_sdk.init(
+                dsn=sentry_dsn,
+                environment=app.config.get("SENTRY_ENVIRONMENT", "production"),
+                traces_sample_rate=app.config.get("SENTRY_TRACES_SAMPLE_RATE", 0.1),
+                profiles_sample_rate=app.config.get("SENTRY_PROFILES_SAMPLE_RATE", 0.1),
+            )
+            sentry_sdk.set_tag("service", "buildman")
+            sentry_sdk.set_tag("buildman", buildman_name)
 
     run_build_manager()
--- a/config-tool/pkg/lib/editor/package.json
+++ b/config-tool/pkg/lib/editor/package.json
@@ -40,7 +40,7 @@
     "moment-timezone": "^0.5.35",
     "ng-metadata": "^4.0.1",
     "node-forge": "^0.10.0",
-    "raven-js": "^3.1.0",
+    "@sentry/browser": "^7.100.0",
     "restangular": "^1.6.1",
     "rxjs": "5.5.7",
     "showdown": "^1.6.4",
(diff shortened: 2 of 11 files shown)


buildman/builder.py
```diff
@@ -31,6 +30,24 @@
 DEFAULT_CONTROLLER_PORT = 8686
 
 
+def initialize_sentry():
+    """Initialize Sentry if configured."""
+    # Define buildman name for general use
+    buildman_name = "%s:buildman" % socket.gethostname()
```
deshpandevlab, line 36: this change is in order for builder to build without sentry

util/saas/exceptionlog.py
```diff
@@ -26,7 +27,25 @@ def init_app(self, app):
...
+                    sentry_sdk.init(
+                        dsn=sentry_dsn,
+                        integrations=[FlaskIntegration()],
+                        environment=app.config.get("SENTRY_ENVIRONMENT", "production"),
+                        traces_sample_rate=app.config.get("SENTRY_TRACES_SAMPLE_RATE", 0.1),
+                        profiles_sample_rate=app.config.get("SENTRY_PROFILES_SAMPLE_RATE", 0.1),
+                    )
+                    sentry = FakeSentry()
+                except Exception as e:
+                    import logging
```
harishsurf: nit: do you want to move imports to top of the file to keep it consistent with the other files?

util/saas/exceptionlog.py
```diff
@@ -26,7 +27,25 @@ def init_app(self, app):
...
+            if sentry_dsn:
+                try:
+                    sentry_sdk.init(
+                        dsn=sentry_dsn,
+                        integrations=[FlaskIntegration()],
+                        environment=app.config.get("SENTRY_ENVIRONMENT", "production"),
+                        traces_sample_rate=app.config.get("SENTRY_TRACES_SAMPLE_RATE", 0.1),
+                        profiles_sample_rate=app.config.get("SENTRY_PROFILES_SAMPLE_RATE", 0.1),
+                    )
+                    sentry = FakeSentry()
```
harishsurf: I think we shouldn't be creating a FakeSentry() obj here?  The `sentry_sdk.init()` should already be doing it?

config-tool/pkg/lib/editor/package.json
```diff
@@ -40,7 +40,7 @@
     "moment-timezone": "^0.5.35",
     "ng-metadata": "^4.0.1",
     "node-forge": "^0.10.0",
-    "raven-js": "^3.1.0",
+    "@sentry/browser": "^7.100.0",
```
harishsurf, line 43: nit: would be a good idea to make sure config-tool editor UI loads fine after this dep bump

requirements.txt
```diff
@@ -102,7 +102,7 @@ python-redis-lock==3.7.0
 python-swiftclient==4.4.0
 pytz==2019.3
 PyYAML==6.0.2
-raven==6.10.0
+sentry-sdk[flask]==1.40.0
```
harishsurf, line 105: curious, do we need `[flask]` bit? not sure if it's compatible with downstream builds

harishsurf (commented):
looks good! left few comments

workers/test/test_quotatotalworker.py
```diff
@@ -40,3 +41,100 @@ def assert_mock_run_backfill(namespace_id):
         worker = QuotaTotalWorker()
         worker.backfill()
         assert len(expected_calls) == 0
+
+
+def test_worker_name_defined_outside_sentry_condition():
```
bcaton85: nit: these tests were put in the `test_quotatotalworker.py` file but the change effects all workers. Is there another file that this should be in? or separate it into it's own file?

util/saas/exceptionlog.py
deshpandevlab: good catch. This has been how we initialize sentry today, We initialize fake object even when sentry is successfully initialized globally. But this should be simplified. updated

requirements.txt
deshpandevlab: yeah, added flask to capture the details of the http errors. Without it we may be fine but wanted to keep it to the auto capture request details in the error. Removed it.

util/saas/exceptionlog.py
deshpandevlab: yep! weird thing cursor did even after nudging multiple times. Fixing

workers/test/test_quotatotalworker.py
deshpandevlab: yeah moved the tests thanks

harishsurf (approved):
lgtm

PR #4171, opened 2025-08-05
Create a dedicated template for ROSA deployment

lechuk47:


syed:
can you use https://github.com/quay/quay/blob/master/deploy/openshift/quay-py3-deploy-template.yaml instead. This will allow us to also create canary deployments

lechuk47:
> can you use https://github.com/quay/quay/blob/master/deploy/openshift/quay-py3-deploy-template.yaml instead. This will allow us to also create canary deployments

I made some changes as discussed: 
* Change the object names like in the deploy-template. 
* Added deployment port definitions. They are optional, but I think it's worth keeping them updated. 
* Cluster-ip service is kept. It's used to scrape metrics. 
* Minor nits 

PTAL.

PR #4179, opened 2025-08-08
deps: removing unused raven dependencies (PROJQUAY-9198)

deshpandevlab:
- The issue was that the webpack bundle was trying to load `@sentry/browser` as an external dependency, but the Sentry global variable wasn't available when sentry_public_dsn was not set (it is not set by default).
- We moved the FakeSentry wrapper to load BEFORE the main scripts, so it's available when the webpack bundle tries to access window.Sentry.
- Both the old Raven library and new Sentry SDK were being loaded simultaneously causing conflict for the console page loading.

### Testing
reproduced locally: 

Screenshot 2025-08-08 at 12 08 49 PM

- loaded the page locally, tested loading all the nested pages. The error is resolved now
Screenshot 2025-08-08 at 10 56 49 AM

- Running nightly tests


harishsurf (approved):
lgtm if you tested it out locally

PR #4193, opened 2025-08-11
db: enable WAL mode and other PRAGMA stmts for SQLite to avoid db locking during concurrent writes (PROJQUAY-8758)

harishsurf:
Enable `WAL` mode and other PRAGMA statements for SQLite backed database to avoid db locking during concurrent writes. 
Tested the changes against OMR by running `oc mirror`, during mirroring and deleting the repo being mirrored. Repository GC worker tries to clean up the repo while the mirroring is also writing to the repo. Since both processes are in contention to write to the same table, sqlite throws database is locked. This error is not seen in after setting the database to WAL

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -454,7 +454,11 @@ def _db_from_url(
     drivers = _SCHEME_DRIVERS[parsed_url.drivername]
     driver = drivers.driver
     # Skip connection pooling for SQLite - use single connection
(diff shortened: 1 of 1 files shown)


bcaton85 (approved):
Looks good. Ideally these parameters would be configurable for things like timeout. That way if the user needs to tune or add they're own parameters they have the capability to do so. Or if these parameters cause issues (which i don't believe they will) they can remove them

PR #4198, opened 2025-08-15
tracing: improving otlp handling (PROJQUAY-8902) 

michaelalang:
* Include Opentelemetry handling for Setup to avoid hardcoding one Receiver (dynatrace)
* Include externalldap Trace creation
* Include S3 boto call Trace creation


michaelalang:
with removing the hardcoded endpoint configuration we could instead utilize sidecar injection from the Red Hat Build of OpenTelemetry operator to get a localhost accessible Collector which can handle multiple exporters(dynatrace,tempo...) and processors(filter,batch,tail-sampling,...)  

Examples:
Dynatrace Exporter
```
spec:
  extraEnvs:
    - name: DT_ENDPOINT
      valueFrom:
        secretKeyRef:
          key: DT_ENDPOINT
          name: dynatrace
    - name: API_TOKEN
      valueFrom:
        secretKeyRef:
          key: API_TOKEN
          name: dynatrace
  config:
    exporters:
      otlphttp/dynatrace:
        # NOTE signal specific paths will be appended 
        # the url in secret shall therfor be 'https://{tenant-id}.live.dynatrace.com/api/v2/otlp'
        endpoint: ${env:DT_ENDPOINT} 
        headers:
          Authorization: "Api-Token ${env:API_TOKEN}"
```

OpenShift Tempo Product Exporter
```
spec:
  config:
    exporters:
      otlp/tempo:
        auth:
          authenticator: bearertokenauth
        endpoint: tempo-observe-gateway.tempo.svc.cluster.local:8090
        headers:
          X-Scope-OrgID: tenant-name
        tls:
          ca_file: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
    extensions:
      bearertokenauth:
        filename: /var/run/secrets/kubernetes.io/serviceaccount/token
    service:
      extensions:
        - bearertokenauth
```

Tail sampling to remove single spans created by autoinstrumentation which do not give context
```
spec:
  processors:
      tail_sampling:
        decision_cache:
          non_sampled_cache_size: 100000
          sampled_cache_size: 100000
        decision_wait: 10s
        expected_new_traces_per_sec: 100
        num_traces: 100
        policies:
          - name: drop-single-span-traces
            span_count:
              max_spans: 2000
              min_spans: 2
            type: span_count
         - name: rate-limit-for-dynatrace
           type: rate_limiting
           rate_limiting:
             spans_per_second: 16  # 16 * 60 to stay below 1000
```

Filter to remove unnecessary Quay API calls from being sampled 
```
spec:
    processors:
      filter/health:
        error_mode: ignore
        traces:
          span:
            - IsMatch(attributes["http.route"], "/(health|ready|livez|startupz).*")
            - IsMatch(attributes["http.target"], "/(health|ready|livez|startupz).*")
      filter/v1_endpoints:
        error_mode: ignore
        traces:
          span:
            - IsMatch(attributes["http.route"], "/(v1/|_).*")
            - IsMatch(attributes["http.target"], "/(v1/|_).*")
```

Extending traces with k8s specific attributes 
```
spec:
  env:
    - name: KUBE_NODE_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName
    processors:
      k8sattributes:
        auth_type: serviceAccount
        extract:
          labels:
            - from: pod
              key: app.kubernetes.io/component
              tag_name: app.label.component
          metadata:
            - k8s.pod.name
            - k8s.pod.uid
            - k8s.deployment.name
            - k8s.namespace.name
            - k8s.node.name
            - k8s.pod.start_time
        filter:
          node_from_env_var: KUBE_NODE_NAME
        passthrough: false
        pod_association:
          - sources:
              - from: resource_attribute
                name: k8s.pod.ip
          - sources:
              - from: resource_attribute
                name: k8s.pod.uid
          - sources:
              - from: connection
```

michaelalang:
Showing some example of the traces that Quay now generates 

traces-example-01
traces-example-02
traces-example-05
traces-example-04
traces-example-03

endpoints/v1/registry.py
```diff
@@ -4,6 +4,10 @@
 from functools import wraps
 from time import time
 
+from util.metrics.otel import trace
```
kleesc, line 7: It's probably not worth tracing the `v1/` endpoints. It's deprecated and all users should be using v2

app.py
```diff
@@ -368,9 +369,27 @@ def load_user(user_uuid):
...
+            if all([status_code >= 200, status_code < 400]):
+                span.set_status(StatusCode.OK)
+            else:
+                span.set_status(StatusCode.ERROR)
+            tparent = get_traceparent(span)
+            response_headers.append(("traceparent", tparent))
+
     FlaskInstrumentor().instrument_app(
         app,
         excluded_urls=app.config.get("OTEL_TRACING_EXCLUDED_URLS", None),
```
kleesc, line 390: Do we need to keep this list, now that there is explicitly a decorator for the endpoints?

util/config/schema.py
```diff
@@ -1609,15 +1609,36 @@
...
                 },
-                "dt_api_token": {
+                "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT": {
                     "type": "string",
-                    "description": "token for dynatrace api",
-                    "x-example": "sometoken",
+                    "description": "url for OTLP api environment overwrites",
+                    "x-example": "https://dynatrace-api.example:443",
+                },
+                "OTEL_EXPORTER_OTLP_HEADERS": {
```
kleesc, line 1622: Can you still configure it without a collector or sidecar? Could we still put the previous `dt_api_url` and `dt_api_token` in `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_HEADERS` and have backward compatible behavior?

endpoints/v1/registry.py
michaelalang: let me know if I should remove the `endpoints/v1/registry.py` decoration or not ... I don't mind keeping it even if it's deprecated ... (one never knows who/what still calls it :D)

here's what I get with 3.15 on the v1 endpoint 
```
GET /api/v1/messages
GET /api/v1/repository
GET /api/v1/user/
GET /api/v1/user/apptoken
GET /api/v1/user/notifications
```

app.py
michaelalang: Not sure what you mean with `list`. The hooks have been added to ensure, quay does not only receive and consider but also return traceparent headers accordingly for tracking through various infrastructure points.

the attributes added will increase the information level of the traces with the headers available by each request.

util/config/schema.py
michaelalang: Yes the configuration provide in the above comments show's how this is done as well as I have also recorded the equal functionality between sidecar/sidecar-less tracing integration.
Screenshot From 2025-08-31 19-31-44

michaelalang: a mapped version of the previous `dt_api_url` and `dt_api_token` would be 

```
OTEL_CONFIG:
  OTEL_EXPORTER_OTLP_ENDPOINT: https://<tenant>.live.dynatrace.com/api/v2/otlp
  OTEL_EXPORTER_OTLP_PROTOCOL: http
  OTEL_TRACES_SAMPLER_ARG: 1.0
  OTEL_EXPORTER_OTLP_HEADERS:
    Authorization: Api-Token xxxxxx
  service_name: quay
```

here's proof that those settings will work (just changed it to no sidecar with the above configuration)
Screenshot From 2025-09-11 20-10-06

endpoints/v1/registry.py
kleesc: The endpoints you have above are Quay's API and are valid. I meant endpoints starting with `/v1/...`
I think it's fine. We shouldn't be getting too many traces from these anyways.

app.py
kleesc: `OTEL_TRACING_EXCLUDED_URLS` tells the flask instrumentor which endpoints not to trace, but it looks like we're explicitly adding `@tracer.start_as_current_span` to every endpoints we care about. It doesn't do anything to leave it as is. Just wondering if it's still needed.

kleesc:
@michaelalang should be good to merge. LGTM

michaelalang:
@kleesc I modified the header part in util/metrics/otel once more as the environ parsing would only add a string and not a dict. 
This last minute change is of no other purpose and results as follows in testing

```
# test1 multiple headers
$ OTEL_EXPORTER_OTLP_HEADERS='x-auth-sentry: sentry sentry_key=xxx,x-tenant=something' python test.py 
{'x-auth-sentry: sentry sentry_key': 'xxx', 'x-tenant': 'something'}

# test2 multiple headers space between separator
$ OTEL_EXPORTER_OTLP_HEADERS='x-auth-sentry: sentry sentry_key=xxx, x-tenant=something' python test.py 
{'x-auth-sentry: sentry sentry_key': 'xxx', 'x-tenant': 'something'}

# test3 no headers at all
$ python test.py 
{}
```

kleesc (approved):
LGTM

michaelalang:
@kleesc @lechuk47, I removed the explizit trace context from upload_chunk in my PR but I cannot restart the iteration/evaluation process as the state of the PR is merged ...

PR #4232, opened 2025-08-29
fix: resolve Sentry/OpenTelemetry integration conflicts (PROJQUAY-9198)

deshpandevlab:
- Configure Sentry to use minimal integrations when OTEL_TRACING is enabled to prevent instrumentation conflicts that broke exception capture

- There is also an issue where connection pool gets full and sentry receives discarded connection error
```
[27744] [WARNING] [urllib3.connectionpool] Connection pool is full, discarding connection: app.getsentry.com. Connection pool size: 1
```

This PR fixes this issue by increasing connection pool size

### DIFF
--- a/workers/worker.py
+++ b/workers/worker.py
@@ -10,7 +10,12 @@ from threading import Event
 
 import sentry_sdk
 from apscheduler.schedulers.background import BackgroundScheduler
+from sentry_sdk.integrations.flask import FlaskIntegration
+from sentry_sdk.integrations.logging import LoggingIntegration
+from sentry_sdk.integrations.sqlalchemy import SqlalchemyIntegration
+from sentry_sdk.integrations.stdlib import StdlibIntegration
 
+import features
 from app import app
 from data.database import UseThenDisconnect
 from util.log import logfile_path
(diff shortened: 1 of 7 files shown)


deshpandevlab:
codepatch for builder is mostly imports and logging

workers/worker.py
```diff
@@ -67,11 +72,43 @@ def __init__(self):
...
+                        logger.info(
+                            "OpenTelemetry enabled - using minimal Sentry integrations for worker"
+                        )
+
                     sentry_sdk.init(
                         dsn=sentry_dsn,
                         environment=app.config.get("SENTRY_ENVIRONMENT", "production"),
                         traces_sample_rate=app.config.get("SENTRY_TRACES_SAMPLE_RATE", 0.1),
                         profiles_sample_rate=app.config.get("SENTRY_PROFILES_SAMPLE_RATE", 0.1),
+                        integrations=integrations,
```
cubismod, line 101: Which integrations was sentry including prior to this change?

cubismod: Answered my own question checking the docs: https://docs.sentry.io/platforms/python/configuration/options/#integrations. My next question: did you identify which of the default integrations/auto-enabling integrations was breaking sentry? Regardless, it's probably best practice to disable integrations that we know we'll never use.

deshpandevlab: this change explicitely disabled all the default integrations: like StdlibIntegration, DedupeIntegration, ExcepthookIntegration, LoggingIntegration  etc. Along with that it also included FlaskIntegration, SqlalchemyIntegration (conditional) etc

cubismod (approved):

PR #4235, opened 2025-09-02
proxy: Drop error 403 token requests for anonymous pulls (PROJQUAY-9012)

sharpenedblade:
ghcr.io returns 403 when attempting an anonymous pull, which causes the proxy to attempt authentication even when authentication is disabled. This change ignores 403 status codes when requesting anonymous pulls, similar to the existing behavior for 401 responses.

Fixes https://issues.redhat.com/browse/PROJQUAY-9012, the code is from the original issue poster. I tested the code myself and it appears to work and not break anything.

### DIFF
--- a/proxy/__init__.py
+++ b/proxy/__init__.py
@@ -206,7 +206,7 @@ class Proxy:
 
         # ignore fetching a token when validating proxy cache config to allow anonymous pulls from registries,
         # since the repo name is not known during the initial proxy configuration
-        if resp.status_code == 401 and auth is None and self._repo is None:
+        if (resp.status_code == 401 or resp.status_code == 403) and auth is None and self._repo is None:
             return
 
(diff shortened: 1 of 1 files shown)


harishsurf:
@sharpenedblade , seems like the CI didn't get triggered. Could you push another commit/ rebase PR and may be that kicks it off?

sharpenedblade:
I rebased onto the latest `master` but it looks like CI still isn't running. The CI page says the workflows need approval from a maintainer, so I think its a permissions issue.

harishsurf:
@sharpenedblade, you can ignore the cypress test failure, but could you add a test for codecov check that's failing?

sharpenedblade:
Sorry, but I don't know how to test that code path since the existing 401 case is not covered in the unit tests.

PR #4241, opened 2025-09-04
added sentry initialization and logging

deshpandevlab:


util/saas/exceptionlog.py
```diff
@@ -37,6 +37,9 @@ def init_app(self, app):
             sentry_dsn = app.config.get("SENTRY_DSN", "")
             if sentry_dsn:
                 try:
+                    logger = logging.getLogger(__name__)
+                    logger.info("Initializing Sentry with DSN: %s", sentry_dsn[:50] + "...")
```
lechuk47:
```suggestion
                    logger.info("Initializing Sentry with DSN: %s...", sentry_dsn[:50])
```

NIT: I would also consider reducing the DSN length to 8/10.

util/saas/exceptionlog.py
```diff
@@ -84,6 +99,24 @@ def init_app(self, app):
         app.extensions["sentry"] = sentry
         return sentry
 
+    def test_sentry_connection(self):
```
bcaton85: if this is only for tests do we need this in a non-tests file?

test/test_exceptionlog_sentry.py
```diff
@@ -200,3 +200,87 @@ def test_sentry_extensions_existing(self):
         assert "existing" in mock_app.extensions
         assert "sentry" in mock_app.extensions
         assert mock_app.extensions["sentry"] is sentry.state
+
+    def test_sentry_connection_test_with_real_sentry(self):
+        """Test that test_sentry_connection returns True when Sentry is properly configured."""
+        mock_app = MagicMock()
+        mock_app.config.get.side_effect = lambda key, default=None: {
+            "EXCEPTION_LOG_TYPE": "Sentry",
+            "SENTRY_DSN": "https://test@sentry.io/123",
```
bcaton85: do these tests actually reach out to sentry? unit tests should be self contained and not rely on external services

cubismod: Yeah I'm also kind of confused as to the purpose of these tests. I don't think we can recreate the current issues with sentry in a unit test.

util/saas/exceptionlog.py
deshpandevlab: great catch! thanks fixed it

util/saas/exceptionlog.py
deshpandevlab: nope. You are right. I removed the related tests and the file as well

test/test_exceptionlog_sentry.py
deshpandevlab: Wanted to try a way to test if this will actually fix the integration. I removed the tests.

PR #4243, opened 2025-09-05
ui: Add Quota management functionality to UI (PROJQUAY-8887)

mfrances17:
This PR adds Quota management functionality to the new Quay UI.

All UI components and functionality that existed in the legacy UI should now be working in the new React-based UI with Patternfly 5. 

Screen shots:
TBD when moving from draft to PR

### DIFF
--- a/web/cypress/e2e/quota.cy.ts
+++ b/web/cypress/e2e/quota.cy.ts
@@ -1,0 +1,602 @@
+/// <reference types="cypress" />
+
+describe('Organization Quota Management', () => {
+  beforeEach(() => {
+    // Handle pre-existing CanceledError from organization page navigation
+    cy.on('uncaught:exception', (err) => {
+      if (
+        err.message.includes('canceled') ||
+        err.message.includes('CanceledError')
+      ) {
+        return false;
+      }
+      return true;
+    });
+
+    cy.exec('npm run quay:seed');
+    cy.request('GET', `${Cypress.env('REACT_QUAY_APP_API_URL')}/csrf_token`)
+      .then((response) => response.body.csrf_token)
+      .then((token) => {
+        cy.loginByCSRF(token);
+      });
+    cy.intercept('GET', '/config', {fixture: 'config.json'}).as('getConfig');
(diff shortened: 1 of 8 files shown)


harishsurf (dismissed):
Tested out the quota functionality. Overall looks good, just had few comments:
1. When a user is not superuser, the quota tab should be readonly and shouldn’t allow editing quota configs. Currently new UI allows it, but fails at the API level.
2. Adding multiple “reject” quota policy type shouldn’t be allowed. We can only have one “reject” policy type
3. In Angular, I see it displays [(%) quota consumed] (https://github.com/quay/quay/blob/master/static/directives/repo-list-table.html#L90) per repository along with its size, where are react only shows the repo size
4. In superuser panel, the old UI has a total registry size displayed along with a calculate button. Not sure if this was implemented in a separate PR, but wanted to call it out. 

Feel free to add in the fixes in same/ or separate PR. Thank you for your work!

mfrances17:
Great review... thanks! I've fixed the merge conflicts and addressed all of these issues:

> 1. When a user is not superuser, the quota tab should be readonly and shouldn’t allow editing quota configs. Currently new UI allows it, but fails at the API level.

This should now be fixed:
quota-not-superuser

> 2. Adding multiple “reject” quota policy type shouldn’t be allowed. We can only have one “reject” policy type

I did originally address this by catching it and displaying a notification, but now I've added some code to prevent the action entirely if a reject is already defined (which is better UX than the original UI did):

prevent-reject

> 3. In Angular, I see it displays [(%) quota consumed]  per repository along with its size, where are react only shows the repo size

I added this functionality... note the subtleties in this screen cap, `clair-jwt` shows 268.47KiB (26%) because it has a quota, the two `hello-worlds` have no quota so don't display a (%), `prometheus` isn't defined and has no quota defined, and `repo1` is undefined and has a quota defined). 

repo-size-and-quota-percents

> 4. In superuser panel, the old UI has a total registry size displayed along with a calculate button. Not sure if this was implemented in a separate PR, but wanted to call it out.

Added this functionality... this screen cap is after I clicked Calculate and then refreshed:
org-after-caclulate

PR #4284, opened 2025-09-15
feat(auth): add global read-only superuser permission plumbing and tests

HammerMeetNail:
Topic: global-ro-superuser-v1-01-foundation


HammerMeetNail:
Closing as requested.

PR #4286, opened 2025-09-15
feat(app tokens): expose limited app token visibility for global readonly superusers

HammerMeetNail:
Topic: global-ro-superuser-v1-03-endpoints-user-token
Relative: global-ro-superuser-v1-02-auth-middleware


HammerMeetNail:
Closing as requested.

PR #4290, opened 2025-09-16
feat(org/team endpoints): allow safe read access for global readonly superusers

HammerMeetNail:
Topic: global-ro-superuser-v1-06-endpoints-org-team
Relative: global-ro-superuser-v1-05-endpoints-user


HammerMeetNail:
Closing as requested.

PR #4295, opened 2025-09-16
fix: Fix Sentry transport KeyError in exception logging initialization (PROJQUAY-9198)

deshpandevlab:
Fix Sentry transport KeyError in exception logging initialization

**Root cause:**

Custom transport configuration using `sentry_sdk.transport.make_transport()` was incompatible with the current Sentry SDK version. It did not recognize the transport keyword.

**Testing:**
- Verified Sentry initialization completes successfully
- Confirmed exception capture works correctly
- Tested both OTEL-enabled and OTEL-disabled scenarios
- OTEL/Sentry conflict resolution logic preserved and working
Screenshot 2025-09-16 at 2 32 54 PM

### DIFF
--- a/buildman/builder.py
+++ b/buildman/builder.py
@@ -72,14 +72,6 @@ def initialize_sentry():
                     integrations=integrations,
                     default_integrations=False,
                     auto_session_tracking=True,
-                    # Fix connection pool issues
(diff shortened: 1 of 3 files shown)


syed:
why is the `package-lock.json` file changing?

bcaton85 (approved):

PR #4304, opened 2025-09-22
fix(new ui): Move tag gives note when inputting existing tag name (PROJQUAY-9422)

LiZhang19817:
This PR will fix the new UI of https://issues.redhat.com/browse/PROJQUAY-9422

**Now here're the expected behavior on new UI:**
- Check if an image tag is existed or not
- Update to change the head message to "**Move** tag to manifest *****"
- Display a note that the new image tag name is already existed, and will move the tag to another manifest.

image

### DIFF
--- a/web/src/routes/RepositoryDetails/Tags/TagsActionsAddTagModal.tsx
+++ b/web/src/routes/RepositoryDetails/Tags/TagsActionsAddTagModal.tsx
@@ -57,9 +57,8 @@ export default function AddTagModal(props: AddTagModalProps) {
         tagName,
         true,
       );
-      const existingTag = tagsResponse.tags.find(
-        (tag) => tag.name === tagName,
-      );
+
+      const existingTag = tagsResponse.tags.find((tag) => tag.name === tagName);
 
       if (existingTag) {
         setExistingTagInfo({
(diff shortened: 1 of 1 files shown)


web/src/routes/RepositoryDetails/Tags/TagsActionsAddTagModal.tsx
```diff
@@ -1,60 +1,127 @@
...
+  const checkTagExists = async (tagName: string) => {
+    if (!tagName.trim()) {
+      setExistingTagInfo(null);
+      return;
+    }
+
+    setIsCheckingTag(true);
+    try {
+      // Call existing tags API to check if tag exists
+      const tagsResponse = await getTags(
```
bcaton85, line 52: My only concern here is the list tags operation is quite expensive. I don't think it would cause any issues on quay.io since this isn't ran very often but it's something we probably shouldn't add if we don't need to. Does the old UI perform this same operation?

bcaton85: Just tested on the old ui, it does not make these requests

bcaton85 (changes requested):
Thank you for the contribution! this does work and it was a good catch finding this. The only concern is making the list tags call after typing, which is an already expensive API. The old ui doesn't do this, can we copy the functionality from the old ui? This can be resolved by passing through the existing tag list through the [table component](https://github.com/quay/quay/blob/master/web/src/routes/RepositoryDetails/Tags/TagsTable.tsx#L229)

PR #2632, opened 2024-01-24
sts: Add STS support for S3 (PROJQUAY-6362)

jonathankingfc:
- Add STS authentication to s3
- Add STS validation to config tool
- PR automatically refreshes token on expiry

PR #4312, opened 2025-09-25
feat: add IRSA support when creating Quay.io clients (PROJQUAY-9228)

cubismod:
[PROJQUAY-9228](https://issues.redhat.com/browse/PROJQUAY-9228)

Add support for using IRSA (IAM roles for service accounts) with [boto3's automatic credential](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#assume-role-with-web-identity-provider) refreshing to Quay.io. 

IRSA is supported on ROSA clusters, see https://aws.amazon.com/blogs/containers/fine-grained-iam-roles-for-red-hat-openshift-service-on-aws-rosa-workloads-with-sts/ for more information. This PR adds IRSA support in a pretty simple way, by creating clients without an access key or secret key but passing the following environment variables: `AWS_WEB_IDENTITY_TOKEN_FILE`, `AWS_ROLE_ARN`, & `AWS_ROLE_SESSION_NAME` to the container. From there, boto3 will handle assuming the role and refreshing sessions. 

The bulk of this PR is fixing up some bugs and oversights in the validator Golang code and adding proper support for testing with Localstack.

## Testing

This was tested locally with Localstack mocking AWS services.

**Assisted with Claude Code**

### DIFF
--- a/config-tool/pkg/lib/shared/storage_validators.go
+++ b/config-tool/pkg/lib/shared/storage_validators.go
@@ -203,7 +203,7 @@ func ValidateStorage(opts Options, storageName string, storageType string, args 
 				newError := ValidationError{
 					Tags:       []string{"DISTRIBUTED_STORAGE_CONFIG"},
 					FieldGroup: fgName,
-					Message:    "No access key or secret key were provided. Attempted to fetch credentials from default chain and failed.",
+					Message:    fmt.Sprintf("No access key or secret key were provided. Attempted to fetch credentials from default chain and failed: %v", err),
 				}
 				errors = append(errors, newError)
 				return false, errors


storage/cloud.py
```diff
@@ -1219,13 +1219,20 @@ def __init__(
         maximum_chunk_size_gb=None,
         signature_version="s3v4",
     ):
+        sts_kwargs = {}
+        if endpoint_url:
+            sts_kwargs["endpoint_url"] = endpoint_url
```
lechuk47: Isn't `endpoint_url` the S3 bucket URL? If that's the case, it shouldn't be part of the client parameters. 

Why the region is needed now?

lechuk47: The configuration for this class is hard to understand. 

* If `sts_role_arn` is set, the client will assume the role via an assumeRole call via an STS Client.
* The credentials or the STS client can come from the default supported configuration places, or be set with `sts_user_access_key` and `sts_user_secret_key`. 

Thinking about the ROSA/IRSA case, the Role set `sts_role_arn` needs to be assumed by the POD IAM Role; therefore will require additional AWS configuration.

lechuk47:
We should try to generalize the AWS credentials usage as much as we can. For example, the `S3StorageClass` class does not use the `use_sts` attribute; it relies on the IAM credentials being empty or not. The other cases use this specific attribute. 

Also, the `S3StorageClass` (and possibly others) can assume another role to interact with S3, so `use_sts` might be misleading.

storage/cloud.py
cubismod: No, `endpoint_url` is referring to the S3 host (s3.amazonaws.com, localstack.localhost, etc). Good point, region isn't needed. Yeah I agree, this class is confusing. Most of it was done in https://github.com/quay/quay/pull/2632 and I tried to maintain backwards compatibility with that setup since customers use it. Should I make a separate class specifically for IRSA?

cubismod: I'm going to do a separate class for IRSA

cubismod: I also reverted the changes to S3 storage.

cubismod:
Working on fixing the validator code now.

storage/cloud.py
```diff
@@ -1255,3 +1255,31 @@ def __init__(
...
+            context,
+            storage_path,
+            s3_bucket,
+            s3_region=s3_region,
+            endpoint_url=endpoint_url,
+            maximum_chunk_size_gb=maximum_chunk_size_gb,
+            signature_version=signature_version,
+            s3_access_key=None,
+            s3_secret_key=None,
+        )
```
kleesc: What is this class doing differently from `S3Storage`?

data/logs_model/logs_producer/kinesis_stream_logs_producer.py
```diff
@@ -71,14 +72,19 @@ def __init__(
...
+            self._producer = boto3.client(
+                "kinesis", use_ssl=True, region_name=self._aws_region, config=client_config
+            )
+        else:
+            self._producer = boto3.client(
+                "kinesis",
+                use_ssl=True,
+                region_name=self._aws_region,
+                aws_access_key_id=self._aws_access_key,
+                aws_secret_access_key=self._aws_secret_key,
```
kleesc: Would these aws keys have a value set if sts is used in the first place?

storage/cloud.py
cubismod: Nothing, actually. I created this class specifically so that I could update the validator code and not break existing validation for S3Storage.

data/logs_model/logs_producer/kinesis_stream_logs_producer.py
cubismod: Actually, no, this condition probably isn't needed in this case since those values default to `None`.

storage/cloud.py
cubismod: Removed in [b207187](https://github.com/quay/quay/pull/4312/commits/b207187ccc82afcccdd11fccebe1c38e8a77ad9a)

config-tool/pkg/lib/shared/storage_validators.go
```diff
@@ -19,6 +20,76 @@ import (
...
+
+		awsConfig.Endpoint = aws.String(stsEndpointURL)
+		if !isSecure {
+			awsConfig.DisableSSL = aws.Bool(true)
+		}
+	} else {
+		// Default to AWS S3
+		s3Endpoint = "s3.amazonaws.com"
+		isSecure = true
+	}
```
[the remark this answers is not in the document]
cubismod: Resolved in https://github.com/quay/quay/pull/4312/commits/ed72e7e267bb928d2aaf49d09f2ffbcec4326411

config-tool/pkg/lib/shared/storage_validators.go
```diff
@@ -174,10 +243,31 @@ func ValidateStorage(opts Options, storageName string, storageType string, args
...
+				Message:    "Invalid endpoint configuration: " + err.Error(),
+			})
+			return false, errors
+		}
+
 		var credentials *sts.Credentials
 		// Prefer using web tokens to authenticate and fallback to access and secret keys
 		if webIdentityTokenFile != "" {
-			sess := session.Must(session.NewSession())
+			sess := session.Must(session.NewSession(awsConfig))
```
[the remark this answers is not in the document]
cubismod: Fixed in [5425e53](https://github.com/quay/quay/pull/4312/commits/5425e53c2c992c9fbaacfa9c515f77690168130c)

PR #4318, opened 2025-09-29
db: Added TagPullStatistics and ManifestPullStatistics tables with migrations (PROJQUAY-8414)

deshpandevlab:
Added TagPullStatistics and ManifestPullStatistics tables with migrations

Enhancement Document: [link](https://github.com/quay/enhancements/blob/main/enhancements/last-pull-metrics.md)

schema:

```
class TagPullStatistics(BaseModel):
    repository = ForeignKeyField(Repository)
    tag_name = CharField(max_length=255)
    tag_pull_count = BigIntegerField(default=0)
    last_tag_pull_date = DateTimeField(default=datetime.utcnow)
    current_manifest_digest = CharField(max_length=255, null=True)

class ManifestPullStatistics(BaseModel):
    repository = ForeignKeyField(Repository)  
    manifest_digest = CharField(max_length=255)
    manifest_pull_count = BigIntegerField(default=0)
    last_manifest_pull_date = DateTimeField(default=datetime.utcnow)
```

### Testing
- migrated database locally and manually CRUD data
- added unit tests for the data model

### DIFF
--- a/web/cypress/test/quay-db-data.txt
+++ b/web/cypress/test/quay-db-data.txt
@@ -67,6 +67,8 @@ ALTER TABLE IF EXISTS ONLY public.repositoryautoprunepolicy DROP CONSTRAINT IF E
 ALTER TABLE IF EXISTS ONLY public.repositoryautoprunepolicy DROP CONSTRAINT IF EXISTS fk_repositoryautoprunepolicy_namespace_id_user;
 ALTER TABLE IF EXISTS ONLY public.repositoryauthorizedemail DROP CONSTRAINT IF EXISTS fk_repositoryauthorizedemail_repository_id_repository;
 ALTER TABLE IF EXISTS ONLY public.repositoryactioncount DROP CONSTRAINT IF EXISTS fk_repositoryactioncount_repository_id_repository;
+ALTER TABLE IF EXISTS ONLY public.tagpullstatistics DROP CONSTRAINT IF EXISTS fk_tagpullstatistics_repository_id_repository;
+ALTER TABLE IF EXISTS ONLY public.manifestpullstatistics DROP CONSTRAINT IF EXISTS fk_manifestpullstatistics_repository_id_repository;
 ALTER TABLE IF EXISTS ONLY public.repository DROP CONSTRAINT IF EXISTS fk_repository_visibility_id_visibility;
 ALTER TABLE IF EXISTS ONLY public.repository DROP CONSTRAINT IF EXISTS fk_repository_namespace_user_id_user;
 ALTER TABLE IF EXISTS ONLY public.repository DROP CONSTRAINT IF EXISTS fk_repository_kind_id_repositorykind;


web/cypress/test/quay-db-data.txt
```diff
@@ -67,6 +67,9 @@ ALTER TABLE IF EXISTS ONLY public.repositoryautoprunepolicy DROP CONSTRAINT IF E
 ALTER TABLE IF EXISTS ONLY public.repositoryautoprunepolicy DROP CONSTRAINT IF EXISTS fk_repositoryautoprunepolicy_namespace_id_user;
 ALTER TABLE IF EXISTS ONLY public.repositoryauthorizedemail DROP CONSTRAINT IF EXISTS fk_repositoryauthorizedemail_repository_id_repository;
 ALTER TABLE IF EXISTS ONLY public.repositoryactioncount DROP CONSTRAINT IF EXISTS fk_repositoryactioncount_repository_id_repository;
+-- Drop pull statistics foreign keys first
```
harishsurf: nit: you might have to add in queries for this to drop FK and also the indexes, before you can drop table

data/database.py
```diff
@@ -1479,6 +1483,45 @@ class Meta:
...
+    current_manifest_digest = CharField(max_length=255, null=True)
+
+    class Meta:
+        database = db
+        read_only_config = read_only_config
+        indexes = (
+            # Composite unique constraint for (repository, tag_name)
+            (("repository", "tag_name"), True),
+            # Indexes for query performance
+            (("last_tag_pull_date",), False),
```
harishsurf: do we need indexes for each of the columns? we probably could get away with just keeping the `("repository", "tag_name")` index as that's the one we would heavily use when querying. Since this feature is going to be write heavy fewer indexes would significantly improve write performance

data/database.py
```diff
@@ -1479,6 +1483,45 @@ class Meta:
...
+    last_manifest_pull_date = DateTimeField(default=datetime.utcnow)
+
+    class Meta:
+        database = db
+        read_only_config = read_only_config
+        indexes = (
+            # Composite unique constraint for (repository, manifest_digest)
+            (("repository", "manifest_digest"), True),
+            # Indexes for query performance
+            (("last_manifest_pull_date",), False),
```
harishsurf: same comment for index applies here as well

data/model/pull_statistics.py
```diff
@@ -0,0 +1,236 @@
...
+                    )
+
+                    # Update existing record
+                    existing.tag_pull_count += update["pull_count"]
+                    existing.last_tag_pull_date = max(existing.last_tag_pull_date, last_pull)
+                    existing.current_manifest_digest = update["manifest_digest"]
+                    try:
+                        existing.save()
+                        rows_affected += 1
+                    except Exception as save_error:
```
harishsurf: nit: keeping it consistent with rest of the code
```suggestion
                    except Exception as save_error:
                        logger.error(f"Failed to save existing tag pull statistics record: {save_error}")
                        raise
```

data/model/pull_statistics.py
```diff
@@ -0,0 +1,236 @@
...
+    Raises:
+        PullStatisticsException: If bulk operation fails
+    """
+    if not tag_updates:
+        return 0
+
+    rows_affected = 0
+
+    with db_transaction():
+        for update in tag_updates:
```
harishsurf: This loop still does individual get/update operations in a loop rather than true bulk operations. Wondering if we can use [insert_many()](https://github.com/quay/quay/blob/master/data/model/storage.py#L412) here but we also do an upsert here, so may be not?

data/migrations/versions/7078c84d14e8_add_pull_statistics_tables.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # Create indexes for TagPullStatistics
+    op.create_index(
+        "tagpullstatistics_repository_id",
+        "tagpullstatistics",
+        ["repository_id"],
+        unique=False,
+    )
+
+    op.create_index(
```
harishsurf: same comment for indexes, also note that migration has a lot more indexes than the table defined in database.py. We usually keep both consistent

harishsurf (commented):
Overall looks good. Added some comments

data/database.py
deshpandevlab: agreed. It would be expensive to update every index on insert. Will update it. We can always add an index later

data/model/pull_statistics.py
deshpandevlab: we could actually. My plan was to add it later. But updated this PR to optimally upsert the data. The change first checks whether the record is already existing, if yes then update the record 1 by 1 if no then does the bulk insert like insert_many(). The latest commit would have this change

data/database.py
deshpandevlab: updated the indexes

data/database.py
deshpandevlab: updated the indexes

data/migrations/versions/7078c84d14e8_add_pull_statistics_tables.py
deshpandevlab: thanks for pointing out. It is fixed in the latest commit

harishsurf (dismissed):
Thanks for addressing reviews! change lgtm!

harishsurf (dismissed):

deshpandevlab:
The cypress tests are failing for the existing UI changes. This PR is ready to be merged.

@bcaton85 is helping bypass the cypress tests and merge the PR.

PR #4321, opened 2025-10-01
feat: Added Redis database and Redis flush worker (PROJQUAY-7176)

deshpandevlab:
Added Redis database and Redis flush worker

Enhancement Document: [link](https://github.com/quay/enhancements/blob/main/enhancements/last-pull-metrics.md)

Role of redis scanner vs redis worker:

Time 0: Docker pull → Redis key created
Time 5: RedisScanner reads key → returns formatted data
Time 5: Background worker processes formatted data
Time 5: Database updated successfully  
Time 5: Original Redis key deleted

### Testing

accessed redis locally:
- `make local-dev-up`
-  `docker exec -it quay-redis redis-cli ping`

Tested config tool locally

```
(string) (len=5) "Redis": (*redis.RedisFieldGroup)({
        BuildlogsRedis: (*redis.BuildlogsRedisStruct)({
                Password: (string) "",
                Port: (int) 6379,
                Host: (string) (len=10) "quay-redis",
                Ssl: (bool) false
        }),
        UserEventsRedis: (*redis.UserEventsRedisStruct)({
                Password: (string) "",
                Port: (int) 6379,
                Host: (string) (len=10) "quay-redis",
                Ssl: (bool) false
        }),
        PullMetricsRedis: (*redis.PullMetricsRedisStruct)({
                Password: (string) "",
                Port: (int) 6379,
                Host: (string) (len=10) "quay-redis",
                Ssl: (bool) false,
                Db: (int) 1
        })
})
```


workers/pullstatsredisflushworker.py
```diff
@@ -0,0 +1,326 @@
...
+
+from app import app
+from workers.worker import Worker
+
+logger = logging.getLogger(__name__)
+
+# Configuration constants
+POLL_PERIOD = app.config.get("REDIS_FLUSH_WORKER_POLL_PERIOD", 300)  # 5 minutes
+BATCH_SIZE = app.config.get("REDIS_FLUSH_WORKER_BATCH_SIZE", 1000)
+REDIS_SCAN_COUNT = app.config.get("REDIS_FLUSH_WORKER_SCAN_COUNT", 100)
```
deshpandevlab, line 37: todo: we should make this configurable for customers

config-tool/pkg/lib/fieldgroups/redis/redis.go
```diff
@@ -28,6 +29,15 @@ type BuildlogsRedisStruct struct {
...
 }
 
+// PullMetricsRedisStruct represents the PullMetricsRedisStruct config fields
+type PullMetricsRedisStruct struct {
+	Password string `default:"" validate:"" json:"password,omitempty" yaml:"password,omitempty"`
+	Port     int    `default:"" validate:"" json:"port,omitempty" yaml:"port,omitempty"`
+	Host     string `default:"" validate:"" json:"host,omitempty" yaml:"host,omitempty"`
+	Ssl      bool   `default:"false" validate:"" json:"ssl,omitempty" yaml:"ssl,omitempty"`
+	Db       int    `default:"1" validate:"" json:"db,omitempty" yaml:"db,omitempty"`
+}
```
cubismod, line 39: Could all of these be consolidated into one type? They're pretty much all duplicates

workers/pullstatsredisflushworker.py
```diff
@@ -0,0 +1,327 @@
...
+            keys: List of Redis keys to process
+
+        Returns:
+            Tuple of (tag_updates, manifest_updates, processed_keys)
+        """
+        tag_updates = []
+        manifest_updates = []
+        processed_keys = set()
+
+        for key in keys:
```
cubismod, line 210: `SCAN` in the earlier function can possibly return duplicate values, see https://redis.io/docs/latest/commands/scan/#scan-guarantees

so I think you should de-duplicate the keys with a set before iterating over them otherwise you might get multiple manifest/tag updates for the same keys

config-tool/pkg/lib/fieldgroups/redis/redis.go
deshpandevlab: yeah didn't want to disturb the existing structure. @harishsurf pointed out that config tool is deprecated. Trying to get team's opinion if we need the config tool change

workers/pullstatsredisflushworker.py
deshpandevlab: > A given element may be returned multiple times. It is up to the application to handle the case of duplicated elements

great catch. will add a dedupe may be a set on keys before iterating

config-tool/pkg/lib/fieldgroups/redis/redis.go
cubismod: gotcha, no worries if you don't wanna disrupt what works already

workers/pullstatsredisflushworker.py
deshpandevlab: @cubismod thanks, Addressed the concern in the latest commit

config-tool/pkg/lib/fieldgroups/redis/redis.go
deshpandevlab: keeping the current flow as it is. will add a todo item to clean the config tool

lechuk47:
is the config-tool binary file a mistake?

deshpandevlab:
> is the config-tool binary file a mistake?

yeah! removed it in latest commit

workers/pullstatsredisflushworker.py
```diff
@@ -0,0 +1,432 @@
...
+        super(RedisFlushWorker, self).__init__()
+        self.redis_client = None
+        self._initialize_redis_client()
+        self.add_operation(self._flush_pull_metrics, POLL_PERIOD)
+
+    def _initialize_redis_client(self):
+        """Initialize Redis client for pull metrics."""
+        try:
+            # Get Redis configuration
+            redis_host = app.config.get("PULL_METRICS_REDIS_HOST", "localhost")
```
harishsurf: as per schema.py, `host`, `port` etc are nested properties of `PULL_METRICS_REDIS`.`PULL_METRICS_REDIS_HOST` wouldn't exist  in `app.config`

You can access them like: `redis_config = app.config.get("PULL_METRICS_REDIS", {})` - here `redis_config` would have all the nested properties

workers/pullstatsredisflushworker.py
```diff
@@ -0,0 +1,432 @@
...
+            redis_port = app.config.get("PULL_METRICS_REDIS_PORT", 6379)
+            redis_db = app.config.get("PULL_METRICS_REDIS_DB", 0)
```
harishsurf: the default would be db=1? as defined in config-tool [here](https://github.com/quay/quay/pull/4321/files#diff-524fe77e47f0480dc771e22e7a23be7677d7418fc0a006c8cc1d3bf62e96d9d3R38)

config-tool/pkg/lib/fieldgroups/redis/redis_validator.go
harishsurf, line 73: unrelated to your change:  I think this should be `fmt.Sprintf("%d", fg.UserEventsRedis.Port)`

config-tool/pkg/lib/fieldgroups/redis/redis_validator.go
```diff
@@ -32,7 +32,18 @@ func (fg *RedisFieldGroup) Validate(opts shared.Options) []shared.ValidationErro
 	}
 
 	// Check for user events host
-	if ok, err := shared.ValidateRequiredString(fg.BuildlogsRedis.Host, "USER_EVENTS_REDIS.HOST", "Redis"); !ok {
+	if ok, err := shared.ValidateRequiredString(fg.UserEventsRedis.Host, "USER_EVENTS_REDIS.HOST", "Redis"); !ok {
+		errors = append(errors, err)
+	}
+
+	// Check for pull metrics config
+	if ok, err := shared.ValidateRequiredObject(fg.PullMetricsRedis, "PULL_METRICS_REDIS", "Redis"); !ok {
```
harishsurf: This implies config-tool validator requires `PULL_METRICS_REDIS` to be present. Instead we should only validate if pull metrics feature is provided in config.yaml, something like:
```
if fg.PullMetricsRedis != nil {
    if ok, err := shared.ValidateRequiredString(fg.PullMetricsRedis.Host, ...); !ok {
        errors = append(errors, err)
    }
    // ... rest of validation
}
```

workers/pullstatsredisflushworker.py
```diff
@@ -0,0 +1,432 @@
...
+                return False
+
+            if last_pull_timestamp < 0:
+                logger.debug(
+                    f"RedisFlushWorker: Key {key} has invalid timestamp: {last_pull_timestamp}"
+                )
+                return False
+
+            manifest_digest = metrics_data.get("manifest_digest", "")
+            if not manifest_digest or not manifest_digest.startswith("sha256:"):
```
harishsurf: we have `parse_digest()` helper method for validating digest. [Example usage](https://github.com/quay/quay/blob/master/image/oci/manifest.py#L515-L520)

workers/pullstatsredisflushworker.py
```diff
@@ -0,0 +1,432 @@
...
+                decode_responses=True,
+                socket_connect_timeout=redis_connection_timeout,
+                socket_timeout=redis_connection_timeout,
+            )
+
+            # Test connection
+            self.redis_client.ping()
+            logger.info("RedisFlushWorker: Initialized Redis client for pull metrics")
+
+        except Exception as e:
```
harishsurf, line 88: nit: Generic exception catching hides specific errors. something like: 
```
except redis.ConnectionError as e:
      logger.warning(f"RedisFlushWorker: Redis connection failed (will retry): {e}")
      self.redis_client = None
  except redis.AuthenticationError as e:
      logger.critical(f"RedisFlushWorker: Redis authentication failed - check config: {e}")
      self.redis_client = None
      raise  # Fatal - worker should not start with bad credentials
  except redis.RedisError as e:
      logger.error(f"RedisFlushWorker: Redis initialization error: {e}")
      self.redis_client = None
```

workers/pullstatsredisflushworker.py
```diff
@@ -0,0 +1,432 @@
...
+
+            logger.info(f"RedisFlushWorker: Processing {len(pull_event_keys)} Redis keys")
+
+            # Process keys and aggregate data
+            (
+                tag_updates,
+                manifest_updates,
+                cleanable_keys,
+                database_dependent_keys,
+            ) = self._process_redis_events(pull_event_keys)
```
harishsurf, line 117: Note: we should define a TTL for keys to avoid risk of redis filling up with unprocessed keys when writes to db don't succeed

workers/pullstatsredisflushworker.py
```diff
@@ -0,0 +1,432 @@
...
+
+                # Break if we've scanned through all keys
+                if cursor == 0:
+                    break
+
+            # Convert set back to list and limit results
+            keys_list = list(keys_set)
+            return keys_list[:limit]  # Ensure we don't exceed the limit
+
+        except Exception as e:
```
harishsurf, line 187: nit: similar generic exception comment

config-tool/pkg/lib/fieldgroups/redis/redis_validator.go
deshpandevlab: PULL_METRICS_REDIS is independent of the feature flag imo. Customer can enable the feature while not configuring the Redis at all. We should add check for both

workers/pullstatsredisflushworker.py
deshpandevlab: yep! makes sense

workers/pullstatsredisflushworker.py
deshpandevlab: Redis will be our single source of truth. We can not wipe the data before we record it to DB in order to prevent data loss. There will be metrics to track Redus health and we can discuss whether we can to increase Redis capacity it the alerts are red.

harishsurf: yeah, the TTL is for quay.io scale. For downstream it should be fine, but good to call it out in docs about the risk of redis filling up in case writes to primary db fails

config-tool/pkg/lib/fieldgroups/redis/redis_validator.go
harishsurf: discussed this over call. decided to add the check

config-tool/pkg/lib/fieldgroups/redis/redis_validator.go
deshpandevlab: there are a few config changes required. I will bundle them into a separate PR

workers/pullstatsredisflushworker.py
deshpandevlab: wondering if this exception handling will be excessive. Adding them as of now, we can remove it in future.

.github/workflows/CI.yaml
deshpandevlab, line 1: i think this file should be rebased from master

```
git diff origin/master --name-only
.github/workflows/oci-distribution-spec.yaml
config-tool/pkg/lib/fieldgroups/redis/redis.go
config-tool/pkg/lib/fieldgroups/redis/redis_fields.go
config-tool/pkg/lib/fieldgroups/redis/redis_test.go
config-tool/pkg/lib/fieldgroups/redis/redis_validator.go
features/__init__.pyi
util/config/schema.py
web/cypress/test/quay-db-data.txt
workers/pullstatsredisflushworker.py
workers/test/test_pullstatsredisflushworker.py
```

workers/pullstatsredisflushworker.py
deshpandevlab: thanks for pointing out. Addressed the comment in the latest commit

deshpandevlab:
the test failures: 
- test_reconciliationworker.py is existing test failure. Unrelated to this change.
- Cypress tests failures are unrelated to this change

harishsurf (approved):
thanks for addressing reviews! 
only the config.py change I called out above is missing, feel free to follow up in your other PR

PR #4323, opened 2025-10-03
sentry: filtering out invalid token format errors (PROJQUAY-9436)

bcaton85:


util/saas/exceptionlog.py
```diff
@@ -6,6 +6,36 @@
 from sentry_sdk.integrations.sqlalchemy import SqlalchemyIntegration
 from sentry_sdk.integrations.stdlib import StdlibIntegration
 
+# Note: Imported inside functions to avoid circular imports where possible
+
+
+def _sentry_before_send_ignore_known(ex_event, hint):
```
deshpandevlab, line 12: what is hint? is it a sentry context?

util/saas/exceptionlog.py
```diff
@@ -6,6 +6,36 @@
...
+    try:
+        exc_info = hint.get("exc_info") if isinstance(hint, dict) else None
+        if exc_info is not None and len(exc_info) >= 1:
+            exc_type = exc_info[0]
+            # Compare by class name to avoid import-time cycles; fall back to isinstance if safe
+            name = getattr(exc_type, "__name__", "")
+            if name in {"InvalidBearerTokenException", "InvalidJWTException"}:
+                return None
+        # Also check mechanism-embedded exception if available
+        mechanism = (ex_event or {}).get("exception", {}).get("values", [])
```
deshpandevlab, line 28: minor: we could get chain like

```
ex_event.get("exception", {}).get("values") or []

```

and then run a check on values being empty to avoid empty values

util/saas/exceptionlog.py
bcaton85: it seems to be the [original error](https://docs.sentry.io/platforms/python/configuration/filtering/#event-hints) most often based on the documentation

util/saas/exceptionlog.py
bcaton85: sorry not understanding, isn't that what the original is doing? it seems like that would fail if `ex_event` is possibly undefined

deshpandevlab: same functionality just easier to read.

```
val = ex_event.get("exception", {}).get("values") or []
if val:
    type = val[0].get("type")
    if type in {"InvalidBearerTokenException", "InvalidJWTException"}:
        return None
```

it won't change the functionality

deshpandevlab (approved):

PR #4402, opened 2025-10-27
fix: Quay new UI global readonly superuser can't see all organizations and image repos (PROJQUAY-6882)

LiZhang19817:
With this fix, now **Quay global readonly superuser** can see all organizations and image repos of other normal users
image


LiZhang19817:
@HammerMeetNail Pls help review this fix.

PR #4444, opened 2025-11-03
uibug: only allow super user to configure the Quota on Super user admin Panel (PROJQUAY-9407)

deshpandevlab:
uibug: only allow super user to configure the Quota on Super user admin Panel

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -2390,21 +2390,6 @@
       "integrity": "sha512-Kvp459HrV2FEJ1CAsi1Ku+MY3kasH19TFykTz2xWmMeq6bk2NU3XXvfJ+Q61m0xktWwt+1HSYf3JZsTms3aRJg==",
       "peer": true
     },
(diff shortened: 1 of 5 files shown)


web/package-lock.json
harishsurf, line 1: this file shouldn't have changed, can you take a look?

harishsurf (changes requested):
1. the Storage quota size dropdown remains disabled when configuring quota for a user record type. This only happens in new UI (image attached below)
2. When i create a new user, the configure quota option doesn't show up for this user. 

I had below config for both: 

```
FEATURE_QUOTA_MANAGEMENT: true
FEATURE_SUPER_USERS: true
FEATURE_SUPERUSERS_FULL_ACCESS: true
```

Image

web/package-lock.json
deshpandevlab: ah! local testing changes got committed. thanks will remove in the latest commit

jbpratt:
@coderabbit full review

PR #4791, opened 2026-01-08
feat(web): add column management to list tables (PROJQUAY-10133)

jbpratt:
Add ManageColumnsModal component that allows users to show/hide
table columns with preferences persisted to localStorage.

Changes:
- Add ManageColumnsModal and ManageColumnsButton components
- Add useColumnManagement hook for localStorage persistence
- Integrate column management into RepositoriesList table
- Integrate column management into OrganizationsList table
- Add Playwright e2e tests for column management on both tables

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

image
image

### DIFF
--- a/web/src/components/ManageColumns/ManageColumnsModal.tsx
+++ b/web/src/components/ManageColumns/ManageColumnsModal.tsx
@@ -1,0 +1,209 @@
+import React, {useState, useMemo, useCallback, useEffect} from 'react';
+import {
+  Modal,
+  ModalVariant,
+  Button,
+  Alert,
+  Split,
+  SplitItem,
+  DataList,
+  DataListItem,
+  DataListItemRow,
+  DataListItemCells,
+  DataListCell,
+  DataListCheck,
+  Title,
+  Divider,
+} from '@patternfly/react-core';
+import type {ColumnConfig, ManageColumnsModalProps} from './types';
+
+const MAX_COLUMNS = 9;
+
+export const ManageColumnsModal: React.FC<ManageColumnsModalProps> = ({
+  isOpen,
+  onClose,
+  columns,
+  onSave,
+  maxVisibleColumns = MAX_COLUMNS,
+  title = 'Manage columns',
+}) => {
+  // Local state for editing (only committed on Save)
+  const [editedColumns, setEditedColumns] = useState<ColumnConfig[]>(columns);
+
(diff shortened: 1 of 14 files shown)


web/src/components/ManageColumns/ManageColumnsModal.tsx
```diff
@@ -0,0 +1,209 @@
...
+              </DataListItem>
+            );
+          })}
+        </DataList>
+      )}
+    </div>
+  );
+
+  return (
+    <Modal
```
harishsurf, line 154: When the modal opens, I see below accessibility warning in the browser console: 
```
Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: <button.pf-v5-c-button pf-m-plain>
Ancestor with aria-hidden: <div#root> 
```

web/src/routes/OrganizationsList/ColumnNames.ts
```diff
@@ -10,4 +12,77 @@ const ColumnNames = {
...
+  {
+    id: 'name',
+    title: ColumnNames.name,
+    isVisible: true,
+    isDefault: true,
+    isDisabled: true, // Name column cannot be hidden
+    sortIndex: 0,
+  },
+  {
+    id: 'adminEmail',
```
harishsurf, line 31: some of these column names are specific to superuser. When I log in as a normal user, these non-existent columns might be misleading

harishsurf (commented):
tested out, looks good. Just a few minor comments

web/src/routes/OrganizationsList/ColumnNames.ts
jbpratt: Woops, thought I fixed them - they shouldn't show for non superusers

jbpratt: They are properly hidden! 

image

jbpratt:
holding until pf6 to use https://www.patternfly.org/component-groups/helpers/column-management-modal/

PR #4797, opened 2026-01-08
fix(storage): disable MD5 content validation in FIPS mode (PROJQUAY-7743)

jbpratt:
When FEATURE_FIPS is enabled, Azure SDK's stage_block() method fails
because it uses MD5 for content validation, which is blocked by OpenSSL
in FIPS mode. This change disables MD5 validation when FIPS is enabled.

HTTPS transport already provides cryptographically strong integrity
verification via TLS, making the application-layer MD5 check redundant.

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/storage/azurestorage.py
+++ b/storage/azurestorage.py
@@ -24,6 +24,7 @@ from azure.storage.blob import (
     generate_blob_sas,
 )
 
+import features
 from storage.basestorage import BaseStorage
 from util.registry.filelike import READ_UNTIL_END, LimitingStream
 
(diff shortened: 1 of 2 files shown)


storage/azurestorage.py
```diff
@@ -264,7 +265,12 @@ def stream_upload_chunk(self, uuid, offset, length, in_fp, storage_metadata, con
 
                 try:
                     # TODO (kleesc): Look at doing this multithreaded
-                    self._blob(upload_blob_path).stage_block(block_id, buf, validate_content=True)
+                    # In FIPS mode, disable MD5 content validation as MD5 is not FIPS-compliant.
+                    # HTTPS transport already provides integrity verification.
+                    validate_content = not getattr(features, "FIPS", False)
+                    self._blob(upload_blob_path).stage_block(
+                        block_id, buf, validate_content=validate_content
+                    )
```
ibazulic, line 273: The problem is not in SSL, the hash check is to verify that the blob has been properly written. MD5 hash, just as any other validation, will differ between badly uploaded blob and a properly uploaded blob. This checks if no corruption happened during upload. Maybe not use MD5 but something that is FIPS compliant? 

https://learn.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.blobclient?view=azure-python#azure-storage-blob-blobclient-stage-block

PR #4872, opened 2026-01-15
chore: Deprecate TLS 1.0 and TLS 1.1 support (PROJQUAY-9620)

deshpandevlab:
- Update IsValidTLS() to only accept TLSv1.2 and TLSv1.3 for PostgreSQL connections
- Update error messages to reflect supported TLS versions
- Update SSL_PROTOCOLS examples and descriptions in schema documentation

TLSv1 and TLSv1.1 are deprecated due to known security vulnerabilities.


harishsurf (commented):
[this method](https://github.com/quay/quay/blob/master/util/security/ssl.py#L51) still seems to use TLS1.0

deshpandevlab:
> [this method](https://github.com/quay/quay/blob/master/util/security/ssl.py#L51) still seems to use TLS1.0

I think this shouldnt validate the actual network connection but yes it needs to be fixed. Added the change. Thanks for catching this.

PR #5262, opened 2026-03-03
QUAYIO-1583: cache: cache child manifest queries

kleesc:


data/registry_model/registry_oci_model.py
```diff
@@ -108,20 +108,27 @@ def _get_expiry_label_for_manifest(self, manifest):
...
+                    child = database.Manifest.get(id=child_manifest_id)
+                    child_manifest = Manifest.for_manifest(child, self._legacy_image_id_handler)
+                    child_labels = {
+                        label.key: label.value
+                        for label in self.list_manifest_labels(child_manifest, key_prefix="quay")
+                    }
+                    child_label_dicts.append(child_labels)
+                except database.Manifest.DoesNotExist:
+                    logger.warning(f"Child manifest {child_manifest_id} not found")
+                    continue
```
cubismod, line 136: Echoing coderabbit's comment here: https://github.com/quay/quay/pull/5262#pullrequestreview-3882851300

This is a change in behavior that can result in an incorrect expiration time if one child manifest is missing. I'd suggest returning `None` rather than continuing.

kleesc: Keeping continue is the closest the the previous behavior since it's doing the intersection of all labels. We were previously using `select()`, which would return the manifest objects directly, or an empty list if not found. This change does it by getting the ids first. In theory, this exception should never be raised, but in the case it does, I think we'd still want the labels from the other child manifests.

data/model/oci/tag.py
```diff
@@ -128,6 +128,38 @@ def get_child_manifests(repo_id: int, manifest_id: int):
...
+        model_cache: Cache instance to use for caching
+
+    Returns:
+        List of child manifest IDs
+    """
+    from data.cache import cache_key
+
+    def load_child_manifests():
+        # Query and convert to list of IDs
+        children = get_child_manifests(repo_id, manifest_id)
```
Sunandadadi, line 150: the child manifests query is a pretty simple select on an indexed foreign key. I think it should be a cheap query.

data/cache/cache_key.py
```diff
@@ -93,3 +93,14 @@ def for_manifest_referrers(repository_id, manifest_digest, cache_config):
...
+
+
+def for_manifest_children(repository_id, manifest_id, cache_config):
+    """
+    Returns a cache key for the child manifests of a manifest list.
+
+    Since manifests are immutable (content-addressed), child relationships
+    never change once created, so this can be cached with a long TTL.
+    """
+    cache_ttl = cache_config.get("manifest_children_cache_ttl", "3600s")  # 1 hour default
```
Sunandadadi, line 105: Since child manifests are immutable, does it make sense to invalidate?

data/registry_model/registry_oci_model.py
cubismod: Got it, thanks for explaining. Okay by me then.

data/cache/cache_key.py
cubismod: Do you mean setting an indefinite TTL?

data/registry_model/registry_oci_model.py
```diff
@@ -108,20 +108,27 @@ def _get_expiry_label_for_manifest(self, manifest):
...
-                child_labels = {
-                    label.key: label.value
-                    for label in self.list_manifest_labels(child_manifest, key_prefix="quay")
-                }
-                child_label_dicts.append(child_labels)
+            for child_manifest_id in child_manifest_ids:
+                # Load the child manifest object
+                try:
+                    child = database.Manifest.get(id=child_manifest_id)
+                    child_manifest = Manifest.for_manifest(child, self._legacy_image_id_handler)
```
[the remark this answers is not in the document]
Sunandadadi: I think this makes sense. what do you think about either bulk look up or caching here?

data/cache/cache_key.py
Sunandadadi: sorry please strike my comment.

PR #5347, opened 2026-03-10
PROJQUAY-8440: fix(proxy): serve cached images when upstream registry is unavailable

jbpratt:
## Summary

Fixes [PROJQUAY-8440](https://issues.redhat.com/browse/PROJQUAY-8440): Pulling cached images from Quay's proxy cache fails when the upstream registry is unavailable, even though the images are already cached locally.

### Root Cause

When a pull request comes in for a proxy cache organization:

1. `inject_registry_model()` creates `ProxyModel()` which creates `Proxy()`
2. `Proxy.__init__()` immediately calls `_authorize()`, which makes an HTTPS request to the upstream registry's `/v2/` endpoint
3. If the upstream is unreachable, `UpstreamRegistryError` is raised **during model construction** — before the manifest/tag lookup logic ever executes
4. The existing fallback-to-cache logic in `get_repo_tag()` and `lookup_manifest_by_digest()` is **never reached**

Additionally, no request timeouts were configured, so connections to unreachable upstreams would hang for 60+ seconds (OS TCP timeout), causing nginx 504 Gateway Timeout errors.

### Fix

- **Defer authorization**: Moved `_authorize()` from `Proxy.__init__()` to a lazy `_ensure_authorized()` method called on first request. This allows `ProxyModel` to be constructed without network calls, so the existing fallback logic can catch `UpstreamRegistryError` and return cached tags. Validation mode (`validation=True`) retains eager auth for config validation.
- **Add request timeouts**: Added default `(5s connect, 30s read)` timeout to all upstream HTTP requests via `_safe_request()`, preventing indefinite blocking.

### Files Changed

| File | Change |
|------|--------|
| `proxy/__init__.py` | Defer auth to first request, add request timeouts |
| `proxy/test_proxy.py` | Update existing tests for deferred auth, add 3 new tests |
| `endpoints/v2/test/test_manifest_pullthru.py` | Add integration test for cached pull with upstream down |

## Test plan

- [x] All 128 existing proxy cache tests pass (proxy, pullthru, registry_proxy_model)
- [x] New test: `test_deferred_auth_no_network_on_construction` — verifies no network calls during `Proxy()` construction
- [x] New test: `test_validation_mode_auth_is_eager` — verifies `Proxy(validation=True)` still authorizes eagerly
- [x] New test: `test_request_timeout_applied` — verifies default timeout is applied to requests
- [x] New test: `test_returns_cached_tag_when_upstream_is_down` — end-to-end test that caches a manifest, then verifies it's returned when upstream is unreachable

### DIFF
--- a/data/registry_model/registry_proxy_model.py
+++ b/data/registry_model/registry_proxy_model.py
@@ -66,7 +66,7 @@ from image.oci import OCI_IMAGE_INDEX_CONTENT_TYPE, OCI_IMAGE_MANIFEST_CONTENT_T
 from image.shared import ManifestException
 from image.shared.interfaces import ManifestInterface
 from image.shared.schemas import parse_manifest_from_bytes
-from proxy import Proxy, UpstreamRegistryError
+from proxy import Proxy, UpstreamAuthError, UpstreamRegistryError
 from util.bytes import Bytes
 
 logger = logging.getLogger(__name__)
@@ -291,6 +291,8 @@ class ProxyModel(OCIModel):
             )
         except ManifestDoesNotExist:
             raise
+        except UpstreamAuthError:
+            raise ManifestDoesNotExist("upstream authentication failed")
         except UpstreamRegistryError:
             # when the upstream fetch fails, we only return the tag if
             # it isn't yet expired. note that we don't bump the tag's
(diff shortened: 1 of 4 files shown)


data/registry_model/registry_proxy_model.py
shaonrh, line 1: This file has missing codecov, not sure why it's not showing up in CI. @jbpratt is that expected?

shaonrh (approved):

PR #5385, opened 2026-03-12
Go rewrite db

harishsurf:


tools/export_schema_ddl.py
```diff
@@ -0,0 +1,387 @@
+#!/usr/bin/env python3
```
deshpandevlab, line 1: can we have tests added for the export_schema_ddl? this will boost our confidence in the edge cases

deshpandevlab:
The pipeline has four phases. python models -> the sqlAlchemy metadata -> SQLite DDL -> slqc. The transtion between the phases is just guarded by schema_sync parity check. A bug or semantic mismatch in any layer produces silent incorrectness in our Go code. The drift/parity check only verifies textual equality of the output SQL files. It cannot catch semantic issues/miggration inconsistencies or race conditions of schema migrations eg an update/inconsistency in the python → SQLAlchemy translation.

One way to address it could be a drift check at every phase transition

PR #5413, opened 2026-03-16
poc: add database migration rollback POC and tooling (QUAYIO-1605)

deshpandevlab:
- Add dummy migration (poc_rollback_demo table) to test downgrade flow
- Add tools/db_rollback.py automation script with dry-run, status, history, and safe downgrade with pre-flight checks
- Add database_rollback_sop.md documenting manual and automated rollback procedures for stage/prod environments
- Fix pre-commit config: rename invalid stage 'pre-commit' to 'commit'

### DIFF
--- a/agent_docs/database_rollback_sop.md
+++ b/agent_docs/database_rollback_sop.md
@@ -1,0 +1,275 @@
+# Database Migration Rollback — Standard Operating Procedure
+
+## Overview
+
+This document covers how to safely roll back (downgrade) an Alembic database migration in Quay. It was created as part of a POC using a dummy `poc_rollback_demo` table and the `tools/db_rollback.py` automation script.
+
+## Table of Contents
+
+1. [Background](#background)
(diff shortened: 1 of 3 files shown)


ibazulic:
Very quick question: how does this work for migrations that remove tables (and therefore data from those tables) from Quay? Let's say you remove a table in 3.17 that Quay code in 3.16 requires to function. A cleanup, for instance. It doesn't even have to be a deletion, it can be an alteration of an existing table where data set changes. One Quay version requires one data set, the other requires a different data set and they both use the same table to store data.

kleesc:
This will work for simple rollbacks, but becomes harder to enforce manually when you have more involved data migrations. IMO this would fit more into something similar to the quay-operator. You would need the existing migration status and do any validation, backward compatibility checks, etc, ... And rolling back is sometimes not an option (e.g whenever we fully cutover to a new data model)

PR #5526, opened 2026-03-24
NO-ISSUE: fix: prevent deadlock in test_has_garbage under parallel execution (PROJQUAY-11089)

deshpandevlab:
The RetryOperationalError._execute_sql_original retry logic was calling self.close() after catching OperationalError, but this fails when its inside a transaction (e.g. the initialized_db test fixture's savepoint wrapper). Masking the real error with "Attempting to close database while transaction is open". This vchange skip the `close-and-reconnect` path when in_transaction() is true, as reconnecting mid-transaction is never valid. We now use GC per repo instead.

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -285,6 +285,9 @@ class RetryOperationalError(object):
                 # https://github.com/PyMySQL/PyMySQL/blob/main/pymysql/connections.py#L1354
                 raise
 
+            if self.in_transaction():
+                raise
+
             if not self.is_closed():
                 self.close()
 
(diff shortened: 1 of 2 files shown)


shaonrh (approved):

ibazulic:
IMHO we should enable Redis with tests instead of changing tests. The current suite seems to pass without issues locally when Redis is available and locking does what it's designed to do. What would it take to try and test with Redis completely enabled for the full test suite? (having Redis is also  closer to real production environments where we expect it).

PR #5547, opened 2026-03-26
NO-ISSUE: feat(secscan): add V4SecurityScanner V2 with SELECT FOR UPDATE SKIP LOCKED

lechuk47:
## Summary

This PR introduces `V4SecurityScanner2`, an alternative security scanner implementation that uses database-level row locking (`SELECT FOR UPDATE SKIP LOCKED`) for distributed coordination between worker pods.

**Enable with:** `SECURITY_SCANNER_V4_MODEL_VERSION: 2`

---

### 1. Distributed Coordination via `SELECT FOR UPDATE SKIP LOCKED`
- Workers atomically claim manifests by locking Manifest rows in the database
- `SKIP LOCKED` ensures workers get disjoint sets — zero duplicate work
- Locks are held only during the claim transaction (SELECT + ManifestSecurityStatus update), not during Clair API calls

### 2. Concurrent Clair API Calls
- Uses `ThreadPoolExecutor` with configurable concurrency (default: 4 workers)
- Phases: Filter manifests (serial) → Call Clair API (parallel) → Write results (serial)
- Significant throughput improvement over V1's serial API calls

### 3. Priority-Ordered Processing
- High priority (always run): `not_indexed` → `stale_in_progress` → `failed`
- Low priority (capacity-permitting): `needs_reindex`
- Chunk-based claiming within a total `batch_size` cap

### 4. Crash Recovery
- `stale_in_progress_query` reclaims manifests stuck `IN_PROGRESS` for > 30 minutes (configurable via `SECURITY_SCANNER_V4_IN_PROGRESS_TIMEOUT`)
- V1 has no such recovery mechanism

### 5. Enhanced Observability
- Prometheus metrics with labels:
  - `secscan_manifests_claimed{query_type}` — tracks claims by query type
  - `secscan_manifests_skipped{reason}` — tracks skipped manifests (manifest_list, no_layers, artifact)
  - `secscan_index_errors{error_type}` — tracks Clair API errors
- Detailed logging at key phases (claim, batch start/finish, cycle start/finish)

---

## Why V2 is Better Than V1

| Aspect | V1 | V2 |
|---|---|---|
| **Coordination** | Random block allocation + optional Redis lock + best-effort preemption checks | Database-level `SELECT FOR UPDATE SKIP LOCKED` |
| **Duplicate work** | Frequent — multiple pods process the same manifests | Zero — workers get disjoint sets |
| **Clair API calls** | Serial (one at a time) | Concurrent (`ThreadPoolExecutor`) |
| **Crash recovery** | None | `stale_in_progress_query` after 30 minutes |
| **State management** | `ScanToken` tracked across calls | Stateless — no token needed |
| **Metrics** | Basic (`secscan_result_duration`) | Labeled metrics by query type, reason, error type |

**Performance impact:** In multi-pod deployments, V2 eliminates the duplicate work that V1 suffers from. The concurrent Clair calls provide additional throughput gains.

---

### What gets locked?
- **Manifest table rows** (not `ManifestSecurityStatus`)
- Only the rows returned by queries (`_not_indexed_query`, `_stale_in_progress_query`, etc.)

### Lock duration
- Locks are held **only during the claim transaction**:
  ```python
  with db_transaction():
      candidates = db_for_update(query, skip_locked=SKIP_LOCKED)  # Lock acquired
      _mark_batch_in_progress(candidates)                        # Update ManifestSecurityStatus
      # Transaction commits → locks released
  
  # Clair API calls happen here (locks already released)
  _index_batch_concurrent(candidates)
  ```
- Duration: **milliseconds** (just a SELECT + UPDATE/INSERT, no external API calls)

### Impact on other operations

**Garbage Collection (GC):**
- GC does **not** use `FOR UPDATE` on Manifest rows (only on Tag rows)
- GC's plain `SELECT` and `DELETE` operations:
  - Regular `SELECT` — not blocked by row locks (PostgreSQL MVCC)
  - `DELETE` — would block if lock is held, but only for milliseconds

**No existing contention:**
- Grep shows no other code in the codebase does `FOR UPDATE` on Manifest rows
- No deadlock risk (queries are ordered by `Manifest.id`, and `SKIP LOCKED` prevents waiting)

**Worst case scenario:**
- A GC delete on a manifest blocks for milliseconds while a worker holds the lock
- If GC deletes a manifest mid-indexing, the scan result is silently discarded (manifest was garbage anyway)

---

### Supported
- **PostgreSQL 9.5+** (native `SKIP LOCKED` support)
- **MySQL 8.0+** (native `SKIP LOCKED` support)

### Unsupported
- **SQLite** — `SKIP LOCKED` is a no-op (via `null_for_update` in `database.py`)
- **MySQL 5.7** — `SKIP LOCKED` syntax error

### Fallback Behavior
If `SECURITY_SCANNER_V4_MODEL_VERSION=2` is set with an unsupported database:
```
ERROR: SECURITY_SCANNER_V4_MODEL_VERSION=2 requires PostgreSQL or MySQL 8+. 
       Detected database driver 'sqlite'. Falling back to model version 1.
```
The proxy (`SecurityScannerModelProxy`) automatically uses V1 instead.

---

## Configuration

| Config Key | Default | Description |
|---|---|---|
| `SECURITY_SCANNER_V4_MODEL_VERSION` | `1` | Set to `2` to enable V2 |
| `SECURITY_SCANNER_V4_BATCH_SIZE` | `100` | Max manifests per indexing cycle |
| `SECURITY_SCANNER_V4_CONCURRENCY` | `4` | ThreadPoolExecutor workers for Clair API calls |
| `SECURITY_SCANNER_V4_IN_PROGRESS_TIMEOUT` | `1800` (30 min) | Stale IN_PROGRESS threshold |
| `SECURITY_SCANNER_V4_REINDEX_THRESHOLD` | `86400` (1 day) | Minimum time before reindexing |

---

## Testing

- [x] Syntax check: `python3 -m py_compile data/secscan_model/secscan_v4_model_v2.py`
- [x] Import check: `python3 -c "from data.secscan_model.secscan_v4_model_v2 import V4SecurityScanner2"`
- [ ] Integration test with PostgreSQL (requires live DB)
- [ ] Performance test: multi-pod scenario comparing v1 vs v2 duplicate work

---

## Future Work

- [ ] Backfill tests for v2 with PostgreSQL test DB
- [ ] Prometheus dashboard for v2 metrics
- [ ] Consider MySQL 5.7 deprecation (update CI to MySQL 8)

### DIFF
--- a/data/secscan_model/secscan_v4_model_v2.py
+++ b/data/secscan_model/secscan_v4_model_v2.py
@@ -1,6 +1,6 @@
 import logging
 from concurrent.futures import ThreadPoolExecutor, as_completed
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 from peewee import IntegrityError, fn
 
@@ -151,7 +151,7 @@ class V4SecurityScanner2(SecurityScannerInterface):
 
     def _mark_batch_in_progress(self, candidates):
         """Mark claimed manifests as IN_PROGRESS via upsert."""
-        now = datetime.utcnow()
+        now = datetime.now(timezone.utc)
         candidate_ids = [c.id for c in candidates]
 
         # Update existing ManifestSecurityStatus rows
@@ -208,7 +208,7 @@ class V4SecurityScanner2(SecurityScannerInterface):
 
     def _upsert_status(self, manifest_id, repo_id, index_status, indexer_hash, error_json):
         """Upsert ManifestSecurityStatus (update-first, create-on-miss)."""
-        now = datetime.utcnow()
+        now = datetime.now(timezone.utc)
 
         rows_updated = (
             ManifestSecurityStatus.update(
@@ -505,13 +505,13 @@ class V4SecurityScanner2(SecurityScannerInterface):
             chunk_size,
         )
 
-        reindex_threshold = datetime.utcnow() - timedelta(
+        reindex_threshold = datetime.now(timezone.utc) - timedelta(
             seconds=self.app.config.get(
                 "SECURITY_SCANNER_V4_REINDEX_THRESHOLD",
                 DEFAULT_SECURITY_SCANNER_V4_REINDEX_THRESHOLD,
             )
         )
-        stale_threshold = datetime.utcnow() - timedelta(
+        stale_threshold = datetime.now(timezone.utc) - timedelta(
             seconds=self.app.config.get(
                 "SECURITY_SCANNER_V4_IN_PROGRESS_TIMEOUT", DEFAULT_IN_PROGRESS_TIMEOUT
             )
--- a/data/secscan_model/secscan_v4_model_v2.py
+++ b/data/secscan_model/secscan_v4_model_v2.py
@@ -99,7 +99,7 @@ class V4SecurityScanner2(SecurityScannerInterface):
     def _not_indexed_query(self):
         """Manifests with no ManifestSecurityStatus row (never scanned)."""
         return (
-            Manifest.select()
+            Manifest.select(can_use_read_replica=True)
             .where(
                 ~fn.EXISTS(
                     ManifestSecurityStatus.select().where(
@@ -113,7 +113,7 @@ class V4SecurityScanner2(SecurityScannerInterface):
     def _stale_in_progress_query(self, stale_threshold):
         """Manifests stuck in IN_PROGRESS (crash recovery)."""
         return (
-            Manifest.select()
+            Manifest.select(can_use_read_replica=True)
             .join(ManifestSecurityStatus)
             .where(
                 ManifestSecurityStatus.index_status == IndexStatus.IN_PROGRESS,
@@ -125,7 +125,7 @@ class V4SecurityScanner2(SecurityScannerInterface):
     def _failed_query(self, reindex_threshold):
         """Failed manifests past retry threshold."""
         return (
-            Manifest.select()
+            Manifest.select(can_use_read_replica=True)
             .join(ManifestSecurityStatus)
             .where(
                 ManifestSecurityStatus.index_status == IndexStatus.FAILED,
@@ -137,7 +137,7 @@ class V4SecurityScanner2(SecurityScannerInterface):
     def _needs_reindex_query(self, indexer_hash, reindex_threshold):
         """Manifests with outdated indexer hash (Clair DB updated)."""
         return (
-            Manifest.select()
+            Manifest.select(can_use_read_replica=True)
             .join(ManifestSecurityStatus)
             .where(
                 ManifestSecurityStatus.index_status != IndexStatus.MANIFEST_UNSUPPORTED,
@@ -191,9 +191,19 @@ class V4SecurityScanner2(SecurityScannerInterface):
         """
         Atomically claim a batch of manifests using SELECT FOR UPDATE SKIP LOCKED.
         Returns list of claimed Manifest rows.
+
+        Two-step process:
+        1. Find candidate IDs on read replica (query_fn uses can_use_read_replica=True)
+        2. Claim those IDs on primary with FOR UPDATE SKIP LOCKED
         """
+        # Step 1: Find candidate IDs on read replica (outside transaction)
+        candidate_ids = [m.id for m in query_fn().limit(batch_size)]
+        if not candidate_ids:
+            return []
+
+        # Step 2: Claim on primary with FOR UPDATE SKIP LOCKED
         with db_transaction():
-            query = query_fn().limit(batch_size)
+            query = Manifest.select().where(Manifest.id.in_(candidate_ids))
             candidates = list(db_for_update(query, skip_locked=SKIP_LOCKED))
             if not candidates:
                 return []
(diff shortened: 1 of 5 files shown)


data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -0,0 +1,716 @@
...
+                self._index_batch_concurrent(claimed, reindex_threshold)
+
+        # Re-indexing only if enabled AND we have remaining capacity
+        if remaining > 0 and self.app.config.get("SECURITY_SCANNER_V4_ENABLE_REINDEXING", True):
+            # Even smaller chunks for reindex (lower priority)
+            reindex_chunk = min(chunk_size // 5, remaining)
+            if reindex_chunk == 0:
+                reindex_chunk = 1
+
+            query_fn = lambda: self._needs_reindex_query(indexer_hash, reindex_threshold)
```
kleesc, line 564: Could we just add this to the priority above? I don't think this should be optional, as we _want_ to reindex in the case of an indexer state update.

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -0,0 +1,716 @@
...
+                "SECURITY_SCANNER_V4_IN_PROGRESS_TIMEOUT", DEFAULT_IN_PROGRESS_TIMEOUT
+            )
+        )
+
+        indexer_hash = indexer_state.get("state", "")
+        remaining = batch_size
+
+        # High-priority queries always run
+        high_priority_queries = [
+            ("not_indexed", lambda: self._not_indexed_query()),
```
kleesc, line 535: Will this scale - this will be a full scan of the manifest table every time (even with the skip lock?). We were previously selecting random blocks based off ids. This was also the reason behind a separate `perform_indexing_recent_manifests`

data/secscan_model/secscan_v4_model_v2.py
lechuk47: From a UX point of view, I think non-indexed manifests should have more priority. This does not mean it's optional, it means that reindexing will happen when all manifests have an index report, or better said, when all pending manifests are assigned to a worker.

data/secscan_model/secscan_v4_model_v2.py
lechuk47: This should use a INDEX SCAN, not a FULL TABLE SCAN. This query can run in a read replica so it shouldn't be a major problem. I will double check the query and how this behaves though.

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -0,0 +1,726 @@
...
+                    continue
+
+            indexable.append((candidate, manifest, layers))
+
+        if not indexable:
+            return
+
+        # Phase 2: Concurrent Clair API calls
+        results = {}
+        with ThreadPoolExecutor(max_workers=concurrency) as executor:
```
cubismod, line 383: We're creating a new threadpool each time a batch is scanned, would make more sense to move this over to `__init__` to re-use the pool

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -0,0 +1,726 @@
...
+                try:
+                    report, state = future.result()
+                    results[candidate.id] = (candidate, manifest, report, state, None)
+                except InvalidContentSent:
+                    results[candidate.id] = (candidate, manifest, None, None, "invalid_content")
+                    secscan_index_errors.labels(error_type="invalid_content").inc()
+                except LayerTooLargeException:
+                    results[candidate.id] = (candidate, manifest, None, None, "layer_too_large")
+                    secscan_index_errors.labels(error_type="layer_too_large").inc()
+                except APIRequestFailure:
```
cubismod, line 407: I'd suggest adding a broader `except Exception` in case of a `ConnectionError` and then mark the manifest as failed so scans don't get stuck `IN_PROGRESS` Until stale recovery on failure.

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -0,0 +1,726 @@
...
+                    "Indexing manifest [%d] %s/%s@%s",
+                    manifest._db_id,
+                    candidate.repository.namespace_user,
+                    candidate.repository.name,
+                    manifest.digest,
+                )
+                future = executor.submit(self._call_clair_index, manifest, layers)
+                future_to_candidate[future] = (candidate, manifest)
+
+            for future in as_completed(future_to_candidate):
```
cubismod, line 396: You should add a timeout to `as_completed`.

data/secscan_model/secscan_v4_model_v2.py
cubismod: Even with the index scan, that table is pretty large and will require a lot of iteration to find unscanned items.

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -0,0 +1,726 @@
...
+        return None
+
+    def perform_indexing_recent_manifests(self, batch_size=None):
+        """
+        No-op in v2. The main perform_indexing loop handles all manifests efficiently.
+        With SKIP LOCKED, new manifests are picked up immediately by the not_indexed query.
+        """
+        pass
+
+    # ===== Methods copied from V4SecurityScanner (no concurrency issues) =====
```
cubismod, line 593: Can you move these duplicated methods to a base class, something like `V4SecurityScannerBase`? Then inherit them in both v1 & v2 models

PR #5727, opened 2026-04-14
PROJQUAY-10795: Fix Load More Logs button flashing during data fetch

quay-devel:
## Summary

- The "Load More Logs" button was flashing (appearing briefly then disappearing) during log data fetching
- Root cause: the intersection observer fires asynchronously after render, so the button appeared for one render cycle before the observer triggered `fetchNextPage()` and `isFetchingNextPage` became `true`
- Fix: expand the footer visibility condition from `hasNextPage` to `hasNextPage || isFetchingNextPage`, keeping the spinner visible throughout the entire fetch cycle
- Added `data-testid` attributes to `load-more-spinner` and `load-more-button` for testability

## Jira
[PROJQUAY-10795](https://redhat.atlassian.net/browse/PROJQUAY-10795)

## Test Plan
- [ ] Jest tests verify spinner shows when `isFetchingNextPage=true` (with or without `hasNextPage`)
- [ ] Jest tests verify button shows only when `hasNextPage=true` and not fetching
- [ ] Manual: scroll to bottom of Usage Logs — spinner shows smoothly without button flash

🤖 Generated with Claude Code

### DIFF
--- a/web/__mocks__/fileMock.js
+++ b/web/__mocks__/fileMock.js
@@ -1,0 +1,1 @@
+module.exports = 'test-file-stub';
--- a/web/babel.config.js
+++ b/web/babel.config.js
@@ -1,0 +1,7 @@
+module.exports = {
+  presets: [
+    ['@babel/preset-env', {targets: {node: 'current'}}],
+    ['@babel/preset-react', {runtime: 'automatic'}],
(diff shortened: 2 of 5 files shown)


quay-devel:
Reopening from fork quay-devel/quay instead.

PR #5763, opened 2026-04-15
NO-ISSUE: chore(web): add multi-model PR review workflow

HammerMeetNail:
## Summary
- add repository-local `review-pr` skills and reviewer agents for both Claude Code and Cursor so PR reviews can fan out across three independent reviewers and then synthesize the results into one report
- pin Cursor reviewers to explicit model-specific agents (`Gemini 3.1 Pro`, `GPT-5.4 High`, and `Claude 4.6 Opus Thinking`) while keeping Claude Code-compatible reviewer definitions for the Claude path
- tighten the orchestration instructions so the parent agent performs only minimal setup before issuing all reviewer launches in one batch, and require the final report to compare reviewer contributions, call out unique findings, and note whether the run used true parallelism or degraded staggered concurrency

## Why
- the original `review-pr` command was effectively a single-agent workflow, which made it hard to test whether multiple reviewers were actually contributing distinct findings
- Cursor UI labels based on generic reviewer names were misleading, making it difficult to verify which model was being used for each reviewer
- earlier verification showed that child reviewers could be spawned with different models, but the launches were sometimes staggered; the updated prompts make the intended fan-out behavior explicit and make the final report self-describe the value of the multi-agent run

## Test plan
- [x] Run Cursor CLI in `stream-json` mode and confirm the workflow loads the Cursor-specific `review-pr` skill override
- [x] Verify the review workflow spawns three distinct reviewer subagents with unique `agentId` values in the NDJSON event stream
- [x] Confirm the reviewer launches resolve to the expected model IDs:
  - `pr-review-gemini-3-1-pro` -> `gemini-3.1-pro`
  - `pr-review-gpt-5-4-high` -> `gpt-5.4-high`
  - `pr-review-claude-4-6-opus-high-thinking` -> `claude-4.6-opus-high-thinking`
- [x] Confirm the synthesized review instructions now require `Reviewer Comparison` and `Incremental Value` sections in the final output
- [ ] Re-run the full review workflow after this change and verify that all three reviewer `started` events are emitted before the first reviewer completes, proving true batched fan-out instead of staggered execution

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.claude/agents/pr-review-haiku.md
+++ b/.claude/agents/pr-review-haiku.md
@@ -1,0 +1,43 @@
+---
+name: pr-review-haiku
+description: Fast PR reviewer for concrete bugs, regressions, and missing tests. Use as one of the three parallel reviewers for review-pr.
+tools: Read, Grep, Glob, Bash
+model: claude-haiku-4-5
+permissionMode: plan
+maxTurns: 12
+---
+
+You are the fast reviewer in a peer-review swarm.
+
+Your job is to independently review a GitHub pull request and report only
+evidence-backed findings. You are intentionally optimized for breadth and
+speed, so prefer catching obvious correctness issues, regressions, risky edge
+cases, and missing tests over deep architectural speculation.
+
+Rules:
+- Never edit files or make state-changing shell commands.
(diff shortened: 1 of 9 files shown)


HammerMeetNail:
Too expensive

PR #5817, opened 2026-04-27
PROJQUAY-11027: fix(mirror): preserve future syncs when canceling org mirror

shaonrh:
## Summary
- Cancel no longer kills future syncs: `update_sync_status_to_cancel` preserves `sync_start_date` instead of clearing it, and `release_org_mirror_config` transitions CANCEL to SUCCESS with a scheduled next sync
- Repo-level CANCEL recovery: `schedule_org_mirror_repos_for_sync` now picks up both NEVER_RUN and CANCEL repos, transitioning them to SYNC_NOW with restored retries on the next discovery cycle
- UI copy fix: Cancel confirmation dialog updated to indicate future scheduled syncs will continue

## Root Cause
When a user cancelled an active org mirror sync, `update_sync_status_to_cancel` set `sync_start_date=None` and `retries=0`. After the worker processed the cancel, the config stayed in CANCEL state with no scheduled date, permanently stopping all future syncs until the user manually triggered sync-now.

## Test plan
- [x] Unit tests for `update_sync_status_to_cancel` preserving `sync_start_date`
- [x] Unit test for `release_org_mirror_config` scheduling next sync after cancel
- [x] End-to-end test: cancel, release, time passes, config becomes eligible again
- [x] Regression test: no re-cancel loop (status is SUCCESS not CANCEL on next pickup)
- [x] Repo-level test: CANCEL repos transition to SYNC_NOW on next discovery, SKIP untouched
- [x] API test: sync-cancel endpoint preserves `sync_start_date`
- [x] Worker test: updated assertion for cancel to SUCCESS transition

### DIFF
--- a/data/model/org_mirror.py
+++ b/data/model/org_mirror.py
@@ -1139,10 +1139,8 @@ def release_org_mirror_config(
     Release an org mirror config after discovery and update its status.
 
     Calculates next sync_start_date based on sync_interval.
-    Decrements retries on failure, resets on success.
-
-    If discovery is cancelled, the job will not be attempted until manual
-    sync-now is triggered by the user.
+    Decrements retries on failure, resets on success or cancel.
+    Cancel is transitioned to SUCCESS after scheduling the next sync.
 
     Args:
         org_mirror_config: The OrgMirrorConfig to release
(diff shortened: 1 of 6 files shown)


shaonrh:
Superseded by #5818.

PR #5827, opened 2026-04-29
PROJQUAY-11403: fix(ui): fix manifest track line on unrelated pages

LiZhang19817:
## Summary
- Fix manifest track visualization (vertical colored lines) appearing on pages where no visible tags share a manifest digest
- Compute tracks from only the current page's visible tags (`props.tags`) instead of all tags (`props.allTags`)
- Use local row indices instead of global offsets since tracks are now per-page

## Root Cause
In `TagsTable.tsx`, `useManifestTracks` received all tags across all pages. When two tags sharing a digest were far apart (e.g., global index 5 and index 3000), `entryByIndex` was populated for every index between them, rendering a continuous "middle" line on every intermediate page — even though neither matching tag was visible.

### Image tags with different digest:
image

### Image tags with the same digest:
image

### clicking a dot still selects all tags sharing that manifest across all pages
image

## Test plan
- [x] Existing `useManifestTracks` unit tests pass (17/17)
- [x] Navigate to a repository with many tags (3000+) and verify no spurious vertical lines on pages where all visible tags have unique digests
- [x] Verify track lines/dots appear correctly on pages where 2+ visible tags share a digest
- [x] Verify clicking a dot still selects all tags sharing that manifest across all pages

JIRA: https://redhat.atlassian.net/browse/PROJQUAY-11403

### DIFF
--- a/web/src/routes/RepositoryDetails/Tags/TagsTable.tsx
+++ b/web/src/routes/RepositoryDetails/Tags/TagsTable.tsx
@@ -571,11 +571,8 @@ export default function TagsTable(props: TableProps) {
 
   // Calculate manifest tracks for visual grouping of tags sharing the same digest
   const {tracks, getTrackEntry, getLineClass, trackCount} = useManifestTracks(
-    props.allTags,
+    props.tags,
   );
-
-  // Page offset converts local row index to global index for track lookup
-  const pageOffset = (props.page - 1) * props.perPage;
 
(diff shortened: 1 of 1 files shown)


LiZhang19817:
@jbpratt @bcaton85 Pls review and merge the fix, now the pre-merge testing is passed.

jbpratt (approved):

PR #5850, opened 2026-04-30
PROJQUAY-11433: fix(gc): batch repository state updates on namespace deletion

quay-devel:
## Summary
- Replace N individual `repo.save()` calls during namespace deletion with a single bulk `UPDATE repository SET state=MARKED_FOR_DELETION WHERE namespace_user=<user>` before the purge loop
- Guard `purge_repository()` from re-issuing a redundant per-repo save when the state is already set

## Root Cause / Rationale
Jaeger traces showed `UPDATE "repository" SET "namespace_user_..."` queries taking 14.7 s and 5.7 s during org/user deletion. The root cause was `delete_user()` calling `gc.purge_repository()` per repo, each of which issued an individual `repo.save()` inside a transaction with cascading deletes across 20+ tables — creating N long-held locks for namespaces with many repositories.

## Changes
- `data/model/user.py`: bulk-update all namespace repos to `MARKED_FOR_DELETION` in one `Repository.update(...).execute()` call before the `purge_repository` loop
- `data/model/gc.py`: skip the individual `repo.save()` inside `purge_repository()` when the state is already `MARKED_FOR_DELETION` (eliminates the redundant round-trip for the namespace deletion path; no behaviour change for the GC worker path)
- `data/model/test/test_user.py`: two new tests — one verifying repos are pre-marked before `purge_repository` is called, one verifying deletion works correctly when some repos are already in `MARKED_FOR_DELETION` state

No migration needed — no schema changes.

## Test Plan

## JIRA
https://redhat.atlassian.net/browse/PROJQUAY-11433

## Backport
- [ ] Backport required: 
- [x] No backport needed

## Automation
- **Session ID**: session-a0f1c781-a4ee-4831-a2fd-e4203e49f110

### DIFF
--- a/data/model/gc.py
+++ b/data/model/gc.py
@@ -81,8 +81,11 @@ def purge_repository(repo, force=False):
     assert repo.state == RepositoryState.MARKED_FOR_DELETION or force
 
     # Update the repo state to ensure nothing else is written to it.
-    repo.state = RepositoryState.MARKED_FOR_DELETION
-    repo.save()
+    # Skip the individual save when the state is already set (e.g. after a bulk
+    # UPDATE from delete_user()) to avoid N per-repo round-trips.
+    if repo.state != RepositoryState.MARKED_FOR_DELETION:
+        repo.state = RepositoryState.MARKED_FOR_DELETION
+        repo.save()
 
     # GC to remove the images and storage.
     _purge_repository_contents(repo)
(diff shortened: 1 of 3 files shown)


quay-devel:
## Ready for Review

CI: 29/31 passed | CodeRabbit: NONE | Unresolved threads: CR=0 Human=0

cc @quay/downstream

jbpratt (approved):

Marcusk19 (approved):

PR #5858, opened 2026-05-01
PROJQUAY-10575: fix(mirror): fix cancel mirror log showing garbled text

shaonrh:
## Summary
- Backend: Fix if/if/else to if/elif/else in perform_mirror() finally block to prevent a spurious success log after FAIL. Pass actual tags instead of hardcoded Not applicable for stdout/stderr.
- Frontend: Add null-guards to mirror log description renderers so null/undefined metadata fields are omitted instead of rendered as literal strings.

## Test plan
- Cancel a running mirror sync and verify the usage log shows clean text
- Verify successful and failed mirror syncs still log correctly
- Confirm no regression in tag-level success/failure log entries

Fixes: [PROJQUAY-10575](https://redhat.atlassian.net/browse/PROJQUAY-10575)

### DIFF
--- a/web/src/hooks/UseLogDescriptions.tsx
+++ b/web/src/hooks/UseLogDescriptions.tsx
@@ -330,16 +330,28 @@ export function useLogDescriptions() {
       return `Mirror started for ${metadata.message}`;
     },
     repo_mirror_sync_success: function (metadata: Metadata) {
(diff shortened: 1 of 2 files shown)


quay-devel:
Superseded by #5916 (same fix rebased on latest master with added regression tests).

PR #5868, opened 2026-05-04
PROJQUAY-11403: fix(ui): fix manifest track line on unrelated pages [redhat-3.17]

quay-devel:
Cherry-pick of #5827 onto `redhat-3.17`.

## Summary
- Fixes manifest track line appearing on pages that don't contain either of the two tags sharing a digest
- Computes manifest tracks from only the current page's visible tags instead of all tags across all pages

## Original PR
#5827


LiZhang19817:
@bcaton85 @jbpratt pls help review and merge this PR to cherry-pick to Quay 3.17

PR #5898, opened 2026-05-05
PROJQUAY-9246: fix(oauth): use client_id for bulk selection identity

quay-devel:
## Summary
- Fix bulk select/delete of OAuth applications in the new UI
- OAuth apps with duplicate names caused selection state to only retain the last app per name, resulting in only one app being deleted when "select all" was used

## Root Cause / Rationale
The `OAuthApplication` database model allows duplicate `name` values (`CharField()` with no uniqueness constraint). The frontend selection logic in `OAuthApplicationsList.tsx` used `name` for identity comparison in both the `onSelectPermission` handler (filtering by `p.name`) and the `isSelected` check (`p.name === oauthApplication.name`).

When multiple apps share a name, selecting one replaces the other in the `selectedOAuthApplications` state array, while `isSelected` (using `.some()`) still shows both as checked. The user sees all checkboxes checked but only the last app with each name is actually in the selection array, so bulk delete only removes that one.

## Changes
- Changed selection identity comparison from `name` to `client_id` in `onSelectPermission` filter
- Changed `isSelected` check from `name` to `client_id` in both table render sections (drawer and non-drawer views)
- `client_id` is the actual unique identifier for OAuth applications and is already used by the delete API endpoint

## Test Plan

Existing OAuth application tests (10 tests in `UseOAuthApplications.test.ts` and `OAuthApplicationResource.test.ts`) pass.

## JIRA
https://redhat.atlassian.net/browse/PROJQUAY-9246

## Backport
- [ ] Backport required
- [x] No backport needed

## Automation
- **Session ID**: session-71ed56d0-9739-4554-9645-989562125772


quay-devel:
Demo environment has been provisioned and is available in Ambient session `session-cd425630-e95a-49a6-9067-fbc300f13e05`.

quay-devel:
Demo environment has been provisioned and is available in Ambient session `session-61229cce-ae9f-4939-be57-8811f227b8de`.

bcaton85 (approved):

jbpratt (approved):

PR #5906, opened 2026-05-05
PROJQUAY-11475: chore(web): migrate frontend package manager from npm to pnpm

quay-devel:
## Summary
- Migrate the `web/` React frontend from npm to pnpm for faster installs, stricter dependency resolution, and better disk efficiency
- Keep npm + `package-lock.json` for the downstream hermetic build (`Dockerfile.downstream`) since Cachi2 does not support pnpm as a prefetch type

## Root Cause / Rationale
pnpm offers significant advantages over npm for frontend development:
- **Faster installs**: pnpm's content-addressable store avoids redundant downloads
- **Disk efficiency**: Hard links to a global store instead of duplicating packages
- **Stricter dependency resolution**: Catches phantom dependency issues (uses `node-linker=hoisted` for webpack compatibility)
- **Better lockfile**: `pnpm-lock.yaml` is more readable and diff-friendly than `package-lock.json`

The downstream hermetic build via Konflux/Cachi2 remains on npm because Cachi2 only supports npm and yarn for JavaScript prefetching — not pnpm. Both lockfiles are kept in sync.

## Changes
- **`web/package.json`**: Add `packageManager` field, add `pnpm.overrides` alongside existing `overrides`
- **`web/.npmrc`**: Replace `legacy-peer-deps=true` with pnpm-compatible `strict-peer-dependencies=false` and `node-linker=hoisted`
- **`web/pnpm-lock.yaml`**: New lockfile generated from `pnpm import` (converted from `package-lock.json`)
- **`Dockerfile`**: `build-ui` stage installs pnpm and uses `pnpm install --frozen-lockfile` / `pnpm run build`
- **`Dockerfile.downstream`**: Unchanged — stays on npm for Cachi2 hermetic build compatibility
- **`web/Dockerfile`**: Updated to use pnpm
- **GitHub Actions workflows**: All 4 web workflows (`web-unit-tests`, `web-playwright-ci`, `web-ci`, `CI.yaml`) updated to use `pnpm/action-setup` with store caching
- **`.github/actions/setup-quay/action.yaml`**: React build step uses pnpm
- **`.pre-commit-config.yaml`**: Exclude `pnpm-lock.yaml` from ESLint
- **Documentation**: `web/AGENTS.md` and `agent_docs/workflow.md` updated to reference pnpm commands

## Test Plan
- [x] Frontend tests pass (Playwright/Cypress)

**Verification performed:**
- `pnpm install --frozen-lockfile` succeeds
- `pnpm run build` (webpack production build) succeeds with 0 errors
- All 1064 unit tests pass across 144 test files via `pnpm test`
- `npm install --package-lock-only --legacy-peer-deps` still regenerates `package-lock.json` correctly for downstream builds

## JIRA
https://redhat.atlassian.net/browse/PROJQUAY-11475

## Backport
- [x] No backport needed

## Automation
- **Session ID**: session-056c9f56-48ed-4b22-8a65-4e79cf2a7d63

### DIFF
--- a/.github/workflows/web-playwright-ci.yaml
+++ b/.github/workflows/web-playwright-ci.yaml
@@ -52,21 +52,37 @@ jobs:
             exit 1
           fi
 
-      - name: Set up Node.js with npm caching
+      - name: Set up Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: '22'
-          cache: 'npm'
-          cache-dependency-path: web/package-lock.json
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+        with:
+          version: 10
+          run_install: false
+
+      - name: Get pnpm store directory
+        id: pnpm-cache
+        shell: bash
+        run: echo "STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
+
+      - name: Cache pnpm store
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('web/pnpm-lock.yaml') }}
+          restore-keys: ${{ runner.os }}-pnpm-store-
 
       - name: Install dependencies
-        run: cd web && npm install
+        run: cd web && pnpm install --frozen-lockfile
 
       - name: Get Playwright version
         id: playwright-version
         run: |
           cd web
-          PLAYWRIGHT_VERSION=$(node -e "console.log(require('./package-lock.json').packages['node_modules/@playwright/test'].version)")
+          PLAYWRIGHT_VERSION=$(node -e "console.log(require('./node_modules/@playwright/test/package.json').version)")
           echo "version=$PLAYWRIGHT_VERSION" >> $GITHUB_OUTPUT
 
       - name: Cache Playwright browsers
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,7 @@ repos:
         entry: web/node_modules/.bin/eslint --fix
         language: system
         files: ^web/
-        exclude: ^web/cypress/test/|^web/cypress/fixtures|^web/Containerfile|^web/Dockerfile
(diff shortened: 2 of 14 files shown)


jbpratt:
@quay-devel we would need some way to keep these in sync, this doesn't sound ideal so far

quay-devel:
Good call — I've added a CI guard in the latest push. The `web-unit-tests` workflow now runs `npm install --package-lock-only --legacy-peer-deps` and fails if `package-lock.json` has a diff. So if someone updates deps via pnpm and forgets to regenerate the npm lockfile, CI will catch it before merge.

The workflow for contributors:
1. `pnpm add/remove/update <pkg>` (updates `pnpm-lock.yaml`)
2. `npm install --package-lock-only --legacy-peer-deps` (regenerates `package-lock.json`)
3. Commit both lockfiles

The dual-lockfile is needed because Cachi2 (Konflux's hermetic prefetch) only supports npm/yarn — not pnpm. Once Cachi2 adds pnpm support, we can drop `package-lock.json` entirely.

.github/workflows/web-playwright-ci.yaml
```diff
@@ -52,21 +52,37 @@ jobs:
...
         with:
           node-version: '22'
-          cache: 'npm'
-          cache-dependency-path: web/package-lock.json
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
+        with:
+          version: 10
+          run_install: false
```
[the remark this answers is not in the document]
quay-devel: This is an upstream issue with the `pnpm/action-setup` action itself — it ships with a Node.js 20 runtime, which GitHub Actions is deprecating. We're already on the latest release (v4.1.0, SHA-pinned). The fix needs to come from the pnpm/action-setup maintainers updating their action to Node.js 22. Nothing actionable on our side — we'll get the fix automatically when they release a new version and we update the pin.

.pre-commit-config.yaml
jbpratt, line 1: @quay-devel can we add a pre-commit hook to ensure the package-log.json is kept in sync?

quay-devel:
Added a pre-commit hook (`npm-lockfile-sync`) that automatically regenerates and stages `package-lock.json` whenever `pnpm-lock.yaml` or `package.json` is in the staged changeset. So the sync is now fully automatic for anyone with pre-commit hooks installed.

There's also a CI validation step (`npm ci --dry-run`) in the `web-unit-tests` workflow as a safety net.

.pre-commit-config.yaml
quay-devel: Done — added an `npm-lockfile-sync` pre-commit hook in the latest push. When `pnpm-lock.yaml` or `package.json` is staged, it auto-runs `npm install --package-lock-only --legacy-peer-deps` and stages the updated `package-lock.json`. Completely transparent to the developer.

harishsurf (approved):

PR #5907, opened 2026-05-05
PROJQUAY-11288: fix(nginx): return 502 instead of 404 when backend is unreachable

quay-devel:
## Summary

- **Bug:** When Quay's backend storage is unreachable, nginx returns a misleading `404 Not Found` instead of `502 Bad Gateway`
- **Root cause:** The `error_page 502` directive in `server-base.conf.jnj` used `{{static_dir}}/502.html` which nginx treated as a URI. The internal redirect matched the `/` location block whose `root` was `{{static_dir}}/patternfly/`, causing the file path to resolve as `/quay-registry/static/patternfly/quay-registry/static/502.html` — a doubled, non-existent path
- **Fix:** Replace with a URI-based redirect (`error_page 502 =502 /502.html`) and a dedicated internal location block with the correct `root` directive pointing to `{{static_dir}}`

## Changes

- `conf/nginx/server-base.conf.jnj` — Fix error_page directive and add internal location block
- `conf/init/test/test_nginx_conf_create.py` — Add 10 regression tests for error page configuration

## Test plan

- [x] Verified rendered nginx config resolves 502.html to the correct path
- [x] 10 new unit tests pass covering:
  - error_page uses URI, not absolute filesystem path
  - dedicated location block exists with correct root
  - location is marked as internal
  - 502 status code is preserved with `=502`
  - root directive adapts to different static_dir values
  - no patternfly path contamination in 502 location block
- [x] No regressions in existing tests

## Rollback

Revert this single commit to restore previous behavior.

Fixes: https://issues.redhat.com/browse/PROJQUAY-11288

### DIFF
--- a/.claude/scripts/validate-pr-title.sh
+++ b/.claude/scripts/validate-pr-title.sh
@@ -13,22 +13,28 @@ if [ -z "$CMD" ]; then
   exit 0
 fi
 
-TITLE=""
+# Use python3 shlex to correctly parse shell argv — portable and handles all
+# quoting/escaping forms. CMD is passed as argv[1] to avoid stdin conflict.
+TITLE=$(python3 -c '
+import shlex, sys
 
-# Strategy 1: --title 'message' (single quotes)
-if [ -z "$TITLE" ]; then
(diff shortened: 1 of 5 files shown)


quay-devel:
Superseded by a new PR after syncing the fork with upstream.

PR #5927, opened 2026-05-07
PROJQUAY-11217: fix(ui): show empty state instead of spinner when repository search has no results

quay-devel:
## Summary

- When a user searches for a repository name that doesn't exist, the repository list table displayed a permanent loading spinner instead of an empty state message
- Fixed by gating the `<Spinner>` on the `loading` flag from `useRepositories`; when loading is complete and the filter matched nothing, "No repositories found" is shown instead

## Root Cause / Rationale

`RepositoriesList.tsx:400` unconditionally rendered `<Spinner size="lg" />` whenever `filteredRepos.length === 0`, without consulting the `loading` flag. When a search returned no results, `loading` was already `false` and the spinner was never dismissed.

The two empty-state guards are now complementary:
- `RepositoriesList.tsx:299` — `if (!loading && !repos?.length)` → full `<Empty>` component for "org has no repos at all"
- `RepositoriesList.tsx:400` — `filteredRepos.length === 0 && !loading` → inline "No repositories found" for "search matched nothing"

## Changes

- `web/src/routes/RepositoriesList/RepositoriesList.tsx` — Added `loading` guard inside the `filteredRepos.length === 0` branch (lines 400–411)
- `web/src/routes/RepositoriesList/RepositoriesList.test.tsx` — New file: 3 unit tests covering spinner/empty/populated states
- `web/vitest.config.ts` — Added `resolveId` hook to stub `.scss` imports as virtual modules in the test environment (prevents SCSS compilation errors via the `Breadcrumb → OrganizationsList → Organizations.scss` import chain)

## Test Plan

- [x] Unit tests added/updated — 3 regression tests in `RepositoriesList.test.tsx`
- [x] Frontend tests pass — full Vitest suite (146 files, 1073 tests) passes
- [ ] Registry tests pass (`make registry-test`) — backend-only, not affected
- [ ] Type checking passes (`make types-test`) — no type changes
- [ ] Manual testing performed — no running Quay instance available in CI environment

**Manual verification steps:**
1. Navigate to any organization's repository list
2. Type a search term that matches no repositories
3. Confirm "No repositories found" appears instead of a spinner

## JIRA
https://issues.redhat.com/browse/PROJQUAY-11217

## Backport
- [x] No backport needed (no Target Version set on JIRA ticket)

## Notes

The "No repositories found" text rendered in `<Td>` is intentionally minimal to fix the regression. A follow-up improvement could use a full PatternFly `EmptyState` component (with icon, title, and body) to match the UX of the truly-empty org path at line 299. Tracked separately.

## Automation
- **Session ID**: session-c568011e-deff-4e47-9185-15d34568e746

### DIFF
--- a/web/src/routes/RepositoriesList/RepositoriesList.tsx
+++ b/web/src/routes/RepositoriesList/RepositoriesList.tsx
@@ -398,12 +398,17 @@ export default function RepositoriesList(props: RepositoriesListProps) {
           </Thead>
           <Tbody data-testid="repository-list-table">
             {filteredRepos.length === 0 ? (
-              // Repo table loading icon
-              <Tr>
-                <Td>
-                  <Spinner size="lg" />
-                </Td>
-              </Tr>
+              loading ? (
+                <Tr>
+                  <Td>
+                    <Spinner size="lg" />
+                  </Td>
+                </Tr>
+              ) : (
+                <Tr>
+                  <Td>No repositories found</Td>
+                </Tr>
+              )
             ) : (
               paginatedRepositoryList.map((repo, rowIndex) => (
                 <Tr key={rowIndex}>
--- a/web/src/routes/RepositoriesList/RepositoriesList.tsx
+++ b/web/src/routes/RepositoriesList/RepositoriesList.tsx
@@ -400,13 +400,13 @@ export default function RepositoriesList(props: RepositoriesListProps) {
             {filteredRepos.length === 0 ? (
               loading ? (
                 <Tr>
-                  <Td>
+                  <Td colSpan={5}>
                     <Spinner size="lg" />
                   </Td>
                 </Tr>
               ) : (
                 <Tr>
-                  <Td>No repositories found</Td>
+                  <Td colSpan={5}>No repositories found</Td>
                 </Tr>
               )
             ) : (
--- a/web/playwright/e2e/repository/repositories-list.spec.ts
+++ b/web/playwright/e2e/repository/repositories-list.spec.ts
@@ -501,5 +501,33 @@ test.describe('Repositories List', {tag: ['@repository']}, () => {
       const count = await rows.count();
       expect(count).toBeGreaterThanOrEqual(2);
     });
+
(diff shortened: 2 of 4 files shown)


web/src/routes/RepositoriesList/RepositoriesList.tsx
```diff
@@ -398,12 +398,17 @@ export default function RepositoriesList(props: RepositoriesListProps) {
           </Thead>
           <Tbody data-testid="repository-list-table">
             {filteredRepos.length === 0 ? (
-              // Repo table loading icon
-              <Tr>
-                <Td>
-                  <Spinner size="lg" />
-                </Td>
-              </Tr>
+              loading ? (
```
jbpratt, line 401: Could we add an e2e test?

quay-devel: Good call — added a Playwright E2E test to `repositories-list.spec.ts` in the `search functionality` describe block (commit `7b34d568`). The test creates an org with one repo, waits for it to appear, searches for a term that won't match, and asserts no `progressbar` role and that 'No repositories found' is visible. It also clears the search and verifies the repo reappears.

web/src/routes/RepositoriesList/RepositoriesList.tsx
```diff
@@ -398,12 +398,17 @@ export default function RepositoriesList(props: RepositoriesListProps) {
...
+                <Tr>
+                  <Td>
+                    <Spinner size="lg" />
+                  </Td>
+                </Tr>
+              ) : (
+                <Tr>
+                  <Td>No repositories found</Td>
+                </Tr>
```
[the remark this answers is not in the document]
quay-devel: Fixed in commit `53a8f956` — added `colSpan={5}` to both the spinner `<Td>` and the 'No repositories found' `<Td>`. This table has at most 5 columns (checkbox + Name + Visibility + Size + Last Modified, where checkbox and Size are conditional), matching the pattern used elsewhere in the codebase (e.g. `RobotAccountsList.tsx`).

jbpratt (approved):

harishsurf (approved):

PR #5931, opened 2026-05-08
PROJQUAY-5196: fix(auth): exclude superusers from restricted user checks

quay-devel:
## What does this PR do?

When `FEATURE_RESTRICTED_USERS` is enabled, superusers — including those
identified via `LDAP_SUPERUSER_FILTER` — are incorrectly treated as restricted
users. This blocks superusers from administrative operations (creating orgs,
pushing images) unless they are also explicitly listed in
`RESTRICTED_USERS_WHITELIST`.

This PR adds a superuser guard in `UserManager.is_restricted_user()` so that
superusers are never considered restricted, regardless of whitelist or LDAP
filter configuration.

## Why is this change being made?

Fixes [PROJQUAY-5196](https://issues.redhat.com/browse/PROJQUAY-5196). The bug
was originally reported against Quay v3.8.3 and is still present on master. Two
prior PRs (#1883, #2967) attempted the same fix but were not merged.

## How was this tested?

Added 5 new test functions (11 parametrized cases) covering:
- Superuser exclusion from restriction (config-based and LDAP-based)
- Feature flag interaction (`RESTRICTED_USERS` off, `SUPER_USERS` off)
- Integration tests with real `ConfigUserManager` and `FederatedUserManager`

## Checklist

- [x] Bug fix addresses root cause
- [x] Regression tests added
- [x] Existing tests unmodified (no breakage)
- [x] Minimal change (2 lines of production code)
- [x] Backport to v3.18.0 required

### DIFF
--- a/data/users/__init__.py
+++ b/data/users/__init__.py
@@ -409,6 +409,9 @@ class UserManager(object):
         if not features.RESTRICTED_USERS:
             return False
 
+        if self.is_superuser(username):
+            return False
+
         return self.state.is_restricted_user(username)
 
     def is_superuser(self, username):
--- a/data/users/__init__.py
+++ b/data/users/__init__.py
@@ -360,7 +360,7 @@ class UserAuthentication(object):
             else:
                 password = decrypted
 
-        (result, err_msg) = self.state.verify_and_link_user(username_or_email, password)
+        result, err_msg = self.state.verify_and_link_user(username_or_email, password)
         if not result:
             return (result, err_msg)
 
--- a/endpoints/api/test/test_repository.py
+++ b/endpoints/api/test/test_repository.py
@@ -267,8 +267,9 @@ def test_create_repo_with_restricted_users_enabled_as_superuser(namespace, expec
 
 
 def test_create_repo_with_restricted_users_enabled_as_normal_user(app):
-    with patch("features.RESTRICTED_USERS", FeatureNameValue("RESTRICTED_USERS", True)), patch.object(
-        usermanager, "is_superuser", return_value=False
+    with (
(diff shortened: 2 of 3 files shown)


data/users/__init__.py
```diff
@@ -409,6 +409,9 @@ def is_restricted_user(self, username):
         if not features.RESTRICTED_USERS:
             return False
 
+        if self.is_superuser(username):
```
jbpratt, line 412: Is this a cached call or does this make an ldap call to superusers on every is_restricted_user check?

quay-devel: Good question. For config-backed deployments (`ConfigUserManager`), `is_superuser()` is a shared-memory array lookup — effectively O(1), no external calls.

For LDAP deployments (`FederatedUserManager`), `is_superuser()` delegates to `LDAPUsers.is_superuser()` which uses `_cached_permission_check()` — an in-process TTL cache (`LDAPCache`, `OrderedDict`-backed with configurable TTL). So the first call per TTL window hits LDAP; subsequent calls are served from cache. This is the same caching already used by the existing `is_superuser()` calls elsewhere in the request path (e.g., `SuperUserPermission` checks in endpoint decorators).

Net: no new LDAP calls are introduced — the same cached lookup that would already run for superuser permission checks now also short-circuits the restricted-user check.

jbpratt: The cache is a feature to be enabled, it isn't on by default

jbpratt: @quay-devel The cache is a feature to be enabled, it isn't on by default

jbpratt:
@quay-devel check the review thread

PR #5932, opened 2026-05-08
PROJQUAY-9883: fix(web): fix service keys filter not working for multi-word key names

quay-devel:
## Summary

- Fix Superuser Service Keys search filter that had no effect when text was typed
- Multi-word key names (containing spaces) now filter correctly
- Pagination now resets to page 1 when a search query is applied

## Root Cause / Rationale

`val.trim()` was applied inside the `TextInput` `onChange` handler on every
keystroke. Because the input is a controlled React component (`value={search.query}`),
trimming the space character prevented spaces from ever entering the query
string. Typing "Alpha Key" produced query state `"AlphaKey"`, which matches
nothing — making the filter appear completely broken for any multi-word name.

Single-word queries were unaffected (trim has no effect on non-space characters).
The filter hook logic (`UseServiceKeys.ts`) was correct and already covered by
unit tests; the bug was entirely in the UI layer's `onChange` handler.

Secondary: the hook did not call `setPage(1)` when `search.query` changed, so
users on page 2+ who applied a filter saw an empty table.

## Changes

- `web/src/routes/Superuser/ServiceKeys/ServiceKeys.tsx` — remove `val.trim()`
  from `onChange`; pass `val` directly to `setSearch`
- `web/src/hooks/UseServiceKeys.ts` — apply `.trim()` at filter evaluation time
  (`search.query.trim().toLowerCase()`) so leading/trailing whitespace is
  tolerated without breaking input; add `useEffect` to reset `page` to 1 on
  `search.query` change
- `web/src/routes/Superuser/ServiceKeys/ServiceKeys.test.tsx` (new) — 4
  component-level regression tests, including multi-word search
- `web/src/hooks/UseServiceKeys.test.ts` — add page-reset regression test

## Test Plan

**Regression test:** `ServiceKeys filter > supports multi-word key names in search`
fails without the fix (DOM shows `value="AlphaKey"` after typing "Alpha Key")
and passes with it. Verified by reverting and re-running.

Full vitest suite: 1087/1087 pass.

To manually verify:
1. Log in as a superuser
2. Navigate to Superuser → Service Keys
3. Type a multi-word key name (e.g., part of "My Auth Key") in the filter input
4. Confirm the list narrows correctly

## JIRA

https://issues.redhat.com/browse/PROJQUAY-9883

## Backport

- [ ] Backport required: `redhat-3.16` (verify Target Version on ticket — label `quay-3.16-qe-ui-bugs` suggests 3.16 scope; run `/cherrypick redhat-3.16` after merge if confirmed)
- [ ] No backport needed

---

## Automation

Generated by Amber (Quay bugfix workflow). Session: `session-c6e560c5-14fd-4569-af4d-31af011206c8`

### DIFF
--- a/web/src/hooks/UseServiceKeys.test.ts
+++ b/web/src/hooks/UseServiceKeys.test.ts
@@ -8,6 +8,7 @@ import {
   createServiceKey,
   deleteServiceKey,
   approveServiceKey,
+  IServiceKey,
 } from 'src/resources/ServiceKeysResource';
 
 vi.mock('src/resources/ServiceKeysResource', () => ({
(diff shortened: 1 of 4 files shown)


quay-devel:
## Ready for Review

CI: 1/1 passed | CodeRabbit: NONE | Unresolved threads: CR=0 Human=0

cc @quay/downstream

quay-devel:
Demo environment has been provisioned and is available in Ambient session `session-9ed07f22-cb61-4dcd-a7ae-094325a736b6`.

bcaton85 (approved):

jbpratt (approved):

PR #5964, opened 2026-05-12
PROJQUAY-11345: chore(deps): Bump pillow and follow-redirects

aroyoredhat:
Bumping pillow to version 12.2.0 and follow-redirects to version 1.16.0


nindsimonv (dismissed):

web/package-lock.json
jbpratt, line 1: @aroyoredhat we introduced `pnpm` recently, please ensure that lock file is kept up to date - this one is just for konflux at this point

aroyoredhat: Thanks I was not aware of it, I just ran

$ pnpm install --no-frozen-lockfile

I grabbed it from the test errors but not sure if you run a different command to regenerate the file

nindsimonv (approved):

PR #6029, opened 2026-05-19
PROJQUAY-11588: feat(oauth): allow any authenticated user to authorize OAuth apps

dronenb:
### Problem

Quay's OAuth authorization flow (`/oauth/authorize`) currently requires the authorizing user to be an **organization administrator** of the org that owns the OAuth app, or to have been explicitly pre-assigned a token by an org admin. This prevents third-party integrations (Backstage/Developer Hub, CI/CD tools, security scanners, etc.) from working because users outside the app-owning org cannot authorize the app.

This is inconsistent with how OAuth works on GitHub, GitLab, and other platforms, where any authenticated user can authorize any registered OAuth app regardless of who created it.

### Solution

Introduces a new feature flag `FEATURE_PUBLIC_OAUTH_APPS` (default: `True`) that, when enabled, removes the org-admin requirement from the OAuth **authorization** flow. Any authenticated user can authorize any registered OAuth app.

- App **management** (create, edit, delete) remains restricted to org admins — unchanged.
- The existing `FEATURE_ASSIGN_OAUTH_TOKEN` mechanism is preserved as a complementary opt-in flow.
- Setting `FEATURE_PUBLIC_OAUTH_APPS: false` in `config.yaml` restores the previous restrictive behavior.

### Changes

| File | Change |
|------|--------|
| `features/__init__.pyi` | Declare `PUBLIC_OAUTH_APPS` feature flag |
| `config.py` | Set default `True`, add to `CLIENT_WHITELIST` |
| `endpoints/web.py` | Wrap org-admin check on `/oauth/authorize` with feature flag guard |
| `data/model/oauth.py` | Wrap org-admin check in `get_token_response()` with feature flag guard; add `import features` |
| `data/model/test/test_oauth_public_apps.py` | 5 new test cases |

### Security Considerations

- **No scope escalation:** OAuth tokens are user-scoped. A token can only access resources the authorizing user already has permission to access. Opening authorization to all users does not grant any additional privileges beyond what the user themselves can do.
- **Invalid client IDs are still rejected** regardless of the feature flag.
- App management (CRUD) continues to require org admin privileges, preventing unauthorized modification of OAuth app settings (redirect URIs, secrets, etc.).

### Testing

- `test_non_admin_blocked_when_feature_disabled` — non-org-admin gets `unauthorized_client` when flag is off
- `test_non_admin_allowed_when_feature_enabled` — non-org-admin successfully gets a token when flag is on
- `test_org_admin_allowed_regardless_of_feature_flag` — org admin works in both states
- `test_token_assignment_still_works_when_feature_enabled` — token assignment consumed correctly alongside the flag
- `test_invalid_client_id_rejected_regardless_of_feature` — bad client_id always rejected

All existing OAuth tests (`data/model/test/test_oauth.py` — 21 tests) continue to pass with no regressions.

### DIFF
--- a/config.py
+++ b/config.py
@@ -999,7 +999,7 @@ class DefaultConfig(ImmutableConfig):
     ROBOTS_WHITELIST: Optional[List[str]] = []
 
     FEATURE_ASSIGN_OAUTH_TOKEN = True
-    FEATURE_PUBLIC_OAUTH_APPS = True
+    FEATURE_PUBLIC_OAUTH_APPS = False
     DEFAULT_NAMESPACE_AUTOPRUNE_POLICY: Optional[Dict[str, str]] = None
 
     # Allows users to set up notifications on image expiry, can remove flag once feature is tested
--- a/config.py
+++ b/config.py
@@ -54,6 +54,7 @@ CLIENT_WHITELIST = [
     "FOOTER_LINKS",
     "UI_DELAY_AFTER_WRITE_SECONDS",
     "FEATURE_ASSIGN_OAUTH_TOKEN",
+    "FEATURE_PUBLIC_OAUTH_APPS",
     "FEATURE_IMAGE_EXPIRY_TRIGGER",
     "FEATURE_AUTO_PRUNE",
     "DEFAULT_NAMESPACE_AUTOPRUNE_POLICY",
@@ -998,6 +999,7 @@ class DefaultConfig(ImmutableConfig):
     ROBOTS_WHITELIST: Optional[List[str]] = []
 
     FEATURE_ASSIGN_OAUTH_TOKEN = True
+    FEATURE_PUBLIC_OAUTH_APPS = True
     DEFAULT_NAMESPACE_AUTOPRUNE_POLICY: Optional[Dict[str, str]] = None
 
     # Allows users to set up notifications on image expiry, can remove flag once feature is tested
--- a/data/model/test/test_oauth_public_apps.py
+++ b/data/model/test/test_oauth_public_apps.py
@@ -24,7 +24,7 @@ def setup_app():
     return application, org
 
 
-def test_non_admin_blocked_when_feature_disabled(initialized_db):
+def test_non_admin_token_blocked_when_feature_disabled(initialized_db):
     """Non-org-admin cannot get a token when FEATURE_PUBLIC_OAUTH_APPS is disabled."""
     application, org = setup_app()
     non_admin = model.user.get_user("freshuser")
@@ -47,7 +47,29 @@ def test_non_admin_blocked_when_feature_disabled(initialized_db):
     assert "error=unauthorized_client" in response.headers["Location"]
 
 
-def test_non_admin_allowed_when_feature_enabled(initialized_db):
+def test_non_admin_code_blocked_when_feature_disabled(initialized_db):
+    """Non-org-admin cannot get a code when FEATURE_PUBLIC_OAUTH_APPS is disabled."""
+    application, org = setup_app()
+    non_admin = model.user.get_user("freshuser")
+
+    assert not is_org_admin(non_admin, org)
+
+    with patch("data.model.oauth.url_for", return_value="http://foo/bar/baz"):
+        with patch("data.model.oauth.features") as mock_features:
+            mock_features.PUBLIC_OAUTH_APPS = False
+            provider = MockDatabaseAuthorizationProvider(non_admin)
+            response = provider.get_authorization_code(
+                "code",
+                application.client_id,
+                REDIRECT_URI,
+                scope=READ_REPO.scope,
+            )
+
+    assert response.status_code == 302
+    assert "error=unauthorized_client" in response.headers["Location"]
+
+
+def test_non_admin_token_allowed_when_feature_enabled(initialized_db):
     """Non-org-admin can get a token when FEATURE_PUBLIC_OAUTH_APPS is enabled."""
     application, org = setup_app()
     non_admin = model.user.get_user("freshuser")
@@ -70,7 +92,30 @@ def test_non_admin_allowed_when_feature_enabled(initialized_db):
     assert "access_token=" in response.headers["Location"]
 
 
-def test_org_admin_allowed_regardless_of_feature_flag(initialized_db):
+def test_non_admin_code_allowed_when_feature_enabled(initialized_db):
+    """Non-org-admin can get a code when FEATURE_PUBLIC_OAUTH_APPS is enabled."""
+    application, org = setup_app()
+    non_admin = model.user.get_user("freshuser")
+
+    assert not is_org_admin(non_admin, org)
+
+    with patch("data.model.oauth.url_for", return_value="http://foo/bar/baz"):
+        with patch("data.model.oauth.features") as mock_features:
+            mock_features.PUBLIC_OAUTH_APPS = True
+            provider = MockDatabaseAuthorizationProvider(non_admin)
+            response = provider.get_authorization_code(
+                "code",
+                application.client_id,
+                REDIRECT_URI,
+                scope=READ_REPO.scope,
+            )
+
+    assert response.status_code == 302
+    assert "error" not in response.headers["Location"]
+    assert "code=" in response.headers["Location"]
+
+
+def test_org_admin_token_allowed_regardless_of_feature_flag(initialized_db):
     """Org admin can always get a token, regardless of the flag."""
     application, org = setup_app()
     admin_user = model.user.get_user("devtable")
@@ -92,6 +137,29 @@ def test_org_admin_allowed_regardless_of_feature_flag(initialized_db):
         assert response.status_code == 302
         assert "error" not in response.headers["Location"]
         assert "access_token=" in response.headers["Location"]
+
+
+def test_org_admin_code_allowed_when_feature_disabled(initialized_db):
+    """Org admin can always get a code, regardless of the flag."""
+    application, org = setup_app()
+    admin_user = model.user.get_user("devtable")
+
+    assert is_org_admin(admin_user, org)
+
+    with patch("data.model.oauth.url_for", return_value="http://foo/bar/baz"):
+        with patch("data.model.oauth.features") as mock_features:
+            mock_features.PUBLIC_OAUTH_APPS = False
+            provider = MockDatabaseAuthorizationProvider(admin_user)
+            response = provider.get_authorization_code(
+                "code",
+                application.client_id,
+                REDIRECT_URI,
+                scope=READ_REPO.scope,
+            )
+
+    assert response.status_code == 302
+    assert "error" not in response.headers["Location"]
+    assert "code=" in response.headers["Location"]
 
 
 def test_token_assignment_still_works_when_feature_enabled(initialized_db):
@@ -123,8 +191,8 @@ def test_token_assignment_still_works_when_feature_enabled(initialized_db):
     assert model.oauth.get_token_assignment(token_assignment.uuid, non_admin, org) is None
 
 
-def test_invalid_client_id_rejected_regardless_of_feature(initialized_db):
-    """Invalid client_id is always rejected, even with feature enabled."""
+def test_invalid_client_id_rejected_locally_regardless_of_feature(initialized_db):
+    """Invalid client_id is always rejected locally, even with feature enabled."""
     non_admin = model.user.get_user("freshuser")
 
     with patch("data.model.oauth.url_for", return_value="http://foo/bar/baz"):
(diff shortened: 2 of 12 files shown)


jbpratt:
Hey @dronenb ! Thanks for the contribution, I'll give a full review asap. Initial pass looks good to me. Could we add some basic e2e coverage to `web/playwright/e2e/organization/oauth-applications.spec.ts`?

/ok-to-test

dronenb:
@jbpratt I added an E2E test, let me know what you think

web/playwright/e2e/organization/oauth-applications.spec.ts
```diff
@@ -180,4 +181,32 @@ test.describe('OAuth Applications', {tag: ['@organization']}, () => {
       expect(updatedApp?.client_secret).not.toBe(secretBefore);
     }
   });
+
+  test('non-admin user can authorize an OAuth app (PUBLIC_OAUTH_APPS)', async ({
```
jbpratt: @dronenb let's add a tag here filtering on your added feature flag (`feature:PUBLIC_OAUTH_APPS`), you'll need to update the auto skip fixture to include this flag for type safety. That way the tests can pass be skipped if the feature flag is disabled

features/__init__.pyi
```diff
@@ -223,6 +223,7 @@ REFERRERS_API: FeatureNameValue
 
 SUPERUSERS_FULL_ACCESS: FeatureNameValue
 ASSIGN_OAUTH_TOKEN: FeatureNameValue
+PUBLIC_OAUTH_APPS: FeatureNameValue
```
jbpratt, line 229: Need to add this to the util/config/schema.py and the corresponding Go test file for schema drift (we are planning a rewrite 😄 )

config.py
```diff
@@ -968,6 +969,7 @@ def create_transaction(db):
     ROBOTS_WHITELIST: Optional[List[str]] = []
 
     FEATURE_ASSIGN_OAUTH_TOKEN = True
+    FEATURE_PUBLIC_OAUTH_APPS = True
```
tlwu2013: Please change the default to `False`:

```python
FEATURE_PUBLIC_OAUTH_APPS = False

File: util/config/schema.py [hunk 4]
```diff
@@ -1745,6 +1745,11 @@
             "description": "Allows organization administrators to assign OAuth tokens to other users",
             "x-example": False,
         },
+        "FEATURE_PUBLIC_OAUTH_APPS": {
+            "type": "boolean",
+            "description": "When enabled, any authenticated user can authorize OAuth applications regardless of organization membership",
```
tlwu2013: Please help expand the description to include a note about security considerations:

> "description": "When enabled, any authenticated user can authorize OAuth applications regardless of organization membership. Administrators should review the security considerations in the documentation before enabling.",

util/config/schema.py
```diff
@@ -1745,6 +1745,11 @@
...
+            "x-example": True,
```
tlwu2013: Also changed `"x-example": True` to `"x-example": False` to match the recommended default.

tlwu2013:
Thanks for this PR.  We've reviewed the use cases (i.e., Backstage + internal CLI tool for onboarding) and are prepared to accept this feature.

Two changes are identified:

1. *default value*: `FEATURE_PUBLIC_OAUTH_APPS` is default to `False`, not `True`
- All existing deployments aren't affected on upgrade
- Admins should explicitly opt-in after understanding the security implications
- Security-by-default posture is maintained

2. *Config schema documentation*: please add a note pointing admins to review security considerations

I've added line-level comments on the specific locations.

Once these changes are made (and passed testing), to me, the change will be good to go.

dronenb:
I have rebased the PR and made the requested changes - should be g2g now

endpoints/web.py
```diff
@@ -745,13 +745,16 @@ def request_authorization_code():
...
 
-    # check if user is org admin, if not check for user_assignment_id, then check that user belongs that assignment, if none exit with 401
-    if (
-        not is_org_admin(current_user.db_user(), oauth_app.organization)
-        and get_token_assignment(assignment_uuid, current_user.db_user(), oauth_app.organization)
-        is None
-    ):
-        abort(403)
+    # If PUBLIC_OAUTH_APPS is disabled, require org admin or a pre-assigned token.
+    if not features.PUBLIC_OAUTH_APPS:
```
jbpratt, line 758: @dronenb I think we need a new helper and a check here for `is_org_member` (and in `oauth.py`), this opens it up to anyone 🙀

web/playwright/e2e/organization/oauth-applications.spec.ts
```diff
@@ -180,4 +181,32 @@ test.describe('OAuth Applications', {tag: ['@organization']}, () => {
       expect(updatedApp?.client_secret).not.toBe(secretBefore);
     }
   });
+
+  test('non-admin user can authorize an OAuth app (PUBLIC_OAUTH_APPS)', {tag: ['@feature:PUBLIC_OAUTH_APPS']}, async ({
```
jbpratt: @dronenb while we are at it lets also add a test to validate non-org members can't create them - you can toggle this on in `local-dev/stack/config.yaml` for CI (pre-commit will block it, just `--no-verify` around it)

endpoints/web.py
dronenb: @jbpratt This is intentional — authorizing an OAuth app should be open to any authenticated user, not just org members. Here's why:

How OAuth authorization works on major platforms:

GitHub: Any authenticated user can authorize any registered OAuth App. You don't need to be a member of the org that owns the app. The app's org is just the "publisher" — authorization grants the app access to the authorizing user's resources, scoped by the requested permissions.
GitLab: Same model — any user can authorize any OAuth application regardless of who registered it.
Bitbucket, Azure DevOps, etc.: Same pattern.
Why this is safe:

Authorization doesn't escalate privileges. The resulting token can only access what that user already has access to, bounded by the requested scopes. A non-org-member authorizing the app gets a token that can only touch their own repos/resources.
The org-admin restriction on app management is preserved. Only org admins can create, edit, or delete OAuth apps — that's unchanged.
The OAuth 2.0 spec (RFC 6749) defines the Resource Owner (the user) as the entity who grants access. Requiring org membership to authorize is non-standard and breaks legitimate integrations.
The use case driving this:

Tools like Backstage, internal CLI tools, and CI/CD scanners need to onboard users across the entire Quay instance — not just members of the org that registered the app. The current restriction makes it impossible for these tools to work without pre-assigning tokens to every user, which doesn't scale, or adding every user into an org, which also doesn't scale.

An `is_org_member` check would still break these use cases since users of the consuming tool typically aren't members of the infrastructure org that registered the OAuth app.

web/playwright/e2e/organization/oauth-applications.spec.ts
dronenb: done

dronenb:
This PR has been rebased and feedback addressed - PTAL when you can. Thanks!

jbpratt:
> This PR has been rebased and feedback addressed - PTAL when you can. Thanks!

@dronenb legend, thank you! I'll take a look ASAP. pre-commit failures should be addressed, postgres integration is flaky which I'm working on separately)

dronenb:
ESLint issue should be fixed, which will fix the `pre-commit` CI errors.

dronenb:
@jbpratt gentle nudge on this - would love to get this merged as soon as we can. Let me know if there is any additional feedback!

jbpratt:
@dronenb oops so sorry, reviewing now and we can try to get this merged asap!

jbpratt:
Pushed follow-up commit `ee7c96e2b`.

Changes:
- Gate code grants with `FEATURE_PUBLIC_OAUTH_APPS`.
- Validate assigned OAuth tokens against the exact request before issuance/consumption.
- Return local OAuth errors before redirecting invalid-client failures.

Verification:
- `pre-commit run --files ...` passed, including Black, isort, and flake8.
- Targeted pytest passed: `91 passed`.

@dronenb please take a look at https://github.com/quay/quay/pull/6029/commits/ee7c96e2be4dc5aba827c92136c4e3ec06a79f42 to ensure I haven't deviated from your original goal!

jbpratt:
resolving the Go lint issue in a separate PR, it slipped thru - I'll rebase this after it is fixed

bcaton85:
**Default value inconsistency**: The JIRA ticket and PR description both say `FEATURE_PUBLIC_OAUTH_APPS` defaults to `True`, but `config.py:1002` sets it to `False`. The code is correct -- removing the org-admin requirement should be an explicit opt-in, not a surprise on upgrade. Please update the JIRA and PR description to reflect the actual default of `False`.

Note: the endpoint test `test_request_authorization_code_assignment_must_match_app` also depends on the default being `False`.

data/model/test/test_oauth_public_apps.py
```diff
@@ -0,0 +1,211 @@
...
+            response = provider.get_token_response(
+                "token",
+                "nonexistent_client_id",
+                REDIRECT_URI,
+                scope=READ_REPO.scope,
+            )
+
+    assert response.status_code == 400
+    assert "Location" not in response.headers
+    assert "unauthorized_client" in response.raw.getvalue()
```
bcaton85, line 211: The code flow tests verify that `get_authorization_code()` returns a redirect with `code=`, but there is no test that exchanges that code for an access token via POST `/oauth/access_token`. Adding an integration test for the full `code -> token` exchange would confirm the generated authorization codes are actually redeemable and that scope/client_secret validation works end-to-end.

data/model/test/test_oauth_public_apps.py
```diff
@@ -0,0 +1,211 @@
...
+            response = provider.get_authorization_code(
+                "code",
+                application.client_id,
+                REDIRECT_URI,
+                scope=READ_REPO.scope,
+            )
+
+    assert response.status_code == 302
+    assert "error" not in response.headers["Location"]
+    assert "code=" in response.headers["Location"]
```
bcaton85, line 115: This test verifies `code=` appears in the redirect, but does not check that the `state` parameter is echoed back. RFC 6749 Section 4.1.2 requires `state` to be returned unchanged -- it is the client's CSRF protection. Consider passing `state="xyzzy"` and asserting `state=xyzzy` appears in the redirect Location.

web/playwright/e2e/organization/oauth-applications.spec.ts
```diff
@@ -180,4 +181,54 @@ test.describe('OAuth Applications', {tag: ['@organization']}, () => {
...
+      ).toBeVisible();
+
+      // Click "Authorize Application" to grant access
+      await page.getByRole('button', {name: 'Authorize Application'}).click();
+
+      // Should redirect to the local OAuth handler with the token in the URL fragment
+      await page.waitForURL(/\/oauth\/localapp/);
+      const url = page.url();
+      expect(url).toContain('access_token=');
+    },
```
bcaton85, line 228: This E2E test covers the implicit flow (`response_type=token`), but the authorization code flow (`response_type=code`) -- which is what server-side integrations like Backstage and CI/CD tools would use -- has no E2E coverage. Consider adding a parallel test with `response_type=code` that verifies the redirect contains `code=` instead of `access_token=`.

data/model/test/test_oauth_public_apps.py
jbpratt: Addressed in `35391cdad`: added `test_public_oauth_code_grant_can_be_exchanged_for_token`, including wrong-secret rejection plus successful `/oauth/access_token` exchange and persisted token assertions.

web/playwright/e2e/organization/oauth-applications.spec.ts
jbpratt: Addressed in `35391cdad`: added Playwright `response_type=code` coverage for public OAuth apps, asserting the redirect contains `code=` and not `access_token=`.

data/model/test/test_oauth_public_apps.py
jbpratt: Addressed in `35391cdad`: `test_authorize_success` now passes `state="xyzzy"` and asserts it is echoed on code redirects. The public code-grant exchange test also asserts state on the issued redirect.

jbpratt:
@bcaton85 pushed `35391cdad` for the test coverage comments. On the default-value note: code and tests intentionally use `FEATURE_PUBLIC_OAUTH_APPS = False`; the PR/JIRA text still needs to be updated to reflect the opt-in default.

bcaton85:
we're all just robots talking to each other

jbpratt:
> we're all just robots talking to each other

jbpratt (approved):

jbpratt:
Thanks again @dronenb - this will go out in 3.18 by the end of the month (roughly)!

PR #6050, opened 2026-05-26
PROJQUAY-11411: test(oci): add integration tests for cosign signatures

dtzhou2:
Added 8 test cases for signature workflow verifcation in test/registry/test_cosign_signatures.py with helper functions in test/registry/cosign_test_helpers.py. The test cases start from the basics of pushing an image with a signature and verifying that the signature returns back on a pull as well as pushing a signature artifact and checking that you can discover it via the API. Additionally reformatted the protocols to support artifacts natively instead of only using images.

### DIFF
--- a/test/registry/test_cosign_signatures.py
+++ b/test/registry/test_cosign_signatures.py
@@ -1,0 +1,396 @@
+# pylint: disable=W0401, W0621, W0613, W0614
+"""
+Integration tests for cosign signatures, OCI referrers API, and OCI artifacts.
+
+Run:
+  TEST=true PYTHONPATH="." pytest test/registry/test_cosign_signatures.py -v
+  make registry-test  # includes this file
+"""
+
+import json
+
+import pytest
+
+from test.fixtures import *
+from test.registry.cosign_test_helpers import (
+    COSIGN_ARTIFACT_TYPE,
+    IN_TOTO_ARTIFACT_TYPE,
+    IN_TOTO_LAYER_TYPE,
+    build_cosign_signature_manifest,
+    build_helm_chart_manifest,
+    build_in_toto_artifact_manifest,
+    build_oci_artifact_manifest,
+    delete_manifest,
+    get_referrers,
+    parse_and_validate_referrers_index,
+    referrer_digests,
+)
+from test.registry.fixtures import *
+from test.registry.liveserverfixture import *
+from test.registry.protocol_fixtures import basic_images, jwk, minimal_oci_artifact
+from test.registry.protocol_v2 import V2Protocol
+from test.registry.protocols import Artifact, ProtocolOptions
+
+pytestmark = pytest.mark.oci
+
+# OCI artifact pushes often reuse config/layer blobs already present in the repo.
+_ARTIFACT_PUSH_OPTIONS = ProtocolOptions()
+_ARTIFACT_PUSH_OPTIONS.skip_head_checks = True
+
+
+@pytest.fixture(params=["oci"])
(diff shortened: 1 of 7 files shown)


test/registry/test_cosign_signatures.py
```diff
@@ -0,0 +1,396 @@
...
+
+    lookup_cached_referrers_for_manifest must store dicts (see lookup_cached_manifest_by_digest);
+    until that is implemented, bypass the cache layer here.
+    """
+    from data.registry_model.registry_oci_model import OCIModel
+
+    def lookup_referrers_uncached(self, model_cache, repository_ref, manifest, artifact_type=None):
+        return self.lookup_referrers_for_manifest(repository_ref, manifest, artifact_type)
+
+    monkeypatch.setattr(OCIModel, "lookup_cached_referrers_for_manifest", lookup_referrers_uncached)
```
bcaton85: why are we using this monkeypatch?

test/registry/test_cosign_signatures.py
```diff
@@ -0,0 +1,396 @@
...
+        "cosign-helm-chart",
+        str(chart.digest),
+        credentials=("devtable", "password"),
+    ).manifest
+    assert pulled.config_media_type == "application/vnd.cncf.helm.config.v1+json"
+    # blob_digests includes the config blob and each layer blob.
+    assert len(list(pulled.blob_digests)) == 2
+
+
+def test_in_toto_artifact_type_push_pull(pusher, liveserver_session, app_reloader):
```
bcaton85: PROJQUAY-11411 Test 7 requires verifying that a custom artifact type can be registered and stored. This test uses the already registered application/vnd.in-toto+json type from the default ALLOWED_OCI_ARTIFACT_TYPES, so it would still pass if custom type registration were broken. Please register a unique test artifact/config media-type combination and verify that it can be pushed and pulled.

test/registry/test_cosign_signatures.py
```diff
@@ -0,0 +1,396 @@
...
+        liveserver_session,
+        "devtable",
+        "cosign-in-toto-artifact",
+        str(artifact.digest),
+        credentials=("devtable", "password"),
+    ).manifest
+    assert pulled.artifact_type == IN_TOTO_ARTIFACT_TYPE
+
+
+def test_signature_deletion_removes_referrer(
```
bcaton85: This does not cover PROJQUAY-11411 Test 8. The criterion asks to delete the base image and verify signature preservation, or to exercise signature GC behavior. This test instead deletes the signature tag/digest, confirms the referrer remains before GC, and confirms the base image is still pullable. The function name also says the referrer is removed even though the assertion verifies the opposite.

bcaton85 (commented):
Just need some clarifications and possible changes. PROJQUAY-605 also needs to be replaced with PROJQUAY-11411

test/registry/test_cosign_signatures.py
dtzhou2: Resolved and recreated the last test case. Requirements were not cleraly stated in my version of the ticket

test/registry/test_cosign_signatures.py
dtzhou2: Recreated test case using custom application type

test/registry/test_cosign_signatures.py
dtzhou2: Deleted monkey patch and tested it to verify that it was uneeded

bcaton85 (dismissed):

bcaton85 (approved):

PR #6057, opened 2026-05-27
PROJQUAY-11235: fix(templates): remove hardcoded external CDN references

quay-devel:
## Summary
- Remove hardcoded external CDN references from standalone HTML templates that bypass the `USE_CDN` config flag
- Replace with `external_styles` Jinja variable in 3 templates, remove entirely from static `502.html`
- Add 8 regression tests to prevent reintroduction

## Root Cause / Rationale
Four standalone templates (`generatedtoken.html`, `message.html`, `error.html`, `static/502.html`) hardcoded `<link>` tags pointing to `fonts.googleapis.com` and `netdna.bootstrapcdn.com`. These templates predate the `USE_CDN` / `external_libraries.py` localization mechanism and were never updated. Since `USE_CDN=False` is the default for all enterprise deployments, these external requests always fail in air-gapped environments, causing 30-40 second page load timeouts.

The main app template (`base.html`) already handles this correctly via `{% for style_url in external_styles %}`, and `render_page_template()` already passes `external_styles` to all templates.

## Changes
- `templates/generatedtoken.html` — replaced 3 hardcoded CDN links with `external_styles` loop
- `templates/message.html` — replaced 3 hardcoded CDN links with `external_styles` loop
- `templates/error.html` — replaced 3 hardcoded CDN links with `external_styles` loop
- `static/502.html` — removed 3 CDN links (page uses only inline CSS, no Bootstrap/FA dependency)
- `endpoints/test/test_template_external_cdn.py` — 8 regression tests (new file)

## Test Plan
- [x] Type checking passes
- [ ] Frontend tests pass

### Regression test details
- `TestNoHardcodedCDN`: 8 tests verify no CDN URLs remain and Jinja templates use `external_styles`
- `TestTemplateRendering`: 3 Flask rendering tests verify templates render with local style URLs (requires full fixture, runs in CI)
- Tests confirmed to **fail without the fix** (3 CDN matches per file on pre-fix code)
- Existing nginx config tests (10/10) and Playwright 502 test continue to pass

## JIRA
https://issues.redhat.com/browse/PROJQUAY-11235

## Backport
- [ ] Backport required
- [x] No backport needed

## Automation
- **Session ID**: session-b20b63d0-379f-4456-b887-cf7372ad9974

### DIFF
--- a/endpoints/test/test_template_external_cdn.py
+++ b/endpoints/test/test_template_external_cdn.py
@@ -71,77 +71,78 @@ class TestNoHardcodedCDN:
 try:
     from test.fixtures import *
 
-    class TestTemplateRendering:
-        @pytest.mark.parametrize(
-            "template_name",
-            [
-                "generatedtoken.html",
-                "message.html",
-            ],
-        )
-        def test_template_renders_with_local_styles(self, template_name, app):
-            from flask import render_template
-
-            local_styles = [
-                "/static/ldn/font-awesome.css-abc123",
-                "/static/ldn/bootstrap.min.css-def456",
-            ]
-
-            config_set = {
-                "BRANDING": {"logo": "/static/img/quay_logo.png"},
-            }
-
-            with app.app_context():
-                html = render_template(
-                    template_name,
-                    external_styles=local_styles,
-                    config_set=config_set,
-                    onprem=True,
-                    message="Test message",
-                )
-
-            for style_url in local_styles:
-                assert (
-                    style_url in html
-                ), f"{template_name} did not render local style URL {style_url}"
-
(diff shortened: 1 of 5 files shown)


quay-devel:
Addressed CodeRabbit feedback in c57a944: replaced the silent `try/except ImportError: pass` with `pytest.mark.skipif` on the `TestTemplateRendering` class. File-content tests (`TestNoHardcodedCDN`, 8 tests) always run; rendering tests cleanly skip when Flask fixtures are unavailable.

jbpratt (approved):

sridipta (approved):

PR #6067, opened 2026-05-28
PROJQUAY-11636: feat(ui): implement invite code flow for new React UI

quay-devel:
## Summary

Implement the invite code flow for the new React UI (`FEATURE_INVITE_ONLY_USER_CREATION`) and fix email invite support on the team members page.

## Quay Local Validation Testing
image
image
image

### Invite Code Flow (new feature for React UI)

- **Root cause:** The new React UI had no `/confirminvite` route and never forwarded `invite_code` to the backend, making `FEATURE_INVITE_ONLY_USER_CREATION` non-functional in the new UI.
- Add a `/confirminvite` route (`ConfirmInvite` component) that accepts the invite for already-authenticated users, or shows sign-in/create-account options for unauthenticated users with the `code` query param preserved.
- Forward `invite_code` in `POST /api/v1/signin` and `POST /api/v1/user/` when a `code` param is present in the URL.
- Show the "Create account" link on the Sign In page even when `INVITE_ONLY_USER_CREATION` is enabled, provided an invite code is in the URL.
- Show an invitation banner on the Create Account page when arriving via an invite link.
- Add `/confirminvite` to the axios auth-page list to prevent 401 redirect loops for unauthenticated users.

### Email Invite Support on Team Members Page (bug fixes)

- **Bug 1 — Blank "Team member" column:** The backend returns `{email, kind: "invite"}` with no `name` field for raw email invites. The UI only rendered `teamMember.name`, leaving the column blank. Fixed by adding `email?` to the `ITeamMember` interface (and making `name` optional) and using a `getMemberDisplayName()` helper that falls back to `email` across table rendering, sorting, filtering, selection, and deletion.
- **Bug 2 — No invite email sent:** The new UI only called `PUT /members/{username}` (for existing users). The separate `PUT /invite/{email}` endpoint was never called. Added `inviteTeamMemberByEmailForOrg()` resource function, `useInviteTeamMemberByEmail` hook, and an "Invite by email" section in the Add Member drawer (gated behind the `MAILING` feature flag).
- **Bug 3 — No success notification:** The old UI showed "E-mail address X was invited to join the team" but the new UI showed nothing. Added success/error alerts matching the old UI message format, using per-call mutation callbacks to avoid stale closure on the email input.
- **Delete email invites:** Added `deleteTeamMemberEmailInviteForOrg()` resource and `useDeleteEmailInvite` hook. The delete handler now routes email invites (`kind === 'invite'`) to `DELETE /invite/{email}` instead of `DELETE /members/{name}`.

### Nginx & Backend

- Fix nginx `server-base.conf.jnj`: change `/confirm` to `/confirm$` to prevent `/confirminvite` from matching the Flask proxy rule (so it reaches the React SPA).
- Fix `endpoints/web.py`: use `request.values.get("code", "")` to gracefully handle missing code param instead of raising KeyError.

### Test Fixtures & Infrastructure

- Set React UI cookie (`setReactUICookie`) in the base `unauthenticatedPage` Playwright fixture (consistent with all authenticated fixtures).
- Add `inviteTeamMemberByEmail` / `deleteTeamEmailInvite` methods to Playwright `ApiClient` and `teamEmailInvite` helper to `TestApi` with auto-cleanup.

## Files Changed (31 files, +1503 / -77)

| Area | Files |
|------|-------|
| **New components** | `ConfirmInvite.tsx` |
| **Routing** | `NavigationPath.tsx`, `StandaloneMain.tsx`, `App.tsx` |
| **Auth flow** | `Signin.tsx`, `CreateAccount.tsx`, `axios.ts` |
| **API resources** | `AuthResource.ts`, `UserResource.ts`, `TeamResources.ts`, `MembersResource.ts` |
| **Hooks** | `UseMembers.ts`, `UseCreateAccount.ts` |
| **Team members UI** | `AddNewTeamMemberDrawer.tsx`, `ManageMembersList.tsx` |
| **Backend** | `endpoints/web.py`, `conf/nginx/server-base.conf.jnj` |
| **Vitest unit tests** | `ConfirmInvite.test.tsx`, `ManageMembersList.test.tsx`, `Signin.test.tsx`, `CreateAccount.test.tsx`, `UseMembers.test.ts`, `MembersResource.test.ts`, `AuthResource.test.ts`, `TeamResources.test.ts`, `UserResource.test.ts` |
| **Playwright E2E tests** | `confirm-invite.spec.ts` (9 tests), `team-members.spec.ts` (4 tests) |
| **Playwright infra** | `fixtures.ts`, `utils/api/client.ts` |
| **Backend tests** | `test_endpoints.py` |

## Test Plan

**Invite code flow:**
- [x] Enable `FEATURE_INVITE_ONLY_USER_CREATION: true` in Quay config
- [x] Invite a user to a team (sends `/confirminvite?code=xxx` email link)
- [x] Click the invite link as an unauthenticated user → `ConfirmInvite` page shows sign-in / create-account options
- [x] Click "Sign in" → arrives at `/signin?code=xxx`, signs in → invite is accepted, redirected to org/team page
- [x] Click "Create account" → arrives at `/createaccount?code=xxx`, sees invitation banner, creates account → invite is accepted
- [x] Click the invite link as an already-authenticated user → invite is accepted immediately, redirected to org/team page

**Email invite on team members page:**
- [x] Navigate to New UI → Organization → Teams and membership → team page
- [x] Verify existing email-invited members show their email address in the "Team member" column
- [x] Use "Add new member" drawer → enter email → click "Invite" → verify success alert and invite email is sent
- [x] Delete an email-invited member → verify correct API endpoint called and success alert shown
- [x] Switch to "Invited" tab → search by email address → verify filtering works

**Automated tests:**
- [x] All vitest unit tests pass
- [x] Playwright E2E tests pass for confirm-invite flow (9 tests)
- [x] Playwright E2E tests pass for email invite flow (4 tests, `@feature:MAILING`)

Fixes: https://redhat.atlassian.net/browse/PROJQUAY-11636

### DIFF
--- a/web/playwright/e2e/auth/confirm-invite.spec.ts
+++ b/web/playwright/e2e/auth/confirm-invite.spec.ts
@@ -1,0 +1,160 @@
+import {test, expect} from '../../fixtures';
+
+/**
+ * Confirm Invite tests verify the /confirminvite route added for
+ * PROJQUAY-11636. Tests that need a real invite code use the
+ * @feature:INVITE_ONLY_USER_CREATION tag.
+ */
+
+test.describe('Confirm Invite Page', {tag: ['@auth']}, () => {
+  test('shows error when no invite code is in the URL', async ({
+    unauthenticatedPage,
+  }) => {
+    await unauthenticatedPage.goto('/confirminvite');
+
+    await expect(
+      unauthenticatedPage.getByTestId('confirm-invite-error'),
+    ).toBeVisible({
+      timeout: 10_000,
+    });
+    await expect(
+      unauthenticatedPage.getByText(/No invite code/i),
+    ).toBeVisible();
+  });
+
+  test('shows sign-in and create-account options for unauthenticated user', async ({
+    unauthenticatedPage,
+  }) => {
+    await unauthenticatedPage.goto('/confirminvite?code=somefakecode');
+
+    await expect(
+      unauthenticatedPage.getByTestId('confirm-invite-unauthenticated'),
+    ).toBeVisible({timeout: 10_000});
+
+    await expect(
+      unauthenticatedPage.getByTestId('confirm-invite-signin-btn'),
+    ).toBeVisible();
+    await expect(
+      unauthenticatedPage.getByTestId('confirm-invite-create-account-btn'),
+    ).toBeVisible();
+  });
+
+  test('sign-in and create-account buttons preserve invite code in URL', async ({
+    unauthenticatedPage,
+  }) => {
+    await unauthenticatedPage.goto('/confirminvite?code=myinvitecode');
+
+    await expect(
+      unauthenticatedPage.getByTestId('confirm-invite-signin-btn'),
+    ).toBeVisible({timeout: 10_000});
+
+    const signinHref = await unauthenticatedPage
+      .getByTestId('confirm-invite-signin-btn')
+      .getAttribute('href');
+    expect(signinHref).toContain('code=myinvitecode');
+
+    const createHref = await unauthenticatedPage
+      .getByTestId('confirm-invite-create-account-btn')
+      .getAttribute('href');
+    expect(createHref).toContain('code=myinvitecode');
+  });
+
+  test('sign-in button navigates to /signin with code param', async ({
+    unauthenticatedPage,
+  }) => {
+    await unauthenticatedPage.goto('/confirminvite?code=nav-test-code');
+
+    await expect(
+      unauthenticatedPage.getByTestId('confirm-invite-signin-btn'),
+    ).toBeVisible({timeout: 10_000});
+
+    await unauthenticatedPage.getByTestId('confirm-invite-signin-btn').click();
+
+    await expect(unauthenticatedPage).toHaveURL(/\/signin\?code=nav-test-code/);
+  });
+
+  test('create-account button navigates to /createaccount with code param', async ({
+    unauthenticatedPage,
+  }) => {
+    await unauthenticatedPage.goto('/confirminvite?code=nav-create-code');
+
+    await expect(
+      unauthenticatedPage.getByTestId('confirm-invite-create-account-btn'),
+    ).toBeVisible({timeout: 10_000});
+
+    await unauthenticatedPage
+      .getByTestId('confirm-invite-create-account-btn')
+      .click();
+
+    await expect(unauthenticatedPage).toHaveURL(
+      /\/createaccount\?code=nav-create-code/,
+    );
+
+    await expect(
+      unauthenticatedPage.getByTestId('invite-code-alert'),
+    ).toBeVisible();
+    await expect(
+      unauthenticatedPage.getByText("You've been invited to join a team"),
+    ).toBeVisible();
+  });
+
+  test('authenticated user with invalid code sees error', async ({
+    authenticatedPage,
+  }) => {
+    await authenticatedPage.goto('/confirminvite?code=invalid-code-xyz');
+
+    await expect(
+      authenticatedPage.getByTestId('confirm-invite-error'),
+    ).toBeVisible({timeout: 10_000});
+  });
+
+  test('create-account page shows invite banner when code is present', async ({
+    unauthenticatedPage,
+  }) => {
+    await unauthenticatedPage.goto('/createaccount?code=testcode');
+
+    await expect(
+      unauthenticatedPage.getByTestId('invite-code-alert'),
+    ).toBeVisible({
+      timeout: 10_000,
+    });
+    await expect(
+      unauthenticatedPage.getByText("You've been invited to join a team"),
+    ).toBeVisible();
+  });
+
+  test(
+    'create-account page preserves invite code in sign-in link',
+    {
+      tag: '@auth:Database',
+    },
+    async ({unauthenticatedPage}) => {
+      await unauthenticatedPage.goto('/createaccount?code=link-code');
+
+      const signinLink = unauthenticatedPage.getByRole('link', {
+        name: 'Sign in',
+      });
+      await expect(signinLink).toBeVisible({timeout: 10_000});
+
+      const href = await signinLink.getAttribute('href');
+      expect(href).toContain('code=link-code');
+    },
+  );
+
+  test(
+    'signin page shows create-account link when invite code is present',
+    {tag: '@feature:INVITE_ONLY_USER_CREATION'},
+    async ({unauthenticatedPage}) => {
+      await unauthenticatedPage.goto('/signin?code=somecode');
+
+      await expect(
+        unauthenticatedPage.getByTestId('signin-create-account-link'),
+      ).toBeVisible({timeout: 10_000});
+
+      const href = await unauthenticatedPage
+        .getByTestId('signin-create-account-link')
+        .getAttribute('href');
+      expect(href).toContain('code=somecode');
+    },
+  );
+});
--- a/web/playwright/e2e/organization/team-members.spec.ts
+++ b/web/playwright/e2e/organization/team-members.spec.ts
@@ -1,4 +1,4 @@
-import {test, expect} from '../../fixtures';
+import {test, expect, uniqueName} from '../../fixtures';
 import {TEST_USERS} from '../../global-setup';
 import type {Page} from '@playwright/test';
 
@@ -256,3 +256,111 @@ test.describe('Manage Team Members', {tag: ['@organization']}, () => {
     );
   });
 });
+
+test.describe(
+  'Team Email Invites',
+  {tag: ['@organization', '@feature:MAILING']},
+  () => {
+    test('email invite shows in team members table', async ({
+      authenticatedPage,
+      api,
+    }) => {
+      const org = await api.organization('testorg');
+      const team = await api.team(org.name, 'emailshow');
+      const email = `${uniqueName('show')}@example.com`;
+
+      await api.teamEmailInvite(org.name, team.name, email);
+
+      await navigateToManageTeamMembers(authenticatedPage, org.name, team.name);
+
+      await expect(authenticatedPage.getByTestId(email)).toBeVisible({
+        timeout: 10_000,
+      });
+      await expect(
+        authenticatedPage
+          .getByTestId(email)
+          .locator('xpath=ancestor::tr')
+          .locator('[data-label="Account"]'),
+      ).toContainText('(Invited)');
+    });
+
+    test('can invite by email via drawer', async ({authenticatedPage, api}) => {
+      const org = await api.organization('testorg');
+      const team = await api.team(org.name, 'emaildrawer');
+      const email = `${uniqueName('drawer')}@example.com`;
+
+      await navigateToManageTeamMembers(authenticatedPage, org.name, team.name);
+
+      await authenticatedPage.getByTestId('add-new-member-button').click();
+
+      const emailInput = authenticatedPage.getByTestId('email-invite-input');
+      const inviteBtn = authenticatedPage.getByTestId('invite-by-email-btn');
+
+      // Button should be disabled with no input
+      await expect(inviteBtn).toBeDisabled();
+
+      await emailInput.fill(email);
+      await expect(inviteBtn).toBeEnabled();
+      await inviteBtn.click();
+
+      const alert = authenticatedPage
+        .locator('.pf-v6-c-alert.pf-m-success')
+        .last();
+      await expect(alert).toContainText('was invited to join the team', {
+        timeout: 10_000,
+      });
+      await expect(alert).toContainText(email);
+    });
+
+    test('invited tab shows only invited members', async ({
+      authenticatedPage,
+      api,
+    }) => {
+      const org = await api.organization('testorg');
+      const team = await api.team(org.name, 'emailfilter');
+      const email = `${uniqueName('filter')}@example.com`;
+
+      await api.teamMember(org.name, team.name, TEST_USERS.user.username);
+      await api.teamEmailInvite(org.name, team.name, email);
+
+      await navigateToManageTeamMembers(authenticatedPage, org.name, team.name);
+
+      await authenticatedPage.getByTestId('Invited').click();
+
+      const paginationInfo = authenticatedPage
+        .locator('.pf-v6-c-pagination__total-items')
+        .first();
+
+      await expect(paginationInfo).toContainText('1 - 1 of 1', {
+        timeout: 10_000,
+      });
+      await expect(authenticatedPage.getByTestId(email)).toBeVisible();
+    });
+
+    test('can revoke email invite via delete icon', async ({
+      authenticatedPage,
+      api,
+    }) => {
+      const org = await api.organization('testorg');
+      const team = await api.team(org.name, 'emailrevoke');
+      const email = `${uniqueName('revoke')}@example.com`;
+
+      await api.teamEmailInvite(org.name, team.name, email);
+
+      await navigateToManageTeamMembers(authenticatedPage, org.name, team.name);
+
+      await expect(authenticatedPage.getByTestId(email)).toBeVisible({
+        timeout: 10_000,
+      });
+
+      await authenticatedPage.getByTestId(`${email}-delete-icon`).click();
+      await authenticatedPage.getByTestId(`${email}-del-btn`).click();
+
+      await expect(
+        authenticatedPage.locator('.pf-v6-c-alert.pf-m-success').last(),
+      ).toContainText('Successfully revoked invitation', {
+        timeout: 10_000,
+      });
+    });
+  },
+);
--- a/web/src/hooks/UseCreateAccount.ts
+++ b/web/src/hooks/UseCreateAccount.ts
@@ -64,7 +64,10 @@ export function useCreateAccount() {
 
           // If user has prompts (e.g., confirm_username, enter_name, enter_company), redirect to updateuser
           if (user.prompts && user.prompts.length > 0) {
-            navigate('/updateuser');
+            const updateUrl = inviteCode
+              ? `/updateuser?code=${encodeURIComponent(inviteCode)}`
+              : '/updateuser';
+            navigate(updateUrl);
             return {success: true};
           }
 
--- a/web/src/hooks/UseCreateAccount.ts
+++ b/web/src/hooks/UseCreateAccount.ts
@@ -20,13 +20,14 @@ export function useCreateAccount() {
     username: string,
     password: string,
     email: string,
+    inviteCode?: string,
   ) => {
     setIsLoading(true);
     setError(null);
 
     try {
       // Create the user account
-      const response = await createUser(username, password, email);
+      const response = await createUser(username, password, email, inviteCode);
 
       // Clear CSRF token after account creation (session state changed)
       GlobalAuthState.csrfToken = null;
@@ -39,7 +40,7 @@ export function useCreateAccount() {
 
       // Auto-login after successful account creation
       try {
-        const loginResponse = await loginUser(username, password);
+        const loginResponse = await loginUser(username, password, inviteCode);
 
         if (loginResponse.success === true) {
           // Login successful, set auth state
@@ -64,6 +65,12 @@ export function useCreateAccount() {
           // If user has prompts (e.g., confirm_username, enter_name, enter_company), redirect to updateuser
           if (user.prompts && user.prompts.length > 0) {
             navigate('/updateuser');
+            return {success: true};
+          }
+
+          // If there's an invite code, redirect to confirminvite to accept it
+          if (inviteCode) {
+            navigate(`/confirminvite?code=${encodeURIComponent(inviteCode)}`);
             return {success: true};
           }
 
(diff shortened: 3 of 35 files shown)


quay-devel:
Addressed all CodeRabbit feedback in commit 6d6b4144:

1. **UseCreateAccount auto-login** — `inviteCode` is now forwarded to the auto-login `loginUser` call so invite context is preserved after account creation.
2. **URL-encode invite code** — `encodeURIComponent(code)` is now used when building the `/api/v1/teaminvite/` path.
3. **Sign In links in CreateAccount** — both Sign In links now carry the invite code when present (via a `signinUrl` variable).
4. **Test mocks `as any`** — replaced with `Partial<IUserResource>` using the imported type.
5. **React.FC annotation** — skipped intentionally: this codebase consistently uses `export function Foo()` declarations (see `CreateAccount.tsx`, `Signin.tsx`). Aligning with project convention.

LiZhang19817 (dismissed):

LiZhang19817:
@bcaton85 @jbpratt @harishsurf Pls help review and approve this fix.

web/playwright/e2e/auth/confirm-invite.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+import {BASE_URL} from '../../utils/config';
+
+/**
+ * Confirm Invite tests verify the /confirminvite route added for
+ * PROJQUAY-11636. Tests that need a real invite code use the
+ * @feature:INVITE_ONLY_USER_CREATION tag.
+ *
+ * The built-in anonPage fixture does not set the React UI
+ * cookie, so /confirminvite would load the old Angular UI. We extend
+ * with a custom anonPage that sets the cookie first.
```
jbpratt: Is it worth fixing the existing fixture rather than adding a test specific one?

LiZhang19817:
I've identified three related bugs:

  1. Blank team member column: Backend returns email (not name) for raw email invites, but the UI only renders teamMember.name
  2. No invite email sent: The new UI only calls the /members/{name} endpoint (for existing users). It never calls the
  /invite/{email} endpoint, so email invites can't be created through the new UI
  3. No notification banner: The old UI shows "E-mail address test002@redhat.com was invited to join the team" — the new UI has no
  equivalent since the invite flow is missing

LiZhang19817:
New Gap in E2E testing in CI:

Gaps — no E2E coverage for

  1. Email invite via drawer UI — opening "Add new member", typing email in the invite input (data-testid="email-invite-input"),
  clicking Invite (data-testid="invite-by-email-btn"), verifying success alert
  2. Email-invited member display — verifying the email address appears in the "Team member" column (the blank name bug fix)
  3. Invited tab filter — clicking the "Invited" toggle, verifying email-invited members appear
  4. Revoking email invite via UI — clicking delete icon on an email-invited member, confirming, verifying success
  5. MAILING feature flag gating — verifying the email invite section is hidden when MAILING is disabled

LiZhang19817 (dismissed):

LiZhang19817:
@jbpratt @bcaton85 Pls help review and approve the fix of this PR.

web/playwright/e2e/auth/confirm-invite.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  });
+
+  test('create-account page preserves invite code in sign-in link', async ({
+    unauthenticatedPage,
+    quayConfig,
+  }) => {
+    test.skip(
+      quayConfig?.config?.AUTHENTICATION_TYPE !== 'Database',
+      'Create account page requires Database auth',
+    );
```
jbpratt: You can add a `tag` to the test of `@auth:Database` for auto skipping

web/playwright/e2e/auth/confirm-invite.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    const href = await signinLink.getAttribute('href');
+    expect(href).toContain('code=link-code');
+  });
+
+  test(
+    'signin page shows create-account link when invite code is present',
+    {tag: '@feature:INVITE_ONLY_USER_CREATION'},
+    async ({unauthenticatedPage, quayConfig}) => {
+      test.skip(...skipUnlessFeature(quayConfig, 'INVITE_ONLY_USER_CREATION'));
```
jbpratt:
```suggestion
```
duplicate (the feature tag does this)

web/playwright/e2e/organization/team-members.spec.ts
```diff
@@ -256,3 +256,126 @@ test.describe('Manage Team Members', {tag: ['@organization']}, () => {
...
+      const org = await api.organization('testorg');
+      const team = await api.team(org.name, 'emailshow');
+      const email = `${uniqueName('show')}@example.com`;
+
+      try {
+        await api.teamEmailInvite(org.name, team.name, email);
+      } catch {
+        test.skip(true, 'Email team invitations not enabled');
+        return;
+      }
```
jbpratt: we don't want to blindly skip/fail silently here

web/playwright/e2e/organization/team-members.spec.ts
```diff
@@ -256,3 +256,126 @@ test.describe('Manage Team Members', {tag: ['@organization']}, () => {
...
+      const team = await api.team(org.name, 'emailfilter');
+      const email = `${uniqueName('filter')}@example.com`;
+
+      await api.teamMember(org.name, team.name, TEST_USERS.user.username);
+      try {
+        await api.teamEmailInvite(org.name, team.name, email);
+      } catch {
+        test.skip(true, 'Email team invitations not enabled');
+        return;
+      }
```
jbpratt: same as above

web/playwright/e2e/organization/team-members.spec.ts
```diff
@@ -256,3 +256,126 @@ test.describe('Manage Team Members', {tag: ['@organization']}, () => {
...
+      const org = await api.organization('testorg');
+      const team = await api.team(org.name, 'emailrevoke');
+      const email = `${uniqueName('revoke')}@example.com`;
+
+      try {
+        await api.teamEmailInvite(org.name, team.name, email);
+      } catch {
+        test.skip(true, 'Email team invitations not enabled');
+        return;
+      }
```
jbpratt: same as above

web/src/routes/ConfirmInvite/ConfirmInvite.tsx
```diff
@@ -0,0 +1,143 @@
...
+      },
+    },
+  );
+
+  useEffect(() => {
+    if (!code || userLoading || user === undefined) return;
+    if (!user.anonymous) {
+      doAcceptInvite();
+    }
+  }, [code, user, userLoading, doAcceptInvite]);
```
jbpratt: **[Major]** `doAcceptInvite` can fire multiple times

The `useEffect` includes `doAcceptInvite` in its dependency array. Since it comes from `useMutation` with a closure over `code`, its identity can change across renders, re-triggering the effect. There's also no guard against re-firing when the mutation is already in-flight or has already succeeded — if `user` object identity changes (e.g. refetch), this fires again.

Suggestion: add a ref guard and remove `doAcceptInvite` from deps:

```tsx
const hasAttemptedAccept = useRef(false);

useEffect(() => {
  if (!code || userLoading || user === undefined || hasAttemptedAccept.current) return;
  if (!user.anonymous) {
    hasAttemptedAccept.current = true;
    doAcceptInvite();
  }
}, [code, user, userLoading]);
```

web/src/routes/ConfirmInvite/ConfirmInvite.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const {isLoading: configLoading} = useQuayConfigWithLoading();
+  const {user, loading: userLoading, error: userError} = useCurrentUser(!!code);
+
+  const {mutate: doAcceptInvite, isLoading: isAccepting} = useMutation(
+    () => acceptTeamInvite(code),
+    {
+      onSuccess: (resp) => {
+        navigate(`/organization/${resp.org}/teams/${resp.team}`);
+      },
+      onError: (err: any) => {
```
jbpratt: **[Minor]** `err: any` bypasses TypeScript safety

Consider typing as `AxiosError` or `unknown` with proper narrowing:

```tsx
import {AxiosError} from 'axios';

onError: (err: AxiosError<{message?: string}>) => {
  setErrorMessage(
    err?.response?.data?.message || 'Invalid or expired invite code.',
  );
}
```

web/src/routes/ConfirmInvite/ConfirmInvite.tsx
```diff
@@ -0,0 +1,143 @@
...
+      {isAccepting && (
+        <div style={{textAlign: 'center', padding: '40px'}}>
+          <Spinner
+            size="xl"
+            data-testid={
+              isAccepting
+                ? 'confirm-invite-accepting'
+                : 'confirm-invite-loading'
+            }
+          />
```
jbpratt: **[Nit]** Redundant conditionals — dead code

The outer block is already guarded by `{isAccepting && ...}` (line 81), so:
- The ternary `isAccepting ? 'confirm-invite-accepting' : 'confirm-invite-loading'` always resolves to `'confirm-invite-accepting'`
- The inner `{isAccepting && <p>}` is always true

Simplify to:
```tsx
<Spinner size="xl" data-testid="confirm-invite-accepting" />
<p style={{marginTop: '16px'}}>Accepting invitation…</p>
```

web/src/routes/OrganizationsList/Organization/Tabs/TeamsAndMembership/TeamsView/ManageMembers/ManageMembersList.tsx
```diff
@@ -223,12 +235,39 @@ export default function ManageMembersList(props: ManageMembersListProps) {
     }
   }, [errorDeleteTeamMember]);
 
+  useEffect(() => {
+    if (successDeleteEmailInvite) {
+      setIsDeleteModalForRowOpen(false);
+      addAlert({
+        variant: AlertVariant.Success,
+        title: `Successfully revoked invitation for: ${memberToBeDeleted.memberName}`,
+      });
```
jbpratt, line 256: **[Minor]** Missing mutation reset calls

Neither `resetDeleteEmailInvite()` nor the existing `resetDeleteTeamMember()` are called after success/error effects. Without resetting, the success/error boolean stays `true` and these effects can fire on subsequent re-renders (e.g. after a query refetch), potentially producing duplicate alert toasts.

Consider calling the reset function at the end of each effect handler.

web/src/routes/Signin/Signin.tsx
```diff
@@ -131,14 +131,17 @@ export function Signin() {
...
     return (
       !inReadOnlyMode &&
       quayConfig.features?.USER_CREATION === true &&
       quayConfig.config?.AUTHENTICATION_TYPE === 'Database' &&
       shouldShowDirectLogin() &&
-      !quayConfig.features?.INVITE_ONLY_USER_CREATION &&
+      (!quayConfig.features?.INVITE_ONLY_USER_CREATION || hasInvite) &&
       !inAccountRecoveryMode
     );
   };
```
jbpratt, line 151: **[Minor]** Conflicting UI messages when invite code is present

When `INVITE_ONLY_USER_CREATION` is true and an invite code is in the URL, both `showCreateAccount()` (returns true because `hasInvite` is true) and the "Invitation required to sign up" banner will display simultaneously. The banner becomes misleading when the user *does* have an invite.

Consider gating the invitation-required message on `!hasInvite`.

web/src/hooks/UseMembers.ts
```diff
@@ -93,9 +95,10 @@ export function useFetchMembers(orgName: string) {
 }
 
 export interface ITeamMember {
-  name: string;
+  name?: string;
+  email?: string;
   kind: string;
```
jbpratt, line 100: **[Minor]** Both `name` and `email` are optional with no enforcement

If both are absent, `getMemberDisplayName()` returns `''`, which would break display, sorting, and filtering. The backend contract should guarantee at least one is present, but a defensive runtime check or a discriminated union type would make this safer.

A simple guard in `getMemberDisplayName`:
```ts
function getMemberDisplayName(member: ITeamMember): string {
  const name = member.name || member.email;
  if (!name) console.warn('TeamMember has neither name nor email:', member);
  return name || '(unknown)';
}
```

jbpratt (commented):
## Code Review

Consolidated findings as inline comments, ordered by severity. Overall the PR is well-structured with strong test coverage — two items should be addressed before merge.

sudipshil9862:
@LiZhang19817 @jbpratt
what worked:- 
- navigating to http://localhost:9000/confirminvite?code=testcode123 gives a page contains "Confirm invitation", "Sign in or create an account to accept this invitation"
- clicking on the "Create account" button -> then navigates to /createaccount?code=testcode123 gives a blue infor banner that "You've been invited to join a team - Create your account to accept the invitation"
- in sign in, http://localhost:9000/confirminvite gives "No invite code found in the URL"

improvement needed:
- in invite by email address section -> type email address -> hit enter  => page directly redirect to repositories section, it should hit the invite button next to it.
image

https://github.com/user-attachments/assets/fb1f8305-2460-458d-b3ab-25d52131bcdd


- no option to verify valid email id.

sudipshil9862:
also tested email invite tested using mailpit

image

https://github.com/user-attachments/assets/29f4bb50-3203-4ac0-9bfb-e0011b3c6f70

sudipshil9862:
so user needs to manually enable the FEATURE_MAILING: true and FEATURE_INVITE_ONLY_USER_CREATION: true in local-dev/stack/config.yaml
and for testing purpose i used `docker compose up -d mailpit` for geting test mails

LiZhang19817:
> so user needs to manually enable the FEATURE_MAILING: true and FEATURE_INVITE_ONLY_USER_CREATION: true in local-dev/stack/config.yaml and for testing purpose i used `docker compose up -d mailpit` for geting test mails

Correct

LiZhang19817:
**Fixed this minor issue:**

in invite by email address section -> type email address -> **hit enter** => page directly redirect to repositories section, it should hit the invite button next to it.
image

LiZhang19817:
@jbpratt @bcaton85 The local validation testing is passed, pls help review and approve this PR.

LiZhang19817 (dismissed):

web/src/routes/Signin/Signin.tsx
```diff
@@ -178,7 +182,7 @@ export function Signin() {
   ) => {
     e.preventDefault();
     try {
-      const response = await loginUser(username, password);
+      const response = await loginUser(username, password, inviteCode);
```
jbpratt, line 189: **Critical: Existing users who sign in via invite link never join the team.**

`inviteCode` is correctly passed to `loginUser()` here, but the backend's `handle_invite_code()` only marks the user as verified — it does NOT accept the team invite. The actual acceptance requires a separate `PUT /api/v1/teaminvite/<code>` call via the `/confirminvite` page.

After login succeeds (lines 206-218 below this hunk), the component redirects to `/updateuser`, `redirect_url`, or `/organization` — it **never** redirects to `/confirminvite?code=<code>`. The invite is silently dropped for all existing users.

By contrast, `UseCreateAccount.ts:72-74` correctly redirects to `/confirminvite?code=...` after signup.

**Suggested fix** — add invite code redirect after the prompts check (around line 210), before the `redirect_url` fallback:
```typescript
// If there's an invite code, redirect to confirminvite to accept it
if (inviteCode) {
  navigate(`/confirminvite?code=${encodeURIComponent(inviteCode)}`);
  return;
}

const redirectUrl = searchParams.get('redirect_url');
```

web/src/routes/Signin/Signin.tsx
```diff
@@ -131,21 +131,25 @@ export function Signin() {
     );
   };
 
+  const inviteCode = searchParams.get('code') || undefined;
```
jbpratt, line 138: **Critical: OAuth/SSO login path loses the invite code entirely.**

`inviteCode` is extracted here, but `startExternalLogin()` (called on line 78 for SSO auto-redirect, and in the external login buttons) does NOT receive the invite code. After OAuth completes, the user lands at `/` with no memory of the invite.

This breaks the invite flow for all SSO-only deployments.

**Suggested fix:** Thread the invite code through the OAuth flow — e.g. via session storage:
```typescript
if (inviteCode) {
  sessionStorage.setItem('pendingInviteCode', inviteCode);
}
startExternalLogin(singleProvider, redirectUrl);
```
Then in the OAuth callback handler, retrieve and redirect:
```typescript
const pendingCode = sessionStorage.getItem('pendingInviteCode');
if (pendingCode) {
  sessionStorage.removeItem('pendingInviteCode');
  navigate(`/confirminvite?code=${encodeURIComponent(pendingCode)}`);
}
```

web/src/hooks/UseCreateAccount.ts
```diff
@@ -67,6 +68,12 @@ export function useCreateAccount() {
             return {success: true};
```
jbpratt, line 71: **Critical: Invite code is lost when user has pending prompts.**

If the newly created user has prompts (e.g. `confirm_username`, `enter_name`), this line redirects to `/updateuser` without the invite code. After the user completes the prompts flow, they're redirected away and never reach `/confirminvite`.

The invite code redirect on line 72 below is correct — but it's **unreachable** when prompts exist because this block returns first.

**Suggested fix:** Pass the invite code through to `/updateuser`:
```typescript
if (user.prompts && user.prompts.length > 0) {
  const updateUrl = inviteCode
    ? `/updateuser?code=${encodeURIComponent(inviteCode)}`
    : '/updateuser';
  navigate(updateUrl);
  return {success: true};
}
```
Then update `/updateuser` to forward the code to `/confirminvite` after completion.

jbpratt (commented):
## Review: Critical flow bugs in invite code threading

Three issues where the invite code is silently dropped, meaning users who follow an invite link will never actually join the team.

Findings identified via multi-model code review (Opus + Sonnet + Haiku).

web/src/routes/Signin/Signin.tsx
LiZhang19817: Valid catch — fixed. Added the invite code redirect after the prompts check, before the `redirect_url` fallback:

```typescript
if (inviteCode) {
  navigate(`/confirminvite?code=${encodeURIComponent(inviteCode)}`);
  return;
}
```

This matches the pattern already in `UseCreateAccount.ts:72–74`.

web/src/routes/Signin/Signin.tsx
LiZhang19817: Agree this is a real gap for SSO-only deployments — however it's a **pre-existing limitation**, not a regression introduced by this PR.

I traced the old Angular UI's flow: the `confirm-invite.js` page embeds the signin-form directive inline and sets `$scope.redirectUrl = window.location.href` (the `/confirminvite?code=…` page itself). After direct login the form reloads the confirm-invite page which then calls `acceptOrganizationTeamInvite`. But when a user clicks an external login button from that same page, the OAuth redirect goes through the backend `_perform_login()` which hard-redirects to `/updateuser` or `/` — the `redirectUrl` on the Angular scope is lost. So the OAuth invite flow was already broken in the old UI too.

The sessionStorage approach is the right fix for this, but it touches the OAuth callback handling and `StandaloneMain`/`UpdateUser` — scope that goes beyond this PR. I'll track it as a follow-up ticket so SSO-only deployments get proper invite support.

The direct-login path (which is the primary flow when `AUTHENTICATION_TYPE: Database` + `INVITE_ONLY_USER_CREATION` are configured) is now fixed — see the reply on the sibling comment.

web/src/hooks/UseCreateAccount.ts
LiZhang19817: Valid point. In the old Angular UI this wasn't an issue because the signup form is embedded *inside* the `/confirminvite` page — after signup, `UserService.load()` triggers the same page's `ConfirmInviteCtrl` watcher which calls `acceptOrganizationTeamInvite`. No navigation away from the page means the invite code is never lost.

In the new React UI, we navigate to a separate `/updateuser` page, so the code does need to survive that hop.

The fix is to thread the invite code through `/updateuser`:

1. `UseCreateAccount.ts` — pass invite code as query param when navigating to `/updateuser`
2. `UpdateUser.tsx` — after prompts complete, check for the `code` param and redirect to `/confirminvite`

Will fix in the next push.

LiZhang19817: Update: all three parts of the fix are now applied:

1. **`UseCreateAccount.ts`** — passes invite code to `/updateuser?code=...` when prompts exist
2. **`Signin.tsx`** — same fix for the existing-user sign-in path (had the same gap)
3. **`UpdateUser.tsx`** — checks `searchParams.get('code')` in both:
   - The `onSuccess` callback (after prompts complete)
   - The `useEffect` guard (defensive — prevents a race where `invalidateQueries(['user'])` re-fires the effect with empty prompts before the `onSuccess` navigate takes effect)

Both paths redirect to `/confirminvite?code=...` which calls `acceptTeamInvite`.

LiZhang19817:
@jbpratt @bcaton85 Pls help review and merge this PR.

PR #6077, opened 2026-05-29
PROJQUAY-12444, PROJQUAY-12445: fix(config-tool/validators): block SSRF in SMTP/LDAP validators and fix CWE-598 in GitLab OAuth

aroyoredhat:
## Summary

- Block SSRF in `ValidateEmailServer`: add `isSafeHost()` guard that resolves hostnames and rejects RFC 1918 / loopback / link-local / cloud-metadata IPs before dialing SMTP
- Block SSRF in `ValidateLDAPServer`: same guard applied to the URI host before `ldap.DialURL`; guard fails closed if the URI cannot be parsed
- Fix CWE-598 in `ValidateGitLabOAuth`: move `client_id` and `client_secret` from URL query string to POST body with correct `Content-Type` header

## Root Cause / Rationale

`config-tool/pkg/lib/shared/validators.go` passed user-supplied hostnames directly to `net.DialTimeout` (SMTP) and `ldap.DialURL` (LDAP) with no IP-range filtering, allowing an attacker with config-editor access to probe internal network services from the Quay pod (CVE-2026-10052 / CWE-918). The GitLab OAuth validator concatenated `client_secret` into the URL query string, exposing it in server and proxy logs (CWE-598).

## Changes

| File | What changed |
|------|-------------|
| `config-tool/pkg/lib/shared/validators.go` | Added `privateRanges` var and `isSafeHost()` helper; SSRF guard in `ValidateEmailServer`; fail-closed SSRF guard in `ValidateLDAPServer`; GitLab OAuth credentials moved to POST body |
| `config-tool/pkg/lib/shared/validators_test.go` | Added 4 regression tests: `TestIsSafeHost`, `TestValidateEmailServerRejectsPrivateHost`, `TestValidateLDAPServerRejectsPrivateHost`, `TestValidateGitLabOAuthCredentialsInBody` |

## Test Plan

- [ ] Integration/registry tests pass *(N/A for config-tool shared package — no live SMTP/LDAP in CI)*
- [x] Type checking passes *(`go vet` clean)*
- [x] Manual testing performed *(SSRF rejection verified via regression tests)*
- [ ] Frontend tests pass *(N/A — backend only)*
- [ ] Migration tested *(N/A — no schema changes)*

```bash
cd config-tool
go test ./pkg/lib/shared/... \
  -run "TestIsSafeHost|TestValidateEmailServerRejectsPrivateHost|TestValidateLDAPServerRejectsPrivateHost|TestValidateGitLabOAuthCredentialsInBody" \
  -v
```

Expected: all 4 new tests pass.

**Operator note:** Deployments using RFC 1918 addresses for SMTP or LDAP servers will see config *validation* failures after this patch. The runtime connections in Quay itself are not affected — only the config-tool validation step.

## JIRA

https://issues.redhat.com/browse/PROJQUAY-11645

## Backport

- [x] No backport needed *(Target Version not set on JIRA ticket)*

## Automation

**Issue:** https://issues.redhat.com/browse/PROJQUAY-11645

**Changelog:** Fixed SSRF in SMTP/LDAP config validators and CWE-598 credential exposure in GitLab OAuth validator (CVE-2026-10052, [PROJQUAY-11645](https://redhat.atlassian.net/browse/PROJQUAY-11645))

**Docs:** No user-facing documentation changes required. Operators who configure SMTP or LDAP using private IP addresses (RFC 1918) will see validation errors from config-tool — this is expected behavior post-patch; runtime connections are unaffected.

**Testing:** Run `go test ./pkg/lib/shared/... -run "TestIsSafeHost|TestValidateEmailServer|TestValidateLDAP|TestValidateGitLab" -v` from the `config-tool/` directory and confirm all 4 new tests pass. The httptest-based GitLab test captures the live HTTP request and verifies credentials are in the POST body, not the URL.

**Details:** Three issues fixed in `config-tool/pkg/lib/shared/validators.go`:
1. `ValidateEmailServer` — `net.DialTimeout` now guarded by `isSafeHost()` rejecting RFC 1918, loopback, and cloud-metadata addresses.
2. `ValidateLDAPServer` — `ldap.DialURL` guarded by same helper; URI parse failure is fail-closed (returns error rather than silently bypassing the guard).
3. `ValidateGitLabOAuth` — `client_secret` moved from URL querystring to POST body.

Known out-of-scope gaps: `ValidateGitHubOAuth` and the `gitlabEndpoint` HTTP dial path connect to user-supplied URLs without host filtering — same SSRF class, not called out in the CVE text, tracked for a follow-up.


quay-devel:
The **Cypress Tests** failure is pre-existing and unrelated to this PR. The same test (`enableGithubTrigger` `cy.wait()` timeout) is failing on `master` as of this commit: confirmed via `/repos/quay/quay/commits/master/check-runs`. This change touches only `config-tool/pkg/lib/shared/validators.go` and its test file — no frontend code.

config-tool/pkg/lib/shared/validators.go
```diff
@@ -31,6 +31,50 @@ import (
...
+		"::1/128",        // IPv6 loopback
+		"fc00::/7",       // IPv6 unique local
+		"fe80::/10",      // IPv6 link-local
+	}
+	nets := make([]*net.IPNet, len(blocks))
+	for i, b := range blocks {
+		_, nets[i], _ = net.ParseCIDR(b)
+	}
+	return nets
+}()
```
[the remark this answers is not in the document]
quay-devel: Valid finding —  is a well-known SSRF bypass on Linux where connections to  route to localhost. Added to  and to the  blocked list in the follow-up commit (a5c7582).

config-tool/pkg/lib/shared/validators.go
```diff
@@ -731,6 +776,14 @@ func ValidateElasticSearchCredentials(url, accessKey, accessSecret string) bool
 // ValidateEmailServer validates that the provided smtp server is valid
 func ValidateEmailServer(opts Options, mailServer string, mailPort int, useTLS bool, useAuth bool, username string, password string, fgName string) (bool, ValidationError) {
 
+	if err := isSafeHost(mailServer); err != nil {
```
bcaton85, line 779: **should-fix:** This check validates one DNS resolution, but the SMTP dial below still uses the original hostname:

```go
net.DialTimeout("tcp", mailServer+":"+strconv.Itoa(int(mailPort)), 3*time.Second)
```

That leaves a DNS-rebinding/TOCTOU bypass: an attacker-controlled hostname can resolve to a public IP during `isSafeHost(mailServer)`, then resolve to `127.0.0.1`, `169.254.169.254`, or an RFC1918 address during the actual dial.

For this SSRF guard to be more robust, we should resolve once, validate all returned IPs, and dial only the validated IP literal(s), while preserving the original hostname for SMTP/TLS identity (`smtp.NewClient` / `ServerName`).

config-tool/pkg/lib/shared/validators.go
```diff
@@ -31,6 +31,51 @@ import (
...
+// privateRanges is the set of CIDR blocks that validators must never dial.
+var privateRanges = func() []*net.IPNet {
+	blocks := []string{
+		"0.0.0.0/8",      // "this" network — routes to localhost on Linux
+		"127.0.0.0/8",    // loopback
+		"10.0.0.0/8",     // RFC 1918
+		"172.16.0.0/12",  // RFC 1918
+		"192.168.0.0/16", // RFC 1918
+		"169.254.0.0/16", // link-local / cloud metadata (AWS, GCP, Azure)
+		"::1/128",        // IPv6 loopback
```
bcaton85, line 43: **should-fix:** Since this denylist includes `0.0.0.0/8` to prevent local-connect SSRF bypasses, it should also include the IPv6 unspecified address `::/128`.

`isSafeHost("::")` currently succeeds, and `ldap://[::]:389` can reach a local wildcard listener via IPv6 in the same spirit as the `0.0.0.0` bypass. Adding `::/128` here plus a regression test for `ldap://[::]:389` would close that gap.

config-tool/pkg/lib/shared/validators.go
```diff
@@ -875,6 +936,24 @@ func ValidateLDAPServer(opts Options, ldapUri, ldapAdminDn, ldapAdminPasswd, lda
...
+	// Fail closed: any parse failure is treated as an unsafe URI.
+	ldapParsed, parseErr := url.Parse(ldapUri)
+	if parseErr != nil {
+		return false, ValidationError{
+			Tags:       []string{"LDAP_URI"},
+			FieldGroup: fgName,
+			Message:    "LDAP_URI: " + parseErr.Error(),
+		}
+	}
+	if err := isSafeHost(ldapParsed.Host); err != nil {
```
bcaton85, line 949: **should-fix:** Same DNS-rebinding concern as the SMTP path: `isSafeHost(ldapParsed.Host)` validates one resolution, but `ldap.DialURL(ldapUri, ...)` below dials the original URI and may perform a fresh DNS lookup.

For this guard to be more robust, resolve once, validate all returned addresses, and dial LDAP using validated IP literal URL(s). For `ldaps://`, preserve the original hostname in `tls.Config.ServerName` so certificate verification still uses the configured LDAP hostname rather than the IP literal.

bcaton85, line 949: **should-fix:** `ldapParsed.Host` includes IPv6 brackets and possibly the port. This can reject valid bracketed IPv6 LDAP URIs that `ldap.DialURL` would otherwise understand.

For example, `url.Parse("ldap://[2001:db8::1]")` has `Host == "[2001:db8::1]"`; because there is no port, `net.SplitHostPort` in `isSafeHost` fails and then attempts to resolve the bracketed string. Consider using `ldapParsed.Hostname()` and `ldapParsed.Port()` for validation/rewriting instead of passing `Host` directly.

config-tool/pkg/lib/shared/validators_test.go
```diff
@@ -1,9 +1,125 @@
...
+		w.Header().Set("Content-Type", "application/json")
+		w.Write([]byte(`{"error":"invalid_grant"}`))
+	}))
+	defer ts.Close()
+
+	result := ValidateGitLabOAuth("test-client-id", "super-secret", ts.URL+"/")
+	if !result {
+		t.Fatal("ValidateGitLabOAuth should return true for invalid_grant response")
+	}
+	if strings.Contains(got.rawQuery, "client_secret") {
```
bcaton85, line 106: **should-fix:** This test verifies `client_secret` is absent from the query string, but the PR/JIRA says both `client_id` and `client_secret` were moved out of URL parameters and into the POST body.

Please also assert that `client_id` is not present in `got.rawQuery` — or simply assert `got.rawQuery == ""` — so a regression leaving `client_id` in the URL is caught.

config-tool/pkg/lib/shared/validators.go
```diff
@@ -31,6 +31,51 @@ import (
 	"golang.org/x/oauth2"
 )
 
+// privateRanges is the set of CIDR blocks that validators must never dial.
+var privateRanges = func() []*net.IPNet {
```
bcaton85, line 35: **blocker**: Will excluding any of these ranges cause deployments to OCP to fail?

bcaton85 (changes requested):
Some discussion points. Is this covered by a e2e test? We're our own QE now so this has to be ready to ship on merge

PR #6086, opened 2026-06-02
RFE-4378: feat(notifications): add FEATURE_SKIP_EMAIL_CONFIRMATION flag

ranjithrajaram:
Trying to address RFE https://redhat.atlassian.net/browse/RFE-4378

Add configuration flag to allow bypassing email confirmation requirement for repository notifications in trusted enterprise environments.

Changes:
- Add FEATURE_SKIP_EMAIL_CONFIRMATION flag (default: False for security)
- Modify EmailMethod.validate() to skip confirmation check when enabled
- Add configuration schema definition with security warnings
- Add test coverage for both enabled and disabled states

Security:
- Default: False (maintains current secure behavior)
- Explicit opt-in required via config.yaml
- Only recommended for trusted enterprise deployments with internal emails

Use case:
Enterprise customers with 1000+ repositories who need to configure email notifications via API without manual confirmation for trusted internal email addresses.

When FEATURE_SKIP_EMAIL_CONFIRMATION is False (default):
- Email notifications require confirmation (current behavior)
- Users must click confirmation link in email

When FEATURE_SKIP_EMAIL_CONFIRMATION is True:
- Email notifications can be created without confirmation
- Removes protection against email bombing and unauthorized disclosure
- Only use in trusted enterprise environments

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -139,3 +139,4 @@ bin/
 .claude/implementation
 .claude/poll-state/
 .claude/session-state/
+.claude/scheduled_tasks.lock
--- a/config.py
+++ b/config.py
@@ -357,6 +357,12 @@ class DefaultConfig(ImmutableConfig):
     # Feature Flag: Whether emails are enabled.
     FEATURE_MAILING = True
 
+    # Feature Flag: Whether to skip email confirmation requirement for repository notifications.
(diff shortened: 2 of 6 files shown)


ranjithrajaram:
@coderabbitai resume

ranjithrajaram:
Will reopen a new PR

PR #6095, opened 2026-06-03
PROJQUAY-11576: feat(ldap): add queue-based admin connection pool

jbpratt:
## Summary

- Fix generator leak in `check_group_lookup_args` — the generator from `iterate_group_members` held an open LDAP connection until GC
- Add `LDAPConnectionPool` with queue-based admin connection reuse — lazy init (safe with Gunicorn `preload_app`), rootDSE health checks on idle connections, soft lifetime rotation with jitter, overflow connections when pool exhausted
- Add bind-and-revert for user password verification — verifies credentials by rebinding on a pooled connection then restoring admin identity (follows the [UnboundID SDK `bindAndRevertAuthentication`](https://docs.ldap.com/ldap-sdk/docs/javadoc/com/unboundid/ldap/sdk/LDAPConnectionPool.html) pattern)
- Polymorphic `verify_bind` on both `LDAPConnectionPool` and `LDAPConnectionBuilder` — no isinstance dispatch

### Configuration

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| `LDAP_CONNECTION_POOLING` | bool | `true` | Kill switch — set `false` to fall back to per-request connections |
| `LDAP_POOL_SIZE` | int | `10` | Max pooled connections per worker process |
| `LDAP_POOL_MAX_WAIT` | float | `5.0` | Seconds to wait before creating overflow connection |
| `LDAP_POOL_CONNECTION_LIFETIME` | int | `300` | Max connection age before rotation |

### Benchmark Results (10 concurrent users, 30s, local dev with 389 DS)

| Metric | pool + bind-revert | no pool |
|---|---|---|
| LDAP connections created | **7** | 178 |
| Throughput (req/s) | 2.6 | 2.51 |
| Latency p50 (ms) | 3728 | 3899 |

The key metric is **96% fewer LDAP connections created** — in production with TLS over a network, each saved connection avoids a full TCP handshake + TLS negotiation + SASL bind.

### Security Invariants

- **User credentials never persist in the pool** — bind-and-revert uses the password once, then immediately rebinds as admin
- **If admin rebind fails, connection is discarded** — never returned to pool in wrong identity
- **Anonymous bind rejected before pool** — empty password guard fires first
- **Pool holds only admin credentials** — constructed with `LDAP_ADMIN_DN` / `LDAP_ADMIN_PASSWD`
- **Socket cleanup on connection creation failure** — no FD leak when bind fails

## Test plan

- [x] 69 unit tests pass (`test/test_ldap.py`) including pool mechanics, edge cases, bind-and-revert
- [x] Existing `TestLDAP` suite runs against pooled path by default (pool is default-on)
- [x] Generator close test verifies leak fix
- [x] Bind-and-revert identity restoration test proves admin identity is restored after user verification
- [x] Edge case tests: pool exhaustion/overflow, lifetime rotation, health check failure, admin rebind failure, connection error propagation
- [x] Kill switch test confirms fallback to `LDAPConnectionBuilder`
- [x] Manual validation via local dev (LDAP login, burst test, pool rotation logs confirmed)
- [x] `make types-test` passes (pre-existing unrelated `rsa` stub error only)
- [ ] CI: existing Playwright `@auth:LDAP` tests cover login/team-sync paths through pool

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -688,10 +688,12 @@ class LDAPUsers(FederatedUsers):
         if err is not None:
             return (False, err)
 
-        if not next(it, False):
-            return (False, "Group does not exist or is empty")
-
-        return (True, None)
+        try:
+            if not next(it, False):
+                return (False, "Group does not exist or is empty")
+            return (True, None)
+        finally:
+            it.close()
 
     def iterate_group_members(self, group_lookup_args, page_size=None, disable_pagination=False):
         try:
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -1,10 +1,14 @@
 import hashlib
 import logging
 import os
+import queue
+import random
 import re
 import sys
 import threading
+import time
 from collections import OrderedDict, namedtuple
+from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from typing import Any, Callable, Optional
 
@@ -140,6 +144,14 @@ class LDAPConnectionBuilder(object):
         self._network_timeout = network_timeout
         self._referrals = int(referrals)
 
+    @property
+    def admin_dn(self) -> str:
+        return self._user_dn
+
+    @property
+    def admin_pw(self) -> str:
+        return self._user_pw
+
     def get_connection(self):
         return LDAPConnection(
             self._ldap_uri,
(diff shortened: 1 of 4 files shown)


data/users/externalldap.py
```diff
@@ -151,6 +163,212 @@ def get_connection(self):
...
+    def verify_bind(self, user_dn: str, password: str) -> bool:
+        """Verify user credentials via a direct (non-pooled) connection."""
+        try:
+            with LDAPConnection(
+                self._ldap_uri,
+                user_dn,
+                password,
+                self._allow_tls_fallback,
+                self._timeout,
+                self._network_timeout,
```
bcaton85, line 175: 🔴 **Critical: Missing `self._referrals` parameter**

This `LDAPConnection` is constructed with 6 positional args, but `get_connection()` (line 155) passes 7 — including `self._referrals`. When `LDAP_CONNECTION_POOLING=false` and `LDAP_FOLLOW_REFERRALS: 0` is configured, the fallback verify path will follow referrals when it shouldn't, creating a correctness divergence between pooled and non-pooled behavior.

```python
with LDAPConnection(
    self._ldap_uri,
    user_dn,
    password,
    self._allow_tls_fallback,
    self._timeout,
    self._network_timeout,
    self._referrals,  # ← add this
):
```

Additionally, if `simple_bind_s` raises a non-`INVALID_CREDENTIALS` exception (e.g., `SERVER_DOWN`), `__exit__` won't run (Python doesn't call `__exit__` when `__enter__` raises), leaking the socket from `ldap.initialize()`. The pool path (`_create_raw_connection`) handles this correctly with explicit `__exit__` in a try/except — consider the same pattern here:

```python
def verify_bind(self, user_dn: str, password: str) -> bool:
    wrapper = LDAPConnection(
        self._ldap_uri, user_dn, password,
        self._allow_tls_fallback, self._timeout,
        self._network_timeout, self._referrals,
    )
    try:
        wrapper.__enter__()
        return True
    except ldap.INVALID_CREDENTIALS:
        return False
    finally:
        try:
            wrapper.__exit__(None, None, None)
        except Exception:
            pass
```

data/users/externalldap.py
```diff
@@ -151,6 +163,212 @@ def get_connection(self):
...
+    conn: Any
+    created_at: float = field(default_factory=time.monotonic)
+    last_used: float = field(default_factory=time.monotonic)
+    is_overflow: bool = False
+    lifetime_jitter: float = field(
+        default_factory=lambda: random.uniform(1.0 - _POOL_LIFETIME_JITTER, 1.0)
+    )
+
+
+class LDAPConnectionPool:
```
bcaton85, line 199: 🟠 **Document gevent monkey-patching dependency**

This pool uses `threading.Lock` and `queue.Queue`, which are safe under `monkey.patch_all()` (all Quay gunicorn configs call it at line 4 before imports). The pool is fork-safe with `preload_app=True` because `_queue` is lazy-initialized post-fork.

However, if anyone imports `data.users.externalldap` in a non-monkey-patched context (test harness, CLI tool, migration script), these become stdlib blocking primitives that block the entire greenlet hub.

Please add a class-level docstring note, e.g.:
```python
class LDAPConnectionPool:
    """Queue-based LDAP admin connection pool with lazy initialization.

    Thread/greenlet safety: relies on gevent monkey-patching (active in all
    Quay gunicorn processes) for cooperative concurrency. The pool is fork-safe
    under preload_app=True because the internal queue is lazily initialized on
    first use within each worker process.
    """
```

bcaton85 (commented):
Looks good, claude gave back only one major issue. Do we need to add e2e tests for this as well?

jbpratt:
> Do we need to add e2e tests for this as well?

The existing Playwright `@auth:LDAP` tests (`ldap-login.spec.ts`, `team-sync-ldap.spec.ts`) cover this implicitly — the pool is default-on, so every LDAP login in CI goes through it. If the pool breaks connections, those tests fail.

Pool mechanics (connection reuse, lifetime rotation, overflow, bind-and-revert identity restoration) are internal behavior with no UI surface — covered by the 69 unit tests including 21 pool-specific edge case tests (87% coverage on `externalldap.py`). Testing those through Playwright would require monitoring LDAP server connection counts via `cn=monitor` from inside the test, which is fragile and not something Playwright is suited for.

bcaton85 (approved):

harishsurf (approved):

PR #6096, opened 2026-06-03
PROJQUAY-11661: gc: Optimize deletion queries

ibazulic:
Previously, we were running deletion queries iteratively on manifests, blobs, child manifests and labels. While this is fine for manifests and blobs, iterative deletion of labels doesn't make much sense. It creates one query per each label and for a lot of labels, this takes a while to execute. 

Changing to bulk deletion shrinks the total number of queries down to 2 instead of the previous `2*n` + `n` transactions speeding up the GC process. Additionally, we no longer collect context per provided manifest, we collect it for all manifests at once, further shrinking the amount of queries GC needs to do.

Finally, so far, there was a chance for the GC process to lock itself in an infinite loop for repositories that have a lot of multiarch images caused by the way batching works. In Quay's case, we collect up to 10 manifest rows from the `Manifest` table to delete. If the only thing we collected are child images without their parent, these would be skipped because of the link in `ManifestChild` table between children and parents; parents entry would still remain. But Quay would then enqueue the same batch of 10 manifests for GC and would enter into an infinite loop. To stop that, we explicitly remove all links between parents and their children from the `ManifestChild` table early during GC process. This is a safe operation because we only delete links for the repo that is being deleted. This allows these manifests to be deleted successfully.

Co-authored-by: Claude [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/data/model/gc.py
+++ b/data/model/gc.py
@@ -186,6 +186,21 @@ def _purge_repository_contents(repo):
     """
     logger.debug("Purging repository %s", repo)
 
+    # We need to break manifest child-parent relationship here. Child images contain ids that are lower than
+    # their parents. If only child images are picked up during chunking, they will not be able to be deleted
+    # because their parent's entry in ManifestChild table will continue to exist. This will cause an endless
+    # loop of constantly enqueueing the same manifests over and over.
+    # The operation is safe because the repository is already marked for deletion and cannot be modified in
+    # any way.
+    ManifestChild.delete().where(ManifestChild.repository == repo).execute()
+
+    # We also delete all UploadedBlob references for the deleted repository at once. This stops potential leaks
+    # later on and also speeds up the GC process becuase we're doing one DELETE at a time instead of doing it
+    # for each individual blob. However, we still have the same _is_blob_orphaned check so this change has no impact
+    # on preserving shared blobs.
+    # The operation is safe because we're only deleting entries related to the deleted repository.
+    UploadedBlob.delete().where(UploadedBlob.repository == repo).execute()
+
     # Purge via all the tags.
     while True:
         found = False
--- a/data/model/gc.py
+++ b/data/model/gc.py
@@ -293,13 +293,31 @@ def garbage_collect_repo(repo):
 
 def _run_garbage_collection(context):
     """
-    Runs the garbage collection loop, deleting manifests, images, labels and blobs in an iterative
-    fashion.
+    Runs the garbage collection loop. Unreferenced Manifests, images and blobs are processed in an iterative
+    fashion, while manifest labels are deleted in bulk to speed up the gc process.
     """
+
     has_changes = True
 
     while has_changes:
         has_changes = False
+
+        manifest_id_list = list(context.manifest_ids)
+        if manifest_id_list:
+            for manifest_blob in ManifestBlob.select().where(
+                ManifestBlob.manifest << manifest_id_list
+            ):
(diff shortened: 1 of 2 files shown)


jbpratt (dismissed):

jbpratt:
@kleesc @bcaton85 could someone give this a second look?

data/model/gc.py
```diff
@@ -293,9 +293,31 @@ def garbage_collect_repo(repo):
...
+
+    while to_discover:
+        batch = list(to_discover)
+        discovered.update(to_discover)
+
+        for manifest_blob in ManifestBlob.select().where(ManifestBlob.manifest << batch):
+            context.add_blob_id(manifest_blob.blob_id)
+
+        for manifest_label in ManifestLabel.select().where(ManifestLabel.manifest << batch):
+            context.add_label_id(manifest_label.label_id)
```
kleesc, line 320: nit: I'm wondering if these should be batched (chunked), for repos with large number of manifests.

ibazulic: @kleesc shouldn't be a problem. The `_run_garbage_collection` function is called from `_purge_repository_contents` which chunks the tags and manifests. So we are always sending up to 10 tags or manifests to be cleaned.

data/model/gc.py
```diff
@@ -550,32 +559,32 @@ def _garbage_collect_manifest(manifest_id, context):
...
-            ManifestLabel.select().where(ManifestLabel.label == label_id).get()
-            return True
-        except ManifestLabel.DoesNotExist:
-            pass
-
-    return False
-
+def _bulk_garbage_collect_labels(context):
+    """
+    Bulk deletes all labels which are part of a specific manifest.
```
bcaton85: will it delete labels for a specific manifest or all manifests in the context?

bcaton85:
Looks good, claude recommended some additional tests:
```
  🔴 No tests for the new code paths

  No existing test creates manifests with labels or manifest list structures. The BFS traversal and bulk label deletion are completely dark in CI. The minimum bar:

  - test_gc_manifest_with_labels — actually exercises Label.delete()
  - test_gc_shared_label_survives_partial_gc — validates the reference check
  - test_gc_manifest_list_with_children — validates the BFS across levels
  - test_gc_manifest_list_partial_gc_child_in_use — validates Phase 1 + Phase 2 interaction
```

data/model/gc.py
ibazulic: @bcaton85 You're correct, it's a bit misleading. It will pick all labels from a specific context. Meaning deleting all labels for up to 10 manifests in context, because of the batching process. I'll correct it.

data/model/test/test_gc.py
```diff
@@ -1562,3 +1564,178 @@ def __exit__(self, *args):
...
+    assert Label.select().where(Label.id == label1.id).exists()
+    assert Label.select().where(Label.id == label2.id).exists()
+
+    # delete tag1, GC manifest1
+    with assert_gc_integrity(expect_storage_removed=True):
+        delete_tag(repo, "tag1", expect_gc=True)
+
+    # verify that label2 survived
+    assert Label.select().where(Label.id == label2.id).exists()
+
```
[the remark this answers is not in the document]
ibazulic: Should be resolved in the new commit.

jbpratt (dismissed):

kleesc (approved):
LGTM

PR #6101, opened 2026-06-04
QUAYIO-1805: feat(secscan): lock-free V2 security scanner indexing job

kleesc:
Add a new V2 security scanner indexer that replaces Redis-based global locking with PostgreSQL FOR UPDATE SKIP LOCKED for work distribution. Multiple worker instances claim disjoint batches of manifests without coordination, enabling horizontal scaling without contention or wasted duplicate scans.

Key changes:
- New V4SecurityScannerV2 indexer — two-phase claiming: optimistic INSERT for unindexed manifests, then SKIP LOCKED SELECT for failed/stale/reindex work. Includes per-manifest vulnerability notification on scan success.
- Interface split — SecurityScannerInterface split into SecurityScannerReadInterface (vulnerability lookups, notifications, GC) and SecurityScannerIndexerInterface (indexing). V2 implements only the indexer interface; reads always go through V1.
- Prometheus metrics — four new histograms/counters for V2 cycle duration, manifests claimed, scan duration, and scan results.
- Dead code removal — register_model_cleanup_callbacks and legacy_api_handler (no callers).

The intent is to deprecate V1's indexing code entirely — the interface split makes that a clean removal without touching the read side

### DIFF
--- a/data/secscan_model/__init__.py
+++ b/data/secscan_model/__init__.py
@@ -11,22 +11,38 @@ from data.secscan_model.interface import (
 from data.secscan_model.secscan_v4_model import NoopV4SecurityScanner
 from data.secscan_model.secscan_v4_model import ScanToken as V4ScanToken
 from data.secscan_model.secscan_v4_model import V4SecurityScanner
+from data.secscan_model.secscan_v4_model_v2 import V4SecurityScannerV2
 
 logger = logging.getLogger(__name__)
 
 
 class SecurityScannerModelProxy(SecurityScannerInterface):
     def configure(self, app, instance_keys, storage):
+        self._model_v2 = None
+
         try:
             self._model = V4SecurityScanner(app, instance_keys, storage)
         except InvalidConfigurationException:
             self._model = NoopV4SecurityScanner()
 
+        if app.config.get("FEATURE_SECURITY_SCANNER_V2", False):
+            self._model_v2 = V4SecurityScannerV2(app, instance_keys, storage)
+
         logger.info("===============================")
         logger.info("Using split secscan model: `%s`", [self._model])
+        if self._model_v2:
+            logger.info("Using V2 secscan indexer: `%s`", self._model_v2)
         logger.info("===============================")
 
         return self
+
+    @property
+    def v2_enabled(self):
+        return self._model_v2 is not None
+
+    def perform_indexing_v2(self, batch_size):
+        if self._model_v2 is not None:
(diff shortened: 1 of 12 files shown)


data/secscan_model/__init__.py
```diff
@@ -11,23 +11,39 @@
...
     def configure(self, app, instance_keys, storage):
+        self._model_v2 = None
+
         try:
             self._model = V4SecurityScanner(app, instance_keys, storage)
         except InvalidConfigurationException:
             self._model = NoopV4SecurityScanner()
 
+        if app.config.get("FEATURE_SECURITY_SCANNER_V2", False):
+            self._model_v2 = V4SecurityScannerV2(app, instance_keys, storage)
```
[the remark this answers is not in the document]
kleesc: Not really an issue, since V2 uses the same endpoint as v1

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -0,0 +1,393 @@
...
+            claimed_mss = self._find_and_claim_batch(
+                remaining, reindex_threshold, stale_threshold, indexer_hash
+            )
+
+        total_claimed = len(claimed_new) + len(claimed_mss)
+        secscan_v2_manifests_claimed.observe(total_claimed)
+
+        if total_claimed == 0:
+            logger.debug("No manifests to index this cycle")
+            return
```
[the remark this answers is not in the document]
kleesc: A default is set if None

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -0,0 +1,394 @@
...
+
+        try:
+            indexer_state = self._secscan_api.state()
+        except APIRequestFailure:
+            logger.exception("Failed to fetch indexer state from Clair")
+            return
+
+        indexer_hash = indexer_state.get("state", "")
+
+        reindex_threshold = datetime.utcnow() - timedelta(
```
cubismod, line 85: nit: `datetime.utcnow()` is deprecated

kleesc: I'll leave this as is, since we haven't updated the rest of the codebase. Especially since it's still in the old secscan model, and we're doing comparison against datetime columns.

cubismod: fair

cubismod (approved):

PR #6103, opened 2026-06-05
RFE-4378: feat(notifications): add FEATURE_SKIP_EMAIL_CONFIRMATION flag

ranjithrajaram:
Trying to address https://redhat.atlassian.net/browse/RFE-4378

Add configuration flag to bypass email confirmation requirement for repository notifications in trusted enterprise environments.

When FEATURE_SKIP_EMAIL_CONFIRMATION is False (default), users must confirm email addresses before receiving notifications. When True, notifications can be sent to any address without confirmation.

Changes:
- Add FEATURE_SKIP_EMAIL_CONFIRMATION to config.py (default: False)
- Add schema definition in util/config/schema.py with security guidance
- Modify EmailMethod.validate() to check features.SKIP_EMAIL_CONFIRMATION
- Add test coverage for both enabled and disabled states
- Add config loading path test to verify flag wiring
- Add FEATURE_SKIP_EMAIL_CONFIRMATION to Go knownUnmapped (Python-only)

Security: Defaults to False for secure-by-default behavior. Only enable in trusted enterprise environments with internal email addresses.

### DIFF
--- a/config.py
+++ b/config.py
@@ -364,6 +364,12 @@ class DefaultConfig(ImmutableConfig):
     # Feature Flag: Whether organizations can share email addresses with other users/orgs.
     FEATURE_ORG_SHARED_EMAIL = False
 
+    # Feature Flag: Whether to skip email confirmation requirement for repository notifications.
+    # When False (default), users must confirm email addresses before they can receive repository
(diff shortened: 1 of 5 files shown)


ranjithrajaram:
@coderabbitai full review

PR #6109, opened 2026-06-07
NO-ISSUE: feat(cmd): add quay migrate command for OMR-to-Go upgrade path

jbpratt:
## Summary

Adds a `quay migrate` CLI command that provides a seamless upgrade path from mirror-registry (OMR) v1.4.x to the new Go-based `quay` binary.

A customer upgrading from OMR runs one command with no arguments:

```bash
./quay migrate
```

The command auto-detects the old OMR installation (systemd units, Podman volumes, config paths), discovers the hostname from the old `config.yaml`, finds the image archive in the binary's directory, and handles the entire upgrade: stop old services → copy data → convert storage format → upgrade schema → deploy new registry → verify health.

Flags exist as overrides if auto-detection gets something wrong, but the happy path is zero arguments.

## How it works

Nine sequential phases:

1. **Detect** — inspects systemd units → Podman volumes → ~/quay-install defaults. Extracts hostname, resolves volume mountpoints, finds image archive next to the binary.
2. **Validate** — DB integrity check, Alembic version whitelist, cert/key pair match, storage dir, target dir is clean.
3. **Stop** — stops old OMR services (quay-app, quay-redis, quay-pod) to free port 8443.
4. **Copy** — copies DB (renamed quay_sqlite.db → quay.db), flattens certs to data-dir root, recursively copies blob storage. Idempotent on re-run.
5. **Convert** — transforms Python Quay's flat storage layout (`sha256/<prefix>/<digest>`) to distribution format (`docker/registry/v2/blobs/sha256/<prefix>/<digest>/data`) using hard-links (zero copy cost). Writes manifest bytes from DB as blobs. Creates repository metadata tree (revision links, tag links, layer links) from DB queries.
6. **Schema Upgrade** — bridge migration bringing any OMR SQLite version to Go binary's target via idempotent SQL (CREATE TABLE IF NOT EXISTS, INSERT OR IGNORE) + Go helpers for ADD COLUMN and index uniqueness fixes.
7. **Install** — chains into existing installer.Run() for Quadlet deployment.
8. **Verify** — healthz check against the live service using custom TLS client.
9. **Cleanup** (opt-in via -cleanup) — removes old units, volumes, Redis secret. Preserves ~/quay-install rootCA. Can be run standalone after a successful migration.

## Design decisions

- **Reuses old TLS certs** — disconnected customers may have distributed the CA. Regenerating breaks trust chains.
- **Copy, don't move** — rollback = delete target data-dir + restart old OMR.
- **Hard-link storage conversion** — O(1) per blob regardless of size. Falls back to copy if hard link fails.
- **Bridge migration with version whitelist** — handles any OMR SQLite version, refuses unknown versions with a clear error.
- **slog for all logging** — structured logging throughout, fmt only for dry-run plan output.
- **No OMR changes** — the old mirror-registry project is not modified.

## E2E test

New e2e-migrate GHA job: downloads offline OMR installer → installs old OMR → pushes test images → runs `quay migrate` with zero args → trusts migrated cert → verifies images pullable via podman pull → runs cleanup.

## Test plan

- [x] 16 migrate package tests + 7 bridge tests + 7 system tests pass
- [x] golangci-lint clean
- [x] Binary builds, `quay migrate -h` shows flags
- [x] Graceful failure when no OMR is installed
- [x] E2E GHA job passes

### DIFF
--- a/.github/workflows/ci-go.yaml
+++ b/.github/workflows/ci-go.yaml
@@ -105,7 +105,7 @@ jobs:
         run: make go-schema-check
 
   e2e-mirror:
-    name: E2E Mirror
+    name: E2E Mirror Install
     runs-on: ubuntu-latest
     needs: [build]
     timeout-minutes: 30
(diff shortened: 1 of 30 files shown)


jbpratt:
@coderabbit resolve

internal/dal/dbcore/bridge.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	if !knownOMRVersions[ver] {
+		return fmt.Errorf(
+			"unknown schema version %q — this tool supports OMR v1.3.x/v1.4.x databases; "+
+				"upgrade to a newer version of this binary or contact support", ver)
+	}
+
+	fmt.Fprintf(w, "Bridging schema from %s to %s\n", ver, TargetVersion)
+
+	if _, err := db.ExecContext(ctx, "PRAGMA foreign_keys = OFF"); err != nil {
```
harishsurf, line 90: **MAJOR: Error message claims v1.3.x/v1.4.x support but the bridge only handles v2.0.x (SQLite) databases**

`knownOMRVersions` contains only Quay 3.12.x era Alembic hashes, which correspond to OMR v2.0.x (SQLite). OMR v1.3.x uses Postgres and will never have a `quay_sqlite.db` file. A customer on v1.3.x would read this error, think they are supported, and not understand why the tool cannot find their database.

The same misleading version appears in `0001_bridge_from_omr.sql` line 3.

**Fix:** Change to: `"this tool supports OMR v2.0.x (SQLite) databases — if you are on OMR v1.3.x, upgrade to v2.0 first"`

.github/workflows/ci-go.yaml
```diff
@@ -241,3 +241,446 @@ jobs:
...
+        with:
+          name: quay-binary
+          path: bin/
+
+      - name: Build container image
+        run: |
+          chmod +x bin/quay
+          podman build -f Dockerfile.mirror -t quay-mirror:test .
+
+      - name: Save image archive next to binary
```
harishsurf, line 267: **MAJOR: E2E test downloads OMR v2.0.3 instead of v2.0.11**

`releases/latest` returns v2.0.3 because v2.0.4–v2.0.11 are all marked as pre-releases on GitHub. The test misses coverage for any schema or storage changes shipped in newer OMR versions.

**Fix:** Pin to a specific version:

```yaml
curl -fsSL -o /tmp/mirror-registry.tar.gz \
  https://github.com/quay/mirror-registry/releases/download/v2.0.11/mirror-registry-offline.tar.gz
```

internal/migrate/copy.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+
+	// Copy blob storage recursively.
+	if m.Source.StoragePath != "" {
+		targetStorage := filepath.Join(m.DataDir, "storage")
+		count, totalBytes, err := copyDirRecursive(m.Source.StoragePath, targetStorage, m.Out)
+		if err != nil {
+			return fmt.Errorf("copy storage: %w", err)
+		}
```
harishsurf, line 70: **MAJOR: No container UID/permission handling for migrated files**

Old OMR runs Python Quay as UID 1001 inside the container, with ACLs (`entity: 1001, permissions: wx`) and the `:U` Podman volume flag for ownership remapping. The migration copies files preserving source permissions (`srcInfo.Mode()`).

The new Go container (`Dockerfile.mirror`) has no `USER` directive so it runs as root — this works today. But if the Dockerfile is later updated to run as a non-root user (security best practice), migrated files owned by UID 1001 become inaccessible.

**Fix:** Not a blocker now, but document this assumption. When non-root container support is added, the migration will need a `chown` or ACL step for the target data directory.

internal/migrate/copy.go
```diff
@@ -0,0 +1,142 @@
...
+
+	markerPath := filepath.Join(m.DataDir, markerFile)
+	if err := os.WriteFile(markerPath, []byte("migration in progress\n"), 0o600); err != nil {
+		return fmt.Errorf("write marker: %w", err)
+	}
+
+	// Copy database (rename quay_sqlite.db → quay.db).
+	if err := checkpointSQLite(ctx, m.Source.DBPath); err != nil {
+		return fmt.Errorf("checkpoint source database: %w", err)
+	}
```
harishsurf, line 38: **MODERATE: `copyFileIdempotent` uses size-only comparison for skip detection**

If a previous migration was interrupted mid-write and the partial file happens to match the source file size, the copy is skipped and the target database could be corrupted. The post-migration integrity check would catch this, but only after the schema upgrade phase has already run against the bad copy.

**Fix:** When the migration marker file exists (indicating an incomplete previous run), always recopy the database file regardless of size match. The marker is already checked in `validateTargetDir`.

internal/dal/dbcore/migrate.go
```diff
@@ -328,7 +330,16 @@ func extractRevisionID(migrationSQL string) (string, error) {
...
 // and discarding empty entries.
 func splitStatements(rawSQL string) []string {
-	raw := strings.Split(rawSQL, ";")
+	var withoutComments strings.Builder
+	for _, line := range strings.Split(rawSQL, "\n") {
+		if strings.HasPrefix(strings.TrimSpace(line), "--") {
+			continue
+		}
+		withoutComments.WriteString(line)
+		withoutComments.WriteByte('\n')
```
harishsurf, line 340: **MODERATE: `splitStatements` does not handle inline comments containing semicolons**

The fix correctly strips full-line `--` comments before splitting on `;`, but inline comments like `SELECT 1; -- comment with ; semicolon` would still break the split. Safe today since `0001_bridge_from_omr.sql` has no inline comments with semicolons, but fragile for future migration files.

**Fix:** Acceptable as-is. Consider adding a comment noting the limitation for future migration authors.

internal/migrate/migrate.go
```diff
@@ -0,0 +1,175 @@
...
+	Image        string
+	ImageArchive string
+
+	SourceRoot    string
+	SourceDB      string
+	SourceStorage string
+	SourceCerts   string
+
+	DryRun      bool
+	Cleanup     bool
```
harishsurf, line 51: **MODERATE: `SkipVerify` field exists on the struct but is never used**

The `Migrator` struct has a `SkipVerify` field, but `migrateData()` does not call a verify step, and the CLI does not expose a `-skip-verify` flag. This is dead code.

**Fix:** Either remove `SkipVerify` from the struct, or wire up the verify phase and add the corresponding CLI flag.

internal/migrate/schema.go
```diff
@@ -0,0 +1,105 @@
...
+	if len(m.Source.UnitFiles) == 0 {
+		slog.Info("no OMR services detected, skipping stop")
+		return nil
+	}
+
+	var scopeArgs []string
+	if m.Source.SystemdScope == scopeUser {
+		scopeArgs = []string{"--user"}
+	}
+
```
harishsurf, line 69: **MAJOR: `stopOldOMR` swallows all errors — migration proceeds with potentially running services**

`systemctl stop` failures are logged as warnings and the function always returns `nil`. If `quay-app` fails to stop (e.g., permission denied, systemd unresponsive), the migration continues to checkpoint and copy the database while the old Quay is still writing to it. The copied database may have uncommitted WAL transactions — corrupted or stale data served after migration, recently pushed images silently lost.

`quay-app` is the critical service — it holds the database and storage open. Redis and pod failures are less severe and can remain as warnings.

**Fix:**

```go
if err := m.Runner.Run(ctx, "systemctl", args...); err != nil {
    if svc == "quay-app" {
        return fmt.Errorf("failed to stop %s (database may be in use): %w", svc, err)
    }
    slog.Warn("failed to stop service (may already be stopped)", "service", svc, "err", err)
}
```

jbpratt: Updated in a2bc72c30. I made this stricter than the suggested patch: if any detected OMR systemd service stop fails, migration now returns an error before checkpoint/copy/schema work proceeds. It still attempts all known OMR stop commands first so the host is left as quiesced as possible, then returns a joined error with the failing service names.

Already-stopped services should continue to work because systemctl stop is idempotent. If no unit files are detected, stopOldOMR still skips stopping so podman-volume/default/manual migration paths continue to work for installs where units were already deleted.

Added tests for quay-app stop failure, redis/pod stop failures, no detected unit files, and all stops succeeding. Validated with `GOCACHE=/tmp/go-build go test ./internal/migrate`, `GOCACHE=/tmp/go-build GOLANGCI_LINT_CACHE=/tmp/golangci-lint-cache golangci-lint run --timeout=5m`, and `git diff --check`.

harishsurf (approved):

PR #6110, opened 2026-06-08
PROJQUAY-11432: fix(api): fix BuildTrigger.to_dict() None check and add superuser build tests

quay-devel:
## Summary

- Fix `BuildTrigger.to_dict()` dead-code logic bug: the original `if not self.trigger and not self.trigger.uuid:` condition would raise `AttributeError` when `self.trigger is None` because `and` doesn't short-circuit on a truthy left operand — it evaluates both sides. Simplified to `if not self.trigger:`.
- Add integration tests verifying superuser build endpoints handle both trigger=None and trigger-present cases correctly.

Note: the primary `superuser_models_pre_oci.py` guard and `RepositoryBuild.to_dict()` trigger guard were already applied to master independently. This PR completes the fix with the remaining logic correction and test coverage.

## Root Cause / Rationale

`RepositoryBuild.trigger` is a nullable FK. `BuildTrigger.to_dict()` contained a broken guard:

```python
# broken: if trigger is None, `not self.trigger` is True, so Python evaluates
# `not self.trigger.uuid` → AttributeError: 'NoneType' has no attribute 'uuid'
if not self.trigger and not self.trigger.uuid:
    return None
```

While the upstream callers now guard before calling `to_dict()`, the defensive fix is still correct and removes confusing dead code.

## Changes

- `endpoints/api/superuser_models_interface.py`: Fix `BuildTrigger.to_dict()` None guard
- `endpoints/api/test/test_superuser.py`: Add integration tests for superuser build endpoints with and without triggers

## Test Plan

- [x] `pytest endpoints/api/test/test_superuser.py::test_get_repository_build_no_trigger -v` — passes
- [x] `pytest endpoints/api/test/test_superuser.py::test_get_repository_build_with_trigger -v` — passes
- [x] All CI checks passing

## JIRA

https://redhat.atlassian.net/browse/PROJQUAY-11432

## Automation
- **Session ID**: session-72226051-2da7-4585-8fce-9bf398c9371b

### DIFF
--- a/endpoints/api/superuser_models_interface.py
+++ b/endpoints/api/superuser_models_interface.py
@@ -55,7 +55,7 @@ class BuildTrigger(
     """
 
     def to_dict(self):
-        if not self.trigger and not self.trigger.uuid:
+        if not self.trigger:
             return None
 
         build_trigger = BuildTriggerHandler.get_handler(self.trigger)
--- a/endpoints/api/test/test_superuser.py
+++ b/endpoints/api/test/test_superuser.py
@@ -1,11 +1,13 @@
 import pytest
 
+from data import model
 from data.database import DeletedNamespace, User
 from endpoints.api.superuser import (
     SuperUserDumpConfig,
     SuperUserList,
     SuperUserManagement,
     SuperUserOrganizationList,
+    SuperUserRepositoryBuildResource,
 )
 from endpoints.api.test.shared import conduct_api_call
 from endpoints.test.shared import client_with_identity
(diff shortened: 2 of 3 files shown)


quay-devel:
@coderabbitai The Docstring Coverage warning: no existing `to_dict()` method in `superuser_models_interface.py` has an individual docstring — they rely on class-level documentation. Adding one only to the method we touched would be inconsistent with the surrounding code. The two new test functions do include docstrings. Resolving as won't-fix for consistency.

LiZhang19817:
close this PR as there's a duplicated PR merged.
https://github.com/quay/quay/pull/6111

quay-devel:
Closing — the fix is fully covered by #6111 ([PROJQUAY-11678](https://redhat.atlassian.net/browse/PROJQUAY-11678)) which was already merged on 2026-06-09. Apologies for the repeated reopens; I was responding to JIRA-unlink notifications without realising the team had already superseded this PR.

PR #6126, opened 2026-06-09
PROJQUAY-11691: fix(permissions): Optimize loading of permissions on UI login

ibazulic:
Loading of permissions, particularly in an LDAP environment, for large sets of repositories and organizations, can be very slow with the current loading techniques. This is mainly because of multiple permission checks and needless password verifications. The optimiziations put here resolve permission loading and speed up UI bootstrap.

On a registry that contains 20 users, 50 organizations and roughly 300 repositories, loading of permissions is dramatically faster with the optimizations rather than without them. For `api/v1/user` endpoint crucial during UI rendering: 

~~~
  ┌─────────────────────┬────────┬───────┬─────────────┐
  │        User         │ Before │ After │ Improvement │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Superuser (50 orgs) │ 14.43s │ 0.29s │ ~50x faster │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Normal user         │ 4.72s  │ 0.30s │ ~16x faster │
  └─────────────────────┴────────┴───────┴─────────────┘

  Overall SQL queries: 4,075 → 3,082 (~25% reduction across the entire session)
~~~

By the same token, LDAP queries:

~~~
  ┌─────────────────┬────────┬───────┬───────────┐
  │     Metric      │ Before │ After │ Reduction │
  ├─────────────────┼────────┼───────┼───────────┤
  │ LDAP operations │ 6,760  │ 4,504 │ 33% fewer │
  ├─────────────────┼────────┼───────┼───────────┤
  │ LDAP binds      │ 563    │ 375   │ 33% fewer │
  ├─────────────────┼────────┼───────┼───────────┤
  │ LDAP searches   │ 2      │ 2     │ same      │
  └─────────────────┴────────┴───────┴───────────┘
~~~

Metrics were collected by comparing the current Quay 3.17.2 with no LDAP caching vs an image built with the optimizations in the PR, but without the LDAP pooling code. With LDAP pooling, the number of LDAP operations would be dramatically lower.

Co-authored-by: Claude [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/auth/permissions.py
+++ b/auth/permissions.py
@@ -282,7 +282,7 @@ class QuayDeferredPermissionUser(Identity):
 
             self._superuser_checked = True
 
-        # Super users can do anything on the repositories so we can simply return here if
+        # Super users can do anything on the registry so we can simply return here if
         # full access is enabled
         if self._superuser_loaded and features.SUPERUSERS_FULL_ACCESS:
             if permission.namespace or permission.repo_name:
@@ -338,7 +338,11 @@ class QuayDeferredPermissionUser(Identity):
                 return super(QuayDeferredPermissionUser, self).can(permission)
 
         # Lazy-load the namespace-wide-only permissions.
-        if perm_namespace and perm_namespace not in self._namespace_wide_loaded:
+        if (
+            perm_namespace
+            and perm_namespace not in self._namespace_wide_loaded
+            and not self._all_namespaces_loaded
+        ):
             self._populate_namespace_wide_provides(user_object, perm_namespace)
             self._namespace_wide_loaded.add(perm_namespace)
 
--- a/auth/permissions.py
+++ b/auth/permissions.py
@@ -5,6 +5,7 @@ from typing import DefaultDict, Optional
 
 from flask_principal import Identity, Permission, identity_changed, identity_loaded
 
+import features
 from app import app, model_cache, usermanager
 from auth import scopes
 from data import model
@@ -86,6 +87,13 @@ class QuayDeferredPermissionUser(Identity):
         self._repositories_loaded = set()
         self._repositories_loaded_from_primary = set()
         self._personal_loaded = False
+
+        self._superuser_loaded = False
+        self._readonly_superuser_loaded = False
+
+        self._superuser_checked = False
+
+        self._all_namespaces_loaded = False
 
         self._scope_set = auth_scopes
         self._user_object = user
@@ -174,6 +182,8 @@ class QuayDeferredPermissionUser(Identity):
             )
             logger.debug("Team added permission: {0}".format(team_grant))
             self.provides.add(team_grant)
+            if namespace_filter is None:
+                self._namespace_wide_loaded.add(team.organization.username)
 
     def _populate_repository_provides(
         self, user_object, namespace_filter, repository_name, for_write=False
@@ -249,28 +259,44 @@ class QuayDeferredPermissionUser(Identity):
             logger.debug("User added permission: {0}".format(repo_grant))
             self.provides.add(repo_grant)
 
-    def _populate_superuser_provides(self, user_object):
-        # Global readonly superusers and regular superusers are mutually exclusive.
-        # If a user is in both lists (misconfiguration), readonly takes precedence.
-        if usermanager.is_global_readonly_superuser(user_object.username):
-            logger.debug("Adding global readonly superuser to user: %s", user_object.username)
-            self.provides.add(_GlobalReadOnlySuperUserNeed())
-        elif (
-            scopes.SUPERUSER in self._scope_set or scopes.DIRECT_LOGIN in self._scope_set
-        ) and usermanager.is_superuser(user_object.username):
-            logger.debug("Adding superuser to user: %s", user_object.username)
-            self.provides.add(_SuperUserNeed())
-
     def can(self, permission):
         logger.debug("Loading user permissions after deferring for: %s", self.id)
         user_object = self._user_object or model.user.get_user_by_uuid(self.id)
         if user_object is None:
             return super(QuayDeferredPermissionUser, self).can(permission)
 
+        # determine if this user is a super user or a normal user
+        # Global readonly superusers and regular superusers are mutually exclusive.
+        # If a user is in both lists (misconfiguration), readonly takes precedence.
+        if not self._superuser_checked:
+            if usermanager.is_global_readonly_superuser(user_object.username):
+                logger.debug("Adding global readonly superuser to user: %s", user_object.username)
+                self.provides.add(_GlobalReadOnlySuperUserNeed())
+                self._readonly_superuser_loaded = True
+            elif (
+                scopes.SUPERUSER in self._scope_set or scopes.DIRECT_LOGIN in self._scope_set
+            ) and usermanager.is_superuser(user_object.username):
+                logger.debug("Adding superuser to user: %s", user_object.username)
+                self.provides.add(_SuperUserNeed())
+                self._superuser_loaded = True
+
+            self._superuser_checked = True
+
+        # Super users can do anything on the repositories so we can simply return here if
+        # full access is enabled
+        if self._superuser_loaded and features.SUPERUSERS_FULL_ACCESS:
+            if permission.namespace or permission.repo_name:
+                return True
+
         # Add the user-specific provides.
         if not self._personal_loaded:
             self._populate_user_provides(user_object)
             self._personal_loaded = True
+
+        # Add all permissions for all namespaces but only if we have direct login (UI)
+        if not self._all_namespaces_loaded and scopes.DIRECT_LOGIN in self._scope_set:
+            self._populate_namespace_wide_provides(user_object, namespace_filter=None)
+            self._all_namespaces_loaded = True
 
         # If we now have permission, no need to load any more permissions.
         if super(QuayDeferredPermissionUser, self).can(permission):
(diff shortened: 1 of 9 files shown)


test/registry/registry_tests.py
```diff
@@ -1702,8 +1702,8 @@ def test_tags_disabled_namespace(
         ("devtable", "buynlarge", "baserepo", "buynlarge/baserepo", None),
         # Unsuccessful mount, unknown repo.
         ("devtable", "devtable", "baserepo", "unknown/repohere", Failures.UNAUTHORIZED_FOR_MOUNT),
-        # Unsuccessful mount, no access.
-        ("public", "public", "baserepo", "public/baserepo", Failures.UNAUTHORIZED_FOR_MOUNT),
+        # Successful mount, super user has access to everything if FEATURE_SUPERUSERS_FULl_ACCESS is set
+        ("public", "public", "baserepo", "public/baserepo", None),
```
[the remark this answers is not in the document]
ibazulic: This addition has been reverted due to other changes.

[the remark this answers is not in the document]
ibazulic: Pardon, that was a different comment. It's resolved in the new commit.

endpoints/v2/v2auth.py
```diff
@@ -257,6 +257,19 @@ def _authorize_or_downscope_request(scope_param, has_valid_auth_context):
...
+    # exit immediately with empty scopes.
+    namespace_exists = model.user.get_namespace_user(namespace) is not None
+    if not namespace_exists and not app.config.get("CREATE_NAMESPACE_ON_PUSH", False):
+        return scopeResult(
+            actions=[],
+            namespace=namespace,
+            repository=reponame,
+            registry_and_repo=registry_and_repo,
+            tuf_root=_get_tuf_root(repository_ref, namespace, reponame),
+        )
```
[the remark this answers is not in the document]
ibazulic: The `v2_auth.py` is very difficult to follow as is and requires major refactoring all the way. This potentially adds 2-3 `SELECT` queries, not a heavy database operation, which is still considerably less than what we had before (other optimizations). Given that the whole granting process requires simplification, I think this is an acceptable tradeoff.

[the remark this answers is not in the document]
ibazulic: No need, I have it in my sights for a future PR. Thank you!

test/registry/registry_tests.py
```diff
@@ -2091,6 +2135,39 @@ def test_login_scopes(
...
+        if not expect_success:
+            return
+
+        # Validate the returned token
+        encoded = response.json()["token"]
+        payload = decode_bearer_header(
+            "Bearer " + encoded, instance_keys, {"SERVER_HOSTNAME": "localhost:5000"}
+        )
+        assert payload is not None
+        assert payload["access"] == expected_access
```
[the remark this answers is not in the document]
ibazulic: This should be addressed in the new commit. Additional tests are added.

ibazulic:
With enabled LDAP pooling and 10 pooled connections:

~~~
  ┌─────────────────┬────────┬───────────┬─────────────────────┬─────────────────┐
  │     Metric      │ Before │ Optimized │ Optimized + Pooling │ Total reduction │
  ├─────────────────┼────────┼───────────┼─────────────────────┼─────────────────┤
  │ /api/v1/user    │ 14.43s │ 0.29s     │ 0.14s               │ ~100x faster    │
  ├─────────────────┼────────┼───────────┼─────────────────────┼─────────────────┤
  │ SQL queries     │ 4,075  │ 3,082     │ 2,214               │ 46% fewer       │
  ├─────────────────┼────────┼───────────┼─────────────────────┼─────────────────┤
  │ LDAP operations │ 6,760  │ 4,504     │ 358                 │ 95% fewer       │
  ├─────────────────┼────────┼───────────┼─────────────────────┼─────────────────┤
  │ LDAP binds      │ 563    │ 375       │ 34                  │ 94% fewer       │
  ├─────────────────┼────────┼───────────┼─────────────────────┼─────────────────┤
  │ Log lines       │ 39,493 │ 35,968    │ 15,689              │ 60% fewer       │
  └─────────────────┴────────┴───────────┴─────────────────────┴─────────────────┘
~~~

This is with a local database but with a remote LDAP server. Real numbers will be a bit higher but the ratio of lookups/queries vs what we had before is what makes the difference here. Even with a remote db, we should still see huge improvements during loading.

auth/permissions.py
```diff
@@ -87,6 +88,13 @@ def __init__(self, uuid, auth_type, auth_scopes, user=None):
         self._repositories_loaded_from_primary = set()
         self._personal_loaded = False
 
+        self._superuser_loaded = False
+        self._readonly_superuser_loaded = False
```
jbpratt, line 95: We never read the value of this it seems

endpoints/v2/v2auth.py
```diff
@@ -358,7 +371,11 @@ def _authorize_or_downscope_request(scope_param, has_valid_auth_context):
                         except model.DataModelException as ex:
                             raise Unsupported(message="Cannot create organization")
 
-                if CreateRepositoryPermission(namespace).can() and user is not None:
+                if (
+                    CreateRepositoryPermission(namespace).can()
+                    and user is not None
+                    and model.user.get_namespace_user(namespace) is not None
```
jbpratt:
```suggestion
                    and namespace_exists
```

test/registry/registry_tests.py
```diff
@@ -2091,6 +2135,158 @@ def test_login_scopes(
...
+    expect_success,
+    v2_protocol,
+    liveserver_session,
+    app_reloader,
+    registry_server_executor,
+    liveserver,
+):
+    """
+    Tests that super users have access to a non-existent namespace/repo if SUPERUSERS_FULL_ACCESS
+    is enabled and CREATE_NAMESPACE_ON_PUSH is .
```
jbpratt:
```suggestion
    is enabled and CREATE_NAMESPACE_ON_PUSH is enabled
```

jbpratt (dismissed):
Nice stuff @ibazulic , few things to resolve

auth/permissions.py
ibazulic: We don't read `self._readonly_superuser_loaded`, true. that was put in place to short circuit read only super users as well. However, later, I thought about it some more, global read only super users usually are used as service accounts, for external workloads such as image scanners via the Docker v2 API. In that case, they do not need optimized loading, as we specify the `namespace/repo_name` during Docker v2 auth. So this was just left over. I would recommend we keep it though, if we decide to implement the same kind of short circuit later.

As for `self._superuser_loaded`, we do use it, see line 287.

endpoints/v2/v2auth.py
ibazulic: Yeah, this is an extra query that was made redundant with the `namespace_exists`. This adds one more query, but given that the whole `v2_auth.py` requires thorough redesign as it's incredibly difficult to follow, I didn't consider it a problem.

test/registry/registry_tests.py
ibazulic: Should be resolved in the new commit.

endpoints/v2/v2auth.py
ibazulic: Removed in the new commit.

auth/permissions.py
jbpratt: Instead of just an un-read var, could we add a comment here mentioning the future improvement?

ibazulic: Resolved in the new commit. Added a comment explaining the flag and future use.

endpoints/v2/v2auth.py
```diff
@@ -358,7 +371,11 @@ def _authorize_or_downscope_request(scope_param, has_valid_auth_context):
                         except model.DataModelException as ex:
                             raise Unsupported(message="Cannot create organization")
 
-                if CreateRepositoryPermission(namespace).can() and user is not None:
+                if (
+                    CreateRepositoryPermission(namespace).can()
+                    and user is not None
+                    and namespace_exists is not None
```
harishsurf: this seems like a dead code? `namespace_exists` is assigned as a bool (on line 263), so so `namespace_exists is not None` is ALWAYS True.

jbpratt:
```suggestion
                    and namespace_exists
```
this is already a bool from L263

endpoints/v2/v2auth.py
harishsurf, line 356: nit: `model.user.get_namespace_user(namespace)` - this call could be replaced by the `namespace_exists` cached value you created on line 263

ibazulic: The whole `v2auth.py` needs to be refactored. This code has existed and I have not touched it. This adds one query that we run, while other changes to the permission model save almost 50% of the queries. Given that the file needs to be simplified, I would leave it as it is for now. It does not impact performance and it does not introduce any bugs.

endpoints/v2/v2auth.py
ibazulic: See my other comment.

ibazulic: Ah, ignore, I'll change. The same issue has been raised by Brady.

ibazulic: Should be resolved in new commit.

ibazulic: Should be resolved in new commit.

jbpratt (approved):

PR #6134, opened 2026-06-10
PROJQUAY-11606: feat(bootstrap): add programmatic OAuth token provisioning

shaonrh:
Enable fully automated Quay deployments by allowing OAuth API tokens to be created programmatically, without requiring UI interaction.

New endpoints:
- `POST /api/v1/bootstrap/token` — create OAuth tokens via Basic/Bearer auth
- `GET /api/v1/bootstrap/tokens` — list bootstrap-created tokens with filters
- `DELETE /api/v1/bootstrap/tokens/<uuid>` — revoke a bootstrap token
- `POST/GET/DELETE /api/v1/organization/<org>/applications/<client_id>/tokens` — per-application token lifecycle management

Gated behind `FEATURE_PROGRAMMATIC_BOOTSTRAP` (default: disabled). Only superusers are authorized to use bootstrap endpoints.


ibazulic:
I don't like the fact that Oauth API sits behind basic auth. If someone gets a super user username/password, then that person can provision for themselves an Oauth bearer token and have full access on everything in the registry. IMHO this is a massive attack vector. I am fine with provisioning tokens through the API, but it has to be protected by another Bearer token. Or there should be a specific auth endpoint where a bearer token would be returned back so that we can provision later tokens. If user initialization is enabled, then the user who sets up Quay can initialize a new user (provided the internal auth method is `Database`) and the endpoint will give the user a bearer token to bootstrap other tokens.

PR #6150, opened 2026-06-11
NO-ISSUE: fix(ci): use --dist loadfile for PG integration tests

jbpratt:
## Summary
- Switch pytest-xdist from `--dist load` (round-robin) to `--dist loadfile` for `py312-psql`
- Ensures tests from the same file run on the same xdist worker, preventing cross-file PostgreSQL deadlocks
- Matches the strategy already used by `py312-sqlite`

## Problem
The PostgreSQL Integration Test CI job fails ~20% of the time on master due to deadlocks between tests in different files (e.g., `test_gc.py` and `test_secscan_v4_model.py`) that write to shared tables while running on different xdist workers.

## Test plan
- [ ] CI PostgreSQL Integration Test passes on this PR
- [ ] Monitor 10+ subsequent CI runs for flake rate reduction


jbpratt:
Closing — `--dist loadfile` breaks legacy test initialization. gw1's first test (`test_v2_endpoint_security.py`, a legacy unittest.TestCase) fails with 'models are missing data' because the xdist worker ordering puts it on a worker that can't see the populated data. Need a different approach.

PR #6154, opened 2026-06-11
QUAYIO-1637: feat(spam): add repository ingress spam detection

HammerMeetNail:
## Summary

Adds opt-in repository-description spam detection to Quay's create and update ingress paths.

Quay now:

- evaluates descriptions locally with a versioned Bayesian classifier artifact;
- requires an HTTP/HTTPS hyperlink before a score can reject a description;
- uses bounded, overlapping token windows to prevent link padding from diluting spam scores;
- supports dry-run and fail-open/fail-closed operation;
- verifies the configured artifact version and optional SHA-256 checksum;
- emits `quay_spam_ingress_decisions_total{action,outcome}` for create/update outcomes; and
- returns HTTP 400 without persisting the rejected create or description update.

Quay never calls `quay-service-tool` on the request path. Classifier training, artifact generation, scanning, review, quarantine, restore, and remediation history remain owned by `quay-service-tool#216`.

## Configuration

| Setting | Purpose |
| --- | --- |
| `FEATURE_SPAM_DETECTION` | Enables ingress evaluation |
| `SPAM_DETECTION_DRY_RUN` | Evaluates and records metrics without rejecting |
| `SPAM_DETECTION_FAIL_OPEN` | Allows writes when the classifier is unavailable |
| `SPAM_DETECTION_CLASSIFIER_PATH` | Local JSON artifact, defaulting to `/conf/spam-detection/classifier.json` |
| `SPAM_DETECTION_CLASSIFIER_VERSION` | Expected artifact version |
| `SPAM_DETECTION_CLASSIFIER_SHA256` | Optional artifact checksum |

## Validation

- Classifier schema, checksum, visibility threshold, repository-name feature, hyperlink gating, and bounded-window tests.
- Create/update tests for disabled, dry-run, enforced, fail-open, and fail-closed behavior.
- Playwright API coverage for the baked artifact path and enforced create/update rejection.
- Quay pre-commit hooks pass for all changed files.

Related enhancements: `quay/enhancements#42`, `quay/enhancements#44`  
JIRA: https://redhat.atlassian.net/browse/QUAYIO-1637

## Backport

- [ ] No backport needed


HammerMeetNail:
@coderabbitai full review

HammerMeetNail:
@coderabbitai review

HammerMeetNail:
Related PRs:
- Enhancement proposal: https://github.com/quay/enhancements/pull/42
- Service-tool training, scanning, quarantine, and remediation workflows: https://github.com/quay/quay-service-tool/pull/216

data/model/spam_ingress.py
```diff
@@ -0,0 +1,205 @@
...
+
+DEFAULT_INGRESS_THRESHOLD = 0.9
+DEFAULT_TOKEN_PATTERN = r"[a-z0-9][a-z0-9_-]*"
+REQUIRED_ARTIFACT_FIELDS = {
+    "version": str,
+    "training_corpus_version": str,
+    "spam_prior": (int, float),
+    "ham_prior": (int, float),
+    "token_spam_counts": dict,
+    "token_ham_counts": dict,
```
deshpandevlab, line 24: nit: `spam_token_total`, `ham_token_total`, and `vocabulary_size` are validated as strict `int`, but `json.loads` deserializes `100.0` as `float`. If the service-tool (or any JSON serialization path) produces these as floats, artifact loading will fail with `SpamIngressUnavailable`.

Should we consider accepting `(int, float)` here and casting to `int` in `BayesianSpamClassifier.__init__`, or documenting in the artifact contract that these fields must be JSON integers (no decimal point).

deshpandevlab (approved):
Looks good — well-scoped to the Quay-side ingress hook, safe defaults, clean module design.

One non-blocking note below on artifact field type validation.

endpoints/api/repository.py
```diff
@@ -69,6 +71,57 @@
...
+        return
+
+    context = spam_ingress.SpamIngressContext(
+        namespace=namespace,
+        repository=repository,
+        description=description,
+        visibility=visibility,
+        action=action,
+    )
+
```
deshpandevlab, line 100: non-blocking: This catches bare `Exception`, which means a programming bug (e.g. `TypeError`, `AttributeError`) in `spam_ingress.py` is indistinguishable from a missing/corrupt artifact in logs — both emit `"Spam detection ingress classifier unavailable"`. Since fail-open must be preserved for both cases (a 500 would block repo create/update), consider a two-tier catch so operators can alert on unexpected errors separately:

```python
try:
    decision = spam_ingress.evaluate_description(context, app.config)
except spam_ingress.SpamIngressUnavailable:
    logger.exception(
        "Spam classifier unavailable; allowing %s for %s/%s",
        action, namespace, repository,
    )
    return
except Exception:
    logger.exception(
        "Unexpected spam classifier error; allowing %s for %s/%s",
        action, namespace, repository,
    )
    return
```

Both paths fail-open, but the distinct log lines let you tell artifact issues from code bugs.

data/model/test/test_spam_ingress.py
```diff
@@ -0,0 +1,158 @@
...
+        "ingress_threshold": 0.9,
+        "ingress_thresholds": {"public": 0.9, "private": 0.98},
+        "feature_config": {
+            "token_pattern": spam_ingress.DEFAULT_TOKEN_PATTERN,
+            "include_repository_name": False,
+        },
+        "training_metrics": {
+            "example_count": 2,
+            "spam_examples": 1,
+            "ham_examples": 1,
```
deshpandevlab, line 30: non-blocking: The test artifact includes `"ingress_thresholds": {"public": 0.9, "private": 0.98}` but no test exercises the visibility-specific threshold path (`_threshold_for_visibility`). A test that classifies the same borderline description as `visibility="public"` vs `visibility="private"` and asserts different `allowed` outcomes would cover this. Without it, a bug that always falls through to the default threshold ships undetected.

Same gap exists for `include_repository_name: true` — implemented but untested.

endpoints/api/test/test_repository.py
```diff
@@ -173,6 +174,180 @@ def test_create_repository(repo_name, extended_repo_names, expected_status, app)
...
+    repo = model.repository.get_repository("devtable", "simple")
+    assert repo.description == original_description
+    context = mock_evaluate.call_args[0][0]
+    assert context.visibility == expected_visibility
+    assert context.action == "update"
+
+
 @pytest.mark.parametrize(
     "has_tag_manifest",
     [
```
deshpandevlab, line 531: non-blocking: The create path has 5 focused tests (disabled, dry-run, enforcement, fail-open, fail-closed) but the update path only has this one enforcement test. Since the update path has different upstream code (extra DB query for visibility, different request parsing), consider adding at least a disabled + dry-run test for update parity. Can be a follow-up.

deshpandevlab (dismissed):
LGTM 

Three non-blocking follow-up items noted inline.

endpoints/api/repository.py
```diff
@@ -381,6 +483,18 @@ def put(self, namespace, repository):
             raise NotFound()
 
         values = request.get_json()
+        repo = repository_model.get_repository(namespace, repository)
```
shaonrh, line 486: Could we use `registry_model.lookup_repository(namespace, repository)` as the single lookup?

The returned `RepositoryReference` provides `is_public` for deriving visibility and can also be passed to `_check_spam_ingress()` for the active-tag check. This would avoid the additional `model.repo_exists()` and `repository_model.get_repository()` queries.

shaonrh (approved):
LGTM

PR #6159, opened 2026-06-12
QUAYIO-1805: feat(secscan): create PENDING ManifestSecurityStatus at push time

kleesc:
Create a ManifestSecurityStatus row with PENDING status when a manifest is pushed, so the V2 indexer picks it up via the normal batch-claim path instead of the expensive LEFT JOIN sweep. Throttle the unindexed sweep to every N cycles (default 60) and reindex newest-first.


data/model/oci/manifest.py
```diff
@@ -286,7 +289,7 @@ def get_or_create_manifest(
     if existing is not None:
         return CreatedManifest(manifest=existing, newly_created=False, labels_to_apply=None)
 
-    return _create_manifest(
+    result = _create_manifest(
```
lechuk47: Why assign a variable instead of returning the method call directly?

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -89,63 +88,59 @@ def perform_indexing(self, start_token=None, batch_size=None):
...
+                )
+            )
+            conditions |= (
+                (
+                    ManifestSecurityStatus.index_status.not_in(
+                        [
+                            IndexStatus.MANIFEST_UNSUPPORTED,
+                            IndexStatus.MANIFEST_LAYER_TOO_LARGE,
+                            IndexStatus.IN_PROGRESS,
+                        ]
```
lechuk47: Is this part of the condition really needed? Are not the other part catching the required rows?

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -89,63 +88,59 @@ def perform_indexing(self, start_token=None, batch_size=None):
...
                     )
                 )
-                .order_by(ManifestSecurityStatus.last_indexed.asc())
+                & (ManifestSecurityStatus.indexer_hash != indexer_hash)
```
lechuk47: Isn't this excluding FAILED rows ?

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -89,63 +88,59 @@ def perform_indexing(self, start_token=None, batch_size=None):
...
-            "Indexing cycle complete: %d manifests in %.1fs", total_claimed, cycle_duration
-        )
+        logger.debug("Indexing cycle complete: %d manifests in %.1fs", len(claimed), cycle_duration)
 
     def _find_and_claim_batch(self, batch_size, reindex_threshold, stale_threshold, indexer_hash):
+        max_retries = self.app.config.get(
+            "SECURITY_SCANNER_V2_MAX_SCAN_RETRIES", DEFAULT_MAX_SCAN_RETRIES
+        )
+
         with db_transaction():
```
lechuk47: Does the SELECT query need to be put in the transaction block?

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -264,10 +225,24 @@ def _index_manifest_by_id(self, manifest_id, repository_id):
...
+                ManifestSecurityStatus.create(
+                    manifest=candidate,
+                    repository=candidate.repository,
+                    error_json=report["err"],
+                    index_status=IndexStatus.COMPLETED,
+                    indexer_hash=state,
+                    indexer_version=IndexerVersion.V4,
+                    metadata_json={},
+                )
+            secscan_v2_scan_result.labels(result="completed").inc()
```
lechuk47: Why we DELETE/CREATE instead of UPDATE?

data/database.py
```diff
@@ -221,6 +222,66 @@ def sorted_model_key(query_fk_tuple):
...
+            "COALESCE(json_extract(%s, '$.%s'), 0) + 1)" % (col, increment_key, col, increment_key),
+            params=[json.dumps(attrs)],
+        )
+
+
+SCHEME_SPECIALIZED_JSON_OPS = {
+    "sqlite": _SqliteJsonOps(),
+}
+
+_pg_json_ops = _PgJsonOps()
```
lechuk47: Is JSONB really needed here? This feels a bit like YAGNI.

data/secscan_model/secscan_v4_model_v2.py
kleesc: Yes, it allows for re-indexing COMPLETED manifests when the security scanner is upgraded (indexer hash changes). The other three conditions cover PENDING, stale IN_PROGRESS, and FAILED, but they don't pick up COMPLETED ROWS after an indexer update.

data/secscan_model/secscan_v4_model_v2.py
kleesc: The not_in list is `[MANIFEST_UNSUPPORTED, MANIFEST_LAYER_TOO_LARGE, IN_PROGRESS]`. FAILED is not in this exclusion list

data/secscan_model/secscan_v4_model_v2.py
kleesc: Yes, the claim pattern is: SELECT rows, then UPDATE them to IN_PROGRESS. Both must be in the same transaction to prevent partial failures (e.g., SELECT succeeds but UPDATE fails, leading to an inconsistent state)

data/secscan_model/secscan_v4_model_v2.py
kleesc: I think that's just how it was before. I'll update it to a single update.

data/database.py
kleesc: I could've added a simple column for the counter alone, but the PR also stores last_error. Both approach requires a migration. The JSONB approach avoids a schema migration for each future metadata field.

cubismod: YAGNI? @lechuk47

data/migrations/versions/f77bb557a345_change_metadata_json_to_jsonb_on_.py
```diff
@@ -0,0 +1,38 @@
...
+def upgrade(op, tables, tester):
+    bind = op.get_bind()
+    if bind.engine.name == "postgresql":
+        op.alter_column(
+            "manifestsecuritystatus",
+            "metadata_json",
+            type_=postgresql.JSONB(),
+            existing_type=sa.Text(),
+            postgresql_using="metadata_json::jsonb",
+        )
```
cubismod: This will exclusively lock the table during the migration

kleesc: Right... I'll move this to a separate int column then. I don't think it's worth the complexity, specifically for this use case.

cubismod (approved):

PR #6165, opened 2026-06-16
PROJQUAY-7306: feat(quota): add quota threshold notifications with external channels

sridipta:
## Summary
- Add quota threshold notification system that alerts namespace owners when storage consumption crosses configurable warning/critical thresholds (e.g. 80%, 90%, 100%)
- Support external notification channels (Slack, email, webhook, etc.) via the existing Quay notification infrastructure
- Include deduplication and throttling state machine to prevent notification storms
- Add periodic background worker to catch quota threshold crossings between pushes
- Add namespace notifications UI tab in both organization and user namespace settings

## Root Cause / Rationale
Quay quota management currently enforces hard limits (reject/warn on push) but provides no proactive alerting when namespaces approach their quota limits. Administrators and org owners need advance warning before hitting quota limits so they can take corrective action (clean up images, request quota increases) before pushes start failing.

## Changes
- **Database**: New models for `NamespaceNotification`, `QuotaNotificationDedup` with Alembic migrations
- **API**: CRUD endpoints for namespace notifications (`/api/v1/organization/{org}/notifications`, `/api/v1/user/notifications`)
- **Model layer**: Notification CRUD, dedup/throttle state machine, email routing via `contact_email`
- **Workers**: Push-triggered notification dispatch, retroactive notifications on limit changes, periodic `quotanotificationworker` for threshold scanning
- **Notification dispatch**: Integration with existing notification worker infrastructure for external channels (Slack, email, webhook, Flowdock, HipChat)
- **Frontend (React/PatternFly)**: Namespace notifications tab in org and user settings with create/delete/list UI
- **Contact email** ([PROJQUAY-6975](https://redhat.atlassian.net/browse/PROJQUAY-6975)): Foundation for email routing — org-level `contact_email` field with API, UI, and notification integration
- **Tests**: Comprehensive pytest coverage for all model/API/worker layers + Playwright E2E tests

## Test Plan
- [x] Integration/registry tests pass
- [x] Type checking passes
- [x] Frontend tests pass
- [x] Migration tested (upgrade and downgrade)

## JIRA
https://redhat.atlassian.net/browse/PROJQUAY-7306

## Backport
- [ ] Backport required
- [x] No backport needed


data/migrations/versions/6715e4719375_add_namespace_notification_tables.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+
+    op.create_index(
+        "quotanotificationstate_namespace_threshold",
+        "quotanotificationstate",
+        ["namespace_id", "threshold_percent"],
+        unique=True,
+    )
+
+    op.bulk_insert(
```
jbpratt, line 90: This migration seeds `externalnotificationevent` with `quota_warning` / `quota_error`, but it never seeds matching `notificationkind` rows. Fresh DBs get those rows from `initdb.py`, so fresh and upgraded installs diverge. On an upgraded install, namespace `quay_notification` delivery fails at `NotificationKind.get(name="quota_warning")`. Add a migration for the two `NotificationKind` rows, with a safe downgrade plan.

endpoints/api/namespacenotification.py
```diff
@@ -0,0 +1,396 @@
...
+        "method": notification.method.name,
+        "config": config,
+        "event_config": event_config,
+        "number_of_failures": notification.number_of_failures,
+    }
+
+
+def _check_org_admin(orgname):
+    permission = AdministerOrganizationPermission(orgname)
+    if permission.can() or SuperUserPermission().can():
```
jbpratt: This grants any superuser org-notification admin access, bypassing the `FEATURE_SUPERUSERS_FULL_ACCESS` boundary used by the quota write APIs. With full access disabled, a regular superuser can create/delete/test callbacks and read returned method `config` for any org. Split read/write permission handling around the same full-access checks used by quota APIs, and decide whether read responses should redact sensitive config.

data/model/namespacequota.py
```diff
@@ -243,6 +267,103 @@ def notify_organization_admins(repository_ref, notification_kind, metadata={}):
...
+
+    threshold_percent = quota_result.get("threshold_percent")
+    if threshold_percent is None:
+        return
+
+    namespace_user = model.user.get_user_or_org(namespace_name)
+    if namespace_user is None:
+        return
+
+    if not should_notify(namespace_user, threshold_percent):
```
jbpratt: The dedup flow is check-then-enqueue-then-record. Concurrent pushes/workers can both pass `should_notify()` before either records state, enqueue duplicate jobs, and one can hit the unique `(namespace, threshold_percent)` constraint after the job is already queued. Replace this with an atomic claim/upsert and enqueue only when the caller wins the claim.

data/model/namespacequota.py
```diff
@@ -243,6 +267,103 @@ def notify_organization_admins(repository_ref, notification_kind, metadata={}):
...
+        namespace_name,
+        event_name,
+        extra_data={
+            "threshold_percent": threshold_percent,
+            "usage_bytes": usage_bytes,
+            "limit_bytes": quota_limit_bytes,
+            "usage_percent": usage_percent,
+        },
+    )
+    record_notification(namespace_user, threshold_percent)
```
jbpratt: `record_notification` runs even when `spawn_namespace_notification()` found no matching namespace notifications. A namespace can cross a threshold before any channel exists, get marked notified, then a later-created channel misses the first alert until cooldown expires. Record state only when at least one enabled notification is queued, or trigger a retroactive quota check when a namespace notification is created.

endpoints/api/namespacenotification.py
```diff
@@ -0,0 +1,396 @@
...
+    def post(self, orgname, uuid):
+        _check_org_admin(orgname)
+        found = _get_or_404(uuid)
+        if found.namespace.username != orgname:
+            raise NotFound()
+
+        event_config = json.loads(found.event_config_json or "{}")
+        event_info = NotificationEvent.get_event(found.event.name)
+        sample_data = event_info.get_sample_data(orgname, None, event_config)
+        notification_data = build_notification_data(found, sample_data)
```
jbpratt: The test endpoint queues a repo-shaped payload. `build_notification_data()` does not include `notification_type: "namespace"`, so `NotificationWorker` defaults to repo handling and looks up this UUID as a `RepositoryNotification`, silently dropping the test. Use `build_namespace_notification_data()` or otherwise include `notification_type: "namespace"` here and in the user test endpoint.

endpoints/api/namespacenotification.py
```diff
@@ -0,0 +1,396 @@
...
+def _validate_create_request(parsed):
+    event_name = parsed["event"]
+    if event_name not in VALID_NAMESPACE_EVENTS:
+        raise request_error(
+            message="Invalid event '%s'. Must be one of: %s"
+            % (event_name, ", ".join(sorted(VALID_NAMESPACE_EVENTS)))
+        )
+
+    try:
+        NotificationMethod.get_method(parsed["method"])
```
jbpratt: This checks that the method exists but never validates method-specific config. Direct API clients can persist invalid webhook/slack/quay configs and unsupported namespace methods, leading to async worker failures or silent no-ops. Mirror repository notification creation by calling `method_handler.validate(namespace, None, config)` and returning 400 on validation failure.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotificationsCreateForm.tsx
```diff
@@ -0,0 +1,486 @@
...
+
+  const isValidUrl = (url: string): boolean => {
+    return url.startsWith('http://') || url.startsWith('https://');
+  };
+
+  const isFormComplete = (): boolean => {
+    if (!event || !method) return false;
+    switch (method.type) {
+      case NamespaceNotificationMethodType.email:
+        return email !== '' && isValidEmail(email);
```
jbpratt: The form requires an email address and submits it in `config`, but namespace email delivery ignores `method_config_dict.email` and resolves recipients from the namespace contact/admin/user email instead. This makes the configured value misleading. Either remove/replace the field with the actual recipient policy, or make namespace email delivery honor and validate the configured address.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
```diff
@@ -0,0 +1,279 @@
...
+              <Td colSpan={6}>
+                <ExpandableRowContent>
+                  <NotificationConfig notification={notification} />
+                </ExpandableRowContent>
+              </Td>
+            </Tr>
+          </Tbody>
+        ))}
+      </Table>
+      <NamespaceNotificationsCreateForm
```
jbpratt, line 282: This non-empty table path drops `isUser={isUser}` when rendering the create form. The empty-state path passes it, so the first user-namespace notification can be created, but creating another one posts to `/api/v1/organization/<username>/notifications` instead of `/api/v1/user/notifications`. Pass `isUser` here and add an E2E case for creating another user namespace notification after one already exists.

jbpratt (dismissed):
Requesting changes for the inline issues.

web/playwright/utils/api/client.ts
```diff
@@ -678,6 +697,219 @@ export class ApiClient {
...
+
+    if (!response.ok()) {
+      const body = await response.text();
+      throw new Error(
+        `Failed to get user namespace notifications: ${response.status()} - ${body}`,
+      );
+    }
+
+    return response.json();
+  }
```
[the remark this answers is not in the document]
sridipta: Fixed.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
```diff
@@ -0,0 +1,279 @@
...
+        orgname={organizationName}
+        isOpen={isCreateOpen}
+        onClose={() => setIsCreateOpen(false)}
+      />
```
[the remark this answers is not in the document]
sridipta: Fixed.

data/model/notification.py
```diff
@@ -318,3 +319,102 @@ def delete_tag_notifications_for_tag(tag):
...
+        if notification.namespace.username != namespace_name:
+            raise InvalidNotificationException(
+                "No namespace notification found with uuid: %s" % uuid
+            )
+        NamespaceNotification.update(number_of_failures=0).where(
+            NamespaceNotification.id == notification.id
+        ).execute()
+        return notification
+    except NamespaceNotification.DoesNotExist:
+        return None
```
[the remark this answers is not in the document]
sridipta: Fixed.

data/model/quota_notification_state.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    if state is None:
+        return True
+
+    if state.last_notified_at is None:
+        return True
+
+    cooldown = timedelta(seconds=_get_cooldown_seconds())
+    return datetime.utcnow() >= state.last_notified_at + cooldown
```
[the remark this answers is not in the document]
sridipta: Fixed.

endpoints/api/namespacenotification.py
```diff
@@ -0,0 +1,396 @@
...
+    except InvalidNotificationMethodException:
+        raise request_error(message="Unknown notification method '%s'" % parsed["method"])
```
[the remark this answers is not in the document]
sridipta: Fixed.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotificationsCreateForm.tsx
```diff
@@ -0,0 +1,486 @@
...
+      props.onClose();
+    }
+  }, [successCreatingNotification]);
+
+  useEffect(() => {
+    if (errorCreatingNotification) {
+      setError(errorCreatingNotification);
+      resetCreatingNotification();
+    }
+  }, [errorCreatingNotification]);
```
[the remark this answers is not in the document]
sridipta: Fixed.

workers/quotanotificationworker.py
```diff
@@ -0,0 +1,130 @@
...
+        if not limits:
+            return
+
+        quota_limit_bytes = quota.limit_bytes
+
+        for limit in limits:
+            threshold_percent = limit.percent_of_limit
+            bytes_allowed = int(quota_limit_bytes * threshold_percent / 100)
+
+            if usage_bytes > bytes_allowed:
```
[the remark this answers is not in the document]
sridipta: Fixed.

web/src/hooks/UseUpdateNamespaceNotifications.ts
```diff
@@ -0,0 +1,110 @@
...
+      },
+    },
+  );
+
+  let errorCreationMessage: string = null;
+  if (errorCreatingNotification != null) {
+    errorCreationMessage = (errorCreatingNotification as any)?.response?.data
+      ?.detail
+      ? (errorCreatingNotification as any)?.response?.data?.detail
+      : 'Unable to create notification';
```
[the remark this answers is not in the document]
sridipta: Fixed.

workers/test/test_quotanotificationworker.py
```diff
@@ -0,0 +1,189 @@
...
+                update_namespace_quota_size(quota, limit_bytes)
+        else:
+            quota = create_namespace_quota(namespace_user, limit_bytes)
+        limit = create_namespace_quota_limit(quota, quota_type, threshold)
+        return quota, limit
+
+    def _create_namespace_notification(self, namespace_user, event_name="quota_warning"):
+        event = ExternalNotificationEvent.get(ExternalNotificationEvent.name == event_name)
+        method = ExternalNotificationMethod.select().first()
+        return NamespaceNotification.create(
```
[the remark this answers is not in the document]
sridipta: Fixed.

conf/init/supervisord_conf_create.py
```diff
@@ -58,6 +58,7 @@ def registry_services():
         "securityscanningnotificationworker": {"autostart": "true"},
         "quotatotalworker": {"autostart": "true"},
         "quotaregistrysizeworker": {"autostart": "true"},
+        "quotanotificationworker": {"autostart": "true"},
```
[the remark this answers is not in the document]
sridipta: The existing test pattern for supervisord workers covers this — the conditional is exercised by toggling the feature flag in the config fixture. No additional test needed.

data/model/quota_notification_state.py
```diff
@@ -0,0 +1,79 @@
...
+            QuotaNotificationState.create(
+                namespace=namespace_user,
+                threshold_percent=threshold_percent,
+                last_notified_at=now,
+                cleared=False,
+            )
+        else:
+            state.last_notified_at = now
+            state.cleared = False
+            state.save()
```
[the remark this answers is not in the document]
sridipta: The worker acquires a GlobalLock before processing, so only one instance runs at a time. The unique index on (namespace, threshold_percent) provides a DB-level backstop. The read-then-write pattern is safe here.

web/playwright/e2e/organization/namespace-notifications-dedup.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+          TEST_USERS.user.username,
+          TEST_USERS.user.password,
+        );
+
+        // Wait 5 seconds — no additional webhook should arrive
+        await new Promise((r) => setTimeout(r, 5000));
+
+        // Should have received exactly 1 webhook (dedup suppressed the second)
+        expect(receiver.getRequests()).toHaveLength(1);
+      } finally {
```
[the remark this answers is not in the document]
sridipta: The 5-second window is pragmatic for an in-process queue flush in a local-dev E2E test. The worker processes synchronously within the test harness, so the sleep is a generous upper bound, not a race-condition window.

web/playwright/e2e/organization/namespace-notifications-delivery.spec.ts
```diff
@@ -0,0 +1,394 @@
...
+            await pushUniqueImage(
+              org.name,
+              'fillrepo',
+              'v2',
+              TEST_USERS.user.username,
+              TEST_USERS.user.password,
+            );
+          } catch {
+            // Expected — push rejected due to quota
+          }
```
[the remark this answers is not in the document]
sridipta: The try/catch here is setup, not the assertion — it pushes an oversized image to trigger the quota rejection. The test asserts on the notification delivery side (webhook/email received), not on the push outcome. The push is expected to fail.

web/playwright/e2e/organization/namespace-notifications-user.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+        // Verify email delivered to user's address
+        const email = await mailpit.waitForEmail(
+          (msg) =>
+            msg.To.some(
+              (to) => to.Address === 'testuser@example.com',
+            ),
+          15_000,
+        );
+        expect(email).not.toBeNull();
```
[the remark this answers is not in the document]
sridipta: The configured email matching the namespace owner's email IS the intended behavior. For user namespaces, the email notification resolves the recipient from User.email — the test verifies that this resolution works correctly. A different email would indicate a bug.

web/src/routes/OrganizationsList/CreateOrganizationModal.tsx
```diff
@@ -80,15 +77,12 @@ export const CreateOrganizationModal = (
   };
 
   const handleEmailInputChange = (value: string) => {
-    setOrganizationEmail(value);
+    setContactEmail(value);
   };
```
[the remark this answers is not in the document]
sridipta: This follows the pre-existing pattern in CreateOrganizationModal — the email field is optional and the backend validates format before persisting. Adding client-side validation here would be a separate concern outside this PR's scope.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotificationsKebab.tsx
```diff
@@ -0,0 +1,192 @@
...
+          <DropdownItem
+            data-testid={`${notification.uuid}-delete-notification`}
+            onClick={() => setIsDeleteModalOpen(true)}
+          >
+            Delete Notification
+          </DropdownItem>
+          <Conditional if={isNamespaceNotificationDisabled(notification)}>
+            <DropdownItem
+              data-testid={`${notification.uuid}-enable-notification`}
+              onClick={() => enableNotifications(notification.uuid)}
```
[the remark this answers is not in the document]
sridipta: The uuid is optional in the shared NamespaceNotification interface because it's absent during creation (before the server assigns one). In the kebab menu context, notifications always come from the list endpoint which guarantees uuid is present. Adding a runtime guard would be dead code.

workers/quotanotificationworker.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+        namespace_id_set = {row[0] for row in namespace_ids}
+
+        if not namespace_id_set:
+            return
+
+        for namespace_user in User.select().where(User.id << list(namespace_id_set)):
+            try:
+                self._check_namespace(namespace_user)
+            except Exception:
```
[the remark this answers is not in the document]
sridipta: The N+1 pattern is acceptable here — this is a background worker running on a 5-minute poll interval, and the namespace count is bounded by the number of namespaces with notification configs (typically single digits). Batching would add complexity for negligible gain.

data/model/notification.py
```diff
@@ -318,3 +319,102 @@ def delete_tag_notifications_for_tag(tag):
...
+    )
+
+    if event_name:
+        query = (
+            query.switch(NamespaceNotification)
+            .join(ExternalNotificationEvent)
+            .where(ExternalNotificationEvent.name == event_name)
+        )
+
+    return query
```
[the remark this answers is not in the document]
sridipta: The EnumField FK lookups hit tiny, rarely-changing tables (ExternalNotificationEvent, ExternalNotificationMethod) that Peewee caches after first access. This is a low-frequency endpoint — the overhead is negligible.

data/migrations/versions/6715e4719375_add_namespace_notification_tables.py
sridipta: e9f3e4dbb979 already seeds NotificationKind with both quota_warning and quota_error, and it runs before our migration (it's earlier in the chain: e9f3e4dbb979 → ... → b1a79fa8e630 → 6715e4719375).

So the NotificationKind rows already exist by the time our migration runs on an upgraded install There's no divergence between fresh and upgraded installs.

endpoints/api/namespacenotification.py
sridipta: Done. replaced the bare SuperUserPermission().can() with allow_if_superuser_with_full_access(), which requires both that the user is a superuser and that FEATURE_SUPERUSERS_FULL_ACCESS is enabled — matching the same boundary used by the quota write APIs.

data/model/namespacequota.py
sridipta: Fixed. Replaced the check-then-enqueue-then-record flow with an atomic `claim_notification()` that claims the dedup slot before enqueuing:

- **No row exists:** creates one inside a transaction; catches `IntegrityError` if a concurrent caller raced us.
- **Row exists, eligible (cooldown expired or cleared):** uses a conditional `UPDATE ... WHERE id = X AND last_notified_at = Y AND cleared = Z` — only succeeds if the row hasn't changed since we read it (optimistic lock).
- **Row exists, not eligible:** returns `False` immediately.

The push-path caller now does `claim_notification()` → enqueue (state recorded atomically *before* enqueuing), instead of `should_notify()` → enqueue → `record_notification()`. The old `should_notify()` and `record_notification()` are preserved for the background worker and test helpers.

data/model/namespacequota.py
sridipta: Fixed. `spawn_namespace_notification()` now returns the count of enqueued notifications. If zero were enqueued (no channels configured), the caller calls `release_claim()` which deletes the state row so the threshold starts fresh when a channel is later created.

Flow is now: `claim_notification()` → `spawn_namespace_notification()` → if nothing enqueued, `release_claim()` and return early.

endpoints/api/namespacenotification.py
sridipta: Fixed. Both test endpoints (org at line 245, user at line 401) now use `build_namespace_notification_data()` which includes `notification_type: "namespace"`. Also cleaned up the import — removed the unused `build_namespace_event_data` and `build_notification_data` imports.

endpoints/api/namespacenotification.py
sridipta: Fixed in commit 4578ea8c3. Added `method_handler.validate(namespace_name, None, config)` call in `_validate_create_request()` before persisting, matching the existing repo notification pattern. Returns 400 on validation failure.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
sridipta: Fixed in commit 4578ea8c3. Added `isUser={isUser}` to the non-empty table path at line 267.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotificationsCreateForm.tsx
sridipta: Fixed. Removed the email input field entirely and replaced it with an info banner that tells the user where notifications will actually be sent:

- **Org namespaces:** "Notifications will be sent to the organization contact email or organization admin emails."
- **User namespaces:** "Notifications will be sent to your account email address."

The `getConfig()` for email method now returns `{}` (no misleading `config.email`). Also updated the expanded row display in the notifications table to show "Sent to namespace contact or admin emails" instead of the old `config.email` value.

data/migrations/versions/6715e4719375_add_namespace_notification_tables.py
sridipta: No code change needed. The `notificationkind` rows for `quota_warning` and `quota_error` are already seeded by the earlier migration `e9f3e4dbb979_quota_management_reporting.py` (line 84), which runs before ours in the chain. That migration was introduced in PR #1048 (PROJQUAY-2936) and exists on master. So upgraded installs already have both `notificationkind` rows by the time our migration runs — no divergence between fresh and upgraded installs.

bcaton85 (commented):
Review of quota notification feature — 7 findings from line-by-line walkthrough.

endpoints/v2/blob.py
```diff
@@ -269,6 +269,8 @@ def start_blob_upload(namespace_name, repo_name):
         "FEATURE_VERIFY_QUOTA", True
     ):
         quota = namespacequota.verify_namespace_quota(repository_ref)
+        if quota["severity_level"] in ("Warning", "Reject"):
+            namespacequota.maybe_trigger_quota_notification(namespace_name, quota)
```
bcaton85, line 273: `maybe_trigger_quota_notification` can raise if Redis is down (via `notification_queue.put`) or if the database is unreachable (via `claim_notification`). This would abort the blob upload with a 500 even for Warning-level quota — a notification side effect should never block image pushes.

Suggest wrapping in try/except, either here at all three call sites or inside `maybe_trigger_quota_notification` itself (preferred — protects all callers):
```python
try:
    namespacequota.maybe_trigger_quota_notification(namespace_name, quota)
except Exception:
    logger.exception("Failed to trigger quota notification for %s", namespace_name)
```

endpoints/api/namespacequota.py
```diff
@@ -453,15 +463,24 @@ def put(self, orgname, quota_id, limit_id):
...
-            "type": quota_limit_data.get("type", previous_type),
-            "threshold_percent": quota_limit_data.get("threshold_percent", previous_threshold),
+            "type": current_type,
+            "threshold_percent": current_threshold,
             "previous_type": previous_type,
             "previous_threshold_percent": previous_threshold,
         }
         log_action("org_change_quota_limit", orgname, metadata)
 
+        model.namespacequota.maybe_trigger_retroactive_notification(
```
bcaton85, line 477: This `maybe_trigger_retroactive_notification` call is outside any try/except. If it raises (e.g., Redis/DB error), the admin gets a 500 even though the quota limit update already committed successfully. The same applies to the calls at lines 273 and 391, which are inside `try/except DataModelException` but would not catch Redis or notification queue errors.

Wrap all three retroactive notification calls in try/except so quota CRUD operations always succeed regardless of notification side effects:
```python
try:
    model.namespacequota.maybe_trigger_retroactive_notification(...)
except Exception:
    logger.exception("Failed to trigger retroactive notification")
```

data/model/namespacequota.py
```diff
@@ -243,6 +267,108 @@ def notify_organization_admins(repository_ref, notification_kind, metadata={}):
...
+    import features
+
+    if not features.QUOTA_NOTIFICATIONS:
+        return
+
+    usage_bytes = get_namespace_size(namespace_name)
+    quota_limit_bytes = quota.limit_bytes
+    bytes_allowed = int(quota_limit_bytes * threshold_percent / 100)
+
+    if usage_bytes <= bytes_allowed:
```
bcaton85: `if usage_bytes <= bytes_allowed: return` means usage exactly at the threshold does NOT trigger a retroactive notification. But the push path (`check_limits`, line 221) uses `if size >= bytes_allowed` — exactly-at DOES trigger. The worker (line 79) also uses `>=`. The retroactive path is the only outlier.

Should be:
```python
if usage_bytes < bytes_allowed:
    return
```
to match the `>=` semantics used by the push path and the worker.

data/model/namespacequota.py
```diff
@@ -243,6 +267,108 @@ def notify_organization_admins(repository_ref, notification_kind, metadata={}):
...
+
+    if not claim_notification(namespace_user, threshold_percent):
+        return
+
+    event_name = "quota_warning" if severity == "Warning" else "quota_error"
+    usage_bytes = quota_result.get("usage_bytes", 0)
+    quota_limit_bytes = quota_result.get("quota_limit_bytes", 0)
+    usage_percent = int(usage_bytes * 100 / quota_limit_bytes) if quota_limit_bytes else 0
+
+    enqueued = spawn_namespace_notification(
```
bcaton85: If `spawn_namespace_notification` raises (e.g., Redis connection error), `release_claim` is never called. The claim row holds `cleared=False` and `last_notified_at=now`, so no notification will be retried for this threshold until the 24-hour cooldown expires — silently suppressing all notifications.

Wrap in try/except:
```python
try:
    enqueued = spawn_namespace_notification(
        namespace_name,
        event_name,
        extra_data={...},
    )
except Exception:
    release_claim(namespace_user, threshold_percent)
    logger.exception("Failed to spawn quota notification for %s", namespace_name)
    return

if not enqueued:
    release_claim(namespace_user, threshold_percent)
    return
```

endpoints/api/namespacenotification.py
```diff
@@ -0,0 +1,411 @@
...
+    try:
+        return model.notification.get_namespace_notification(uuid)
+    except InvalidNotificationException:
+        raise NotFound()
+
+
+@resource("/v1/organization/<orgname>/notifications")
+@show_if(features.QUOTA_NOTIFICATIONS)
+@path_param("orgname", "The name of the organization")
+class OrgNamespaceNotificationList(ApiResource):
```
bcaton85, line 96: The new endpoint classes (`OrgNamespaceNotificationList`, `OrgNamespaceNotification`, `TestOrgNamespaceNotification`, `UserNamespaceNotificationList`, `UserNamespaceNotification`, `TestUserNamespaceNotification`) are not covered in `endpoints/api/test/test_security.py`. That test verifies every API endpoint correctly enforces authorization for unauthenticated, non-member, and non-admin users.

Please add entries for all new endpoints to `test_security.py`.

events/quota_warning.html
```diff
@@ -0,0 +1,6 @@
+Namespace <b>{{ event_data.namespace }}</b> storage usage has reached
+<b>{{ event_data.threshold_percent }}%</b> of its quota limit.
+
+<br><br>
+Current usage: {{ event_data.usage_percent }}%
+({{ event_data.usage_bytes }} bytes of {{ event_data.limit_bytes }} bytes)
```
bcaton85: The template displays raw byte counts: `858993459 bytes of 1073741824 bytes`. These are not meaningful to most users at a glance.

Consider formatting as human-readable units (e.g., `819.0 MiB of 1.0 GiB`). This could be done with a Jinja2 filter or by pre-formatting the values in the event payload before they reach the template. The same applies to `events/quota_error.html`.

bcaton85:
**Namespace GC should explicitly clean up new tables**

When a namespace is garbage collected (`data/model/user.py`, `delete_namespace_via_marker`), `NamespaceNotification` and `QuotaNotificationState` rows are only cleaned up indirectly — via `delete_namespace_quota` if a quota exists, or via the slow FK cascade in `User.delete_instance(recursive=True)`. If a namespace has notification configs but no quota (or the quota was already deleted), the rows are orphaned until the final recursive delete.

Add explicit deletes alongside the existing `QuotaNamespaceSize` cleanup at line 1544:
```python
NamespaceNotification.delete().where(NamespaceNotification.namespace == user).execute()
QuotaNotificationState.delete().where(QuotaNotificationState.namespace == user).execute()
```
Also add both models to `skip_transitive_deletes` in `User.delete_instance` (line 910 of `database.py`) for consistency with `UserOrganizationQuota`, `QuotaLimits`, and other models with direct FK to User.

data/model/quota_notification_state.py
```diff
@@ -0,0 +1,147 @@
...
+    now = datetime.utcnow()
+
+    with db_transaction():
+        state = QuotaNotificationState.get_or_none(
+            QuotaNotificationState.namespace == namespace_user,
+            QuotaNotificationState.threshold_percent == threshold_percent,
+        )
+
+        if state is None:
+            QuotaNotificationState.create(
```
bcaton85: `record_notification` does `get_or_none` then `create` within a transaction but doesn't catch `IntegrityError`. The sibling function `claim_notification` (line 49) correctly handles this case. If two background worker instances call `record_notification` concurrently for the same (namespace, threshold), the second insert will fail on the unique index.

Add `IntegrityError` handling on the create path to match `claim_notification`:
```python
if state is None:
    try:
        QuotaNotificationState.create(...)
    except IntegrityError:
        pass  # another worker beat us — safe to ignore
```

data/model/namespacequota.py
```diff
@@ -92,6 +95,12 @@ def update_namespace_quota_size(quota, limit_bytes):
 
 def delete_namespace_quota(quota):
     with db_transaction():
+        QuotaNotificationState.delete().where(
+            QuotaNotificationState.namespace == quota.namespace,
+        ).execute()
+        NamespaceNotification.delete().where(
```
bcaton85: `NamespaceNotification.delete().where(NamespaceNotification.namespace == quota.namespace)` deletes all notification configs for the namespace, not just quota-related ones. Currently only `quota_warning` and `quota_error` events exist, so this is functionally correct. But if `NamespaceNotification` is reused for future non-quota events, deleting a quota would destroy unrelated notification configs.

Consider filtering by event type:
```python
NamespaceNotification.delete().where(
    NamespaceNotification.namespace == quota.namespace,
    NamespaceNotification.event << [quota_warning_event, quota_error_event],
).execute()
```

data/model/namespacequota.py
```diff
@@ -243,6 +267,108 @@ def notify_organization_admins(repository_ref, notification_kind, metadata={}):
...
+
+    threshold_percent = quota_result.get("threshold_percent")
+    if threshold_percent is None:
+        return
+
+    namespace_user = model.user.get_user_or_org(namespace_name)
+    if namespace_user is None:
+        return
+
+    if not claim_notification(namespace_user, threshold_percent):
```
bcaton85, line 334: When a push exceeds a threshold but no notification channels are configured: `claim_notification` creates a state row → `spawn_namespace_notification` returns 0 → `release_claim` deletes the row. This INSERT/DELETE cycle repeats on every push that exceeds the threshold.

Consider checking for configured channels before claiming:
```python
notifications = list(model.notification.list_namespace_notifications(
    namespace_name, event_name=event_name
))
if not notifications:
    return
```
This avoids unnecessary DB writes for namespaces that have the feature flag on but haven't configured any channels.

workers/quotanotificationworker.py
```diff
@@ -0,0 +1,121 @@
...
+        self.add_operation(self._check_quotas, POLL_PERIOD)
+
+    def _check_quotas(self):
+        if not features.QUOTA_NOTIFICATIONS:
+            return
+
+        namespace_ids = (
+            NamespaceNotification.select(NamespaceNotification.namespace).distinct().tuples()
+        )
+        namespace_id_set = {row[0] for row in namespace_ids}
```
bcaton85: `_check_quotas` materializes all namespace IDs with notification configs into a Python set, then passes the full list to `User.select().where(User.id << list(namespace_id_set))`. For large deployments with thousands of namespaces, this creates a large SQL `IN (...)` clause.

Consider either batching (process N namespace IDs at a time) or replacing the two queries with a single joined query:
```python
for namespace_user in (
    User.select()
    .join(NamespaceNotification, on=(NamespaceNotification.namespace == User.id))
    .distinct()
):
    self._check_namespace(namespace_user)
```

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
```diff
@@ -0,0 +1,282 @@
...
+    paginationProps,
+  } = usePaginatedSortableTable(notifications || [], {
+    columns: {
+      0: (item: NamespaceNotification) => item.title || '(Untitled)',
+      1: (item: NamespaceNotification) => item.event,
+      2: (item: NamespaceNotification) => item.method,
+      3: (item: NamespaceNotification) =>
+        isNamespaceNotificationEnabled(item) ? 'Enabled' : 'Disabled',
+    },
+    initialPerPage: 20,
```
bcaton85, line 150: The component uses `usePaginatedSortableTable` with `initialPerPage: 20` and consumes `paginatedData`, but never renders a PatternFly `<Pagination>` component. The toolbar (lines 193-200) shows an item count but has no page controls. If a namespace has more than 20 notification configs, the rest are invisible and unreachable.

Either add `<Pagination {...paginationProps} />` in the toolbar, or increase `initialPerPage` to a value high enough that pagination isn't needed.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotificationsCreateForm.tsx
```diff
@@ -0,0 +1,483 @@
...
+                data-testid="ns-notification-title"
+                value={title}
+                onChange={(_event, value) => setTitle(value)}
+              />
+            </FormGroup>
+
+            <ActionGroup>
+              <Button
+                data-testid="ns-notification-submit-btn"
+                isDisabled={!isFormComplete()}
```
bcaton85: The submit button is disabled only based on `isFormComplete()`. While the create mutation is in-flight, the button remains clickable, allowing duplicate submissions. The `useMutation` hook exposes `isLoading` — use it to disable the button during submission:
```tsx
isDisabled={!isFormComplete() || isCreating}
```

bcaton85 (commented):
Additional findings from detailed review.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
```diff
@@ -0,0 +1,282 @@
...
+    setExpandedUuids((prev) =>
+      prev.includes(uuid) ? prev.filter((u) => u !== uuid) : [...prev, uuid],
+    );
+
+  if (loading) {
+    return <Spinner />;
+  }
+
+  if (error) {
+    return <>Unable to load notifications</>;
```
bcaton85: When the fetch fails, the component renders `<>Unable to load notifications</>` — a plain text fragment with no styling, error icon, or retry action. Other components in the codebase use `RequestError` or PatternFly `Alert` / `EmptyState` for error states.

Consider using:
```tsx
<RequestError message="Unable to load notifications" />
```

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+function NotificationConfig({
+  notification,
+}: {
+  notification: NamespaceNotification;
+}) {
+  switch (notification.method) {
+    case NamespaceNotificationMethodType.email:
+      return (
+        <FlexItem style={{color: 'grey'}}>
```
bcaton85: Six uses of inline `style={{color: 'grey'}}` and `style={{marginBottom: 0}}` in the `NotificationConfig` component (lines 68, 74, 81, 85, 102). Hardcoded color values don't adapt to dark mode themes. Use PatternFly utility classes instead (e.g., `pf-v6-u-color-200`, `pf-v6-u-mb-0`).

web/playwright/e2e/organization/namespace-notifications.spec.ts
```diff
@@ -0,0 +1,510 @@
...
+
+      // Select method: Email
+      await authenticatedPage
+        .getByTestId('ns-notification-method-dropdown')
+        .click();
+      await authenticatedPage.getByTestId('ns-method-email').click();
+
+      // Fill in email
+      await authenticatedPage
+        .getByTestId('ns-notification-email')
```
bcaton85: This test calls `.getByTestId('ns-notification-email').fill('admin@example.com')`, but the form component doesn't render an input with that testid. When email is selected, it shows an info Alert with testid `ns-notification-email-info` — no input field exists because `isFormComplete()` returns `true` immediately for email.

This test will time out waiting for a nonexistent element. Remove the `.fill()` call. Same issue in `web/playwright/e2e/user/namespace-notifications.spec.ts` line 115.

bcaton85 (commented):
Remaining findings — test issues and FlowdockMethod crash.

workers/notificationworker/models_pre_oci.py
```diff
@@ -25,6 +25,22 @@ def notification(notification_row):
...
+    Uses Repository(namespace_name, None) so existing notification methods
+    that access .repository.namespace_name work without modification.
+    """
+    return Notification(
+        uuid=notification_row.uuid,
+        event_name=NamespaceNotification.event.get_name(notification_row.event_id),
+        method_name=NamespaceNotification.method.get_name(notification_row.method_id),
+        event_config_dict=json.loads(notification_row.event_config_json or "{}"),
+        method_config_dict=json.loads(notification_row.config_json or "{}"),
+        repository=Repository(notification_row.namespace.username, None),
```
bcaton85, line 40: **FlowdockMethod.perform() will crash on namespace notifications**

This line sets `repository=Repository(notification_row.namespace.username, None)`. `FlowdockMethod.perform()` (notificationmethod.py:339) concatenates `notification_obj.repository.namespace_name + " " + notification_obj.repository.name` — the `None` will raise `TypeError`.

Flowdock is deprecated, but the code path is reachable. Guard in FlowdockMethod:
```python
project = notification_obj.repository.namespace_name
if notification_obj.repository.name:
    project += " " + notification_obj.repository.name
```

web/playwright/e2e/organization/namespace-notifications-user.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+      expect(gone).toBeUndefined();
+    });
+
+    test('user cannot manage notifications for another user namespace', async ({
+      userClient,
+      adminClient,
+    }) => {
+      // Create notification as admin for their own namespace
+      const createResp = await adminClient.post(
+        '/api/v1/user/notifications',
```
bcaton85: Wrong API path: `/api/v1/user/notifications` should be `/api/v1/user/namespacenotifications`. The backend registers the endpoint at the latter path. This call and the ones at lines 63 and 69 will return 404.

Change all three occurrences to `'/api/v1/user/namespacenotifications'`.

web/playwright/e2e/organization/namespace-notifications.spec.ts
```diff
@@ -0,0 +1,510 @@
...
+      await authenticatedPage
+        .getByTestId('ns-notification-method-dropdown')
+        .click();
+      await authenticatedPage
+        .getByTestId('ns-method-quay_notification')
+        .click();
+
+      // Select team as recipient in entity search
+      const entitySearch = authenticatedPage.getByTestId(
+        'ns-notification-entity-search',
```
bcaton85: `getByTestId('ns-notification-entity-search')` references a testid that doesn't exist. `EntitySearch` uses `id="entity-search"` on its `MenuToggle` and `id="entity-search-input"` on the text input — neither produces a `data-testid` attribute. This selector will time out. Same issue at line 499.

Either add a `data-testid` prop pass-through to `EntitySearch`, or change the locator to `.locator('#entity-search-input')`.

data/model/quota_notification_state.py
```diff
@@ -0,0 +1,147 @@
...
+DEFAULT_COOLDOWN_SECONDS = 86400
+
+
+def _get_cooldown_seconds():
+    if config.app_config is None:
+        return DEFAULT_COOLDOWN_SECONDS
+    return config.app_config.get("QUOTA_NOTIFICATION_COOLDOWN_SECONDS", DEFAULT_COOLDOWN_SECONDS)
+
+
+def claim_notification(namespace_user, threshold_percent):
```
bcaton85, line 20: **Missing test coverage for `claim_notification` and `release_claim`**

These are the concurrency-critical functions used in the push hot path, but they have zero direct test coverage. All state machine tests in `test_quota_notification_state.py` exercise `should_notify` and `record_notification` instead.

At minimum, add tests for:
- Basic claim succeeds (returns `True`, creates state row)
- Second claim within cooldown is rejected (returns `False`)
- Claim after `cleared=True` succeeds
- `release_claim` deletes state so next claim succeeds
- Concurrent create — `IntegrityError` path returns `False`

data/model/namespacequota.py
bcaton85: **Missing test for spawn failure / release_claim cleanup path**

All tests in `test_quota_trigger.py` mock `spawn_namespace_notification` to return a truthy value. The case where it returns 0 (no channels configured) or raises an exception is never tested — this is the path where `release_claim` is called.

Add tests for:
- `spawn_namespace_notification` returns 0 → claim is released
- `spawn_namespace_notification` raises → claim is released (once the try/except fix is applied)

endpoints/api/namespacenotification.py
```diff
@@ -0,0 +1,411 @@
...
+                "event": new_notification.event.name,
+                "method": new_notification.method.name,
+            },
+        )
+        return _notification_view(new_notification), 201
+
+
+@resource("/v1/user/namespacenotifications/<uuid>")
+@show_if(features.QUOTA_NOTIFICATIONS)
+@path_param("uuid", "The UUID of the notification")
```
bcaton85, line 340: **Missing test for unauthorized DELETE/POST on user namespace notifications**

The test `test_get_other_users_notification` verifies that a user can't GET another user's notification, but there are no equivalent tests for DELETE or POST (reset failures) on another user's notification. The code correctly checks ownership in both paths, but test coverage should verify these are secure.

data/migrations/versions/6715e4719375_add_namespace_notification_tables.py
```diff
@@ -0,0 +1,107 @@
+"""add_namespace_notification_tables
+
+Revision ID: 6715e4719375
+Revises: b1a79fa8e630
```
Marcusk19: nit: this `Revises` is not the same as the `down_revision` below. not sure if these comments are used by alembic but it may be necessary to align them.

data/model/namespacequota.py
sridipta: Done.

data/model/quota_notification_state.py
sridipta: Done.

data/model/namespacequota.py
sridipta: Done.

endpoints/api/namespacequota.py
sridipta: Done.

workers/notificationworker/models_pre_oci.py
sridipta: Done.

data/migrations/versions/6715e4719375_add_namespace_notification_tables.py
sridipta: Done.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotificationsCreateForm.tsx
sridipta: Done.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
sridipta: Done.

web/playwright/e2e/organization/namespace-notifications.spec.ts
sridipta: Done.

web/playwright/e2e/organization/namespace-notifications-user.spec.ts
sridipta: Done.

web/playwright/e2e/organization/namespace-notifications.spec.ts
sridipta: Done.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
sridipta: Done.

data/model/namespacequota.py
sridipta: Done.

data/model/namespacequota.py
sridipta: Done.

workers/quotanotificationworker.py
sridipta: Done.

events/quota_warning.html
sridipta: Done.

web/src/routes/OrganizationsList/Organization/Tabs/Settings/NamespaceNotifications.tsx
sridipta: Done.

endpoints/api/namespacenotification.py
sridipta: Done.

data/model/quota_notification_state.py
sridipta: Done.

endpoints/api/namespacenotification.py
sridipta: Done.

data/model/namespacequota.py
sridipta: Done.

endpoints/v2/blob.py
sridipta: Done.

bcaton85 (dismissed):

jbpratt (dismissed):

bcaton85 (approved):

deshpandevlab (commented):
**Question: `check_limits()` — `>` changed to `>=`**

In `data/model/namespacequota.py`, the threshold comparison was changed from `if size > bytes_allowed` to `if size >= bytes_allowed`. Is this change intentional? This change is the existing behavior.

This doesn't affect QuayIO since `FEATURE_QUOTA_NOTIFICATIONS` is disabled, but wanted to confirm the inclusive boundary is the intended semantic for on-prem deployments.

Non-blocking. just want to make sure this is deliberate.

deshpandevlab (approved):

sridipta:
@deshpandevlab Good question. The change is intentional. CodeRabbit flagged that using > (strictly greater than) in check_limits() misses the exact boundary case — e.g., if usage lands exactly at 80%, 90%, or 100% of the quota, the notification wouldn't fire. It recommended changing to >= so notifications trigger at the threshold, not just above it.

PR #6170, opened 2026-06-16
NO-ISSUE: feat(oci): pluggable OCI storage driver with SQLite-backed distribution compat

jbpratt:
## Summary

Introduces a domain-layer storage architecture for the Go registry rewrite, matching Python Quay's proven pattern: blobs in content-addressed storage, all metadata in the database, no filesystem link files.

### Domain Interfaces (`internal/oci/`)

- **`oci.BlobStore`** — content-addressed blob ops + full upload lifecycle (UploadWriter, UploadReader, PutUploadState, etc.)
- **`oci.MetadataStore`** — read-write registry metadata (repos, manifests, tags, blobs, uploadedblob GC protection)
- Local filesystem implementation with Python-compatible path layout (`sha256/xx/xxxx...`)

### DistDriver — Distribution Compatibility Shim

Path-classifying adapter that implements distribution's `StorageDriver` by routing every call through the two domain interfaces:

| Path type | Backend | Example |
|---|---|---|
| Blob (`*/blobs/<alg>/<xx>/<hex>/data`) | `oci.BlobStore` | Content-addressed blob reads/writes |
| Upload (`*/_uploads/<id>/*`) | `oci.BlobStore` | Upload lifecycle (data, hashstates, startedat) |
| Metadata (`*/repositories/...`) | `oci.MetadataStore` | Tag links, layer links, manifest links → SQLite |

Zero filesystem calls in the adapter — swapping the BlobStore implementation (e.g., to S3) requires no DistDriver changes.

**Key design decisions:**
- Link file writes are no-ops (middleware already writes to SQLite)
- Link file reads query SQLite and return the digest
- `Move(upload→blob)` streams through `BlobStore.Writer`
- Layer link resolution is repo-scoped via `manifestblob` + `uploadedblob`, matching Python's `get_repository_blob_by_digest`

### MetadataStore Read Methods

Extends the metastore with read queries for the DistDriver: `GetRepositoryID`, `GetTagDigest`, `GetManifestDigest`, `BlobLinkedToRepo`, `BlobExists`, `ListTags`, `ListRepositories`, `PutUploadedBlob`, `CleanExpiredUploadedBlobs`.

### Python Quay Comparison

| Concern | Python | Go |
|---|---|---|
| Blob paths | `sha256/xx/xxxx...` via `content_path()` | `sha256/xx/xxxx...` via `oci.BlobStore` |
| Tag resolution | PG query | SQLite query via `MetadataStore.GetTagDigest` |
| Upload staging | `uploads/{uuid}` | `uploads/{uuid}/` (data + state files) |
| Blob GC protection | `UploadedBlob` table | `uploadedblob` table (same schema) |
| Link files | None — all PG | None — all SQLite |

## Test plan

- [x] 189 tests passing with `-race`
- [x] Path classification unit tests (9 path patterns)
- [x] DistDriver routing tests (blob, upload, metadata, move, list, stat, redirect)
- [x] MetadataStore read method tests (8 methods)
- [x] Integration test: full push/pull round-trip verifying Quay path layout
- [x] Boundary test: no distribution imports outside allowed packages


harishsurf (commented):
## Review Findings

6 findings from code review, ranked by severity. Finding 1 is a likely blocker for push/pull when accessed via hostname.

---

*Reviewed with Claude Code*

internal/registry/distribution/middleware/middleware.go
```diff
@@ -90,7 +89,7 @@ func (r *repository) Tags(ctx context.Context) distribution.TagService {
 
 // repoName converts this repository's distribution reference to an oci.RepositoryName.
 func (r *repository) repoName() oci.RepositoryName {
-	full := reference.Path(r.Named())
+	full := r.Named().Name()
```
harishsurf: **CRITICAL: `Named().Name()` includes the hostname, breaking namespace resolution**

`reference.Path(r.Named())` returns `admin/repo`. `r.Named().Name()` returns `localhost:5000/admin/repo`.

When this string is split on the first `/`, namespace becomes `localhost:5000` instead of `admin`, and repo name becomes `admin/repo` instead of `repo`.

This means every push/pull creates repos under a spurious hostname namespace in SQLite. Tag resolution, layer link checks, and manifest lookups all fail because the repository ID lookup uses the wrong namespace/name pair.

**Fix:** Revert to `reference.Path(r.Named())`.

```go
full := reference.Path(r.Named())
```

internal/oci/storage/local/distcompat.go
```diff
@@ -0,0 +1,597 @@
...
+	if ok, err := d.meta.BlobExists(ctx, dgst); err != nil {
+		return "", err
+	} else if ok {
+		return dgst, nil
+	}
+	return "", fmt.Errorf("blob not linked")
+}
+
+// PutContent writes small objects. Blob content is stored by digest,
+// upload state goes to the upload directory. Metadata link writes are
```
harishsurf, line 131: **MAJOR: Global `BlobExists` fallback bypasses repository-scoped access control**

After `BlobLinkedToRepo` returns false, this falls back to `BlobExists` which checks if the blob exists *anywhere* in the registry. This means any repository can reference any blob from any other repo by digest.

The comment says this is for "blobs uploaded but not yet referenced by a manifest" — but the fallback is overly broad. It allows permanent cross-repo blob access, not just during the upload window.

**Fix:** Restrict the fallback to the `uploadedblob` table scope (blobs being actively pushed to this specific repo), or remove the global fallback entirely.

```go
// Instead of global BlobExists, check uploadedblob for this repo only
if ok, err := d.meta.BlobUploadedToRepo(ctx, repoID, dgst); err != nil {
    return "", err
} else if ok {
    return dgst, nil
}
```

internal/oci/storage/local/driver.go
```diff
@@ -0,0 +1,167 @@
...
+	return err
+}
+
+// atomicWriter writes to a temp file, verifies digest, renames on Close.
+type atomicWriter struct {
+	tmp  *os.File
+	dst  string
+	dgst digest.Digest
+	hash digest.Digester
+}
```
harishsurf, line 143: **MAJOR: Hash skipped on partial write with error, causing digest mismatch on retry**

When `w.tmp.Write(p)` returns `n > 0` AND `err != nil` (partial write), the hash is NOT updated for the `n` bytes that were written, because of the `err == nil` guard. If the caller retries with the remaining bytes, the hash accumulator diverges from the file content. `Close()` then rejects a valid blob with a digest mismatch.

**Fix:** Hash whenever bytes are written, regardless of error:

```go
func (w *atomicWriter) Write(p []byte) (int, error) {
    n, err := w.tmp.Write(p)
    if n > 0 {
        _, _ = w.hash.Hash().Write(p[:n])
    }
    return n, err
}
```

internal/oci/storage/local/distcompat.go
```diff
@@ -0,0 +1,597 @@
...
+		}
+		return rc, err
+
+	case pathUpload:
+		id := uploadIDFromPath(path)
+		rc, err := d.blobs.UploadReader(ctx, id, offset)
+		if errors.Is(err, storage.ErrNotExist) {
+			return nil, storagedriver.PathNotFoundError{Path: path}
+		}
+		return rc, err
```
harishsurf, line 175: **MAJOR: `Move()` re-copies blob data instead of using `CommitUpload()` rename**

This reads the full upload via `UploadReader`, streams through `io.Copy` into a new `atomicWriter`, then calls `CancelUpload` to clean up. Meanwhile `CommitUpload()` in `chunked.go` does essentially the same thing more efficiently (it can rename instead of copy on the same filesystem).

For a 2GB blob, this doubles I/O and temporarily requires 4GB disk (upload data + blob copy). On constrained edge nodes running mirror-registry, this could exhaust disk space.

**Fix:** Use `CommitUpload` instead of manual copy + cancel:

```go
if srcKind == pathUpload && dstKind == pathBlob {
    dgst, err := digestFromBlobPath(dst)
    if err != nil {
        return err
    }
    id := uploadIDFromPath(src)
    return d.blobs.CommitUpload(ctx, id, dgst)
}
```

internal/oci/storage/local/distcompat.go
```diff
@@ -0,0 +1,597 @@
...
+			}}, nil
+		}
+		key := uploadStateKey(path)
+		if key != "" {
+			data, err := d.blobs.GetUploadState(ctx, id, key)
+			if errors.Is(err, storage.ErrNotExist) {
+				return nil, storagedriver.PathNotFoundError{Path: path}
+			}
+			if err != nil {
+				return nil, err
```
harishsurf, line 251: **MAJOR: `ListRepositories` full table scan called on every tree traversal step**

Both `isNamespacePrefix()` and `listRepoTree()` call `ListRepositories(ctx)` with no filter. Distribution's `Walk()` calls `Stat` + `List` recursively for GC and catalog enumeration. Each step triggers a full table scan, making Walk O(N²) for N repos.

The TODO acknowledges this is "fine at ~20 repos" but this will cause timeouts at scale and should be addressed before wider adoption.

**Fix:** Add a `ListRepositoriesByPrefix(ctx, prefix)` query:

```sql
SELECT u.username, r.name FROM repository r
JOIN "user" u ON r.namespace_user_id = u.id
WHERE u.username || '/' || r.name LIKE ? || '%'
```

internal/oci/storage/local/register.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/distribution/distribution/v3/registry/storage/driver/base"
+	"github.com/distribution/distribution/v3/registry/storage/driver/factory"
+
+	"github.com/quay/quay/internal/oci"
+)
+
+var (
+	metaMu         sync.RWMutex
+	registeredMeta oci.MetadataStore
+)
```
harishsurf, line 18: **MODERATE: `RegisterMetadataStore` global state silently overwritten on second call**

Package-level `registeredMeta` is overwritten on each `RegisterMetadataStore` call. The `init()` factory closure reads `registeredMeta` at Create-time (not register-time), so if `NewRegistry` is called twice (e.g., in tests with different SQLite databases), all registries share the last-registered store — causing silent test pollution.

**Fix:** Either error if already registered, or use `sync.Once`:

```go
var registerOnce sync.Once

func RegisterMetadataStore(meta oci.MetadataStore) {
    registerOnce.Do(func() {
        registeredMeta = meta
    })
}
```

internal/registry/distribution/middleware/middleware.go
jbpratt: The code already uses `reference.Path(r.Named())` (line 93), not `r.Named().Name()`. `reference.Path()` strips the hostname — for `localhost:5000/admin/repo` it returns `admin/repo`, which splits correctly into namespace=`admin`, name=`repo`.

```go
func (r *repository) repoName() oci.RepositoryName {
	full := reference.Path(r.Named())  // ← already correct
	...
}
```

No change needed.

internal/oci/storage/local/distcompat.go
jbpratt: This is intentional and matches both distribution's blobstore behavior and Python Quay's `get_repository_blob_by_digest`.

During a manifest push, distribution checks layer links *before* the manifest PUT that creates them:

1. Client pushes blob → content-addressable storage (global)
2. Client PUTs manifest → distribution calls `Stat` on each layer link
3. If `BlobLinkedToRepo` fails (no manifest references it yet), the push would 404

The `BlobExists` fallback lets step 2 succeed for blobs that exist in CAS but aren't yet linked to this repo via a manifest — this covers shared base layers and blobs from a prior push. The `uploadedblob` table with TTL provides actual GC protection.

Restricting to `BlobUploadedToRepo` would break the standard push flow for shared layers (e.g., `alpine` base layer pushed once, referenced by many repos).

No change needed.

internal/oci/storage/local/driver.go
jbpratt: Good catch. Fixed in 026c4b8 — changed `err == nil && n > 0` to `n > 0` so the hash accumulator stays in sync with the temp file on partial writes.

```go
n, err := w.tmp.Write(p)
if n > 0 {
    _, _ = w.hash.Hash().Write(p[:n])
}
```

internal/oci/storage/local/distcompat.go
jbpratt: Good catch — fixed in 026c4b8. `Move()` now delegates to `CommitUpload()` which verifies the digest and renames in place, avoiding the doubled I/O and 2x disk usage:

```go
func (d *DistDriver) Move(ctx context.Context, src, dst string) error {
    if classify(src) != pathUpload || classify(dst) != pathBlob {
        return fmt.Errorf("unsupported move: %s → %s", src, dst)
    }
    dgst, err := digestFromBlobPath(dst)
    if err != nil {
        return err
    }
    id := uploadIDFromPath(src)
    return d.blobs.CommitUpload(ctx, id, dgst)
}
```

internal/oci/storage/local/distcompat.go
jbpratt: Agreed — there's a TODO on `listRepoTree` for this already (added in this PR). At mirror-registry scale (~20 repos) the full scan takes microseconds, but for multi-tenant use we'd add a `ListRepositoriesByPrefix` query with a `LIKE` filter. Not addressing it in this PR to avoid interface churn for a scale scenario that's explicitly out of scope.

internal/oci/storage/local/register.go
jbpratt: Fair point. In practice, `RegisterMetadataStore` is only called once during process init, and tests bypass it entirely via `NewDistDriver(blobs, meta)` — so no test pollution occurs today. The `sync.Once` approach would actually be worse: it silently ignores subsequent registrations, which would mask misconfiguration rather than surfacing it.

That said, a defensive guard is reasonable. I'll update to panic on double-register (matching the pattern used by `database/sql.Register` and `distribution/registry/storage/driver/factory.Register`):

```go
func RegisterMetadataStore(meta oci.MetadataStore) {
    metaMu.Lock()
    defer metaMu.Unlock()
    if registeredMeta != nil {
        panic("RegisterMetadataStore called twice")
    }
    registeredMeta = meta
}
```

internal/registry/distribution/middleware/middleware.go
jbpratt: Correcting my earlier reply — you're right, the code does use `r.Named().Name()` which includes the hostname. Fixed in eec3ada to use `reference.Path(r.Named())` which strips it. Good catch.

jbpratt: Investigated further — `Named().Name()` is actually correct here. Distribution creates references internally without a hostname (e.g. `projectquay/clair`), so `reference.Path()` incorrectly strips `projectquay` as a hostname, returning just `clair`.

```
Without host: Name() = "projectquay/clair",  Path() = "clair"      ← wrong
With host:    Name() = "localhost:8443/...",  Path() = "projectquay/clair"
```

Reverted to `Named().Name()` with a comment explaining why. The `reference.Path()` change broke E2E Mirror (all blob layer link checks failed because repos were created under the wrong namespace).

harishsurf (approved):

PR #6181, opened 2026-06-17
NO-ISSUE: fix(workers): fix scheduling race and lock TTL in reconciliation worker

lechuk47:
## Summary

- **Fix first-run delay:** The `Worker` base class scheduled the first job execution with a 0.001s buffer (`start_date = now + 1ms`). APScheduler's `IntervalTrigger.get_next_fire_time()` races this — if the 1ms elapses before evaluation, `start_date` is in the past and the trigger skips to `start_date + interval`. At `RECONCILIATION_FREQUENCY=86400`, this delays the first run by a full day. Increased the buffer to 5 seconds.

- **Fix lock TTL on crash:** The `GlobalLock` TTL was set to `RECONCILIATION_FREQUENCY + LOCK_TIMEOUT_PADDING` (86460s). If the lock-holding process crashes, no other process can acquire the lock for ~24 hours. Decoupled by using `RECONCILIATION_TIMEOUT` (300s) with `auto_renewal=True` — the lock auto-extends while the process is alive via a background thread, but expires within 5 minutes on crash. This pattern is already used by `SecurityWorker`.

- Removed unused `RH_SKUS` and `LOCK_TIMEOUT_PADDING` imports.

### DIFF
--- a/workers/test/test_worker.py
+++ b/workers/test/test_worker.py
@@ -220,7 +220,8 @@ def test_start_schedules_job_with_5s_buffer():
         assert start_date is not None
         assert start_date >= before + timedelta(seconds=4)
         assert start_date <= after + timedelta(seconds=6)
-        assert call_kwargs.kwargs.get("seconds") or call_kwargs[1].get("seconds") == 86400
+        seconds = call_kwargs.kwargs.get("seconds") or call_kwargs[1].get("seconds")
+        assert seconds == 86400
 
 
 def test_start_stagger_workers_adds_random_delay():
--- a/workers/test/test_worker.py
+++ b/workers/test/test_worker.py
@@ -1,3 +1,4 @@
+from datetime import datetime, timedelta
 from unittest.mock import MagicMock, patch
 
 from workers.worker import (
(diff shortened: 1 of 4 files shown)


cubismod (changes requested):
Please address the coderabbit comment around tests, otherwise LGTM.

workers/worker.py
```diff
@@ -201,7 +201,7 @@ def start(self):
 
         self._sched.start()
         for operation_func, operation_sec in self._operations:
-            start_date = datetime.now() + timedelta(seconds=0.001)
+            start_date = datetime.now() + timedelta(seconds=5)
```
deshpandevlab, line 205: This changes the worker base class. And will affect all the workers. It might be OK because all our workers are with short intervals. Should we validate this change for the security workers??

workers/worker.py
```diff
@@ -201,7 +201,7 @@ def start(self):
 
         self._sched.start()
         for operation_func, operation_sec in self._operations:
-            start_date = datetime.now() + timedelta(seconds=0.001)
```
deshpandevlab, line 204: Can we add a small comment explaining why we changed the time delta from 0.001 to 5 in the code snippet?

workers/test/test_worker.py
```diff
@@ -184,3 +185,74 @@ def test_default_sentry_config_values():
...
+        mock_sched.add_job.assert_called_once()
+        call_kwargs = mock_sched.add_job.call_args
+        start_date = call_kwargs.kwargs.get("start_date") or call_kwargs[1].get("start_date")
+        if start_date is None:
+            start_date = call_kwargs[0][2] if len(call_kwargs[0]) > 2 else None
+
+        assert start_date is not None
+        assert start_date >= before + timedelta(seconds=4)
+        assert start_date <= after + timedelta(seconds=6)
+        assert call_kwargs.kwargs.get("seconds") or call_kwargs[1].get("seconds") == 86400
```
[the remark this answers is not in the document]
deshpandevlab: good catch by code rabbit. This will be read as `assert call_kwargs.kwargs.get("seconds") or (call_kwargs[1].get("seconds") == 86400)` let's add correct parentheses

deshpandevlab:
looks good overall, left two comments.

workers/worker.py
lechuk47: This only delays the first run by 5s. 0.001 is too short and the condition might be false on the first run. What kind of validation are you thinking of?

workers/worker.py
lechuk47: With git blame is easy to get to this PR and see the changes. I try to avoid code if not strictly necessary. 
The reason is a lot of times they end up being displaced or no longer true.

cubismod (approved):

deshpandevlab (approved):

PR #6216, opened 2026-06-23
PROJQUAY-11930: fix(build-logs): handle archived logs in superuser build logs page

LiZhang19817:
## Summary

- Fix superuser Build Logs page in the new UI showing "No logs available" for completed builds whose logs have been archived to object storage
- Add a new superuser archive endpoint (`GET /api/v1/superuser/<uuid>/logs/archive`) that streams archived build logs directly from object storage under superuser authorization, bypassing the `/logarchive/<uuid>` route which requires repository-level permissions
- Fix `fetchBuildLogsSuperuser()` in the React frontend to handle the `logs_url` response from the API, routing same-origin archive URLs through the new superuser endpoint
- Fix `validate_email()` to replace regex-based validation with string-split logic, preventing potential ReDoS
- Fix build logs not readable in dark mode by replacing hardcoded colors with PatternFly theme-aware `CodeBlock` component
- Add security test coverage for the new archive endpoint
- Add E2E Playwright test for readonly superuser viewing archived build logs from a private repo they have no membership in
- Add unit tests for archived log fetching in `BuildResource.test.ts`
- Add unit tests for email validation edge cases

## Root Cause

When a build completes, Quay archives logs from Redis to object storage (S3/Azure/GCS) and sets `logs_archived=True`. The `GET /api/v1/superuser/{uuid}/logs` endpoint then returns `{"logs_url": "..."}` instead of inline `{"logs": [...]}`. The new UI's superuser path ignored `logs_url` and evaluated `logsResponse.data.logs || []` → `[]`, causing the "No logs available" message.

Additionally, the fallback `/logarchive/<uuid>` route checks repository-level permissions, which fails for superusers (including readonly superusers) who are not members of the repository's namespace. The new `/api/v1/superuser/<uuid>/logs/archive` endpoint solves this by streaming the archived logs under superuser authorization.

### Backend (`endpoints/api/superuser.py`)
- **New endpoint:** `SuperUserRepositoryBuildLogsArchive` at `/v1/superuser/<build_uuid>/logs/archive`
  - Requires superuser scope + fresh login
  - Validates build exists via `pre_oci_model.get_repository_build()`
  - Streams gzip-decompressed archive from object storage via `log_archive`
  - Returns `application/json` response

### Frontend (`web/src/resources/BuildResource.ts`)
- Updated `fetchBuildLogsSuperuser()` to detect `logs_url` in the logs API response
- Same-origin archive URLs (e.g. `/logarchive/<uuid>`) are routed through the superuser archive endpoint to preserve the permission boundary
- External URLs (e.g. S3 pre-signed URLs) are followed directly
- Use `axios.defaults.baseURL` as the API origin instead of `window.location.origin`, supporting split-domain deployments where the API base URL differs from the UI origin

### Frontend (`web/src/routes/Superuser/BuildLogs/BuildLogs.tsx`)
- Replaced hardcoded `<pre>` with PatternFly `CodeBlock`/`CodeBlockCode` components for dark mode support
- Replaced hardcoded `backgroundColor: '#f5f5f5'` with PatternFly theme-aware background
- Replaced hardcoded timestamp color `#666` with `opacity: 0.6` (works with any text color)
- Replaced hardcoded error color `#c9190b` with PatternFly danger token `var(--pf-t--global--color--status--danger--default)`

### Validation (`util/validation.py`)
- Replaced regex-based `validate_email()` with string-split validation to prevent ReDoS
- Added length check (max 254 chars per RFC 5321)
- Validates exactly one `@`, non-empty local/domain parts, and at least two domain labels

### Tests
- **Security tests** (`endpoints/api/test/test_security.py`): Added permission matrix for the new archive endpoint (unauthenticated, freshuser, reader, devtable, globalreadonlysuperuser)
- **E2E tests** (`web/playwright/e2e/superuser/build-logs.spec.ts`): Added test for readonly superuser viewing archived logs from a private repo; added assertion that logs render (not empty state) in existing test
- **Unit tests** (`web/src/resources/BuildResource.test.ts`): Added tests for same-origin routing, external URL following, and empty archived logs
- **Validation tests** (`util/test/test_validation.py`): Added parametrized tests for email validation edge cases

## JIRA

https://redhat.atlassian.net/browse/PROJQUAY-11930

## Test plan

- [x] Trigger a build, wait for completion (logs get archived)
- [x] Open new UI → Superuser → Build Logs, enter the build UUID
- [x] Verify build logs are displayed instead of "No logs available"
- [x] Verify readonly superuser can view archived logs from a private repo they have no membership in
- [x] Verify build logs are readable in both light and dark mode
- [x] Verify non-superuser build view page still works (no regression)
- [x] Unit tests pass (`vitest`)
- [x] E2E tests pass (`playwright`)
- [x] Security endpoint permission tests pass (`pytest`)


web/playwright/e2e/superuser/build-logs.spec.ts
```diff
@@ -124,6 +124,68 @@ test.describe(
...
 
+    test('displays archived build logs fetched via logs_url', async ({
+      superuserPage,
+      api,
+    }) => {
+      const org = await api.organization('suarchive');
+      const repo = await api.repository(org.name, 'archived');
+      const build = await api.build(org.name, repo.name);
+
+      const archivedLogs = {
```
jbpratt: @LiZhang19817 can we do this without mocking? I imagine we can (we shouldn't mock in our e2e tests unless required)

LiZhang19817: yes, remove the mock testing.

web/src/resources/BuildResource.ts
```diff
@@ -488,15 +488,19 @@ export async function fetchBuildLogsSuperuser(
...
-  // Fetch both build info and logs in parallel
   const [buildResponse, logsResponse] = await Promise.all([
     axios.get(`/api/v1/superuser/${buildUuid}/build`, {signal}),
     axios.get(`/api/v1/superuser/${buildUuid}/logs`, {signal}),
   ]);
 
-  // Merge logs into build object
+  let logs = logsResponse.data.logs || [];
+  if (!isNullOrUndefined(logsResponse.data.logs_url)) {
+    const archived = await fetchArchivedBuildLogs(logsResponse.data.logs_url);
```
jbpratt: This now follows `logs_url`, but if that URL is the same-origin fallback from LocalStorage/RadosGW it leaves the `/api/v1/superuser/<uuid>/logs` permission boundary and calls `/logarchive/<uuid>`, which only checks repository permissions. Global readonly superusers, or `SUPER_USERS` without `FEATURE_SUPERUSERS_FULL_ACCESS`, can pass the superuser API calls and then get a 403 here when they are not repo members, so the page errors instead of rendering archived logs. Can we preserve superuser access for fallback archives, either by serving the archive content through the superuser endpoint or by making the fallback route honor `allow_if_any_superuser()` for build-log reads?

For coverage, a targeted Playwright e2e could use `readonlyPage`: create a private-repo build where the readonly superuser has no repo membership, wait for the build to finish and for `/api/v1/superuser/<uuid>/logs` to return a `/logarchive/<uuid>` fallback URL (the archiver polls every 30s), then load `/build-logs` as `readonlyPage` and assert the logs render with no `build-error-alert`. If the archiver timing is too flaky in CI, this should at least be backed by a backend route/API test for the fallback permission path.

util/validation.py
```diff
@@ -28,10 +28,19 @@ def validate_label_key(label_key):
 
 
 def validate_email(email_address):
-    if not email_address:
+    if not email_address or len(email_address) > 254:
```
bcaton85, line 31: @sridipta is working on a PR to remove unique email address, we'll need to make sure this doesn't conflict with those changes

LiZhang19817: without this change, the CI job will be failed.

endpoints/api/superuser.py
```diff
@@ -1329,6 +1330,44 @@ def get(self, build_uuid):
...
+                pre_oci_model.get_repository_build(build_uuid)
+            except InvalidRepositoryBuildException as e:
+                raise InvalidResponse(str(e))
+
+            try:
+                path = log_archive.get_file_id_path(build_uuid)
+                data_stream = log_archive._storage.stream_read_file(
+                    log_archive._locations, path
+                )
+                return send_file(GzipInputStream(data_stream), mimetype="application/json")
```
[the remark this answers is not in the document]
LiZhang19817: This endpoint is already behind allow_if_any_superuser() + @require_fresh_login — only authenticated superusers can access it. The data never reaches an unauthenticated browser/proxy in the first place.

LiZhang19817:
## Deep Review Summary

### Verdict: The bug is fixed correctly. No blocking issues found.

### How the fix works

**The bug:** When build logs are archived, `GET /api/v1/superuser/{uuid}/logs` returns `{"logs_url": "..."}` instead of `{"logs": [...]}`. The old `fetchBuildLogsSuperuser()` ignored `logs_url` and fell through to `logsResponse.data.logs || []` → empty array → "No logs available".

**The fix has two parts:**

1. **Frontend** (`BuildResource.ts:496-508`): Now detects `logs_url`, and intelligently routes:
   - Same-origin URLs (e.g. `/logarchive/<uuid>`) → superuser archive endpoint (avoids 403)
   - External URLs (e.g. S3 pre-signed) → direct fetch

2. **Backend** (`superuser.py:1333-1366`): New `SuperUserRepositoryBuildLogsArchive` endpoint that streams archived logs from object storage under superuser auth, mirroring the exact same logic as the existing `/logarchive/<uuid>` route (`web.py:441-466`) but with superuser authorization instead of repository-level permissions.

### Review Details

| Aspect | Assessment |
|--------|-----------|
| **Core bug fix** | Correct. `logs_url` is now handled, matching how `fetchBuildLogs()` (non-superuser, line 410) already works |
| **Same-origin detection** | Sound. `get_file_url()` returns either a full `https://` cloud URL or `urljoin(get_app_url(), "/logarchive/<uuid>")` — so checking `startsWith('/')` or `startsWith(window.location.origin)` covers both cases |
| **Backend endpoint security** | Properly decorated with `@require_fresh_login`, `@verify_not_prod`, `@require_scope(scopes.SUPERUSER)`, and `allow_if_any_superuser()` — matches all other superuser build endpoints |
| **Error handling** | Catches `InvalidRepositoryBuildException` (bad UUID) and `IOError` (missing archive) — mirrors the existing `/logarchive/` route |
| **Security tests** | Permission matrix covers unauthenticated, freshuser, reader, devtable, and globalreadonlysuperuser |
| **Email validation** | ReDoS fix is correct — the old regex `[^@]+@[^@]+\.[^@]+` had overlapping character classes. String-split approach is safe and the test coverage is thorough |
| **Private `_storage` access** | Uses `log_archive._storage.stream_read_file()` — technically accessing a private member, but this is the **exact same pattern** used by the existing `/logarchive/` route at `web.py:465`, so it's consistent |

### Minor observations (non-blocking)

1. **CodeRabbit's Cache-Control suggestion** — not needed (endpoint is behind superuser auth with fresh login; no public/proxy caching risk)
2. **No `signal` passthrough on archive fetch** — `fetchArchivedBuildLogs(archiveUrl)` doesn't forward the `AbortSignal`. Minor; the archive fetch is fast (single file read). The non-superuser `fetchBuildLogs()` at line 411 has the same behavior.

### CI Note

The failed `test_deduplicate_recent_vs_full_catalog` in PostgreSQL Integration Test is a pre-existing flaky test (`peewee.InternalError: current transaction is aborted`) — completely unrelated to this PR. Re-triggered.

LiZhang19817:
@jbpratt @bcaton85 Pls help review and approve this PR.

web/src/resources/BuildResource.ts
```diff
@@ -488,15 +488,28 @@ export async function fetchBuildLogsSuperuser(
...
 
-  // Merge logs into build object
+  let logs = logsResponse.data.logs || [];
+  if (!isNullOrUndefined(logsResponse.data.logs_url)) {
+    const logsUrl = logsResponse.data.logs_url;
+    // Same-origin fallback URLs (e.g. /logarchive/<uuid>) check repository
+    // permissions, not superuser permissions. Route through the superuser
+    // archive endpoint instead to preserve the permission boundary.
+    const isSameOrigin =
+      logsUrl.startsWith('/') || logsUrl.startsWith(window.location.origin);
```
jbpratt: Could we classify this against the configured API origin instead of `window.location.origin`? This is mainly for the OpenShift console plugin path: in production the plugin UI runs under the console origin while Axios points at the Quay API origin, so `window.location.origin` is not the right authority. Standalone Quay prod is normally same-origin, but plugin prod and the webpack dev split-origin flow can still receive a Quay fallback archive URL like `https://quay.io/logarchive/<uuid>` or `http://localhost:8080/logarchive/<uuid>`. This check treats that as external and fetches `/logarchive` directly, so readonly superusers without repo membership can still hit the repo-permission 403. A focused unit test with UI origin != API base URL would cover this.

jbpratt (approved):

PR #6160, opened 2026-06-14
PROJQUAY-6975: feat(org): allow single email for multiple organizations

sridipta:
## Summary

Implements the full [PROJQUAY-6975](https://redhat.atlassian.net/browse/PROJQUAY-6975) epic: **Allow Single Email for Multiple Organizations**.

Previously, Quay enforced a unique constraint on `User.email`, requiring each organization to have a distinct email address. This prevented users from reusing the same email across multiple orgs they own.

This PR decouples the internal unique email from the user-facing contact email:

- **Internal email (`User.email`):** Auto-generated UUID for orgs — never shown to users, ensures DB uniqueness
- **Contact email (`OrganizationContactEmail.contact_email`):** Optional, user-facing, non-unique — multiple orgs can share the same contact email

### Changes by story

| Story | JIRA | Scope |
|-------|------|-------|
| Backend CRUD + UUID org email | [PROJQUAY-10589](https://redhat.atlassian.net/browse/PROJQUAY-10589) | `create_organization()` generates UUID, 4 new CRUD functions for contact_email |
| API endpoints | [PROJQUAY-10591](https://redhat.atlassian.net/browse/PROJQUAY-10591) | `org_view()`, create/update endpoints expose `contact_email`, backward-compatible `email` field |
| Email notification routing | [PROJQUAY-10590](https://redhat.atlassian.net/browse/PROJQUAY-10590) | Recovery, invoice, payment-failed emails route via `contact_email` with admin fallback |
| React Create Org modal | [PROJQUAY-10592](https://redhat.atlassian.net/browse/PROJQUAY-10592) | Email field always visible, optional, labeled "Contact Email (Optional)" |
| React Org Settings | [PROJQUAY-10593](https://redhat.atlassian.net/browse/PROJQUAY-10593) | Settings page shows/edits `contact_email`, UUID hidden, clearing deletes the record |
| Unified recovery email | [PROJQUAY-11799](https://redhat.atlassian.net/browse/PROJQUAY-11799) | Single combined recovery email for user + org scenarios |
| Playwright E2E tests | [PROJQUAY-6975](https://redhat.atlassian.net/browse/PROJQUAY-6975) | 15 new E2E tests covering create, settings, lifecycle, API contract, recovery |

### Key design decisions

- **Separate table** (`OrganizationContactEmail`) with one-to-one FK, not a column on `User`
- **Backward-compatible API**: `email` field in requests maps to `contact_email`; GET responses include both fields
[...]

PR #6223, opened 2026-06-23
PROJQUAY-6975: feat(email): allow single email for multiple organizations

Marcusk19:
## Summary

Implements [PROJQUAY-6975](https://redhat.atlassian.net/browse/PROJQUAY-6975): **Allow Single Email for Multiple Organizations**.

Previously, Quay enforced a full unique constraint on `User.email`, requiring each organization to have a distinct email address. This PR replaces that with a **PostgreSQL partial unique index** (`UNIQUE WHERE organization = false`), allowing multiple organizations to share the same email while preserving uniqueness for regular user accounts.

### Approach: Partial Unique Index (supersedes #6160)

PR #6160 used a UUID-based approach with a separate `OrganizationContactEmail` table. That approach had [8 review blockers](https://github.com/quay/quay/pull/6160) because every consumer of `User.email` for orgs (webhooks, reconciliation worker, OAuth avatars, recovery) received a UUID instead of a real email.

This PR takes a simpler approach:
- **Organizations store their email directly in `User.email`** — no separate table, no UUIDs for orgs with emails
- **Partial unique index** enforces email uniqueness only for non-organization users
- **Zero changes needed** to webhooks, reconciliation worker, or OAuth avatar code — `User.email` is still the real email
- **Net -9 lines of code** vs the previous approach

### Changes

| Area | What changed |
|------|-------------|
| **Migration** | Drops full unique index, creates partial unique index using `CREATE INDEX CONCURRENTLY` (no write locks). Includes data backfill from `OrganizationContactEmail` table. Idempotent — safe to retry. |
| **ORM** | `User.email` removes `unique=True` (enforced at DB level via partial index) |
| **Model** | `create_organization()` stores email directly in `User.email`. New `find_organizations_by_email()` function. |
| **User lookups** | `find_user_by_email()`, `create_reset_password_email_code()`, `verify_user()`, `create_user_noverify()` all filter by `organization == False` |
| **API** | `org_view()` returns `email` (not `contact_email`). PUT accepts both for backward compat. |
| **Webhooks** | Uses `validate_email()` to detect UUID placeholders, falls back to admin emails |
| **Frontend** | Labels simplified from "Contact Email" to "Email". Single `email` field throughout. |
| **Tests** | 79+ backend tests pass. Playwright E2E tests updated. Migration safety test script included. |

### Production safety (quay.io scale)

- `CREATE INDEX CONCURRENTLY` — no write locks during index build
- New indexes created BEFORE old one is dropped — no window without email index
- Data backfill UPDATE commits before CONCURRENTLY begins (verified via Alembic source)
- Migration is idempotent — each step checks index existence before acting
- Included `data/migrations/test_migration_safety.py` for pre-deployment validation:
  ```bash
  # Read-only pre-flight check (safe against production):
  python data/migrations/test_migration_safety.py --database-url "postgresql://..." --preflight-only
  ```

## Test plan

- [x] All backend unit tests pass (`pytest` — 79+ tests across 5 files)
- [x] Pre-commit hooks pass
- [x] Migration applies cleanly on local dev (PostgreSQL)
- [x] E2E verified: two orgs created with same email, both display correctly
- [x] Database confirms partial unique index: `UNIQUE ... WHERE (organization = false)`
- [ ] Run migration safety pre-flight against staging
- [ ] Playwright CI E2E tests
- [ ] Verify no duplicate non-org emails exist in production before deploying


endpoints/webhooks.py
```diff
@@ -69,12 +71,27 @@ def _stripe_checkout_log_action(kind, namespace_name, performer_name, ip, metada
                 invoice = stripe.Invoice.retrieve(invoice_id)
                 if invoice:
                     invoice_html = renderInvoiceToHtml(invoice, namespace)
-                    send_invoice_email(
-                        namespace.invoice_email_address or namespace.email, invoice_html
-                    )
+                    recipient = namespace.invoice_email_address
+                    if not recipient and namespace.organization:
+                        if validate_email(namespace.email):
+                            recipient = namespace.email
```
bcaton85, line 77: **Blocker:** This introduces a regression for personal users. The old code was:

```python
send_invoice_email(namespace.invoice_email_address or namespace.email, invoice_html)
```

which falls back to `namespace.email` when `invoice_email_address` is None.

The new code sets `recipient = namespace.invoice_email_address`, then only enters the fallback block when `namespace.organization` is True. For a personal user where `invoice_email_address` is None (`CharField(null=True)`), `recipient` stays None, the org block is skipped, and `if recipient:` silently drops the email.

Fix — add a fallback for non-org users after the org block:
```python
elif not recipient:
    recipient = namespace.email
```

internal/dal/schema/sqlite/seed_data.sql
```diff
@@ -209,8 +209,8 @@ INSERT INTO mediatype VALUES(13,'application/vnd.docker.distribution.manifest.v1
 INSERT INTO mediatype VALUES(14,'application/vnd.docker.distribution.manifest.v1+prettyjws');
 INSERT INTO mediatype VALUES(15,'application/vnd.docker.distribution.manifest.v2+json');
 INSERT INTO mediatype VALUES(16,'application/vnd.docker.distribution.manifest.list.v2+json');
-INSERT INTO mediatype VALUES(17,'application/vnd.oci.image.manifest.v1+json');
-INSERT INTO mediatype VALUES(18,'application/vnd.oci.image.index.v1+json');
+INSERT INTO mediatype VALUES(17,'application/vnd.oci.image.index.v1+json');
```
bcaton85: **Should-fix:** This swaps the IDs of OCI manifest and OCI index media types compared to the old seed data:

```
Old: 17 = manifest, 18 = index
New: 17 = index,    18 = manifest
```

This looks like an unintentional side effect of re-dumping the schema. Any code or test that references these media types by numeric ID on fresh SQLite databases would get the wrong type. Should be reverted to preserve the original ID assignments.

endpoints/api/user.py
```diff
@@ -1116,21 +1117,28 @@ def redact(value):
...
+                )
+            send_combined_recovery_email(email, orgs_with_admins, reset_token=reset_token)
+        elif is_personal_user:
+            send_recovery_email(email, reset_token)
 
-        confirmation_code = model.user.create_reset_password_email_code(email)
-        send_recovery_email(email, confirmation_code)
         return {
             "status": "sent",
         }
```
bcaton85, line 1151: **Should-fix:** This is a breaking API change. The old endpoint returned `{"status": "org", "orgemail": ..., "orgname": ...}` when the email belonged to an organization. Now it always returns `{"status": "sent"}`. Worth calling out as a breaking change in the PR description — external API consumers relying on the `"org"` status type would break silently.

endpoints/test/test_webhooks.py
```diff
@@ -1,11 +1,220 @@
...
+        mock_stripe.Invoice.retrieve.return_value = mock_invoice
+
+        with patch(
+            "endpoints.webhooks.model.user.get_user_or_org_by_customer_id", return_value=user
+        ):
+            with patch(
+                "endpoints.webhooks.renderInvoiceToHtml", return_value="<html>invoice</html>"
+            ):
+                with patch("endpoints.webhooks.send_invoice_email") as mock_send:
+                    resp = self._post_stripe_event(
```
bcaton85, line 108: **Should-fix:** This test sets `invoice_email_address = "personal@example.com"`, so it doesn't exercise the fallback path where `invoice_email_address` is None. That's the exact path that regresses in the production code (see comment on `endpoints/webhooks.py`). Add a test where `invoice_email_address` is None and verify the email is sent to `user.email` instead.

util/useremails.py
```diff
@@ -60,7 +60,7 @@ def send_email(recipient, subject, template_file, parameters, action=None):
     app_title = app.config["REGISTRY_TITLE_SHORT"]
     app_url = get_app_url()
     html, plain = render_email(
-        app_title, app_url, recipient, subject, template_file, parameters, action=None
+        app_title, app_url, recipient, subject, template_file, parameters, action=action
```
bcaton85: **Nit:** This fixes a real bug where the `action` parameter was silently dropped for all email types — the old code hardcoded `action=None` instead of forwarding the caller's value. Worth calling out separately in the PR description (or as its own commit) since it changes behavior for all emails, not just org-related ones.

internal/dal/schema/sqlite/quay_schema.sql
```diff
@@ -1337,3 +1336,5 @@ CREATE TABLE organizationcontactemail (
 );
 CREATE UNIQUE INDEX organizationcontactemail_organization_id ON organizationcontactemail (organization_id);
```
bcaton85, line 1337: **Nit:** The bulk of the changes in this file are cosmetic reordering of `CREATE INDEX` / `CONSTRAINT` lines and a `blobupload.created` default timestamp change — artifacts of re-dumping the schema. Consider regenerating with a stable sort order or splitting the semantic changes (the two new indexes at the bottom) from the noise to make future reviews easier.

bcaton85:
**Should-fix — Reconciler worker: UUID placeholder emails bypass the skip filter**

`workers/reconciliationworker.py:170` checks `if user.email is None or user.email == ""` before sending the email to the marketplace API. Orgs created without an email now get a UUID placeholder (e.g. `a1b2c3d4-...`) in `User.email`, which passes this check and gets sent to `lookup_customer_id()`.

The marketplace API won't find a customer for a UUID string, so these orgs silently fall into the `no_customer_ids` metric bucket. No crash, but it inflates that metric and could mask real lookup failures.

Fix: add a `validate_email()` check so UUID placeholders are skipped with an appropriate metric label:
```python
from util.validation import validate_email

if user.email is None or user.email == "" or not validate_email(user.email):
    reconciliation_users_not_processed.labels(reason="missing_email").inc()
    continue
```

bcaton85:
**Should-fix — Reconciler worker: shared org emails may cause duplicate entitlement assignment**

The reconciler iterates all users/orgs and calls `lookup_customer_id(user.email)` independently for each. If two orgs share the same email (now allowed by this PR), both will resolve to the same marketplace customer ID and both will be reconciled against the same customer account.

Depending on the marketplace API's behavior, this could lead to duplicate entitlement creation or double-counting. Worth confirming with the marketplace team whether this is safe, or whether the reconciler should deduplicate by email before processing.

data/model/organization.py
```diff
@@ -11,29 +11,42 @@
...
+            if effective_email and email_required:
+                if not validate_email(effective_email):
+                    raise InvalidEmailAddressException(
+                        "Invalid email address: %s" % effective_email
+                    )
+
+            # Insert with a placeholder email first, then set organization=True
+            # before applying the real email. This avoids hitting the partial
+            # unique index (WHERE organization = false) during INSERT when the
+            # desired email is already used by a non-org user.
```
bcaton85: this partial index seems to be increasing the complexity in different parts of the code. Is there any concern with this change and future changes that this might cause spaghettifying(the technical term 😆 ) the code?

endpoints/test/test_webhooks.py
Marcusk19: added a `test_charge_succeeded_personal_user_no_invoice_email_falls_back` test case

endpoints/api/user.py
Marcusk19: restored to original api response

internal/dal/schema/sqlite/quay_schema.sql
Marcusk19: this one's tough. it's generated by `make go-schema` I don't think it's possible for me to clean up these cosmetic re-orderings as whenever I try to do that our schema drift check starts to fail

data/model/organization.py
```diff
@@ -108,6 +115,30 @@ def get_admin_users(org):
...
+        .join(Team)
+        .join(TeamRole)
+        .where(Team.organization << org_ids, TeamRole.name == "admin", User.robot == False)
+        .distinct()
+    )
+
+    result = {oid: [] for oid in org_ids}
+    for admin in admins:
+        org_id = admin.teammember.team.organization_id
+        result[org_id].append(admin)
```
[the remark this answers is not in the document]
Marcusk19: is this valid? 
> Peewee's ModelCursorWrapper.process_row() reconstructs the entire join chain from the single SQL query.
  admin.teammember.team.organization_id accesses Peewee's raw FK accessor — it doesn't trigger lazy loading because
  Team.organization is in the .select() and the join chain populates all intermediate instances. No fix needed.

endpoints/webhooks.py
```diff
@@ -69,12 +71,27 @@ def _stripe_checkout_log_action(kind, namespace_name, performer_name, ip, metada
...
+                        else:
+                            for admin in get_admin_users(namespace):
+                                if admin.email:
+                                    send_invoice_email(admin.email, invoice_html)
```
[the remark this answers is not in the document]
Marcusk19: n/a admin is always a user in the database (not an org) so there will not be a placeholder/invalid value as an email

data/model/organization.py
```diff
@@ -210,3 +217,7 @@ def is_org_admin(user, org):
         .where(Team.organization == org, TeamRole.name == "admin", TeamMember.user == user)
         .exists()
     )
+
+
+def find_organizations_by_email(email):
+    return User.select().where(User.email == email, User.organization == True)
```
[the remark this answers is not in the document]
Marcusk19: we already index the email field. this should be sufficient enough right?

endpoints/api/organization.py
```diff
@@ -340,20 +355,28 @@ def put(self, orgname):
...
+                    org.email = new_email if new_email else str(uuid.uuid4())
+                    org.save()
+                    log_action(
+                        "org_change_email",
+                        orgname,
+                        {
+                            "email": new_email,
+                            "namespace": orgname,
+                            "old_email": old_email,
+                        },
```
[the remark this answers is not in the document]
Marcusk19: audit logs have a history of including this data and since it's only accessible to members in the correct org/group/owner this is not a risk imo

data/model/organization.py
Marcusk19: I think the complexity should only be limited to the model here. Outside of that, this should give us the simplest way to achieve shared emails across orgs without having to touch other parts of the code

Marcusk19:
re: Should-fix — Reconciler worker: shared org emails may cause duplicate entitlement assignment

Discussed this on slack. I'm starting to see this as a non-issue. TLDR; the reconciler is only there for keeping track of legacy stripe customers in our marketplace system and these SKUs it creates do not actually grant any additional entitlements. I think it's fine then for legacy users to have multiple SKUS (whether paid/paid or paid/free) on their RH account as a result of this

cubismod (approved):
Created https://redhat.atlassian.net/browse/QUAYIO-1891 on our end to track the migration job.

PR #6224, opened 2026-06-23
PROJQUAY-9856: feat(bootstrap): programmatic bootstrap token provisioning

shaonrh:
Add feature-gated bootstrap OAuth token provisioning for Quay. When `FEATURE_PROGRAMMATIC_BOOTSTRAP` is enabled, startup provisions a superuser-scoped OAuth token for the first configured superuser and writes the token for automation consumers. If the feature is disabled, existing bootstrap tokens for the bootstrap application are revoked.

Token storage supports standalone and Kubernetes deployments. By default, the token JSON is written to `PROGRAMMATIC_TOKEN_PATH` on the local filesystem. When running in Kubernetes and `PROGRAMMATIC_TOKEN_K8S_SECRET` is configured, the token is written to that Kubernetes Secret instead, using `PROGRAMMATIC_TOKEN_K8S_KEY` and `PROGRAMMATIC_TOKEN_K8S_NAMESPACE`.

New API endpoints:
- POST `/api/v1/bootstrap/renew` — rotate the bootstrap token
- GET `/api/v1/organization/<orgname>/applications/<client_id>/tokens` — list OAuth API tokens for an organization application
- POST `/api/v1/organization/<orgname>/applications/<client_id>/tokens` — create an OAuth API token for an organization application
- DELETE `/api/v1/organization/<orgname>/applications/<client_id>/tokens/<token_uuid>` — revoke an OAuth API token

Add OAuth token metadata fields and tracking:
```
- created
- last_accessed
```

Update OAuth token validation to record last_accessed, add action log kinds for OAuth API token creation/revocation, and add UI log descriptions for those events.

Added configuration:
```
- FEATURE_PROGRAMMATIC_BOOTSTRAP=false
- PROGRAMMATIC_TOKEN_PATH=/var/lib/quay/quay-machine-token.json
- PROGRAMMATIC_TOKEN_EXPIRATION=3600
- PROGRAMMATIC_TOKEN_SCOPE="org:admin repo:admin repo:create repo:read repo:write super:user user:admin user:read"
- PROGRAMMATIC_TOKEN_K8S_SECRET=None
- PROGRAMMATIC_TOKEN_K8S_KEY=token.json
- PROGRAMMATIC_TOKEN_K8S_NAMESPACE=None
```


endpoints/api/organization_application_tokens.py
```diff
@@ -0,0 +1,228 @@
...
+        "expires_at": token.expires_at.isoformat() + "Z" if token.expires_at else None,
+        "created": token.created.isoformat() + "Z" if token.created else None,
+        "created_by": None,
+        "last_accessed": token.last_accessed.isoformat() + "Z" if token.last_accessed else None,
+    }
+
+    try:
+        if token.created_by:
+            view["created_by"] = token.created_by.username
+    except (AttributeError, User.DoesNotExist):
```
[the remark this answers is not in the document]
shaonrh: I verified this against the current code data/model/oauth.py and don’t think the N+1 applies here,

 `list_application_tokens()` currently does:

 ```python
   CreatedBy = User.alias()
   query = (
       OAuthAccessToken.select(OAuthAccessToken, CreatedBy)
       .join(CreatedBy, JOIN.LEFT_OUTER, on=(OAuthAccessToken.created_by == CreatedBy.id))
       .switch(OAuthAccessToken)
       .where(OAuthAccessToken.application == application)
   )
 ```

Because CreatedBy is included in the select, Peewee hydrates token.created_by from the join before endpoints/api/organization_application_tokens.py calls token_view().

Added unit tests here https://github.com/quay/quay/pull/6224/changes/3384dcb81be254ceea4002b4cbb8a3e7267c5cf8

data/model/oauth.py
```diff
@@ -333,6 +356,110 @@ def is_org_admin_or_has_token_assignment(self, organization, assignment_uuid):
...
+            .get()
+        )
+    except OAuthApplication.DoesNotExist:
+        return create_application(
+            org,
+            name,
+            application_uri="",
+            redirect_uri="",
+            description=BOOTSTRAP_APP_DESCRIPTION,
+        )
```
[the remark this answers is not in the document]
shaonrh: Adding a global unique constraint on `(organization, name)` is probably not appropriate here. OAuth applications currently do not enforce unique names per org, existing API/model behavior allows duplicates, and adding the constraint could break upgrades for installations that already have duplicate app names.

Existing tests cover the helper’s existing semantics, including duplicate OAuth application names, and cover bootstrap/renew end-state behavior. The specific concurrency safety claim is currently validated by code inspection rather than by a true concurrent PostgreSQL test. The current production call graph shows `get_or_create_application()` is only reached from bootstrap provisioning, and that caller locks the superuser row before invoking it.

endpoints/api/bootstrap.py
```diff
@@ -0,0 +1,95 @@
...
+
+        token_string = parts[1].strip()
+        if not token_string:
+            return {"message": "Empty bearer token"}, 401
+
+        current_token = validate_bootstrap_token(token_string, app.config)
+        if current_token is None:
+            return {"message": "Invalid bootstrap token"}, 401
+
+        if _is_expired(current_token) and not _is_loopback_remote_addr(request.remote_addr):
```
jbpratt: This check relies on `request.remote_addr`, which can be proxy-derived because Quay runs behind nginx and Werkzeug `ProxyFix`. In the default nginx template, `set_real_ip_from 0.0.0.0/0` is used when `TRUSTED_PROXY_CIDR` is unset, so `X-Forwarded-For: 127.0.0.1` may make a remote request appear local. Consider avoiding proxy-derived client IP for the expired-token renewal exception.

AI-assisted review note.

data/model/oauth.py
```diff
@@ -373,6 +526,64 @@ def validate_access_token(access_token):
...
+
+    try:
+        found = (
+            OAuthAccessToken.select(OAuthAccessToken, User, OAuthApplication)
+            .join(User, on=(OAuthAccessToken.authorized_user == User.id))
+            .switch(OAuthAccessToken)
+            .join(OAuthApplication)
+            .where(
+                OAuthAccessToken.token_name == token_name,
+                OAuthApplication.name == get_bootstrap_app_name(app_config),
```
jbpratt: This validates only that the token belongs to an OAuth application with the configured bootstrap app name. Should it also verify the expected bootstrap owner/application, or match the token currently stored in the configured sink? Otherwise a pre-existing app with the reserved name could have a token accepted by the renewal endpoint after upgrade.

AI-assisted review note.

shaonrh: mitigated by owner check + name reservation in the latest commit https://github.com/quay/quay/pull/6224/commits/0de9b2bf87a98f91c3107b18ada68a545cf6031d

endpoints/api/bootstrap.py
shaonrh: Thanks, mitigated by _unproxied_remote_addr() in the latest commit https://github.com/quay/quay/pull/6224/commits/0de9b2bf87a98f91c3107b18ada68a545cf6031d

data/model/oauth.py
```diff
@@ -333,6 +387,134 @@ def is_org_admin_or_has_token_assignment(self, organization, assignment_uuid):
...
+        token_name=token_name,
+        token_code=Credential.from_string(token_code),
+        access_token="",
+        token_type="Bearer",
+        expires_at=expires_at,
+        data=json.dumps({"username": user.username}),
+        created_by=user,
+        created=datetime.utcnow(),
+    )
+    return created, access_token
```
ibazulic: This looks the same as this function:

https://github.com/quay/quay/blob/deb032972d74789188ab54ecd78d891eed44e451/data/model/oauth.py#L483-L504

Can we reuse?

data/model/oauth.py
ibazulic: Tokens are unique, they always are part of one specific application. The token is composed of two parts, the first part indicates an app, while the 2nd part indicates the access token. By providing the full token, we already implicitly say to what app that token corresponds to. So why not reuse `validate_access_token` which we already have?

data/migrations/versions/1763d9d00e83_add_bootstrap_token_fields_to_.py
```diff
@@ -0,0 +1,57 @@
...
+    op.add_column("oauthaccesstoken", sa.Column("created_by_id", sa.Integer(), nullable=True))
+    op.add_column("oauthaccesstoken", sa.Column("last_accessed", sa.DateTime(), nullable=True))
+    op.add_column("oauthaccesstoken", sa.Column("created", sa.DateTime(), nullable=True))
+
+    with op.batch_alter_table("oauthaccesstoken") as batch_op:
+        batch_op.create_foreign_key(
+            "fk_oauthaccesstoken_created_by_id",
+            "user",
+            ["created_by_id"],
+            ["id"],
```
ibazulic: Current database schema has the `authorized_user_id` to use the token. Shouldn't we store the `created_by` information into audit logs instead of the database? According to the rest of the code, it doesn't really give any extra information and does appear like it's just for auditing purposes.

endpoints/api/bootstrap.py
```diff
@@ -0,0 +1,111 @@
...
+@show_if(features.PROGRAMMATIC_BOOTSTRAP)
+class BootstrapTokenRenew(ApiResource):
+    """Rotate the bootstrap token."""
+
+    @nickname("renewBootstrapToken")
+    def post(self):
+        auth_header = request.headers.get("Authorization", "")
+        parts = auth_header.split(" ", 1)
+        if len(parts) != 2 or parts[0].lower() != "bearer":
+            return {"message": "Authorization: Bearer <token> required"}, 401
```
ibazulic: We should return `Unauthorized()` here to be the same as all other API endpoints:

https://github.com/quay/quay/blob/deb032972d74789188ab54ecd78d891eed44e451/endpoints/exception.py#L113-L123

ibazulic: And same for all other API responses. Just for consistency sake.

boot.py
```diff
@@ -93,6 +106,130 @@ def setup_instance_service_key():
...
+
+    token = validate_bootstrap_token(access_token, app.config)
+    if token is None:
+        logger.warning("Bootstrap token sink contains an invalid token; rewriting it")
+        return None
+
+    return token
+
+
+def _provision_bootstrap_token():
```
ibazulic: What happens if both instances or multiple instances attempt to read and write the token at the same time? Will the last one created stick? How is that handled? (I could be missing some context).

data/model/oauth.py
shaonrh: Good point. The query is different today because `validate_bootstrap_token()` combines two things in one DB lookup: the normal OAuth token credential validation (`token_name` lookup + `token_code` match) and the bootstrap-specific constraint that the token belongs to the configured bootstrap application.

`validate_access_token()` is intentionally a lower-level credential verifier: it validates that the opaque token string maps to an `OAuthAccessToken` row and that the secret portion matches. It does not enforce expiration or caller policy; those checks happen at higher layers, e.g. `auth/oauth.py::validate_app_oauth_token()`. That behavior is useful here because bootstrap renewal intentionally needs to recognize an expired-but-valid token first, then apply its own localhost-only renewal policy.

That said, because `OAuthAccessToken.token_name` is unique, it should be acceptable to reuse `validate_access_token()` for the shared credential validation and then perform the bootstrap checks afterward:

- verify `found.application.name == get_bootstrap_app_name(app_config)`
- verify the application owner is in `SUPER_USERS`, when configured

This may lazy-load `found.application` / `found.application.organization` instead of doing the join up front, but this is a single-token validation path, not a list/read path, so the extra lookup is acceptable and keeps the credential validation logic centralized.

data/model/oauth.py
```diff
@@ -127,7 +181,7 @@ def load_authorized_scope_string(self, client_id, username):
             OAuthAccessToken.select()
             .join(OAuthApplication)
             .switch(OAuthAccessToken)
-            .join(User)
+            .join(User, on=(OAuthAccessToken.authorized_user == User.id))
```
bcaton85, line 187: Why is this being added? isn't this the same behavior as before?

data/model/oauth.py
```diff
@@ -333,6 +387,134 @@ def is_org_admin_or_has_token_assignment(self, organization, assignment_uuid):
...
+    max_active_tokens=MAX_TOKENS_PER_APPLICATION,
+):
+    """Create an OAuth access token while atomically enforcing the active-token limit."""
+    with db_transaction():
+        locked_application = db_for_update(
+            OAuthApplication.select().where(OAuthApplication.id == application.id)
+        ).get()
+
+        if count_active_tokens(locked_application) >= max_active_tokens:
+            raise TokenLimitExceeded()
```
bcaton85, line 466: Why are we limiting the number of active tokens?

data/model/oauth.py
```diff
@@ -373,6 +555,73 @@ def validate_access_token(access_token):
...
+                return None
+
+        return found
+    except OAuthAccessToken.DoesNotExist:
+        return None
+
+
+def get_bootstrap_tokens(application):
+    """Return all tokens for the given application."""
+    return list(OAuthAccessToken.select().where(OAuthAccessToken.application == application))
```
bcaton85, line 570: this isn't really narrowing down to bootstrap tokens if it's retrieving all tokens matching a specific app right, should this have a more accurate name?

data/migrations/versions/1763d9d00e83_add_bootstrap_token_fields_to_.py
```diff
@@ -0,0 +1,57 @@
...
+def upgrade(op, tables, tester):
+    op.add_column("oauthaccesstoken", sa.Column("created_by_id", sa.Integer(), nullable=True))
+    op.add_column("oauthaccesstoken", sa.Column("last_accessed", sa.DateTime(), nullable=True))
+    op.add_column("oauthaccesstoken", sa.Column("created", sa.DateTime(), nullable=True))
+
+    with op.batch_alter_table("oauthaccesstoken") as batch_op:
+        batch_op.create_foreign_key(
+            "fk_oauthaccesstoken_created_by_id",
+            "user",
+            ["created_by_id"],
```
bcaton85: would this require a backfill for the existing entries on quay.io?

data/model/oauth.py
ibazulic: @bcaton85 

> Why is this being added? isn't this the same behavior as before?

The migration contains a new foreign key constraint so one has to specify which join should be used. Peewee can figure it out if there's only one FK, but with two it doesn't know what to do.

bcaton85: i see now, thank you!

data/migrations/versions/1763d9d00e83_add_bootstrap_token_fields_to_.py
shaonrh: Good point. `created_by_id` was redundant for the current implementation: all token creation paths set the token’s `authorized_user` to the same user that created the token, and auditing is already handled through the Quay action logs.

I removed the persisted `created_by_id` column/FK from the migration and model. The API still returns the `created_by` response field for compatibility, but it is now derived from `authorized_user.username` instead of a separate DB column.

Addressed in e351a793.

data/model/oauth.py
shaonrh: They should not be collapsed into one public function because they have different contracts and callers depend on those differences. We can extract some of the common functionalities to a separate function, but I don't see any big improvement by doing so.

endpoints/api/bootstrap.py
shaonrh: Thanks for catching this; updated in f04bb7bbe to use Quay’s standard Unauthorized() responses for bootstrap renewal auth failures, with tests updated and passing.

data/model/oauth.py
ibazulic: Can you give an example? What kind of contracts are we talking about here? I'm afraid we will end up in having multiple functions that do almost the same thing which will cause needless maintenance burden later.

shaonrh: > I'm afraid we will end up in having multiple functions that do almost the same thing which will cause needless maintenance burden later

You bring up a good point, Ivan. I have pushed update to extract a common method.

conf/nginx/server-base.conf.jnj
```diff
@@ -11,6 +11,7 @@ add_header X-Frame-Options DENY;
 
 # Proxy Headers
 proxy_set_header X-Forwarded-For $proper_forwarded_for;
+proxy_set_header X-Quay-Original-Remote-Addr $quay_original_remote_addr;
```
bcaton85, line 14: Are these changes relevant to this feature?

data/migrations/versions/1763d9d00e83_add_bootstrap_token_fields_to_.py
```diff
@@ -0,0 +1,46 @@
...
+# revision identifiers, used by Alembic.
+revision = "1763d9d00e83"
+down_revision = "c3d4e5f6a7b8"
+
+import sqlalchemy as sa
+
+
+def upgrade(op, tables, tester):
+    op.add_column("oauthaccesstoken", sa.Column("last_accessed", sa.DateTime(), nullable=True))
+    op.add_column("oauthaccesstoken", sa.Column("created", sa.DateTime(), nullable=True))
```
bcaton85, line 18: do we need any indexes for these? or we doing any searches by these dates?

data/migrations/versions/1763d9d00e83_add_bootstrap_token_fields_to_.py
```diff
@@ -0,0 +1,46 @@
+"""add bootstrap token fields to oauthaccesstoken
+
+Revision ID: 1763d9d00e83
+Revises: c3d4e5f6a7b8
+Create Date: 2026-06-12 20:18:35.819774
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = "1763d9d00e83"
```
bcaton85, line 10: before merging you'll want to confirm this upgrade path is correct, in case any other db migration has been merged in the mean time.

endpoints/api/bootstrap.py
```diff
@@ -0,0 +1,155 @@
...
+        new_record = None
+        try:
+            with db_transaction():
+                lock_bootstrap_token_operation()
+
+                current_token = validate_bootstrap_token(token_string, app.config)
+                if current_token is None:
+                    _raise_invalid_bootstrap_token()
+
+                if _is_expired(current_token) and not _is_loopback_bootstrap_renewal_request(
```
bcaton85, line 119: might be worth leaving a comment on this is required for the acceptance criteria. Maybe not needed if only agents are going to be making changes but it's not immediately obvious why this check is here. Comments may help agents figure this out quicker?

endpoints/api/organization_application_tokens.py
```diff
@@ -0,0 +1,262 @@
...
+    if include_secret and secret is not None:
+        view["token"] = secret
+
+    return view
+
+
+@resource("/v1/organization/<orgname>/applications/<client_id>/tokens")
+@path_param("orgname", "The name of the organization")
+@path_param("client_id", "The OAuth client ID")
+class OrganizationApplicationTokens(ApiResource):
```
bcaton85, line 95: This also exists currently in web.py, are we replacing that implementation? Should that be cleaned up?

local-dev/ldap/ldap-config.yaml
```diff
@@ -10,5 +10,6 @@ LDAP_EMAIL_ATTR: mail
 LDAP_MEMBEROF_ATTR: quayMemberOf
 LDAP_UID_ATTR: uid
 LDAP_URI: ldap://quay-ldap:3389
+LDAP_SUPERUSER_FILTER: (|(uid=admin)(uid=user1)(uid=quayadmin))
```
bcaton85, line 13: not opposed to adding this, but is it relevant or required for this PR?

Dockerfile
```diff
@@ -188,7 +188,7 @@ RUN set -ex\
 # symlink.
 	; ln -s $QUAYCONF /conf\
 # Make a grip of runtime directories.
-	; newdir /certificates "$QUAYDIR" "$QUAYDIR/conf" "$QUAYDIR/conf/stack" /datastorage\
+	; newdir /certificates "$QUAYDIR" "$QUAYDIR/conf" "$QUAYDIR/conf/stack" /datastorage /var/lib/quay\
```
bcaton85, line 191: you'll have to make the same change in the downstream dockerfile

bcaton85 (commented):
For the most part LGTM, just need clarifications here and there

conf/nginx/server-base.conf.jnj
shaonrh: yes, this is to prevent renewal of expired token from anywhere but localhost

PR #6271, opened 2026-06-30
PROJQUAY-11912: feat(go): support anonymous pulls

jbpratt:
Summary
- Allows anonymous pull grants for public repositories when FEATURE_ANONYMOUS_ACCESS is enabled.
- Keeps invalid Basic credentials distinct from missing credentials.
- Adds focused Go auth and registry tests.

Testing
- go test ./internal/auth ./internal/registry/distribution
- go test ./internal/cmd
- go vet ./internal/auth ./internal/cmd ./internal/dal/daldb ./internal/registry/distribution
- golangci-lint run ./internal/auth ./internal/cmd ./internal/dal/daldb ./internal/registry/distribution

### DIFF
--- a/internal/registry/distribution/auth.go
+++ b/internal/registry/distribution/auth.go
@@ -5,29 +5,34 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"strings"
 
-	"golang.org/x/crypto/bcrypt"
-
-	"github.com/distribution/distribution/v3/registry/auth"
+	distauth "github.com/distribution/distribution/v3/registry/auth"
+	"github.com/quay/quay/internal/auth"
 	"github.com/quay/quay/internal/dal/daldb"
 	"github.com/quay/quay/internal/oci"
 )
 
 func init() {
-	if err := auth.Register("quaydb", auth.InitFunc(newAccessController)); err != nil {
+	if err := distauth.Register("quaydb", distauth.InitFunc(newAccessController)); err != nil {
 		slog.Error("failed to register quaydb auth", "err", err)
 	}
 }
 
 type accessController struct {
-	queries *daldb.Queries
-	realm   string
+	authenticator    *auth.BasicAuthenticator
+	queries          *daldb.Queries
+	realm            string
+	libraryNamespace string
+	anonymousAccess  bool
 }
 
-var _ auth.AccessController = &accessController{}
-var _ oci.Authenticator = &accessController{}
+var (
+	_ distauth.AccessController = &accessController{}
+	_ oci.Authenticator         = &accessController{}
+)
 
-func newAccessController(options map[string]interface{}) (auth.AccessController, error) {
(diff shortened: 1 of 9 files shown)


internal/registry/distribution/auth.go
```diff
@@ -38,69 +43,128 @@ func newAccessController(options map[string]interface{}) (auth.AccessController,
...
 	}
 
-	return &auth.Grant{User: auth.UserInfo{Name: username}}, nil
+	return &distauth.Grant{User: distauth.UserInfo{Name: result.Principal.Username}}, nil
 }
 
-func (ac *accessController) Authenticate(r *http.Request, _ ...oci.Access) (*oci.Grant, error) {
-	username, ok := ac.authenticateUser(r)
-	if !ok {
+func (ac *accessController) Authenticate(r *http.Request, access ...oci.Access) (*oci.Grant, error) {
```
harishsurf, line 89: **MODERATE: `canAnonymousPull` returns false when `access` is empty — blocks anonymous `/v2/` ping**

The `/v2/` endpoint (version check) is called by all OCI clients before any push/pull. Distribution calls `Authorized()` with an **empty access list** for `/v2/`. With `len(access) == 0 → return false`, an anonymous `GET /v2/` returns a 401 challenge, forcing the client into a token exchange before it can even check if the registry is reachable.

Anonymous pulls still work (the client follows the challenge, gets a token, then pulls), but it adds an unnecessary round-trip. Python Quay allows anonymous `/v2/` access without a challenge.

**Fix:**

```go
if !ac.anonymousAccess {
    return false
}
if len(access) == 0 {
    return true // /v2/ ping — no specific resource requested
}
```

jbpratt: updated!

jbpratt: I validated this locally against the standalone Go registry and had to revert the `/v2/` no-challenge behavior.

Repro on the prior update:
- `GET /v2/` returned `200 {}` anonymously.
- `podman push localhost:18443/admin/test-image:v1 --authfile ...` then failed during blob existence checks with `authentication required`.
- The server logs showed containers/image accepted the anonymous `/v2/` ping, then sent the next `HEAD /v2/admin/test-image/blobs/...` without credentials.

After restoring the Basic challenge for empty-access `/v2/`:
- anonymous `curl -k https://localhost:18443/v2/` returns `401` with `WWW-Authenticate: Basic realm="localhost"`.
- `podman push` succeeds.
- `podman pull localhost:18443/admin/test-image@sha256:...` succeeds.
- `podman pull localhost:18443/admin/test-image:v1` succeeds.

So keeping the challenge is intentional for Go Quay right now; otherwise credentialed private push/pull clients can skip auth setup after the initial ping and fail on the first scoped request. The anonymous public-pull path is still scoped to explicit `repository:...:pull` access checks.

harishsurf (approved):

PR #6280, opened 2026-06-30
PROJQUAY-12140: feat(bootstrap): add file-based startup provisioning

shaonrh:
Add feature-flagged startup provisioning for a bootstrap OAuth token stored in a local file:

- create or reuse the owner-scoped bootstrap OAuth application
- keep startup idempotent when a bootstrap DB token already exists
- roll back newly created DB token state when local file writes fail
- revoke bootstrap tokens when the feature is disabled
- add Docker runtime directory support and reserved app-name guardrails

### DIFF
--- a/boot.py
+++ b/boot.py
@@ -12,9 +12,21 @@ from jinja2 import Template
 import release
 from _init import CONF_DIR
 from app import app
-from data.model import ServiceKeyDoesNotExist
+from data.model import ServiceKeyDoesNotExist, db_transaction
+from data.model.oauth import (
+    create_bootstrap_application,
+    create_bootstrap_oauth_api_token,
+    delete_applications,
+    get_bootstrap_app_name,
+    get_bootstrap_application_candidates,
+    get_bootstrap_tokens,
+    lock_bootstrap_token_operation,
+    lookup_applications_by_name,
+)
 from data.model.release import set_region_release
 from data.model.service_keys import get_service_key
+from data.model.user import get_user
+from util.bootstrap_token import write_bootstrap_token
 from util.config.database import sync_database_with_config
 from util.generatepresharedkey import generate_key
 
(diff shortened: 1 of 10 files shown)


boot.py
```diff
@@ -93,6 +104,137 @@ def setup_instance_service_key():
...
+                duplicate_applications = bootstrap_applications[1:]
+                if duplicate_applications:
+                    delete_applications(duplicate_applications)
+                    logger.info(
+                        "Deleted %s duplicate bootstrap applications",
+                        len(duplicate_applications),
+                    )
+
+                logger.info("Bootstrap token already provisioned, skipping")
+                return
```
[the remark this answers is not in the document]
shaonrh: Thanks for the review. I’m going to skip this because `BOOTSTRAP_TOKEN_PATH` is node-local and cannot be treated as authoritative in multi-node deployments.

If another node provisioned the token, this node may legitimately not have the file or may have a stale/corrupt local copy. Rotating based on that would cause unnecessary DB token churn and could invalidate a still-valid bootstrap token for the cluster.

The DB is the shared source of truth here, and the inline comment around this return documents that we intentionally do not rotate solely because the local file is missing or malformed.

data/model/oauth.py
```diff
@@ -333,6 +340,165 @@ def is_org_admin_or_has_token_assignment(self, organization, assignment_uuid):
...
+
+def lookup_applications_by_name(org, name):
+    """Look up OAuthApplications by organization and name in newest-first order."""
+    return list(
+        OAuthApplication.select()
+        .where(
+            OAuthApplication.organization == org,
+            OAuthApplication.name == name,
+        )
+        .order_by(OAuthApplication.id.desc())
```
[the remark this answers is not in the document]
shaonrh: Thanks for the callout. I verified the current schema already has an index on `oauthapplication.organization_id` (`oauthapplication_organization_id`), so this lookup is not an unbounded table scan. The new lookup is currently only used from bootstrap token setup/revocation during startup, not on a request path, and the existing org index limits the remaining name filter to OAuth applications owned by the configured bootstrap owner.

Given the startup-only usage and expected small per-owner OAuth application cardinality, I do not think adding a new migration for a composite `(organization_id, name)` index is justified here. If this lookup becomes a hot path or we see deployments with unusually large per-owner OAuth app counts, we can add a non-unique composite index in a follow-up.

boot.py
```diff
@@ -93,6 +105,140 @@ def setup_instance_service_key():
...
+        lock_bootstrap_token_operation()
+
+        bootstrap_applications = []
+        for owner in _get_bootstrap_revocation_owners():
+            applications = lookup_applications_by_name(owner, bootstrap_application_name)
+            for application in applications:
+                if get_bootstrap_tokens(application):
+                    bootstrap_applications.append(application)
+
+        delete_applications(bootstrap_applications)
```
LiZhang19817, line 234: looks like the Token file will not be deleted on revocation.

data/model/oauth.py
```diff
@@ -333,6 +340,211 @@ def is_org_admin_or_has_token_assignment(self, organization, assignment_uuid):
...
+
+    return user.get_user(owner_name)
+
+
+def validate_bootstrap_token(access_token, app_config=None):
+    """Validate a token belongs to the configured canonical bootstrap app."""
+    if app_config is None:
+        app_config = config.app_config or {}
+
+    found = validate_access_token(access_token)
```
LiZhang19817, line 494: Looks like this will not check if the token is already expired.

shaonrh: Correct, this path does not check expiration. The startup bootstrap provisioning step intentionally does not handle expired tokens; expiration handling belongs to the renewal/auth layer.

boot.py
shaonrh: @LiZhang19817 Good catch. Fixed this by deleting the local bootstrap token file during revocation as a best-effort cleanup.

boot.py
```diff
@@ -93,6 +105,150 @@ def setup_instance_service_key():
...
+
+
+def _revoke_bootstrap_tokens():
+    bootstrap_application_name = get_bootstrap_app_name(app.config)
+    with db_transaction():
+        lock_bootstrap_token_operation()
+
+        bootstrap_applications = []
+        for owner in _get_bootstrap_revocation_owners():
+            applications = lookup_applications_by_name(owner, bootstrap_application_name)
```
jbpratt, line 229: When FEATURE_PROGRAMMATIC_BOOTSTRAP is disabled after BOOTSTRAP_APP_NAME or BOOTSTRAP_TOKEN_OWNER has changed, this only searches the current owner/name pair. Bootstrap tokens minted under the previous config are still normal OAuth bearer tokens and remain valid until expiry, so disabling the feature can leave an old bootstrap credential active. Revocation should find tokens by their bootstrap metadata rather than only the current configured application lookup.

shaonrh: Fair point, I have been debating with this one. It does cleanup bootstrap apps for all superusers listed in config if owner doesn't exit. However, I guess, we can do that always. Sending a followup PR

jbpratt (approved):

PR #6350, opened 2026-07-02
PROJQUAY-12148: feat(bootstrap): add bootstrap token renewal

shaonrh:
- Add `POST /api/v1/bootstrap/renew` for exchanging bootstrap Bearer tokens
- Rotate bootstrap tokens in the configured local token store and canonical bootstrap app
- Forward original client addresses through nginx with `X-Quay-Original-Remote-Addr`
- Restrict expired-token renewal to loopback original client addresses

### DIFF
--- a/conf/nginx/http-base.conf.jnj
+++ b/conf/nginx/http-base.conf.jnj
@@ -46,6 +46,11 @@ map $proxy_protocol_addr $proper_forwarded_for {
   default $proxy_protocol_addr;
 }
 
+map $realip_remote_addr $quay_original_remote_addr {
+  ""      $remote_addr;
+  default $realip_remote_addr;
+}
+
 map $http_x_forwarded_proto $proper_scheme {
   default $scheme;
   https   https;
--- a/conf/nginx/rate-limiting.conf.jnj
+++ b/conf/nginx/rate-limiting.conf.jnj
@@ -36,6 +36,13 @@ map $uri:$http_user_agent $miner_tag_block {
     default 0;
 }
 
+# Prefer Quay's original client address for bootstrap renewal limits. Fall back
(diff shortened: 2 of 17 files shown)


LiZhang19817:
## Validation of the four PR goals

| Goal | Status | Evidence |
|------|--------|----------|
| **POST /api/v1/bootstrap/renew** | ✅ Works | Bearer token parsed manually, validated against canonical app, double-checked after lock. `@anon_allowed` + `@show_if` correctly scoped. 500 lines of test coverage. |
| **Rotate in local token store + canonical app** | ✅ Works | New token created via `create_bootstrap_oauth_api_token`, old tokens deleted via `delete_bootstrap_tokens(keep_token_id=new_record.id)`, file written atomically via `os.replace()` in `write_token_file`. All within `db_transaction()` + advisory lock. |
| **Forward original client via nginx** | ✅ Works | `$realip_remote_addr` → `$quay_original_remote_addr` map → `proxy_set_header` at server level. Bootstrap `location =` block has no own `proxy_set_header`, so it inherits. nginx's `proxy_set_header` always overwrites client-provided values. |
| **Restrict expired renewal to loopback** | ✅ Works | Two-layer check: (1) `X-Quay-Original-Remote-Addr` must be loopback, AND (2) raw TCP peer must be internal or unset (Unix socket). Tests cover: IPv4 loopback, IPv6 loopback, IPv4-mapped IPv6, spoofed `X-Forwarded-For`, spoofed nginx header with remote peer, Unix socket + nginx header. |

The security design is solid. Multiple defense layers, correct header handling, well-tested spoofing scenarios. Nothing needs to be fixed before merge from a security standpoint.

LiZhang19817 (approved):

PR #6356, opened 2026-07-03
PROJQUAY-11891: fix(mirroring): clean up org mirror before org deletion

shaonrh:


LiZhang19817:
This issue was fixed
image

LiZhang19817 (approved):

PR #6371, opened 2026-07-04
PROJQUAY-12148: fix(bootstrap): gate expired renewal on nginx local signal

### LINKED ISSUE #6358: Audit nginx set_real_ip_from trust for bootstrap token renewal loopback restriction
## What happened

PR [#6350](https://github.com/quay/quay/pull/6350) introduced a bootstrap token renewal endpoint (`POST /api/v1/bootstrap/renew`) that restricts expired-token renewal to loopback client addresses via the `X-Quay-Original-Remote-Addr` header, which is derived from `$realip_remote_addr` in the nginx config. During [CodeRabbit's first review](https://github.com/quay/quay/pull/6350#pullrequestreview) on 2026-07-02, it flagged a security concern: the existing nginx config may use `set_real_ip_from 0.0.0.0/0` as a fallback, which trusts all upstream sources to set `X-Forwarded-For`. This would allow any external client to spoof their real IP and bypass the loopback-only restriction on expired-token renewal. The human reviewer validated the four functional goals but did not address this security concern. The PR was merged on 2026-07-03.

## What could go better

The CodeRabbit security finding was marked as an outside-diff concern (it relates to existing nginx config, not code introduced by the PR) and was not picked up by the human reviewer. This is a moderate-confidence concern — the actual risk depends on how `set_real_ip_from` is configured in production and whether there are other network-level controls (e.g., load balancer restrictions). However, since the bootstrap token renewal endpoint explicitly relies on client IP checking as a security boundary, the IP trust chain should be verified.

## Proposed change

Audit the `set_real_ip_from` directives in `conf/nginx/http-base.conf.jnj` (and any Jinja template conditionals that generate them) to ensure they only trust known proxy/load-balancer CIDRs rather than `0.0.0.0/0`. If the fallback is necessary for certain deployment topologies, add a comment documenting the security implications and consider an alternative mechanism (e.g., a shared secret header) for the bootstrap renewal loopback check in `endpoints/api/bootstrap.py`.

## Validation criteria

The `set_real_ip_from` directives in the nginx config templates either (a) restrict trust to specific proxy CIDRs, or (b) include documented justification for broader trust with compensating controls. The bootstrap token renewal loopback restriction cannot be bypassed by an external client setting `X-Forwarded-For`.

---
_Generated by retro agent from https://github.com/quay/quay/pull/6350_

shaonrh:
Use a scoped nginx-owned local/remote signal for `/api/v1/bootstrap/renew`.

Valid bootstrap tokens can still renew normally. Expired bootstrap tokens now require local access to the Quay runtime, such as running the request inside the Quay container or pod. Remote requests through a load balancer or ingress are rejected, even if they spoof forwarded headers or `X-Quay-Bootstrap-Renewal-Location`.

Fixes: https://github.com/quay/quay/issues/6358
Related: https://github.com/quay/quay/pull/6350


LiZhang19817:
**Pre-merge testing is passed,** the following is the testing coverage
image

bcaton85 (approved):

PR #6381, opened 2026-07-06
QUAYIO-1891: feat: alembic migration job

cubismod:
Creates a new OpenShift Template intended for database migrations. Currently, we run migrations on startup for Quay.io which can create problems as those migration jobs are linked into the pod's startup/liveness probes. If a migration takes too long, a pod can start failing healthchecks and the migration can be cancelled midway through.

This is a separate job that we can spin up independent of normal Quay pod's lifecycles. 

https://redhat.atlassian.net/browse/QUAYIO-1891

Test plan is to deploy to stage with a no-op migration and evaluate whether that works. To be used with https://github.com/quay/quay/pull/6223 migration.

### DIFF
--- a/deploy/openshift/rosa/quay-migration-job.yaml
+++ b/deploy/openshift/rosa/quay-migration-job.yaml
@@ -4,25 +4,6 @@ kind: Template
 metadata:
   name: quay-migration-job
 objects:
-- apiVersion: rbac.authorization.k8s.io/v1
-  kind: Role
-  metadata:
-    name: ${{NAME}}
-  rules:
-  - apiGroups:
-    - ""
-    resources:
-    - secrets
-    verbs:
-    - get
-    - patch
-    - update
-  - apiGroups:
-    - ""
-    resources:
-    - namespaces
-    verbs:
-    - get
 - apiVersion: v1
   kind: ServiceAccount
   metadata:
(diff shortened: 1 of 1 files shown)


deploy/openshift/rosa/quay-migration-job.yaml
```diff
@@ -0,0 +1,157 @@
...
+    value: "head"
+    displayName: migration version
+    description: Target Alembic revision to migrate to (e.g. head, or a specific revision hash)
+  - name: QUAY_COMPONENT
+    value: "db-migration"
+    displayName: quay component selector value
+
+  # Job attributes
+  - name: BACKOFF_LIMIT
+    value: "6"
```
Sunandadadi: does it make sense to have 6 retries? If a migration fails, I think it will require human intervention

deploy/openshift/rosa/quay-migration-job.yaml
```diff
@@ -0,0 +1,157 @@
...
+    verbs:
+    - get
+    - patch
+    - update
+  - apiGroups:
+    - ""
+    resources:
+    - namespaces
+    verbs:
+    - get
```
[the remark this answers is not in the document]
cubismod: Removed

Sunandadadi:
What triggers this job and where is it defined? Can we use quay-db-jobs for this?

deploy/openshift/rosa/quay-migration-job.yaml
cubismod: It doesn't, set to 0 retries.

cubismod:
> What triggers this job and where is it defined? Can we use quay-db-jobs for this?

This will be added as a new entry to Quay's saas file in App Interface that we will kick off ourselves for now. I'm not automating this at the moment. This isn't in DB Jobs since it relies on the Quay app code to do the migrations.

kleesc (approved):

PR #6385, opened 2026-07-06
PROJQUAY-12762: feat(referrers): implement OCI referrers API for Go registry

harishsurf:
## Summary
- Implements the OCI referrers API (`GET /v2/{name}/referrers/{digest}`) for the Go registry rewrite
- Captures `subject` and `artifactType` from manifest JSON during push via the existing middleware, stores them in the already-existing manifest table columns
- Adds a standalone HTTP handler wired before distribution's catch-all that queries referrers and returns an OCI image index response
- Enables `OCI_TEST_CONTENT_DISCOVERY=1` in the OCI conformance CI job

### Write path (manifest push)
The middleware wraps distribution's `ManifestService.Put()` — after the inner Put succeeds, it parses `subject` and `artifactType` from the manifest JSON payload and writes them to SQLite via the extended `UpsertManifest` query. No schema changes — the `subject` and `artifact_type` columns already exist with proper indexes.

### Read path (referrers query)
A standalone `ReferrersHandler` intercepts `GET /v2/{name}/referrers/{digest}` before distribution gets it. It authenticates via `BasicAuthenticator` (with anonymous pull support for public repos), resolves the repository, queries manifests where `subject = {digest}`, and returns an OCI image index. The query uses `length(manifest_bytes)` for the size field to avoid fetching full manifest content.

Key behaviors per OCI spec:
- Empty referrers → 200 with empty `manifests` array (not 404)
- `?artifactType=` query param filters results; sets `OCI-Filters-Applied: artifactType` header
- Response content type: `application/vnd.oci.image.index.v1+json`

### Files changed

| File | Change |
|------|--------|
| `internal/oci/metadata.go` | Added `Subject`, `ArtifactType` to `ManifestRecord`; added `ReferrerRecord` type and `ListReferrers` to `MetadataStore` |
| `internal/dal/queries/manifest.sql` | Extended `UpsertManifest`; added `ListReferrers` and `ListReferrersByArtifactType` queries |
| `internal/dal/daldb/manifest.sql.go` | Regenerated by sqlc |
| `internal/dal/metastore/metastore_sqlite.go` | Updated `PutManifest`; added `ListReferrers` implementation |
| `internal/registry/distribution/middleware/manifest.go` | Parses subject/artifactType from manifest JSON |
| `internal/cmd/serve.go` | Wires `ReferrersHandler` before distribution catch-all |
| `internal/registry/referrers.go` | **New** — OCI referrers HTTP handler |
| `internal/registry/referrers_test.go` | **New** — 10 handler unit tests |
| `internal/dal/metastore/metastore_sqlite_test.go` | 3 integration tests (real SQLite) |
| `internal/registry/distribution/middleware/middleware_test.go` | 2 subject-parsing tests |
| `.github/workflows/ci-go.yaml` | `OCI_TEST_CONTENT_DISCOVERY=1` |

## Test plan
- [x] 10 handler unit tests: empty list, with results, artifact type filter, OCI-Filters-Applied header, invalid digest, repo not found, path matching, multiple referrers, library namespace, nested repo names
- [x] 2 middleware tests: subject/artifactType parsing on push, no-subject leaves fields empty
- [x] 3 metastore integration tests (real SQLite): empty referrers, subject with referrer, filter by artifact type
- [x] `go test ./internal/...` — all 24 packages pass
- [x] `golangci-lint` — 0 new issues
- [ ] OCI conformance with `OCI_TEST_CONTENT_DISCOVERY=1` passes in CI


internal/registry/referrers.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+	return name[:i], name[i+1:]
+}
+
+func (h *ReferrersHandler) authorize(r *http.Request, namespace, repo string) bool {
+	if h.authenticator == nil {
+		return h.config.AnonymousAccess
+	}
+	result := h.authenticator.Authenticate(r)
+	if result.Authenticated {
```
jbpratt, line 157: Blocking: this grants referrers access to any successfully Basic-authenticated user before checking whether that user can pull this repository. That diverges from the Python v2 referrers route, which authenticates through the registry path and enforces repo read/pull permission. Impact: a valid user with no access to a private repo can list its referrer metadata, while normal Bearer-token registry clients are challenged as Basic. Action: authorize through the existing v2 registry auth path for `repository:<name>:pull`, and add tests for: valid Bearer token with pull succeeds, Basic user without repo access is denied, anonymous public access still works when enabled.

internal/registry/referrers.go
```diff
@@ -0,0 +1,192 @@
...
+	referrers, err := h.store.ListReferrers(r.Context(), repoID, dgst, artifactType)
+	if err != nil {
+		slog.Error("list referrers failed", "repository", name, "digest", dgst, "err", err) //nolint:gosec // structured logging, not injection
+		writeOCIError(w, http.StatusInternalServerError, "UNKNOWN", "internal error")
+		return
+	}
+
+	manifests := make([]ociDescriptor, 0, len(referrers))
+	for _, ref := range referrers {
+		manifests = append(manifests, ociDescriptor{
```
jbpratt, line 118: Blocking: the descriptors returned here are built from `ReferrerRecord`, but that record only carries media type, digest, size, and top-level artifact type. The OCI referrers response needs descriptor annotations from the referenced manifest/index, and OCI image manifests without a top-level `artifactType` need the config descriptor media type as the descriptor `artifactType`. Action: either persist the needed descriptor fields or parse them from `manifest_bytes` before building the index, emit `annotations,omitempty`, compute artifact type based on manifest media type, and add tests for annotations plus `?artifactType=<config media type>` filtering.

internal/registry/distribution/middleware/manifest.go
```diff
@@ -51,6 +52,22 @@ func (ms *manifestService) Put(ctx context.Context, manifest distribution.Manife
 		}
 	}
 
+	// Extract subject and artifactType from the manifest JSON if present.
```
jbpratt: Blocking: this stores `subject`, but the HTTP PUT response path still does not return `OCI-Subject: <subject digest>` after accepting a subject manifest. Clients use that header to know the referrers API is supported; without it they can fall back to the referrers tag schema even though this endpoint exists. Action: add an HTTP-level hook/wrapper around successful manifest PUTs that sets `OCI-Subject` when the manifest has a valid subject, and add an integration test for `PUT /v2/<name>/manifests/<ref>` returning 201 plus the header.

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -476,6 +478,51 @@ func (s *SQLiteStore) ListRepositories(ctx context.Context) ([]oci.RepositoryNam
 	return repos, nil
 }
 
+// ListReferrers returns manifests that reference the given subject digest.
```
jbpratt, line 483: Blocking: this only lists rows where `manifest.subject = ?`. The Python implementation also merges valid descriptors from the fallback referrers tag schema (`sha256-...`) and dedupes/applies `artifactType`. Without that, preexisting fallback-tag referrers, or referrers created by clients that fell back because `OCI-Subject` was absent, disappear from the Go API. Action: load the fallback tag/index for the subject digest, validate that descriptors still refer to this subject, merge with direct subject matches, dedupe by digest, and apply the artifactType filter to the merged set. Add a test that creates only the fallback tag schema and verifies the Go endpoint returns it.

internal/registry/referrers.go
```diff
@@ -0,0 +1,192 @@
...
+		store:         store,
+		authenticator: auth.NewBasicAuthenticator(db),
+		queries:       daldb.New(db),
+		config:        cfg,
+	}
+}
+
+// Matches reports whether the request targets the referrers endpoint.
+func (h *ReferrersHandler) Matches(r *http.Request) bool {
+	return r.Method == http.MethodGet && referrersPathRe.MatchString(r.URL.Path)
```
jbpratt: High: `Matches` requires the digest regexp to match, so `/v2/ns/repo/referrers/notadigest` is not handled here and falls through as a 404. A registry that implements the referrers API should match the referrers route first and return a 400 OCI error for invalid digest syntax. Action: loosen the route matcher to capture the trailing reference, parse with `digest.Parse` inside `ServeHTTP`, return `DIGEST_INVALID`/400 on parse failure, and update the wrapper-level test so invalid digest no longer expects 404.

internal/registry/referrers.go
```diff
@@ -0,0 +1,192 @@
...
+	repoID, err := h.store.GetRepositoryID(r.Context(), oci.RepositoryName{
+		Namespace: namespace,
+		Name:      repo,
+	})
+	if err != nil {
+		writeOCIError(w, http.StatusNotFound, "NAME_UNKNOWN", "repository name not known to registry")
+		return
+	}
+
+	artifactType := r.URL.Query().Get("artifactType")
```
jbpratt: Medium: `Query().Get("artifactType")` loses the difference between an absent filter and a present empty filter. Python treats `?artifactType=` as a filter and returns `OCI-Filters-Applied: artifactType`; this path treats it as no filter and returns all referrers. Action: check parameter presence with `r.URL.Query()["artifactType"]`, pass a nullable/pointer filter through the store layer, and set `OCI-Filters-Applied` whenever the key is present, even if the value is empty. Add a test for `?artifactType=`.

internal/registry/referrers.go
```diff
@@ -0,0 +1,192 @@
...
+
+	referrers, err := h.store.ListReferrers(r.Context(), repoID, dgst, artifactType)
```
jbpratt, line 109: High: this lists referrers without first verifying that the subject manifest exists in the repository. The Python route resolves the subject digest first and returns `MANIFEST_INVALID` when it is missing; this Go path returns an empty 200 index for an unknown subject. Action: add a subject manifest existence lookup before `ListReferrers`, allow hidden manifests as Python does, map misses to the same OCI error/status as Python, and add a test for an existing repo with a nonexistent subject digest.

internal/cmd/serve.go
```diff
@@ -107,11 +108,16 @@ func runServe(ctx context.Context, configPath, dataDir, hostname, addr, adminUse
...
+		LibraryNamespace: resolved.Config.LibraryNamespace,
+		AnonymousAccess:  featureEnabled(resolved.Config.FeatureAnonymousAccess),
+	})
+
 	mux := http.NewServeMux()
 	mux.Handle("/healthz", healthHandler(db))
 	mux.Handle("/metrics", promhttp.Handler())
 	mux.Handle("/api/", api)
-	mux.Handle("/", reg.Handler())
+	mux.Handle("/", registry.WrapWithReferrers(referrersHandler, reg.Handler()))
```
jbpratt: Medium: this installs the Go referrers interceptor unconditionally. The Python endpoint is gated by `FEATURE_REFERRERS_API`, so mixed deployments can expose the route from Go when config says it is disabled. Action: wire the feature flag into Go config and only wrap with `WrapWithReferrers` when it is enabled; otherwise mount the distribution handler directly. Add a serve/router test for both enabled and disabled cases.

internal/registry/referrers.go
```diff
@@ -0,0 +1,192 @@
...
+		w.Header().Set("OCI-Filters-Applied", "artifactType")
+	}
+	w.WriteHeader(http.StatusOK)
+	_ = json.NewEncoder(w).Encode(resp)
+}
+
+func (h *ReferrersHandler) splitName(name string) (namespace, repo string) {
+	i := strings.IndexByte(name, '/')
+	if i < 0 {
+		return h.config.LibraryNamespace, name
```
jbpratt, line 147: Medium: single-component names are always rewritten into the configured library namespace. Python's `repopath` handling only allows implicit library repos when library support is enabled. Action: carry the library-support feature into this config and reject single-component names when it is disabled, or delegate name parsing to shared logic so Go and Python route behavior stay aligned. Add tests for `FEATURE_LIBRARY_SUPPORT=false` with `/v2/repo/referrers/<digest>`.

internal/dal/queries/manifest.sql
```diff
@@ -4,9 +4,12 @@ FROM manifest
 WHERE repository_id = ? AND digest = ?;
 
 -- name: UpsertManifest :one
-INSERT INTO manifest (repository_id, digest, media_type_id, manifest_bytes)
-VALUES (?, ?, ?, ?)
-ON CONFLICT (repository_id, digest) DO UPDATE SET manifest_bytes = excluded.manifest_bytes
+INSERT INTO manifest (repository_id, digest, media_type_id, manifest_bytes, subject, artifact_type)
```
jbpratt: Medium: the Go write path now persists `subject` and `artifact_type`, but it does not set `subject_backfilled` or `artifact_type_backfilled`. Python-created manifests set both flags true after parsing these fields, so Go-created rows remain inconsistent and can be reprocessed by the backfill workers. Action: include both backfill columns in this upsert and set them true for rows parsed by Go, including the no-subject/no-artifact case. Add a metastore test that checks the flags after pushing a manifest with and without subject/artifactType.

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -132,6 +132,8 @@ func (s *SQLiteStore) PutManifest(ctx context.Context, repoID int64, m oci.Manif
 		Digest:        m.Digest.String(),
 		MediaTypeID:   mtID,
 		ManifestBytes: string(m.Content),
+		Subject:       sql.NullString{String: m.Subject.String(), Valid: m.Subject != ""},
```
jbpratt, line 137: High: once a manifest has `Subject`, the Python path protects it with hidden non-expiring tag semantics and GC treats subjects with referrers as in-use. This Go write path only stores the subject column and later only creates a tag when `m.Tag != ""`, so untagged referrer manifests are not protected the same way. Action: mirror the Python hidden-tag/lifecycle behavior for subject manifests and add coverage that an untagged subject manifest remains discoverable and is not removed by delete/GC flows that should only expire visible tags.

jbpratt (changes requested):
Requesting changes. The blocking issues are authorization, OCI referrers semantics, fallback tag-schema compatibility, and Python parity during migration. Inline comments include the concrete fix/test expected for each issue. Broad package-boundary feedback is intentionally left out for separate discussion.

internal/cmd/serve.go
```diff
@@ -107,11 +108,16 @@ func runServe(ctx context.Context, configPath, dataDir, hostname, addr, adminUse
 		return 1
 	}
 
+	referrersHandler := registry.NewReferrersHandler(db, store, registry.ReferrersConfig{
+		LibraryNamespace: resolved.Config.LibraryNamespace,
+		AnonymousAccess:  featureEnabled(resolved.Config.FeatureAnonymousAccess),
+	})
```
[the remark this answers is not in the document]
harishsurf: Fixed — now uses `nil || *ptr` pattern matching distribution's `anonymousAccessEnabled()` so both handlers agree when the flag is unset.

internal/registry/distribution/middleware/manifest.go
```diff
@@ -51,6 +52,22 @@ func (ms *manifestService) Put(ctx context.Context, manifest distribution.Manife
...
+	}
+	if json.Unmarshal(payload, &subj) == nil {
+		if subj.Subject != nil && subj.Subject.Digest != "" {
+			if d, err := digest.Parse(subj.Subject.Digest); err == nil {
+				record.Subject = d
+			}
+		}
+		record.ArtifactType = subj.ArtifactType
+	}
+
```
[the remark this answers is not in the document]
harishsurf: Fixed — the unmarshal struct now includes `Config.MediaType`. After parsing, if `artifactType` is empty and `config.mediaType` is set, we use the config value as fallback per OCI spec.

internal/registry/referrers_test.go
```diff
@@ -0,0 +1,313 @@
...
+		config: ReferrersConfig{LibraryNamespace: "library", AnonymousAccess: true},
+	}
+}
+
+func TestReferrers_EmptyList(t *testing.T) {
+	store := &mockStore{repoID: 1}
+	handler := newTestHandler(store)
+
+	subjectDgst := digest.FromString("subject-manifest")
+	req := httptest.NewRequest(http.MethodGet, "/v2/ns/repo/referrers/"+subjectDgst.String(), http.NoBody)
```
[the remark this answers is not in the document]
harishsurf: Already fixed in commit 56342d0.

internal/registry/referrers.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	repoID, err := h.store.GetRepositoryID(r.Context(), oci.RepositoryName{
+		Namespace: namespace,
+		Name:      repo,
+	})
+	if err != nil {
+		writeOCIError(w, http.StatusNotFound, "NAME_UNKNOWN", "repository name not known to registry")
+		return
+	}
```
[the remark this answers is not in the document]
harishsurf: Fixed — now uses `errors.Is(err, oci.ErrNotExist)` to return 404 only for not-found, 500 for everything else. Added `TestReferrers_RepoLookupError` to cover the generic error path.

internal/registry/referrers.go
harishsurf: The Go rewrite's current auth model has no per-repo ACLs. Distribution's own `Authorized()` method (auth.go) does the same thing — authenticates the user without checking repo-level permissions. PR #6271 (merged) established this pattern. The only repo-level check is for anonymous pulls on public repos, which our handler already does. This is a known limitation of the Go rewrite, not specific to the referrers handler.

internal/registry/referrers.go
harishsurf: Fixed — `ListReferrers` SQL now selects `manifest_bytes`, annotations are parsed from the JSON at query time via `parseAnnotations()`. The `config.mediaType` fallback is handled at write time in the middleware. Verified with OCI conformance test (74/74 pass).

internal/registry/distribution/middleware/manifest.go
harishsurf: Fixed — added `SubjectHeaderMiddleware` that wraps distribution's handler. The middleware stores the subject digest in a mutable context holder during `ManifestService.Put()`, and the HTTP wrapper injects the `OCI-Subject` header before the response status is written. This was the root cause of all 4 OCI conformance failures (Ginkgo aborted setup after the first missing header check).

internal/dal/metastore/metastore_sqlite.go
harishsurf: Fixed — `ListReferrers` now calls `fallbackTagReferrers()` after the subject-column query. It looks up the tag `sha256-<encoded>`, parses the manifest as an OCI index, extracts descriptors, deduplicates by digest against the SQL results, and applies the `artifactType` filter.

internal/registry/referrers.go
harishsurf: Fixed — `Matches()` now uses a loose regex (`/v2/.+/referrers/.+`) to catch all referrers paths. `ServeHTTP()` validates with the strict digest regex and returns 400 `DIGEST_INVALID` when it doesn't match.

internal/registry/referrers.go
harishsurf: Fixed — now uses `r.URL.Query()["artifactType"]` to check parameter presence. `OCI-Filters-Applied` header is set whenever the parameter key is present, even if the value is empty.

internal/registry/referrers.go
harishsurf: Intentionally not fixing. The OCI conformance test (v1.1.1) pushes a referrer pointing to a subject that was never pushed, then expects `GET /v2/{name}/referrers/{subject-digest}` to return 200 with that referrer. Adding a subject existence check would break conformance. Both behaviors are OCI-spec-legal — we chose the conformance-compatible path.

internal/cmd/serve.go
harishsurf: Fixed — added `FEATURE_REFERRERS_API` (default `true`) to config. When disabled, the referrers handler is not registered and requests fall through to distribution.

internal/registry/referrers.go
harishsurf: Fixed — added `FEATURE_LIBRARY_SUPPORT` (default `true`) to config. When disabled, single-component repo names are rejected (returns 404 `NAME_UNKNOWN`) instead of being rewritten to the library namespace.

internal/dal/queries/manifest.sql
harishsurf: Fixed — `UpsertManifest` now includes `subject_backfilled = 1, artifact_type_backfilled = 1` in both the INSERT and ON CONFLICT UPDATE clauses.

internal/dal/metastore/metastore_sqlite.go
harishsurf: Will be addressed in GC PR #6318. That PR's `FindOrphanedManifests` query already protects subject targets (`NOT EXISTS ... WHERE m2.subject = m.digest`), but the referrer manifests themselves need hidden non-expiring tags. A prompt has been prepared for the GC PR session to add this protection.

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -476,6 +480,121 @@ func (s *SQLiteStore) ListRepositories(ctx context.Context) ([]oci.RepositoryNam
...
+		RepositoryID: repoID,
+		Name:         tagName,
+	})
+	if err != nil {
+		return nil
+	}
+	content, err := q.GetManifestContentByDigest(ctx, tagDigest)
+	if err != nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
harishsurf: Fixed — `fallbackTagReferrers` now returns an error. `sql.ErrNoRows` returns `(nil, nil)` (tag absent). All other errors propagate to `ListReferrers` which wraps and returns them. Malformed JSON in the fallback index is treated as absent (with a nolint annotation explaining why).

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -476,6 +480,121 @@ func (s *SQLiteStore) ListRepositories(ctx context.Context) ([]oci.RepositoryNam
...
+		out = append(out, oci.ReferrerRecord{
+			Digest:       m.Digest,
+			MediaType:    m.MediaType,
+			ArtifactType: m.ArtifactType,
+			Size:         m.Size,
+			Annotations:  m.Annotations,
+		})
+	}
+	return out
+}
```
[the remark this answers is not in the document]
harishsurf: Fixed — added `TestListReferrers_FallbackTagSchema` that seeds a `sha256-<digest>` tag with an OCI index manifest, then verifies `ListReferrers` returns the descriptor from it with correct digest, artifactType, and annotations.

jbpratt (approved):

PR #6392, opened 2026-07-07
NO-ISSUE: feat(auth): add robot token authentication

jbpratt:
## Summary
Add static robot token authentication to the Go rewrite while preserving compatibility with robot tokens created by the existing Python/OMR registry.

## Related Issue
NO-ISSUE: Go rewrite parity work for migrated robot credentials.

## Changes
- Split HTTP Basic parsing from credential verification and add a database-backed verifier for users and static robot tokens.
- Add Python encrypted-field compatibility for legacy `RobotAccountToken` rows, including `DATABASE_SECRET_KEY` conversion and AES-CCM decryption without a Pion dependency.
- Enforce repository pull, push, delete, and create-on-push authorization for user and robot principals in the Go distribution path.
- Add pull/push/create repository authz queries and tests, including disabled namespace and missing namespace handling.
- Extend the Go OMR migration E2E job to create a robot through old OMR's session+CSRF API flow, verify it before migration, and reuse the migrated token against Go.

## Testing
- [x] `go test ./internal/credentials/encryptedfield/... ./internal/auth ./internal/repository/dal ./internal/registry/distribution ./internal/config -count=1`
- [x] `go test ./internal/... -count=1`
- [x] `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci-go.yaml")'`
- [x] Extracted `e2e-migrate` workflow run blocks and checked them with `bash -n`
- [x] `git diff --check`
- [x] `git diff --exit-code internal/dal/daldb`

## Notes
- Draft PR for review.
- The migration E2E stores robot token material under `/tmp/e2e-migrate-secrets`, outside uploaded diagnostics.
- JWT/federated robot tokens remain out of scope for this change.

### DIFF
--- a/internal/auth/robot_static_token.go
+++ b/internal/auth/robot_static_token.go
@@ -1,0 +1,192 @@
+package auth
+
+import (
+	"context"
+	"crypto/subtle"
+	"database/sql"
+	"log/slog"
+	"time"
+
+	"github.com/quay/quay/internal/credentials/encryptedfield"
+	"github.com/quay/quay/internal/dal/daldb"
+)
+
+// dummyEncryptedRobotValue keeps early robot auth failures on the decrypt/compare path.
+const dummyEncryptedRobotValue = "v0$$XTxqlz/Kw8s9WKw+GaSvXFEKgpO/a2cGNhvnozzkaUh4C+FgHqZqnA==" // #nosec G101 -- encrypted dummy value, not a credential.
+
+type robotVerifier struct {
+	static *staticRobotTokenVerifier
+}
+
+func newRobotVerifier(db *sql.DB, cfg DatabaseVerifierConfig) Verifier {
+	if db == nil {
+		return nil
+	}
+
+	return &robotVerifier{static: newStaticRobotTokenVerifier(db, cfg)}
+}
+
+func (v *robotVerifier) Verify(ctx context.Context, creds Credentials) Result {
+	if v == nil || v.static == nil {
+		return failedRobotResult(creds.Username)
+	}
+
+	return v.static.Verify(ctx, creds)
+}
+
+type staticRobotTokenVerifier struct {
+	queries                        *daldb.Queries
+	databaseSecretKey              string
+	robotsDisallow                 bool
+	robotsWhitelist                map[string]struct{}
+	featureUserLastAccessed        bool
+	lastAccessedUpdateThresholdSec int
+}
+
+func newStaticRobotTokenVerifier(db *sql.DB, cfg DatabaseVerifierConfig) *staticRobotTokenVerifier {
+	whitelist := make(map[string]struct{}, len(cfg.RobotsWhitelist))
+	for _, username := range cfg.RobotsWhitelist {
+		whitelist[username] = struct{}{}
+	}
+
+	return &staticRobotTokenVerifier{
+		queries:                        daldb.New(db),
+		databaseSecretKey:              cfg.DatabaseSecretKey,
+		robotsDisallow:                 cfg.RobotsDisallow,
+		robotsWhitelist:                whitelist,
+		featureUserLastAccessed:        cfg.FeatureUserLastAccessed,
+		lastAccessedUpdateThresholdSec: cfg.LastAccessedUpdateThresholdSec,
+	}
+}
+
+func (v *staticRobotTokenVerifier) Verify(ctx context.Context, creds Credentials) Result {
+	if v == nil || v.queries == nil {
+		return failedRobotResult(creds.Username)
+	}
+
+	owner, ok := v.validateRequest(creds)
+	if !ok {
+		v.dummySecretMatch(creds)
+		return failedRobotResult(creds.Username)
+	}
+
+	robot, ok := v.enabledRobot(ctx, creds.Username)
+	if !ok {
+		v.dummySecretMatch(creds)
+		return failedRobotResult(creds.Username)
+	}
+	if !v.ownerEnabled(ctx, owner, creds.Username) {
+		v.dummySecretMatch(creds)
+		return failedRobotResult(creds.Username)
+	}
+	if !v.secretMatches(ctx, robot.ID, creds) {
+		return failedRobotResult(creds.Username)
+	}
+
+	v.updateLastAccessed(ctx, robot.ID, robot.LastAccessed, creds.Username)
+
+	return Result{
+		Principal: Principal{
+			ID:       robot.ID,
+			UUID:     robot.Uuid,
+			Username: robot.Username,
+			Email:    robot.Email,
+			Kind:     PrincipalRobot,
+		},
+		Username:      creds.Username,
+		Presented:     true,
+		Authenticated: true,
+	}
+}
+
+func (v *staticRobotTokenVerifier) validateRequest(creds Credentials) (string, bool) {
+	owner, _, ok := parseRobotUsername(creds.Username)
+	if !ok {
+		slog.Debug("robot authentication failed", "username", creds.Username, "reason", "invalid robot username")
+		return "", false
+	}
+	if v.databaseSecretKey == "" {
+		slog.Debug("robot authentication failed", "username", creds.Username, "reason", "missing database secret key")
+		return "", false
+	}
+	if !isASCII(creds.Secret) {
+		slog.Debug("robot authentication failed", "username", creds.Username, "reason", "non-ASCII secret")
+		return "", false
+	}
+	if v.robotsDisallow {
+		if _, ok := v.robotsWhitelist[creds.Username]; !ok {
+			slog.Debug("robot authentication failed", "username", creds.Username, "reason", "robot disallowed")
+			return "", false
+		}
+	}
+	return owner, true
+}
+
+func (v *staticRobotTokenVerifier) enabledRobot(ctx context.Context, username string) (daldb.GetRobotByUsernameRow, bool) {
+	robot, err := v.queries.GetRobotByUsername(ctx, username)
+	if err != nil || !robot.Enabled {
+		slog.Debug("robot authentication failed", "username", username, "reason", "robot missing or disabled")
+		return daldb.GetRobotByUsernameRow{}, false
+	}
+	return robot, true
+}
+
+func (v *staticRobotTokenVerifier) ownerEnabled(ctx context.Context, owner, username string) bool {
+	ownerUser, err := v.queries.GetNamespaceUserByUsername(ctx, owner)
+	if err != nil || !ownerUser.Enabled {
+		slog.Debug("robot authentication failed", "username", username, "reason", "owner missing or disabled")
+		return false
+	}
+	return true
+}
+
+func (v *staticRobotTokenVerifier) secretMatches(ctx context.Context, robotID int64, creds Credentials) bool {
+	encryptedToken, err := v.queries.GetRobotTokenByRobotID(ctx, robotID)
+	if err != nil {
+		slog.Debug("robot authentication failed", "username", creds.Username, "reason", "token row missing")
+		return false
+	}
+
+	token, err := encryptedfield.Decrypt(v.databaseSecretKey, encryptedToken)
+	if err != nil {
+		slog.Debug("robot authentication failed", "username", creds.Username, "reason", "decrypt failed")
+		return false
+	}
+	if subtle.ConstantTimeCompare([]byte(token), []byte(creds.Secret)) != 1 {
+		slog.Debug("robot authentication failed", "username", creds.Username, "reason", "token mismatch")
+		return false
+	}
+	return true
+}
+
+func (v *staticRobotTokenVerifier) dummySecretMatch(creds Credentials) {
+	token, _ := encryptedfield.Decrypt(v.databaseSecretKey, dummyEncryptedRobotValue)
+	subtle.ConstantTimeCompare([]byte(token), []byte(creds.Secret))
+}
+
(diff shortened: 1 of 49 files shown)


internal/repository/dal/authz.go
```diff
@@ -57,3 +61,139 @@ func (a *Authorizer) CanAdminRepository(ctx context.Context, principal *auth.Pri
...
 }
+
+// CanPullRepository reports whether principal can pull repo contents.
+func (a *Authorizer) CanPullRepository(ctx context.Context, principal *auth.Principal, repo *repository.Repository) (bool, error) {
+	if repo == nil {
+		return false, nil
+	}
+	if !repo.NamespaceEnabled {
+		return false, nil
+	}
```
harishsurf, line 72: **CRITICAL: `CanPullRepository` skips repo state check — deleted repos remain pullable by authenticated users**

Checks `NamespaceEnabled` and `Visibility` but never checks `repo.State`. For public repos, returns `true` immediately when `Visibility == "public"` regardless of state. `CanPushRepository` properly gates on state via a switch block, but `CanPullRepository` has no equivalent.

Admin marks a public repo for deletion (state=3). Anonymous pulls correctly return 404 (the SQL excludes state=3). But any authenticated user can still pull because `CanPullRepository` sees `Visibility=public, NamespaceEnabled=true` and returns true.

**Fix:**
```go
if repo.State == repository.StateMarkedForDeletion {
    return false, nil
}
```

internal/credentials/encryptedfield/secret.go
```diff
@@ -0,0 +1,66 @@
...
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"math/big"
+
+	"github.com/google/uuid"
+)
+
+const aesKeySize = 32
+
```
harishsurf, line 13: **CRITICAL: Decimal `DATABASE_SECRET_KEY` values like "256" produce wrong AES key — breaks robot token decrypt after migration**

`fmt.Sprintf("%02x", big.Int(256))` produces `"100"` (3 chars, odd-length). `hex.DecodeString("100")` fails with "odd length hex string". Code falls through to the raw-string branch, producing bytes `[0x32, 0x35, 0x36]` (ASCII of "256") instead of Python's `[0x01, 0x00]`.

Verified with Go:
```
256   -> hex="100"   len=3 decode_err=odd length hex string  ← WRONG KEY
65536 -> hex="10000" len=5 decode_err=odd length hex string  ← WRONG KEY
255   -> hex="ff"    len=2 decode_err=<nil>                  ← correct
```

Python pads odd-length hex with a leading zero: `"100"` → `"0100"` → `[0x01, 0x00]`. Go doesn't.

Any installation using a numeric `DATABASE_SECRET_KEY` that produces odd-length hex will fail to decrypt migrated `robotaccounttoken` rows, silently breaking robot auth after migration.

**Fix:**
```go
hexText := fmt.Sprintf("%02x", i)
if len(hexText)%2 != 0 {
    hexText = "0" + hexText
}
```

internal/registry/referrers.go
```diff
@@ -43,15 +48,25 @@ type ReferrersHandler struct {
...
+		authenticator: auth.NewBasicAuthenticator(auth.NewDatabaseVerifier(db, auth.DatabaseVerifierConfig{
+			DatabaseSecretKey:              config.DatabaseSecretKey,
+			RobotsDisallow:                 config.RobotsDisallow,
+			RobotsWhitelist:                config.RobotsWhitelist,
+			FeatureUserLastAccessed:        config.FeatureUserLastAccessed,
+			LastAccessedUpdateThresholdSec: config.LastAccessedUpdateThresholdSeconds,
+		})),
+		queries: daldb.New(db),
+		config:  config,
 	}
```
harishsurf, line 79: **CRITICAL: Referrers endpoint appears to skip repository pull authorization**

This PR adds robot token support to the referrers handler's authenticator but does not add repository-level authorization (`CanPullRepository` / `authorizeRepositoryAccess`). If the base `authorize()` method only checks authentication (valid credentials) without checking pull permission on the specific repo, any authenticated user or robot can call `GET /v2/<private-repo>/referrers/<digest>` and enumerate referrer manifests on repos they have no access to.

The distribution path (`Authorized()`) now has `authorizeDistributionAccess` for pull/push/delete checks, but the referrers handler bypasses distribution entirely — it's a standalone mux handler. Please verify the base `authorize()` checks repo-scoped pull permission, not just valid credentials.

internal/migrate/copy.go
```diff
@@ -74,6 +82,55 @@ func (m *Migrator) copyData(ctx context.Context) error {
...
+		}
+		return fmt.Errorf("stat source config: %w", err)
+	}
+
+	sourceCfg, err := config.Load(sourcePath)
+	if err != nil {
+		return fmt.Errorf("load source config: %w", err)
+	}
+
+	runtimeCfg := map[string]any{
```
harishsurf, line 100: **MAJOR: `writeRuntimeConfig` drops `SUPER_USERS` and `FEATURE_SUPER_USERS_FULL_ACCESS` from migrated config**

Copies `DATABASE_SECRET_KEY`, `ROBOTS_DISALLOW`, `FEATURE_USER_LAST_ACCESSED`, etc. but does NOT copy `SUPER_USERS`, `FEATURE_SUPER_USERS`, or `FEATURE_SUPER_USERS_FULL_ACCESS`. After migration, `newDefaultConfig()` sets `FeatureSuperUsers: true` but with an empty superuser list and `FullAccess=false`.

A source OMR with `SUPER_USERS: ["ops-admin"]` and `FEATURE_SUPER_USERS_FULL_ACCESS: true` silently loses all superuser privileges after migration. The admin discovers this only when they try to manage repos and get "forbidden."

**Fix:** Add these fields to the config propagation.

internal/auth/robot_static_token.go
```diff
@@ -0,0 +1,181 @@
...
+	if v == nil || v.static == nil {
+		return failedRobotResult(creds.Username)
+	}
+
+	return v.static.Verify(ctx, creds)
+}
+
+type staticRobotTokenVerifier struct {
+	queries                        *daldb.Queries
+	databaseSecretKey              string
```
harishsurf, line 39: **MAJOR: Robot token verification timing oracle enables robot account enumeration**

Unlike `userPasswordVerifier` which always runs `bcrypt.CompareHashAndPassword` against a `dummyHash` for non-existent users to maintain constant timing, the robot verifier short-circuits at multiple points: invalid username format → missing/disabled robot → missing/disabled owner → decrypt+compare. Each exit path has measurably different response times.

An attacker can observe timing differences to enumerate valid robot account names (valid names proceed through DB lookups and decryption, invalid ones return immediately).

**Fix:** Add a dummy decrypt+compare fallback for early rejection paths, analogous to the `dummyHash` pattern in `userPasswordVerifier`. Always perform the full verification path and discard the result on failure.

internal/dal/queries/repository.sql
```diff
@@ -107,6 +108,131 @@ SELECT EXISTS(
     AND tr.name = 'admin'
 );
 
+-- name: UserCanPullRepository :one
+SELECT EXISTS(
+  SELECT 1
+  FROM repository r
+  JOIN "user" ns ON r.namespace_user_id = ns.id
```
harishsurf, line 115: **MAJOR:** Pull/push SQL queries lack `ns.enabled` check on namespace-owner branch. The first UNION branch checks `ns.username = @username` but not `AND ns.enabled = 1`. Go layer has the pre-check so not currently exploitable, but lacks defense-in-depth. Compare with `UserCanCreateRepositoryInNamespace` which correctly checks `ns.enabled`. Fix: add `AND ns.enabled = 1` to the first branch of both `UserCanPullRepository` and `UserCanPushRepository`.

internal/auth/robot_static_token.go
```diff
@@ -0,0 +1,181 @@
...
+	if !isASCII(creds.Secret) {
+		slog.Debug("robot authentication failed", "username", creds.Username, "reason", "non-ASCII secret")
+		return "", false
+	}
+	if v.robotsDisallow {
+		if _, ok := v.robotsWhitelist[creds.Username]; !ok {
+			slog.Debug("robot authentication failed", "username", creds.Username, "reason", "robot disallowed")
+			return "", false
+		}
+	}
```
harishsurf, line 121: **MAJOR:** `updateLastAccessed()` performs a pooled write via `v.queries.UpdateUserLastAccessedIfOlder()` without opening a transaction. Under `MaxOpenConns=1`, this contends with concurrent push middleware transactions. Failures are logged at Debug level and swallowed, so robot auth still succeeds but `last_accessed` timestamps silently stop advancing under write contention.

internal/registry/distribution/auth.go
```diff
@@ -83,6 +117,13 @@ func (ac *accessController) Authorized(req *http.Request, access ...distauth.Acc
 		}
 	}
 
+	if err := ac.authorizeDistributionAccess(req, &result.Principal, access); err != nil {
+		return nil, &challenge{
+			realm: ac.realm,
+			err:   distauth.ErrAuthenticationFailure,
```
harishsurf, line 123: **MAJOR:** Authorization failures are returned as `distauth.ErrAuthenticationFailure` (401 challenge) instead of a 403. A correctly authenticated user who lacks push permission gets "authentication failed" — the client retries credentials indefinitely thinking they are wrong. This is an inherent limitation of distribution's auth model (no 403 path) and matches Python Quay's behavior, but worth noting for user-facing debugging.

internal/repository/dal/authz.go
jbpratt: Fixed in `e820c49cd`: `CanPullRepository` now rejects `StateMarkedForDeletion` before the public-visibility allow path. Added an authz regression for a deleted public repo.

internal/registry/referrers.go
jbpratt: Fixed in `e820c49cd`: the referrers handler now builds a repository authorizer, carries superuser config, and calls `CanPullRepository` after successful authentication. Added referrers tests for valid credentials denied without pull permission and allowed with read permission.

internal/dal/queries/repository.sql
jbpratt: Fixed in `e820c49cd`: added `ns.enabled = 1` to the namespace-owner branch of both `UserCanPullRepository` and `UserCanPushRepository`, with a regression covering disabled namespace owner access.

internal/credentials/encryptedfield/secret.go
jbpratt: Leaving this unchanged for Python compatibility. Current Python `convert_secret_key` does not pad odd hex before `bytearray.fromhex`; for `"256"`, `fromhex("100")` raises and Python falls through to raw string bytes. The existing Python test asserts `256 -> b"256..."`, so changing Go to `[0x01, 0x00]` would make Go incompatible with existing Python-encrypted `RobotAccountToken` rows for those secrets.

internal/migrate/copy.go
jbpratt: Fixed in `e820c49cd`: `writeRuntimeConfig` now preserves `SUPER_USERS`, `FEATURE_SUPER_USERS`, and `FEATURE_SUPERUSERS_FULL_ACCESS`, with `copyData` coverage for the migrated runtime config.

jbpratt:
Addressed the latest CodeRabbit review-body nit in `e820c49cd`: `internal/repository/dal/store_test.go` now covers disabled namespace, enabled namespace, and missing namespace `Store.Get` behavior, including `NamespaceEnabled` and `ErrNotFound` assertions.

internal/auth/robot_static_token.go
jbpratt: Fixed in `e820c49cd`: early robot auth rejection paths now run a dummy encrypted-token decrypt plus constant-time compare before returning, so invalid format, missing robot, and disabled owner paths stay closer to the normal verification path.

internal/credentials/encryptedfield/secret.go
```diff
@@ -0,0 +1,66 @@
...
+	return key, nil
+}
+
+func secretSourceBytes(secret string) ([]byte, error) {
+	if i, ok := new(big.Int).SetString(secret, 10); ok {
+		hexText := fmt.Sprintf("%02x", i)
+		if source, err := hex.DecodeString(hexText); err == nil {
+			return source, nil
+		}
+	}
```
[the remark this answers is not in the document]
jbpratt: Leaving this unchanged intentionally. The current Python implementation uses `bytearray.fromhex("{:02x}".format(big_int))`, and `util/security/test/test_secret.py` explicitly expects `"256"` to fall through to raw string bytes. Matching that behavior is required for migrated token compatibility; changing Go to `i.Bytes()` would break installs whose Python tokens were encrypted with that legacy fallback.

internal/registry/distribution/auth.go
jbpratt: No code change here. Distribution auth returns challenge-style failures for denied scopes, and that matches the existing Python v2 behavior. Treating this as a user-facing debugging caveat rather than changing protocol behavior in this PR.

internal/auth/robot_static_token.go
jbpratt: No code change here. This is a best-effort post-auth timestamp write outside any caller-owned transaction; adding a standalone transaction would still contend with SQLite single-writer behavior and would lengthen the auth path. The code already skips fresh timestamps in Go and keeps the SQL predicate as a race guard, while keeping auth availability independent from this telemetry update.

harishsurf (approved):

PR #6443, opened 2026-07-08
PROJQUAY-12196: fix(buildman): fire build_start notification on phase transition

LiZhang19817:
## Summary
- The `BuildStartEvent` notification event was defined in `notifications/notificationevent.py` and configurable via the UI, but never actually triggered when a build starts
- Added `send_notification("build_start")` in `buildman/manager/ephemeral.py` when build phase transitions to `BUILDING`
- Issue confirmed on master, redhat-3.17, and redhat-3.16 branches

## Test plan
- [ ] Configure a repository notification with event type "Image build started" and method "Webhook POST"
- [ ] Trigger a build on the repository
- [ ] Verify the webhook receives a POST request when the build transitions to the BUILDING phase
- [ ] Verify existing build_success and build_failure notifications still work correctly

Jira: https://redhat.atlassian.net/browse/PROJQUAY-12196

### DIFF
--- a/buildman/manager/ephemeral.py
+++ b/buildman/manager/ephemeral.py
@@ -491,6 +491,9 @@ class EphemeralBuilderManager(BuildStateInterface):
                 build_job.build_uuid, phase, self._build_logs.PHASE, phase_metadata
             )
 
+            if phase == BUILD_PHASE.BUILDING:
+                build_job.send_notification("build_start")
+
         # Check if on_job_complete needs to be called
         if updated and phase in EphemeralBuilderManager.COMPLETED_PHASES:
             executor_name = job_data_json.get("executor_name")
--- a/web/playwright/e2e/repository/notifications.spec.ts
+++ b/web/playwright/e2e/repository/notifications.spec.ts
@@ -890,4 +890,46 @@ test.describe('Repository Notifications', {tag: ['@repository']}, () => {
       }
     },
   );
+
+  test(
+    'build_start notification delivered via webhook',
+    {tag: ['@feature:BUILD_SUPPORT', '@PROJQUAY-12196']},
+    async ({api}) => {
+      test.setTimeout(300_000);
+      const org = await api.organization('buildstartwh');
+      const repo = await api.repository(org.name, 'buildstartrepo');
+
+      const receiver = new WebhookReceiver();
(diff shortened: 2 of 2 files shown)


LiZhang19817:
The `e2e test coverage check` failure is a false positive — this PR is a backend-only change (3 lines added to `buildman/manager/ephemeral.py`) that wires up the existing `build_start` notification event when a build transitions to the `BUILDING` phase. There is no frontend change and no meaningful Playwright e2e test can be written for this server-side notification trigger.

LiZhang19817:
## Code Review Summary

Performed a deep review of the fix. **No regression risks found.**

### Correctness

- `BuildJob.send_notification("build_start")` (`buildjob.py:200-241`) is a generic method that accepts any event `kind`. The `event_data` it builds (`build_id`, `build_name`, `docker_tags`, `trigger_id`, `trigger_kind`, `trigger_metadata`) is fully populated at this point — no extra args needed for `build_start`.
- `BuildStartEvent` inherits from `BaseBuildEvent` (`notificationevent.py:254-279`), whose `should_perform()` checks an optional `ref-regex` filter against `trigger_metadata.ref`. The `trigger_metadata` is present in the event data, so filtering works correctly.

### No duplicate notification risk

- `update_phase_then_close()` (`data/model/build.py:209-229`) uses an atomic SQL `UPDATE ... WHERE phase == current_phase` and returns `True` only if the row was actually updated. A repeated `BUILDING` phase transition returns `False`, so the notification won't fire twice.
- `update_job_phase()` also has an early return at line 482 if the build is already in the desired phase.

### Error handling is safe

- `spawn_notification()` queues work to the notification queue. The build phase has already been persisted by `update_phase_then_close()` before our code runs, so a notification failure cannot break the build flow.

### No impact on other notifications

- `build_queued` is fired in `endpoints/building.py:149` via a separate code path — unaffected.
- `build_success` and `build_failure` are fired in `on_job_complete()` — unaffected.

jbpratt:
@LiZhang19817 you should be able to add a playwright test for this with the build infra and the webhook receiver fixtures

LiZhang19817:
@bcaton85 @jbpratt Pls help review and merge this fix.

jbpratt (approved):

PR #6453, opened 2026-07-08
PROJQUAY-12193: feat(oauth): track token last-used metadata

shaonrh:
- Update OAuth token `last_accessed` after successful non-expired validation
- Use OAuth-specific debounce config independent of user `last-accessed` tracking
- Preserve expired-token return behavior while skipping `last_accessed` writes

### DIFF
--- a/config.py
+++ b/config.py
@@ -755,6 +755,10 @@ class DefaultConfig(ImmutableConfig):
     # token will be updated after the previous update.
     LAST_ACCESSED_UPDATE_THRESHOLD_S = 60
 
+    # Defines the delay required (in seconds) before the last_accessed field of an OAuth access
+    # token will be updated after the previous update.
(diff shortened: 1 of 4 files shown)


shaonrh:
**Impact:**
- Quay and quay.io will start writing `oauthaccesstoken.last_accessed` after successful, non-expired OAuth token validation.
 - DB write load increases: first successful use writes immediately, then at most once per token per 60 seconds by default. This should be negligible.
 - Token metadata views will start showing `last_accessed` for tokens used after deployment; unused tokens stay null.
 - Existing tokens not used after deploy will still show null/never used.

**Not impacted:**
 - No immediate UI/API-visible change unless something already exposes `OAuthAccessToken.last_accessed`.
 - OAuth auth behavior should remain the same for users.
 - Once token-listing API/UI work reads this field, quay.io users/admins will see recently used OAuth tokens with populated `last_accessed`.

jbpratt (approved):

PR #6455, opened 2026-07-08
NO-ISSUE: test(playwright): enforce required usage tags

jbpratt:
## Summary
Generalizes the Playwright usage-tag checker so it can enforce multiple required tags instead of only `@superuser`. The checker is now rule-driven and includes `@webhook` coverage for tests that use the webhook receiver fixture or instantiate `WebhookReceiver` directly.

## Related Issue
NO-ISSUE: Playwright test metadata/tooling cleanup from agent research.

## Changes
- Rename `web/playwright/ensure-superuser-tags.cjs` to `web/playwright/ensure-required-tags.cjs`.
- Drive required-tag enforcement from `REQUIRED_TAG_RULES`, with fixture dependency tracing plus constructor/function helper matchers for future extension.
- Keep `@superuser` enforcement for superuser-backed fixtures and local fixtures backed by them.
- Add `@webhook` enforcement for the `webhook` fixture and direct `WebhookReceiver` usage.
- Backfill `@webhook` on existing webhook receiver tests.
- Update package scripts, web CI, and the Playwright migration guide to use the generalized required-tags command.

## Testing
- [x] `NODE_PATH=/var/home/bpratt/git/quay/quay/web/node_modules node --check web/playwright/ensure-required-tags.cjs` passes
- [x] `NODE_PATH=/var/home/bpratt/git/quay/quay/web/node_modules node playwright/ensure-required-tags.cjs` passes
- [x] `NODE_PATH=/var/home/bpratt/git/quay/quay/web/node_modules npm run test:e2e:check-required-tags` passes
- [x] `git diff --check` passes
- [x] Commit hooks passed for secrets, config checks, ESLint, whitespace, and EOF
- [ ] Not run: full Playwright suite; this is a metadata/checker change and the targeted checker validation passed.

## Notes
- The `npm-lockfile-sync` hook was skipped for the new commit because, in this worktree, it rewrote the root `package-lock.json` into the wrong lockfile shape. Package scripts changed, but dependencies did not.
- Backport not expected.


web/playwright/ensure-required-tags.cjs
```diff
@@ -0,0 +1,471 @@
...
+}
+
+const webRoot = path.resolve(__dirname, '..');
+const testRoot = path.join(webRoot, 'playwright', 'e2e');
+
+function listSpecFiles(dir) {
+  const files = [];
+  for (const entry of fs.readdirSync(dir, {withFileTypes: true})) {
+    const fullPath = path.join(dir, entry.name);
+    if (entry.isDirectory()) {
```
deshpandevlab, line 47: Bug: `.spec.tsx` files are silently ignored.

The `endsWith('.spec.ts')` check here will miss any `.spec.tsx` files. The `FIXTURE_SOURCE_EXTENSIONS` set includes `.tsx`, but it's never reached for those files because this condition fails first.

Consider using a regex match instead:

```js
entry.name.match(/\.spec\.tsx?$/) &&
FIXTURE_SOURCE_EXTENSIONS.has(path.extname(entry.name))
```

This way future `.spec.tsx` test files will also be covered by the checker.

web/playwright/ensure-superuser-tags.cjs
```diff
@@ -0,0 +1,471 @@
...
+
+const fs = require('fs');
+const path = require('path');
+const ts = require('typescript');
+
+const SUPERUSER_TAG = '@superuser';
+const SUPERUSER_FIXTURES = new Set([
+  'superuserApi',
+  'superuserContext',
+  'superuserPage',
```
deshpandevlab: Question: Should `freshUser` always imply `@superuser`?

`freshUser` is unconditionally included in `SUPERUSER_FIXTURES` here. Is this fixture always backed by superuser context (i.e., it always uses admin/superuser APIs to create the user)?

If there's ever a variant that uses self-registration instead of admin creation, this would produce false positives (requiring `@superuser` on tests that don't actually need superuser access). Just want to confirm this is intentionally always superuser-backed by design.

web/playwright/ensure-required-tags.cjs
jbpratt: Fixed in `1f0e6901b` by switching the spec filename check to `SPEC_FILE_PATTERN` (`/\.spec\.tsx?$/`) while keeping the extension filter, so future `.spec.tsx` files are scanned.

web/playwright/ensure-superuser-tags.cjs
jbpratt: Yes, `freshUser` is intentionally superuser-backed today. The fixture destructures `superuserApi`, creates the account with `superuserApi.user("iso")`, then calls `superuserApi.raw.updateUserAsSuperuser(...)` before signing in as the new user. So treating it as `@superuser` matches the current fixture contract unless/until we add a separate self-registration fixture.

LiZhang19817 (approved):

PR #6461, opened 2026-07-09
PROJQUAY-12194: feat(api): add organization oauth token lifecycle

shaonrh:
- `GET /api/v1/organization/<orgname>/applications/<client_id>/tokens`: list app tokens.
- `POST /api/v1/organization/<orgname>/applications/<client_id>/tokens`: create app token.
- `DELETE /api/v1/organization/<orgname>/applications/<client_id>/tokens/<token_uuid>`: revoke app token.
- Adds create/revoke audit log descriptions.


endpoints/api/organization_application_tokens.py
```diff
@@ -0,0 +1,261 @@
...
+                expiration_seconds=expiration,
+                max_active_tokens=MAX_TOKENS_PER_APPLICATION,
+            )
+        except oauth_model.TokenLimitExceeded:
+            return {
+                "message": "Token limit reached: maximum %d non-expired tokens per application"
+                % MAX_TOKENS_PER_APPLICATION
+            }, 400
+
+        log_action(
```
LiZhang19817, line 250: Missing LogEntryKind rows for "create_oauth_api_token" and "revoke_oauth_api_token"

shaonrh: These LogEntryKind rows are already present on master from the OAuth audit foundation work. They are inserted by migration https://github.com/quay/quay/blob/master/data/migrations/versions/d064a4f00d4a_add_oauth_api_token_metadata_and_audit_.py and also seeded in initdb.py for fresh installs. This PR reuses those existing log kinds for the lifecycle endpoints.

endpoints/api/organization_application_tokens.py
```diff
@@ -0,0 +1,297 @@
...
+        scopes.WRITE_REPO: ModifyRepositoryPermission(orgname, ""),
+        scopes.ADMIN_REPO: AdministerRepositoryPermission(orgname, ""),
+        scopes.CREATE_REPO: CreateRepositoryPermission(orgname),
+        scopes.READ_USER: UserReadPermission(user.username),
+        scopes.ADMIN_USER: UserAdminPermission(user.username),
+        scopes.ORG_ADMIN: AdministerOrganizationPermission(orgname),
+        scopes.SUPERUSER: SuperUserPermission(),
+    }
+
+    return all(scope_permissions[scope].can() for scope in requested_scopes)
```
bcaton85: `scope_permissions[scope]` will `KeyError` if a new scope is ever added to `ALL_SCOPES` without a corresponding entry here. Today all 8 scopes are covered, but this is a maintenance trap — the failure mode is a 500 in production.

Consider using `.get()` with a deny-by-default fallback:

```python
perm = scope_permissions.get(scope)
if perm is None:
    return False
```

Or validate completeness at import time with an assertion.

endpoints/api/organization_application_tokens.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+def token_view(token, include_secret=False, secret=None):
+    view = {
+        "uuid": token.uuid,
+        "name": getattr(token, "display_name", None),
```
bcaton85: `getattr(token, "display_name", None)` is unnecessary — `display_name` is a defined Peewee model field with `null=True`, so it always exists on `OAuthAccessToken` instances. `token.display_name` would be clearer and avoids suggesting the field might be absent.

web/playwright/e2e/superuser/build-logs.spec.ts
```diff
@@ -1,6 +1,24 @@
+import type {APIRequestContext} from '@playwright/test';
 import {test, expect} from '../../fixtures';
+import type {QuayConfig} from '../../fixtures';
```
bcaton85: The `refreshSuperuserSession` fix and related changes are unrelated to the OAuth token lifecycle feature. Consider splitting into a separate commit or PR for cleaner history.

data/model/oauth.py
```diff
@@ -4,11 +4,12 @@
 import logging
 import posixpath
 from datetime import datetime, timedelta
+from math import isfinite
 from typing import Any
 from urllib.parse import unquote, urlparse
```
bcaton85, line 11: `MAX_TOKENS_PER_APPLICATION = 1000` is a new hardcoded limit. Existing applications with >1000 active tokens won't break (the check is only at creation time), but they'll be blocked from creating new tokens until they drop below the cap.

Consider removing this limit entirely. If a cap is needed, it should be configurable via `config.yaml` so operators can tune it per deployment. A hardcoded limit risks blocking legitimate tenants with no recourse other than a code change.

shaonrh: Addressed in the latest commit. The hardcoded `1000` cap was removed and replaced with optional config: `OAUTH_APPLICATION_MAXIMUM_TOKEN_COUNT`. If unset or `null`, no token cap is enforced; if set, operators can tune the limit per deployment.

bcaton85 (approved):

cubismod (approved):

PR #6475, opened 2026-07-09
QUAYIO-2047: fix(cache): add null-safety guards and fix typo in referrers serialization

deshpandevlab:
[QUAYIO-2047](https://redhat.atlassian.net/browse/QUAYIO-2047)

## Context

Follow-up to [#6511](https://github.com/quay/quay/pull/6511) which introduced referrers cache serialization. That PR resolved the core serialization issue but introduced two bugs that still cause the original Sentry error in production:

```
TypeError: Object of type Manifest is not JSON serializable
at data/cache/impl.py:333 (RedisDataModelCache.retrieve)
transaction: v2.list_manifest_referrers
```

Confirmed still firing in production as of 2026-08-05 (Sentry event ac86c75c).

### 1. Missing null guard on `internal_manifest_bytes`

When `Manifest.select()` omits the `manifest_bytes` column (documented in `datatypes.py`), `internal_manifest_bytes` is `None`. PR #6511 unconditionally calls `.as_unicode()` on it during serialization and `Bytes.for_string_or_unicode()` during deserialization, which would raise `AttributeError`.

**Fix:** Add null checks on both serialization and deserialization paths.

### 2. Wrong key name: `legacy_image_handler` → `legacy_image_row`

The `inputs` dict has a key `legacy_image_row` (set by the `Manifest` datatype constructor), but #6511 nullifies the non-existent key `legacy_image_handler`. This leaves the real non-serializable `legacy_image_row` object in the dict, which causes `json.dumps()` to fail with `TypeError: Object of type Manifest is not JSON serializable`.

**Fix:** Use the correct key name `legacy_image_row`.

## Test

Adds `test_referrers_cache_serialization_roundtrip` to verify:
- `internal_manifest_bytes` survives the cache roundtrip as a `Bytes` instance
- `get_parsed_manifest()` succeeds on cached referrer objects

Jira: https://issues.redhat.com/browse/QUAYIO-2047

### DIFF
--- a/data/registry_model/registry_oci_model.py
+++ b/data/registry_model/registry_oci_model.py
@@ -284,9 +284,7 @@ class OCIModel(RegistryDataInterface):
         self, model_cache, repository_ref, manifest, artifact_type=None
     ):
         def load_referrers():
-            referrers = self.lookup_referrers_for_manifest(
-                repository_ref, manifest, artifact_type
-            )
+            referrers = self.lookup_referrers_for_manifest(repository_ref, manifest, artifact_type)
             serializable = []
             for r in referrers:
                 d = r.asdict()
--- a/data/registry_model/registry_oci_model.py
+++ b/data/registry_model/registry_oci_model.py
@@ -284,7 +284,21 @@ class OCIModel(RegistryDataInterface):
         self, model_cache, repository_ref, manifest, artifact_type=None
     ):
         def load_referrers():
-            return self.lookup_referrers_for_manifest(repository_ref, manifest, artifact_type)
+            referrers = self.lookup_referrers_for_manifest(
+                repository_ref, manifest, artifact_type
+            )
+            serializable = []
+            for r in referrers:
+                d = r.asdict()
+                if d["internal_manifest_bytes"] is not None:
+                    d["internal_manifest_bytes"] = d["internal_manifest_bytes"].as_unicode()
+                if d.get("inputs") and d["inputs"].get("repository") is not None:
+                    d["inputs"]["repository"] = d["inputs"]["repository"].asdict()
+                if d.get("inputs"):
+                    d["inputs"]["legacy_id_handler"] = None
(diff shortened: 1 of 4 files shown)


data/registry_model/registry_oci_model.py
```diff
@@ -284,13 +284,30 @@ def lookup_cached_referrers_for_manifest(
         self, model_cache, repository_ref, manifest, artifact_type=None
     ):
         def load_referrers():
-            return self.lookup_referrers_for_manifest(repository_ref, manifest, artifact_type)
+            referrers = self.lookup_referrers_for_manifest(repository_ref, manifest, artifact_type)
+            serializable = []
+            for r in referrers:
+                d = r.asdict()
+                if d["internal_manifest_bytes"] is not None:
```
lechuk47: What does this attribute contain? I'm a bit hesitant of caching bytes objects

deshpandevlab: `internal_manifest_bytes` contains the raw manifest JSON content (e.g., the OCI/Docker manifest JSON document) stored as a `Bytes` wrapper object it is not raw binary data.

For caching, we serialize it to a unicode string via `.as_unicode()` before storing, and re-wrap it back via `Bytes.for_string_or_unicode()` on cache read. This is the same pattern used by `lookup_cached_manifest_by_digest` ([lines 1065-1084](https://github.com/quay/quay/blob/cfaa99fd226950b755b90adb688dfcf0efcede68/data/registry_model/registry_oci_model.py#L1065-L1084)):

So what actually gets stored in Redis is the JSON string, no raw bytes are cached.

cubismod (approved):

data/registry_model/registry_oci_model.py
lechuk47: Sorry, my coment was not very clear. I understand the serialization/deserialization process. My concern was about the amount/type of data these fields can contain.

PR #6488, opened 2026-07-13
PROJQUAY-9857: feat(oauth): add API access token management

shaonrh:
Adds the API Access Tokens tab to the OAuth application management modal, replacing the legacy Generate Token tab with token inventory and lifecycle controls.

Includes token list metadata, fixed expiration choices for generated tokens, one-time secret display, copy support, scope truncation with tooltip, and revoke actions behind a row kebab with confirmation.

Adds and updates frontend tests covering the token management UI and OAuth application lifecycle expectations.

Fixes [PROJQUAY-12210](https://redhat.atlassian.net/browse/PROJQUAY-12210) and [PROJQUAY-12211](https://redhat.atlassian.net/browse/PROJQUAY-12211)

Kapture _3


web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
```diff
@@ -0,0 +1,711 @@
...
+        </Button>
+      </ModalFooter>
+    </Modal>
+  );
+
+  const authorizationModal = props.application ? (
+    <GenerateTokenAuthorizationModal
+      isOpen={isAuthorizationModalOpen}
+      onClose={handleAuthorizationModalClose}
+      onConfirm={handleAuthorizationConfirm}
```
jbpratt, line 690: **Please lock confirmation as soon as the request starts.**

The pending state is not passed into this modal, so a double-click—or a second click during a slow request—invokes handleAuthorizationConfirm twice.

The backend accepts duplicate names, which means:

1. Both POST requests mint live bearer tokens.
2. generatedToken retains only the last response.
3. The first token remains active, but its one-time secret is permanently lost.

Please pass the create/assign pending state into the authorization modal, disable confirm/close/cancel while pending, and add a synchronous guard in the handler. The revoke confirmation has the same overlapping-request risk.

web/src/hooks/UseOAuthApplications.ts
```diff
@@ -77,6 +109,138 @@ export function useFetchOAuthApplications(org: string) {
...
+    mutateAsync: createOAuthApplicationTokenMutationAsync,
+    isLoading: creatingOAuthApplicationToken,
+    reset: resetCreateOAuthApplicationToken,
+  } = useMutation(
+    async (params: CreateOAuthApplicationTokenParams) => {
+      return createOAuthApplicationToken(org, clientId, params);
+    },
+    {
+      onSuccess: (createdToken: IOAuthApplicationToken) => {
+        queryClient.invalidateQueries([
```
jbpratt, line 109: **Please decouple one-time secret presentation from token-list refresh state.**

After the POST succeeds, this invalidation starts a GET refresh. If that GET exhausts its retries and fails:

1. The token is already live.
2. APIAccessTokensTab returns RequestError.
3. TokenDisplayModal is never rendered.
4. The user permanently loses the token's one-time secret.

Please render the secret modal independently of list loading/error branches and report background refresh failures separately. A successful create must always expose its returned secret.

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
```diff
@@ -0,0 +1,711 @@
...
+      // Error state is set by the mutation onError callback.
+    }
+  };
+
+  const handleRevokeModalClose = () => {
+    setTokenToRevoke(null);
+    setRevokeError('');
+  };
+
+  const generateButton = (
```
jbpratt: **Please gate mutation controls on an explicit management capability.**

Global readonly superusers are intentionally allowed to view this inventory, but the backend rejects token creation, assignment, and revocation for them.

The current UI still renders Generate and Revoke, so every attempted action produces a guaranteed 403 and a generic failure message.

Please derive a canManage capability from the current user/config and hide or disable Generate, Assign, and Revoke while preserving read-only inventory access.

web/playwright/e2e/organization/oauth-applications.spec.ts
```diff
@@ -66,7 +66,9 @@ test.describe('OAuth Applications', {tag: ['@organization']}, () => {
     await expect(
       page.getByRole('tab', {name: 'OAuth Information'}),
     ).toBeVisible();
-    await expect(page.getByRole('tab', {name: 'Generate Token'})).toBeVisible();
+    await expect(
```
jbpratt, line 69: **The Playwright coverage stops at tab visibility.**

This assertion proves that the API Access Tokens tab renders, but it does not exercise the real browser-to-Quay contract:

- Generate a token through the real POST request.
- Display and copy the one-time secret.
- Refresh and verify the token inventory.
- Revoke it through the real DELETE request.

The component tests mock the hooks, so they cannot catch endpoint, payload, CSRF, invalidation, or modal-lifecycle failures. Per web/playwright/AGENTS.md, full UI workflows should use Playwright with real API interactions.

Could we add one create → secret display → inventory → revoke lifecycle test, with fixture-managed cleanup?

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
```diff
@@ -0,0 +1,711 @@
...
+  return (
+    EXPIRATION_OPTIONS.find((option) => option.id === expirationId) ||
+    EXPIRATION_OPTIONS[EXPIRATION_OPTIONS.length - 1]
+  );
+}
+
+function formatTokenName(token: IOAuthApplicationToken): string {
+  return token.name || '';
+}
+
```
bcaton85, line 96: **should-fix: Acceptance criteria gap — Legacy tokens show blank name instead of "Legacy Token"**

JIRA AC 5 specifies that pre-existing tokens should display `Name: "Legacy Token" (non-editable)`. The current `formatTokenName` returns an empty string for `null` names, and the test explicitly asserts `toBeEmptyDOMElement()`.

If blank is the intended UX, the acceptance criteria should be updated. Otherwise, consider returning `"Legacy Token"` for null names.

web/src/resources/OAuthApplicationResource.ts
```diff
@@ -11,6 +15,121 @@ export async function fetchOAuthApplications(org: string) {
...
+        data.next_page,
+      );
+      return tokens.concat(nextTokens);
+    }
+
+    return tokens;
+  } catch (err) {
+    throw new ResourceError(
+      'Failed to fetch OAuth application tokens',
+      clientId,
```
bcaton85, line 62: **nit: Recursive pagination double-wraps errors**

If page 2 fails, the recursive call throws a `ResourceError`, which is caught by page 1's `catch` and wrapped in another `ResourceError` — producing `ResourceError(ResourceError(AxiosError))`. Harmless but obscures the original error.

Could re-throw `ResourceError` instances directly:
```ts
} catch (err) {
  if (err instanceof ResourceError) throw err;
  throw new ResourceError('Failed to fetch OAuth application tokens', clientId, err);
}
```

bcaton85:
**nit: Component naming mismatch**

`ManageOAuthApplicationDrawer` now renders a `Modal`, but the file name, component name, and prop names (`isDrawerOpen`, `handleDrawerToggle`) still use drawer terminology. A follow-up rename would improve readability.

web/playwright/e2e/superuser/build-logs.spec.ts
```diff
@@ -1,6 +1,24 @@
+import type {APIRequestContext} from '@playwright/test';
 import {test, expect} from '../../fixtures';
+import type {QuayConfig} from '../../fixtures';
+import {TEST_USERS, TEST_USERS_LDAP} from '../../global-setup';
+import {ApiClient} from '../../utils/api';
 import {API_URL} from '../../utils/config';
 
+async function refreshSuperuserSession(
+  request: APIRequestContext,
```
bcaton85: **nit: Unrelated test changes bundled in PR**

The `refreshSuperuserSession` additions fix test stability for LDAP/OIDC auth scenarios but are unrelated to OAuth token management. Consider splitting into a separate commit or noting in the PR description.

bcaton85:
**should-fix: Missing test for ASSIGN_OAUTH_TOKEN feature flag disabled**

All tests mock `ASSIGN_OAUTH_TOKEN: true`. There is no test verifying that the "Assign another user" button is hidden when the feature flag is `false`. This is a non-trivial conditional (`APIAccessTokensTab.tsx:479`) that could regress silently.

bcaton85:
**nit: No test for loading/error state rendering**

No tests exercise the `loading` or `errorLoadingOAuthApplicationTokens` early-return paths (`APIAccessTokensTab.tsx:620-625`). These are simple branches but adding a smoke test would prevent regressions.

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
shaonrh: I actually updated it, I don't really see the point of showing "Legacy Token", this might be confusing to users thinking the token itself has changed in some way.

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
shaonrh: Addressed by passing mutation pending state into the confirmation modal, disabling confirm/cancel/close while pending, and adding synchronous guards for create, assign, and revoke. Added double-click and pending-state tests for both flows.

web/src/hooks/UseOAuthApplications.ts
shaonrh: Addressed by rendering the secret modal independently from the token inventory’s Suspense/error boundary. Background refresh failures now show an inline alert without hiding the generated secret, with regression coverage.

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
shaonrh: Addressed by deriving `canManage` from the current user and registry state. Generate, assign, and revoke controls are hidden for global readonly superusers and readonly registries while inventory remains visible.

web/playwright/e2e/organization/oauth-applications.spec.ts
shaonrh: Added a real browser-to-Quay lifecycle test covering token creation, one-time secret display and copy, page refresh/inventory verification, and revocation using fixture-managed cleanup.

web/src/resources/OAuthApplicationResource.ts
shaonrh: Addressed by rethrowing existing `ResourceError` instances rather than wrapping them again. Added a test confirming a later-page failure preserves the original underlying error.

web/playwright/e2e/superuser/build-logs.spec.ts
shaonrh: Addressed by removing the unrelated `build-logs.spec.ts` changes from the final PR diff.

shaonrh:
@bcaton85 Addressed the feature-flag coverage noted in https://github.com/quay/quay/pull/6488#issuecomment-4958606049. Added a test verifying that **Assign another user** is hidden when `ASSIGN_OAUTH_TOKEN` is disabled while the normal token generation form remains available.

bcaton85 (approved):

PR #6489, opened 2026-07-13
PROJQUAY-11619: fix(ldap): handle single-line trace format in password redaction

LiZhang19817:
## Summary
- Fixes LDAP password redaction in `_LDAPTraceRedactor` to handle **all known trace formats** from python-ldap, closing multiple edge cases where passwords could leak into logs

### Fix 1: Single-line trace format
- The original regex required `\s*\n\s*` between the DN and password, which fails for single-line repr output:
  ```
  (('cn=admin,dc=quay,dc=io', 'password', None, None), {})
  ```
- Changed `\s*\n\s*` to `\s*` so both single-line and multi-line pprint formats are matched

### Fix 2: Split-write buffering
- python-ldap writes the trace in a single `write()` call, but if a future library change or intermediary wrapper splits the header and argument tuple across two calls, the password would leak
- Added `_BIND_HEADER_RE` and `_BIND_TUPLE_COMPLETE_RE` with a `_buf` buffer: when `write()` receives data containing `simple_bind` but the tuple is incomplete (no `None, None), {` yet), it buffers until the next `write()` completes the tuple, then applies redaction to the combined string
- Handles splits at **any point** within the bind tuple, not just at the header boundary
- `flush()` emits any buffered data (with redaction applied) to prevent data loss

### Fix 3: Double-quoted DNs and passwords (unified regex)
- `pprint` wraps strings in double quotes when they contain a single-quote character (e.g., `cn=O'Brien` becomes `"cn=O'Brien,dc=quay,dc=io"`)
- The previous approach used two separate regexes (`_BIND_PW_SINGLE_RE` / `_BIND_PW_DOUBLE_RE`) that both **assumed the DN was single-quoted** (`(('dn',`). When the DN was double-quoted by pprint, neither regex matched and the password leaked
- Replaced with a single unified `_BIND_PW_RE` using a generic Python string literal pattern (`_PY_STR`) that matches both `'...'` and `"..."` for both DN and password fields
- This covers all combinations: normal DN + normal password, DN with `'` + normal password, normal DN + password with `'`, both with `'`, passwords with `"`, and passwords with both quote types

## Test plan
- [x] `test_redacts_simple_password` — basic multi-line redaction
- [x] `test_redacts_complex_password` — complex password with special chars
- [x] `test_redacts_passwords_with_escaped_quotes` — escaped quote in password
- [x] `test_redacts_single_line_trace` — single-line repr format
- [x] `test_redacts_single_line_complex_password` — complex password on single-line
- [x] `test_redacts_dn_with_single_quote` — **NEW**: DN with apostrophe (pprint double-quotes DN)
- [x] `test_redacts_dn_with_quote_and_pw_with_quote` — **NEW**: both DN and password contain `'`
- [x] `test_redacts_password_with_double_quote` — **NEW**: password contains `"`
- [x] `test_redacts_password_with_both_quotes` — **NEW**: password contains both `'` and `"`
- [x] `test_redacts_pprint_real_output` — **NEW**: uses real `pprint.pformat()` for 6 DN/password combos
- [x] `test_redacts_split_write_header_then_args` — split write with header then args
- [x] `test_redacts_split_write_header_plus_partial_dn` — **NEW**: split after DN, before password
- [x] `test_flush_emits_buffered_header` — buffered header emitted on flush
- [x] `test_non_bind_trace_not_buffered` — non-bind traces pass through immediately
- [x] `test_trace_pass_unchanged` — non-bind trace unchanged
- [x] `test_flush_delegates_to_stream` — flush delegation
- [x] Playwright e2e: failed login does not expose password in page content
- [x] Existing multi-line tests still pass (backward-compatible)
- [x] No ReDoS risk — tested with 10K-char DNs/passwords, completes in <1ms

## Related
Follow-up to #5959 — the original fix only handled multi-line pprint trace output.

### DIFF
--- a/data/users/externalldap.py
+++ b/data/users/externalldap.py
@@ -60,8 +60,12 @@ class _LDAPTraceRedactor:
     argument tuple arrives in the next write.
     """
 
-    _BIND_PW_RE = re.compile(
+    _BIND_PW_SINGLE_RE = re.compile(
         r"(SimpleLDAPObject\.simple_bind\s*\n\(\('[^']*',\s*')((?:\\.|[^'\\])*)(')",
+    )
+
+    _BIND_PW_DOUBLE_RE = re.compile(
+        r'(SimpleLDAPObject\.simple_bind\s*\n\(\(\'[^\']*\',\s*")((?:\\.|[^"\\])*)(")',
     )
 
     _BIND_HEADER_RE = re.compile(
@@ -72,17 +76,23 @@ class _LDAPTraceRedactor:
         self._stream = stream or sys.stdout
         self._buf = ""
 
+    @classmethod
+    def _redact(cls, data):
+        data = cls._BIND_PW_SINGLE_RE.sub(r"\1******\3", data)
+        data = cls._BIND_PW_DOUBLE_RE.sub(r"\1******\3", data)
+        return data
+
     def write(self, data):
         data = self._buf + data
         self._buf = ""
         if self._BIND_HEADER_RE.search(data):
             self._buf = data
             return
-        self._stream.write(self._BIND_PW_RE.sub(r"\1******\3", data))
+        self._stream.write(self._redact(data))
 
     def flush(self):
         if self._buf:
-            self._stream.write(self._BIND_PW_RE.sub(r"\1******\3", self._buf))
+            self._stream.write(self._redact(self._buf))
             self._buf = ""
         self._stream.flush()
 
@@ -576,7 +586,7 @@ class LDAPUsers(FederatedUsers):
                 return cached
 
         logger.debug("Looking up LDAP %s with key=%s", log_name, cache_key)
-        (found_user, err_msg) = lookup_fn()
+        found_user, err_msg = lookup_fn()
         result = found_user is not None
 
         if found_user is None:
@@ -719,7 +729,7 @@ class LDAPUsers(FederatedUsers):
                 pairs = []
                 err_msg = None
                 for user_search_dn in self._user_dns:
-                    (pairs, err_msg) = self._ldap_user_search_with_rdn(
+                    pairs, err_msg = self._ldap_user_search_with_rdn(
                         conn,
                         username_or_email,
                         user_search_dn,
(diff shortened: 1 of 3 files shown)


bcaton85 (approved):

LiZhang19817:
## Deep Review: LDAP Password Redaction

Reviewed current PR head: `14b9be35a7f2bd3959db46d46823d55f287824d7` against base `539979691058ffef689052188053d1fbbf538209`.

### Result

No blocking issues found in the latest version of this PR.

The earlier LDAP trace redaction gaps appear fixed in the current head. The PR now handles:

- Single-line `simple_bind` trace output.
- Multi-line `pprint` trace output.
- Passwords represented with single or double quotes.
- DNs represented with single or double quotes, including DNs containing apostrophes such as `cn=O'Brien`.
- Passwords containing apostrophes, double quotes, both quote types, and backslashes.
- Split writes after the header, after tuple open, after DN, and before the password is complete.

### Bug Fix Assessment

The bug appears fixed for the known affected trace formats. The updated `_LDAPTraceRedactor` uses a Python string-literal pattern for both DN and password fields, then redacts only the password field. This closes the original single-line leak and the later quote-format leak cases.

The split-write behavior is also improved. Instead of buffering only when a chunk ends exactly at `SimpleLDAPObject.simple_bind`, the redactor now buffers when a `simple_bind` trace has started but the bind tuple is not complete.

### Regression / New Issue Assessment

No functional regression found in the LDAP auth path. The production change is scoped to debug trace redaction, and normal LDAP bind/search behavior is not changed.

Non-blocking observations:

- The Playwright test checks page body text after failed login, but it does not validate server-side LDAP trace redaction. The meaningful coverage for this bug is the Python unit test coverage.
- In malformed/incomplete debug traces, buffering can retain data until `flush()`. This is low risk because it is under `USERS_DEBUG`, but a bounded buffer would be more defensive.

### Security Assessment

No remaining password-leak issue found in the latest implementation for representative `pprint.pformat()` outputs.

I checked the redaction logic against:

- Normal DN/password.
- Password with apostrophe.
- DN with apostrophe.
- Both DN and password with apostrophe.
- Password with double quote.
- Password with both quote types.
- Backslash-containing DN/password.
- Mid-tuple split writes.
- 10K-character DN/password stress case.

All tested password values were redacted, and the long-input probe completed quickly, so I do not see an obvious ReDoS issue in the current regex.

### CI / Validation

Current PR checks are green, including Python unit/integration tests, Playwright E2E, CodeQL, pre-commit, CodeRabbit, and `all-green`.

I could not run the focused pytest target locally because my temp checkout lacked the `mock` Python dependency, but upstream CI covers the project test suite.

### Recommendation

This PR is acceptable from the LDAP password-redaction/security perspective. I would merge after optionally clarifying or removing the Playwright test if the intent is specifically server-side log-redaction coverage.

PR #6490, opened 2026-07-13
NO-ISSUE: fix(go): update dockerfile and certs gen

jbpratt:
Update dockerfile to build from go-toolset, copy only needed files, and the CA generation to be an authority for self signed certs (required for the OCP provisioning)


Marcusk19 (approved):
lgtm

PR #6542, opened 2026-07-16
PROJQUAY-12324: fix(quadlet): add :U flag to volume mount for Podman 5.x UID remapping

harishsurf:
## Summary
- Add `:U` flag to the Quadlet volume mount (`Volume=%s:/data:Z,U`)
- The Containerfile sets `USER 1001`; without `:U`, Podman 5.x cannot remap host UIDs to match the in-container user, causing permission denied errors on `/data`
- Same fix applied to OMR 2.0.x in [PROJQUAY-9329](https://redhat.atlassian.net/browse/PROJQUAY-9329)

## Test plan
- [ ] Verify `TestQuadletInstallVolumeHasUFlag` passes
- [ ] Deploy on a Podman 5.x host and confirm the container starts without permission errors

### DIFF
--- a/internal/system/quadlet.go
+++ b/internal/system/quadlet.go
@@ -52,7 +52,7 @@ After=network-online.target
 
 [Container]
 Image=%s
-Volume=%s:/data:Z
+Volume=%s:/data:Z,U
 PublishPort=%s:%s
 Exec=%s
 
--- a/internal/system/quadlet_test.go
+++ b/internal/system/quadlet_test.go
@@ -94,6 +94,26 @@ func TestQuadletInstallDoesNotPersistBootstrapCredentials(t *testing.T) {
 	}
 }
 
+func TestQuadletInstallVolumeHasUFlag(t *testing.T) {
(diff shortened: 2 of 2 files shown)


jbpratt:
I don't think `:U` is appropriate for the OMR v2-to-v3 migration.

[mirror-registry PR 259](https://github.com/quay/mirror-registry/pull/259) fixed a different 1.3.x-to-2.x issue where `quay-copy` left the SQLite volume with unusable ownership. It intentionally applied `:U` only to SQLite, not blob storage.

The v3 migrator copies the old data into a new host-owned `/data` tree, and `Dockerfile.mirror` deliberately runs as root so rootless Podman maps the container to the invoking host user. `internal/migrate/copy.go` documents this ownership contract.

Adding `:U` here recursively chowns the entire database and blob tree after migration. It cannot fix access to the source OMR volumes because that copying happens before the Quadlet starts.

I recommend dropping `:U` and adding a Podman 5 v2-to-v3 migration test. If old source volumes are unreadable, fix source access in the migrator using `podman unshare` or a helper container, then copy into host-owned v3 files.

PR #6555, opened 2026-07-17
PROJQUAY-12346: chore(deps): upgrade pyasn1 from 0.6.3 to 0.6.4 to address CVE-2026-59885

rhdmalone:
1. chore(deps): upgrade pyasn1 from 0.6.3 to 0.6.4 to address CVE-2026-59885,CVE-2026-59886
2. Fix requirements-pinned hook to correctly skip comment lines by replacing broken \\#* case pattern with "#"*

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     hooks:
     -   id: requirements-pinned
         name: Check requirements.txt is pinned
-        entry: bash -euo pipefail -c 'while IFS= read -r i; do case "$i" in ""|\\#*|-r*|--*) continue;; esac; if [[ "$i" != *"=="* && "$i" != *"@"* ]]; then echo "$i is not pinned"; exit 1; fi; done < requirements.txt'
+        entry: bash -euo pipefail -c 'while IFS= read -r i; do case "$i" in ""|"#"*|-r*|--*) continue;; esac; if [[ "$i" != *"=="* && "$i" != *"@"* ]]; then echo "$i is not pinned"; exit 1; fi; done < requirements.txt'
         language: system
         files: requirements.txt
         pass_filenames: false
--- a/requirements.txt
+++ b/requirements.txt
@@ -81,7 +81,7 @@ prometheus_client==0.7.1
 protobuf==5.29.5
 psutil==5.9.0
 psycopg2==2.9.9
-pyasn1==0.6.3
+pyasn1==0.6.4
 pyasn1_modules==0.4.2
 py-bitbucket @ git+https://github.com/quay/py-bitbucket.git@3ba167408d92eb4ba2cae78c0047b8f1e8821c67
 PyGithub==2.1.1
--- a/web/playwright/utils/container.ts
+++ b/web/playwright/utils/container.ts
@@ -21,7 +21,7 @@ const execFileAsync = promisify(execFile);
 const REGISTRY_HOST = new URL(API_URL).host;
 
 const BUSYBOX_IMAGE =
-  'quay.io/prometheus/busybox@sha256:467c84e76e5bd720ffd3cb86c11e421e2489ef20868114ba58cd0a9e02e70a9d';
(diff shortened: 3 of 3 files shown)


rhdmalone:
Pre-existing issue:
The comment # Leverage wheels for faster builds already exists on origin/master at line 142 — it pre-dates your change entirely. This is a pre-existing failure in the repo, not caused by your pyasn1 bump.

The requirements-pinned hook runs against the whole file whenever requirements.txt is touched, so it's surfacing a bug that was already there. Your options:

Do nothing — leave a comment on the PR explaining the pre-commit failure is pre-existing and unrelated to this CVE fix. Reviewers can verify with git show origin/master:requirements.txt | grep "Leverage wheels".

Fix it in the same commit — the hook is misreading the comment as a package name. The simplest fix would be to check what the hook script expects (it may require comments to be on a line with a package, not standalone). If that's the case, removing or reformatting the comment would fix it. However, that changes scope beyond the CVE fix.

Option 1 is the cleaner approach for a CVE-only PR — note it in the PR and let the maintainers deal with the pre-existing lint issue separately. 

Both comments fail — CI only reported the first one because the script hits exit 1 immediately. Now let me check if this was already failing on origin/master before your commit:

# Leverage wheels for faster builds is not pinned
# Indirect dependencies is not pinned
Confirmed — origin/master itself has the same two failing lines. This is genuinely a pre-existing bug in the repo, introduced by whoever added those comment lines without updating the hook's case pattern.

The hook's case pattern \\#* only skips lines starting with \# (backslash + hash), not plain # comment lines. It's a bug in the hook itself.

Your PR didn't introduce this. The failure you're seeing in CI is because the requirements-pinned check runs against the full file whenever requirements.txt is touched — and it was already broken in origin/master for the same reason.

The clean thing to do is add a note in your PR explaining this, e.g.:

Note: the requirements-pinned pre-commit failure is pre-existing on origin/master and unrelated to this change. The hook's case pattern \\#* does not correctly skip plain # comment lines in requirements.txt.

cc: @aroyoredhat , @Marcusk19 , @bcaton85

rhdmalone:
test failures:
all-green failing because Playwright tests are failing:

Copilot analysis"
I'll help you find a solution for this failing job. Let me first retrieve the job logs to understand what's happening.The job is a status check that monitors other CI jobs. The `web-ci` job failed, which is required to succeed. Let me get more details about the actual failing job:Based on the logs, the failing job is actually a sentinel/aggregator job that checks the status of other CI jobs. The real failure is in the **`web-ci`** job, which ran 109 Playwright E2E tests.

## Root Cause

The tests are failing due to **`skopeo copy` command failures** when trying to pull a container image from `quay.io/prometheus/busybox`. The error message shows:

```
time="2026-07-20T15:21:31Z" level=fatal msg="initializing source docker://quay.io/prometheus/busybox@sha256:467c84e76e5bd720ffd3cb86c11e421e2489ef20868114ba58cd0a9e02e70a9d: reading manifest sha256:467c84e76e5bd720ffd3cb86c11e421e2489ef20868114ba58cd0a9e02e70a9d in q
```

This appears to be a network/registry connectivity issue during test setup where the test infrastructure attempts to copy the busybox image to a local test registry.

## Why Tests Are Failing

The E2E tests need to set up test images in the local Docker registry (running at `localhost:8080`), but the `skopeo copy` command is failing to read the manifest from the remote source. This causes approximately 109 tests to fail, all of which depend on having test images available.

## Recommended Solutions

1. **Check Registry Connectivity**: Verify that `quay.io` is accessible from the CI runner and there are no network/firewall issues blocking manifest reads.

2. **Add Retry Logic**: Modify the test setup to retry failed `skopeo copy` operations with exponential backoff, as transient network issues are common in CI environments.

3. **Use a Cached/Local Image**: Instead of pulling from `quay.io/prometheus/busybox` during each test run, pre-cache the image in the CI runner or use a simpler local test image.

4. **Increase Timeout**: If the issue is a slow network, increase the timeout for `skopeo copy` operations.

The sentinel job itself is working correctly—it's properly detecting and reporting that the `web-ci` job failed. To fix the PR, you need to address the `web-ci` test infrastructure setup issue.

jbpratt (approved):

PR #6573, opened 2026-07-20
QUAYIO-1708: feat(deploy): add PDBs and dedicated node pool support

Sunandadadi:
## Summary

- Add PodDisruptionBudgets (`maxUnavailable: 1`) for Envoy and RLS — prevents node drains from evicting all pods simultaneously
- Change `nodeSelector` from `part-of: quay` to `workload: envoy` — pins Envoy/RLS pods to dedicated nodes
- Add toleration for `workload=envoy:NoSchedule` taint — allows Envoy pods onto tainted nodes while blocking Quay app pods
- Bump default replicas from 1 to 2 — PDB requires 2+ replicas to be effective

## Why

Envoy sits on the critical path for all quay.io traffic. Dedicated nodes isolate Envoy resource usage from Quay app pods and allow independent scaling. PDBs ensure availability during cluster upgrades and node maintenance.

## How it works

Dedicated envoy nodes are labeled `part-of: quay` (ALB can see them) + `workload: envoy` (nodeSelector target) and tainted `workload=envoy:NoSchedule`. This means:
- ALB registers envoy nodes (they have `part-of: quay`)
- Quay app pods can't schedule on envoy nodes (taint blocks them)
- Envoy/RLS pods only schedule on envoy nodes (nodeSelector)

## Depends on

App-interface MR to add the dedicated node pools to the cluster config (machine pools with labels + taints).

## Test plan

- [ ] Verify envoy nodes have both labels (`part-of=quay`, `workload=envoy`)
- [ ] Verify Envoy/RLS pods schedule on envoy nodes only
- [ ] Verify Quay app pods do NOT schedule on envoy nodes
- [ ] Verify PDB prevents full eviction during node drain

### DIFF
--- a/deploy/openshift/rosa/quay-envoy-deploy-template.yaml
+++ b/deploy/openshift/rosa/quay-envoy-deploy-template.yaml
@@ -139,6 +139,10 @@ objects:
       app: quay-rls
   spec:
     ports:
+    - name: healthcheck
+      protocol: TCP
+      port: 8080
+      targetPort: 8080
     - name: grpc
       protocol: TCP
       port: 8081
@@ -184,10 +188,26 @@ objects:
           image: ${RLS_IMAGE}:${RLS_IMAGE_TAG}
           imagePullPolicy: IfNotPresent
           ports:
+          - containerPort: 8080
+            name: healthcheck
           - containerPort: 8081
             name: grpc
           - containerPort: 9090
             name: metrics
+          livenessProbe:
+            httpGet:
+              path: /healthcheck
+              port: 8080
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            timeoutSeconds: 3
+          readinessProbe:
+            httpGet:
+              path: /healthcheck
+              port: 8080
+            initialDelaySeconds: 5
+            periodSeconds: 5
+            timeoutSeconds: 3
           env:
           - name: USE_STATSD
             value: "false"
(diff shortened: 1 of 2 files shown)


deshpandevlab:
side note: Envoy has both probes defined, but the RLS deployment has none. Can we add? It can done as a follow up item

deshpandevlab (dismissed):

Sunandadadi:
> side note: Envoy has both probes defined, but the RLS deployment has none. Can we add? It can done as a follow up item

Good catch, added probes on RLS in this commit - https://github.com/quay/quay/pull/6573/changes/ffb49059b235c1221650b22b39033eeb37b5ebc9

deploy/openshift/rosa/quay-envoy-deploy-template.yaml
```diff
@@ -171,6 +188,20 @@ objects:
...
+            httpGet:
+              path: /healthcheck
+              port: 8080
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            timeoutSeconds: 3
+          readinessProbe:
+            httpGet:
+              path: /healthcheck
+              port: 8080
```
deshpandevlab, line 207: hey, this health probes hit port 8080, but the RLS container only declares ports 8081 (grpc) and 9090 (metrics). https://github.com/quay/quay/blob/ffb49059b235c1221650b22b39033eeb37b5ebc9/deploy/openshift/rosa/quay-envoy-deploy-template.yaml#L141

I think we need to declare 8080 here with the healthcheck label

Sunandadadi: good point, updated here - https://github.com/quay/quay/pull/6573/changes/c121c521c4ba6c8483231ea52d5771cb3161c9ad

deshpandevlab (dismissed):

deshpandevlab (approved):

PR #6576, opened 2026-07-20
NO-ISSUE: feat(auth): add OCI registry bearer token authentication

jbpratt:
## Summary

Add standards-based Docker Registry v2 Bearer authentication to the Go registry so OCI clients can authenticate users and robot accounts through a scoped JWT token exchange. Preserve the active signing identity and public endpoint configuration when migrating an existing installation.

## Related Issue

NO-ISSUE: this completes the Go registry authentication path and migration continuity work without an associated Jira ticket.

## Changes

- Add persisted RSA/RS256 registry JWT issuance and strict validation with Docker access claims, audience checks, bounded lifetimes, and secure key-file handling.
- Add `/v2/auth` credential exchange for users, robot accounts, and anonymous public pulls with permission-aware scope downscoping and request limits.
- Require Bearer tokens across distribution and referrers endpoints and return standards-compliant challenges that preserve `SERVER_HOSTNAME`, including custom ports.
- Validate and import the approved source signing key during migration, safely resume after interrupted migrations, and preserve runtime ports across migration and Quadlet upgrades.
- Update required CI tag-list checks to use `skopeo list-tags`, exercising the client Bearer flow.
- Add focused positive, negative, migration, port, and security regression tests.

## Testing

- [x] `GOCACHE=/tmp/quay-go-cache make go-test` passes
- [x] `GOCACHE=/tmp/quay-go-cache make go-build` passes
- [x] `GOCACHE=/tmp/quay-go-cache go vet ./...` passes
- [x] `GOCACHE=/tmp/quay-go-cache GOLANGCI_LINT_CACHE=/tmp/quay-golangci-cache golangci-lint run --timeout=5m ./...` passes with 0 issues
- [x] `go mod verify` passes
- [x] `GOCACHE=/tmp/quay-go-cache go mod tidy -diff` reports no diff
- [x] Pre-commit hooks pass
- [ ] GitHub Actions E2E and OCI conformance jobs will run in CI

## Notes

- Target Version: not applicable (`NO-ISSUE`).
- Backport: not expected.
- The implementation was independently reviewed for authentication bypasses, scope-based resource exhaustion, key-file safety, migration resume behavior, and port continuity; confirmed findings are addressed in this change.


internal/registry/jwtauth/key.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil, fmt.Errorf("parse registry JWT signing key %s: %w", path, err)
+	}
+	return key, nil
+}
+
+// ParsePrivateKey parses an RSA private key from one PKCS#1 or PKCS#8 PEM block.
+func ParsePrivateKey(data []byte) (*rsa.PrivateKey, error) {
+	block, rest := pem.Decode(data)
+	if block == nil || len(bytes.TrimSpace(rest)) != 0 {
+		return nil, fmt.Errorf("expected one PEM block")
```
harishsurf, line 90: **CRITICAL: Key directory created with 0750 instead of 0700**

`os.MkdirAll(filepath.Dir(path), 0o750)` creates the parent directory group-readable/executable. The key file itself is properly `0o600`, but the directory lets group members list contents and see the key filename. In OpenShift pods with shared supplemental groups, this is an information leak.

**Fix:** `os.MkdirAll(filepath.Dir(path), 0o700)`

internal/registry/distribution/auth.go
```diff
@@ -85,71 +110,135 @@ func newAccessController(options map[string]interface{}) (distauth.AccessControl
...
 
 func (ac *accessController) Authorized(req *http.Request, access ...distauth.Access) (*distauth.Grant, error) {
-	result := ac.authenticator.Authenticate(req)
-	// Only missing credentials may fall back to anonymous public pulls. Invalid
-	// credentials must fail instead of silently becoming anonymous.
-	if !result.Authenticated && !result.Presented && ac.canAnonymousPull(req, access) {
+	return ac.authorizedBearer(req, access)
+}
+
+func (ac *accessController) Authenticate(r *http.Request, access ...oci.Access) (*oci.Grant, error) {
```
harishsurf, line 130: **MAJOR: `/v2/_catalog` becomes inaccessible with Bearer auth**

The Bearer challenge returns `scope="catalog:*:*"` for catalog access, but the token handler (`parseRequestedScopes`) only accepts `repository:` scoped tokens. Clients receive a challenge they cannot fulfill — the token endpoint returns 400 for the catalog scope.

This breaks `skopeo list-tags`, `podman search`, and any client using `/v2/_catalog` for authenticated registries.

internal/registry/distribution/token_handler.go
```diff
@@ -0,0 +1,274 @@
...
+	policy          repositoryAccessPolicy
+	issuer          tokenIssuer
+	anonymousAccess bool
+}
+
+const (
+	maxRequestedScopeCount = 100
+	maxRequestedScopeBytes = 16 << 10
+)
+
```
harishsurf, line 30: **MAJOR: No rate limiting on `/v2/auth` token exchange endpoint**

The endpoint accepts unlimited Basic auth attempts with no rate limiting, account lockout, or progressive delay. Each request triggers bcrypt (users) or AES-CCM decrypt (robots) plus RSA-2048 signing.

For mirror-registry (not internet-facing), this is lower risk. But any deployment exposed to untrusted networks is vulnerable to credential brute-forcing and CPU exhaustion via anonymous scope requests (each requiring an RSA signature).

internal/registry/distribution/token_handler.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		scopes = append(scopes, fields...)
+	}
+	requested := make([]jwtauth.ResourceActions, 0, len(scopes))
+	for _, scope := range scopes {
+		parts := strings.SplitN(scope, ":", 3)
+		if len(parts) != 3 || parts[0] != repositoryResourceType || parts[1] == "" || parts[2] == "" {
+			return nil, fmt.Errorf("invalid scope %q", scope)
+		}
+		actions := strings.Split(parts[2], ",")
```
harishsurf, line 163: **MAJOR: `parseRequestedScopes` rejects `delete` as a valid scope action**

Only accepts `pull`, `push`, and `*`. The Docker Registry v2 token spec includes `delete` as a valid repository action. Clients requesting `scope=repository:name:pull,push,delete` get HTTP 400 instead of a downscoped response. Using `*` works as a workaround but this breaks spec-conforming clients and CI pipelines that explicitly request delete scope.

internal/registry/jwtauth/token.go
```diff
@@ -0,0 +1,291 @@
...
+		issuer: cfg.Issuer, audience: cfg.Audience, lifetime: cfg.Lifetime,
+		leeway: cfg.Leeway, maxAge: cfg.MaxAge, now: cfg.Now, kid: kid,
+		publicKey: &key.PublicKey, signer: signer,
+	}, nil
+}
+
+// KeyID returns the active signing key ID.
+func (s *Service) KeyID() string { return s.kid }
+
+// Audience returns the registry service/audience string.
```
harishsurf, line 145: **MAJOR: Token validation does not require `jti` claim — enables silent token replay**

`Issue()` generates a random 128-bit `jti` but `hasRequiredClaims()` does not check for it. Intentional for backward compatibility with old OMR tokens (confirmed by test), but means stolen tokens are replayable for their full 1-hour lifetime with no detection mechanism.

Consider time-bounding the backward-compat exception (e.g., only accept jti-less tokens during a migration window) rather than permanently allowing replay.

internal/registry/distribution/token_handler.go
```diff
@@ -0,0 +1,274 @@
...
+	}, nil
+}
+
+func (h *TokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		w.Header().Set("Allow", http.MethodGet)
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+	if requestedService := r.URL.Query().Get("service"); requestedService != "" && requestedService != h.issuer.Audience() {
```
harishsurf, line 72: **MAJOR: Anonymous token requests trigger RSA signing on every request — CPU exhaustion vector**

When anonymous access is enabled, unauthenticated requests to `/v2/auth?scope=repository:anything:pull` get a 200 response with a signed JWT (empty scopes for private repos). Each response costs an RSA-2048 signing operation. An attacker can flood the endpoint to exhaust CPU.

Lower risk for mirror-registry (not internet-facing), but worth considering a lightweight response for anonymous requests with no granted scopes (e.g., return the empty-scope token unsigned, or cache a single empty-scope token).

internal/migrate/jwt_key.go
```diff
@@ -0,0 +1,159 @@
...
+		if !jwtauth.PublicKeysEqual(&existing.PublicKey, &key.PublicKey) {
+			if !replace {
+				return fmt.Errorf("existing native registry JWT key does not match the approved source key")
+			}
+			if err := jwtauth.ReplacePrivateKey(targetPath, key); err != nil {
+				return fmt.Errorf("replace source registry JWT key during resume: %w", err)
+			}
+		}
+		return nil
+	}
```
harishsurf, line 52: **MAJOR: Resume replaces mismatched target signing key with no log entry**

When `replace=true` (resume path) and the existing target key does not match the approved source key, `ReplacePrivateKey` is called silently. No log message indicates the replacement happened. An operator who manually placed a different key at the target path during a recovery attempt discovers broken token continuity with no audit trail.

**Fix:** Add `slog.Warn("replacing mismatched target JWT signing key during migration resume")` before the `ReplacePrivateKey` call.

.github/workflows/ci-go.yaml
```diff
@@ -380,6 +380,11 @@ jobs:
           echo '{"auths":{"localhost:8443":{"auth":"'$(echo -n "admin:$PASSWORD" | base64)'"}}}' \
             > ~/.docker/config.json
 
+      - name: Install registry test tools
+        run: |
+          sudo apt-get update
```
harishsurf, line 385: MAJOR: E2E migration test does not verify the migrated signing key is actually used. skopeo follows the Bearer challenge but the test does not assert the migrated key fingerprint matches the target. If LoadOrCreate generates a NEW key instead of the migrated one, the test passes but old OMR tokens would fail. Adding a source-vs-target key fingerprint comparison would catch this.

internal/registry/jwtauth/token.go
jbpratt: Investigated; no change. Requiring `jti` does not prevent replay by itself; enforcement requires persistent or shared one-time-use or revocation state. Validation already bounds audience and `iat`, `nbf`, and `exp`, while accepting tokens without `jti` is required for Python-to-Go signing-key continuity. A time-bounded cutoff would need a durable migration epoch and is a separate product and security design.

internal/registry/distribution/token_handler.go
jbpratt: Investigated; no change. Quay intentionally models registry delete as push permission. `jwtAccessFromDistribution` normalizes Distribution `delete` access to `push` before creating the Bearer challenge, and `Claims.allow` applies the same mapping during authorization. This server therefore challenges for `repository:<name>:push`, not `delete`. The Python `/v2/auth` endpoint has the same pull, push, and wildcard scope profile, so accepting explicit delete only in Go would diverge from that compatibility contract.

internal/migrate/jwt_key.go
jbpratt: Fixed in 53bf20f35. A `slog.Warn` now runs immediately before atomically replacing a mismatched target signing key during migration resume and includes the target path. The resume test captures the log and asserts both the warning and the successful approved-key replacement.

.github/workflows/ci-go.yaml
jbpratt: Fixed in 53bf20f35. The migration E2E job now fingerprints the DER-encoded public key from source `quay.pem` before migration and target `registry-jwt.pem` afterward, and fails if they differ. Read or parse failures also have explicit diagnostics. This catches `LoadOrCreate` accidentally generating a new signing identity.

internal/registry/jwtauth/key.go
jbpratt: Investigated; not changing this call. `registry-jwt.pem` is written directly into the registry data directory, which installer and migration intentionally create as `0750`; `MkdirAll(..., 0700)` would not tighten an existing directory and would make fresh direct calls inconsistent with the data-directory contract. The key and temporary file remain `0600`, and the filename is a fixed public constant, so group listing does not expose key material.

internal/registry/distribution/token_handler.go
jbpratt: Investigated; no in-process limiter is being added in this PR. This Go binary is scoped to single-site and offline mirror-registry deployments, and the existing Python `/v2/auth` path also relies on deployment-layer abuse controls. A correct limiter needs shared policy and state, or ingress/proxy enforcement; a per-process map here would be bypassable and changes product deployment policy. This should be a separate hardening feature rather than an auth correctness patch.

internal/system/hostname.go
```diff
@@ -0,0 +1,14 @@
...
+	"net"
+	"strings"
+)
+
+// HostnameWithoutPort removes a valid port and IPv6 brackets from a hostname.
+func HostnameWithoutPort(hostname string) string {
+	if host, _, err := net.SplitHostPort(hostname); err == nil {
+		hostname = host
+	}
+	return strings.Trim(hostname, "[]")
```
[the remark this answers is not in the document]
jbpratt: Fixed in 53bf20f35. `HostnameWithoutPort` now returns `(string, error)`, rejects empty, mismatched, or stray brackets and nonnumeric or out-of-range ports, validates bracketed input as IPv6, and propagates errors through Quadlet updates and TLS setup. Regression coverage includes each malformed form.

internal/registry/distribution/token_handler.go
jbpratt: Investigated; no change in this PR. An unsigned response would not be a valid Bearer token, and returning a signed token for the authorized intersection, including empty access, is token-service behavior. Caching would deliberately reuse one `jti` and still would not address authenticated Basic brute force. Direct anonymous pulls for public repositories already bypass token issuance; deployment-level rate limiting is the coherent control for untrusted networks.

internal/registry/distribution/auth.go
jbpratt: Investigated; this is intentional rather than a Bearer regression. `authorizeDistributionAccess` explicitly rejects `registry:catalog:*`, backed by `TestAuthorized_CatalogIsDeniedUntilQuayFilteredCatalogExists`, because the upstream distribution catalog is unfiltered. `skopeo list-tags` uses `/v2/<repo>/tags/list`, not `_catalog`, and is covered by the green E2E job. Supporting search or catalog needs a filtered Quay catalog implementation first.

internal/system/hostname.go
```diff
@@ -0,0 +1,65 @@
...
+		if host == "" {
+			return "", fmt.Errorf("hostname in %q is empty", hostname)
+		}
+		if strings.HasPrefix(hostname, "[") && !validIPv6Hostname(host) {
+			return "", fmt.Errorf("invalid bracketed IPv6 hostname %q", hostname)
+		}
+		if err := validateHostnamePort(port); err != nil {
+			return "", fmt.Errorf("invalid port in hostname %q: %w", hostname, err)
+		}
+		return host, nil
```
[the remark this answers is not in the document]
jbpratt: Fixed in 86152389a. The successful `net.SplitHostPort` branch now rejects any hostname containing brackets unless it is valid bracketed IPv6. Added port-qualified regressions for stray opening and closing brackets; targeted tests and the full Go linter pass.

.github/workflows/ci-go.yaml
jbpratt: Follow-up: run 29845265970 exposed that the E2E assertion used the stack config path for the key. Commit 8ef76876d aligns it with Quay runtime and migrator defaults at `/quay-registry/conf/quay.pem`; workflow YAML and the extracted shell block both validate.

harishsurf (approved):

PR #6577, opened 2026-07-20
PROJQUAY-9998: feat(secscan): add retry limiting via metadata_json for v1 and v2 scanners

kleesc:
Track scan retry count in metadata_json to prevent permanently-failing manifests from consuming indexing cycles indefinitely. After exceeding the configurable max retries (default 3), failed manifests are skipped. No migration required, uses the existing metadata_json JSONField.

Rollout https://github.com/quay/quay/pull/6633 first to canary

### DIFF
--- a/config.py
+++ b/config.py
@@ -553,7 +553,7 @@ class DefaultConfig(ImmutableConfig):
     SECURITY_SCANNER_V4_REINDEX_THRESHOLD = 300
 
     # Maximum number of scan retries per indexer hash before a manifest is skipped.
-    SECURITY_SCANNER_MAX_SCAN_RETRIES = 3
+    SECURITY_SCANNER_MAX_SCAN_RETRIES = 5
 
     # Security scanner V2 worker (lock-free, uses FOR UPDATE SKIP LOCKED)
     FEATURE_SECURITY_SCANNER_V2 = False
--- a/config.py
+++ b/config.py
@@ -552,6 +552,9 @@ class DefaultConfig(ImmutableConfig):
     # Minimum number of seconds before re-indexing a manifest with the security scanner.
     SECURITY_SCANNER_V4_REINDEX_THRESHOLD = 300
 
+    # Maximum number of scan retries per indexer hash before a manifest is skipped.
+    SECURITY_SCANNER_MAX_SCAN_RETRIES = 3
+
     # Security scanner V2 worker (lock-free, uses FOR UPDATE SKIP LOCKED)
     FEATURE_SECURITY_SCANNER_V2 = False
     SECURITY_SCANNER_V2_INDEXING_INTERVAL = 30
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -505,6 +505,7 @@ class V4SecurityScanner(SecurityScannerInterface):
 
             if exhausted_manifest_ids:
                 ManifestSecurityStatus.update(
+                    index_status=IndexStatus.SCAN_RETRIES_EXHAUSTED,
                     last_indexed=datetime.utcnow(),
                 ).where(
                     ManifestSecurityStatus.manifest_id.in_(exhausted_manifest_ids),
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -52,6 +52,7 @@ from util.secscan.v4.api import (
     ClairSecurityScannerAPI,
     InvalidContentSent,
     LayerTooLargeException,
+    Non200ResponseException,
 )
 from util.secscan.validator import V4SecurityConfigValidator
 
@@ -638,6 +639,36 @@ class V4SecurityScanner(SecurityScannerInterface):
             except InvalidContentSent as ex:
                 mark_manifest_unsupported(manifest)
                 logger.exception("Failed to perform indexing, invalid content sent")
+                continue
+            except Non200ResponseException as ex:
+                try:
+                    mss = ManifestSecurityStatus.get(ManifestSecurityStatus.manifest == candidate)
+                    metadata = mss.metadata_json or {}
+                    if metadata.get("last_failed_hash") != current_indexer_hash:
+                        metadata["retry_count"] = 1
+                    else:
+                        metadata["retry_count"] = metadata.get("retry_count", 0) + 1
+                except ManifestSecurityStatus.DoesNotExist:
+                    metadata = {"retry_count": 1}
+                metadata["last_failed_hash"] = current_indexer_hash
+
+                ManifestSecurityStatus.update(
+                    index_status=IndexStatus.FAILED,
+                    indexer_hash="server_error",
+                    error_json={
+                        "error": "non-200 response",
+                        "status_code": ex.response.status_code,
+                    },
+                    metadata_json=metadata,
+                    last_indexed=datetime.utcnow(),
+                ).where(
+                    ManifestSecurityStatus.manifest == candidate,
+                    ManifestSecurityStatus.index_status == IndexStatus.IN_PROGRESS,
+                ).execute()
+                logger.exception(
+                    "Failed to perform indexing, security scanner returned %s",
+                    ex.response.status_code,
+                )
                 continue
             except APIRequestFailure as ex:
                 ManifestSecurityStatus.update(
--- a/data/secscan_model/secscan_v4_model.py
+++ b/data/secscan_model/secscan_v4_model.py
@@ -374,7 +374,7 @@ class V4SecurityScanner(SecurityScannerInterface):
             reindex_threshold=reindex_threshold,
         )
 
-        self._index(iterator, reindex_threshold)
+        self._index(iterator, reindex_threshold, indexer_state.get("state", ""))
 
     def perform_indexing(self, start_token=None, batch_size=None):
         try:
@@ -408,13 +408,13 @@ class V4SecurityScanner(SecurityScannerInterface):
             reindex_threshold=reindex_threshold,
         )
 
-        self._index(iterator, reindex_threshold)
+        self._index(iterator, reindex_threshold, indexer_state.get("state", ""))
 
         return ScanToken(max_id + 1)
 
-    def _index(self, iterator, reindex_threshold):
+    def _index(self, iterator, reindex_threshold, current_indexer_hash):
         max_retries = self.app.config.get(
-            "SECURITY_SCANNER_V4_MAX_SCAN_RETRIES", DEFAULT_MAX_SCAN_RETRIES
+            "SECURITY_SCANNER_MAX_SCAN_RETRIES", DEFAULT_MAX_SCAN_RETRIES
         )
 
         def mark_manifest_unsupported(manifest):
@@ -496,7 +496,10 @@ class V4SecurityScanner(SecurityScannerInterface):
             exhausted_manifest_ids = []
             for row in retry_exhausted_query:
                 metadata = row.metadata_json or {}
-                if metadata.get("retry_count", 0) >= max_retries:
+                retry_count = metadata.get("retry_count", 0)
+                if metadata.get("last_failed_hash") != current_indexer_hash:
+                    retry_count = 0
+                if retry_count >= max_retries:
                     preempted_ids.add(row.manifest_id)
                     exhausted_manifest_ids.append(row.manifest_id)
 
@@ -542,6 +545,11 @@ class V4SecurityScanner(SecurityScannerInterface):
         for candidate, abt, num_remaining, should_skip in batched_iterator_with_preemption_check(
             iterator
         ):
+            if should_skip:
+                logger.debug("Manifest %d preempted by another worker (batch check)", candidate.id)
+                abt.set()
+                continue
+
             manifest = ManifestDataType.for_manifest(candidate, None)
             if manifest.is_manifest_list:
                 mark_manifest_unsupported(manifest)
@@ -561,8 +569,6 @@ class V4SecurityScanner(SecurityScannerInterface):
                 mark_manifest_unsupported(manifest)
                 continue
 
-            # We need to verify that the image layers are actual container images. Since Docker v2 schema 1 images
-            # cannot be anything other than container images, for them specifically we should skip the layer check.
             if manifest.media_type not in DOCKER_SCHEMA1_CONTENT_TYPES:
                 if not _has_container_layers(layers):
                     logger.info(
@@ -575,12 +581,6 @@ class V4SecurityScanner(SecurityScannerInterface):
                     )
                     mark_manifest_unsupported(manifest)
                     continue
-
-            # Check preemption status (computed via bulk query in batch)
-            if should_skip:
-                logger.debug("Manifest %d preempted by another worker (batch check)", candidate.id)
-                abt.set()
-                continue
 
             logger.debug(
                 "Indexing manifest [%d] %s/%s@%s"
@@ -639,18 +639,10 @@ class V4SecurityScanner(SecurityScannerInterface):
                 logger.exception("Failed to perform indexing, invalid content sent")
                 continue
             except APIRequestFailure as ex:
-                try:
-                    mss = ManifestSecurityStatus.get(ManifestSecurityStatus.manifest == candidate)
-                    metadata = mss.metadata_json or {}
-                except ManifestSecurityStatus.DoesNotExist:
-                    metadata = {}
-                metadata["retry_count"] = metadata.get("retry_count", 0) + 1
-
                 ManifestSecurityStatus.update(
                     index_status=IndexStatus.FAILED,
                     indexer_hash="api_failure",
                     error_json={"error": str(ex)},
-                    metadata_json=metadata,
                     last_indexed=datetime.utcnow(),
                 ).where(
                     ManifestSecurityStatus.manifest == candidate,
@@ -756,9 +748,13 @@ class V4SecurityScanner(SecurityScannerInterface):
                 try:
                     mss = ManifestSecurityStatus.get(ManifestSecurityStatus.manifest == candidate)
                     metadata = mss.metadata_json or {}
+                    if metadata.get("last_failed_hash") != current_indexer_hash:
+                        metadata["retry_count"] = 1
+                    else:
+                        metadata["retry_count"] = metadata.get("retry_count", 0) + 1
                 except ManifestSecurityStatus.DoesNotExist:
-                    metadata = {}
-                metadata["retry_count"] = metadata.get("retry_count", 0) + 1
+                    metadata = {"retry_count": 1}
+                metadata["last_failed_hash"] = current_indexer_hash
 
                 ManifestSecurityStatus.update(
                     index_status=IndexStatus.FAILED,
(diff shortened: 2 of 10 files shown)


data/secscan_model/secscan_v4_model.py
```diff
@@ -611,11 +639,18 @@ def batched_iterator_with_preemption_check(iterator, batch_size=20):
                 logger.exception("Failed to perform indexing, invalid content sent")
                 continue
             except APIRequestFailure as ex:
-                # Mark as FAILED instead of deleting to preserve scan history
+                try:
+                    mss = ManifestSecurityStatus.get(ManifestSecurityStatus.manifest == candidate)
+                    metadata = mss.metadata_json or {}
+                except ManifestSecurityStatus.DoesNotExist:
+                    metadata = {}
+                metadata["retry_count"] = metadata.get("retry_count", 0) + 1
```
jbpratt: Blocking: `APIRequestFailure` covers connection errors, timeouts, and non-2xx responses, so this increments the same counter used as a permanent stop condition. If the state endpoint works but indexing fails three times during a transient or partial Clair outage, the manifest is never retried after recovery. The V2 `_mark_failed` path has the same behavior. Please keep transport/service failures retryable with backoff, or track them separately from terminal manifest/index errors and provide an automatic re-arm path.

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -154,15 +159,35 @@ def _find_and_claim_batch(self, batch_size, reindex_threshold, stale_threshold,
...
 
-            row_ids = [r.id for r in rows]
             now = datetime.utcnow()
+
+            eligible = []
+            exhausted_ids = []
+            for r in rows:
+                if (
+                    r.index_status == IndexStatus.FAILED
+                    and (r.metadata_json or {}).get("retry_count", 0) >= max_retries
```
jbpratt: The selector intentionally includes rows whose `indexer_hash` differs from the current Clair state, but this unconditional exhaustion check filters them afterward. A manifest exhausted under one indexer version therefore remains permanently failed after an upgrade that may fix the cause. Please scope the retry budget to the indexer state, or reset/ignore exhaustion when the current hash changes.

jbpratt (commented):
Targeted deep review of the retry-limiting behavior. The highest-risk issue is permanent exhaustion after transient scanner failures; the remaining inline comments cover recovery, batching cost, configuration semantics, and test coverage.

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -154,15 +159,35 @@ def _find_and_claim_batch(self, batch_size, reindex_threshold, stale_threshold,
             if not rows:
                 return []
 
-            row_ids = [r.id for r in rows]
             now = datetime.utcnow()
+
+            eligible = []
+            exhausted_ids = []
+            for r in rows:
```
jbpratt, line 167: `LIMIT batch_size` has already been applied before this Python filtering. An old exhausted window can therefore return no eligible work even when valid rows follow it; for example, 5,000 exhausted rows with a batch size of 50 take roughly 100 worker cycles to drain. Please exclude exhausted rows before the limit, or continue claiming windows until the eligible batch is filled, and add a mixed exhausted/eligible regression test.

data/secscan_model/secscan_v4_model.py
```diff
@@ -479,7 +484,30 @@ def batch_preemption_check(candidate_ids):
...
+                ManifestSecurityStatus.metadata_json,
+            ).where(
+                ManifestSecurityStatus.manifest_id.in_(candidate_ids),
+                ManifestSecurityStatus.index_status == IndexStatus.FAILED,
+            )
+            exhausted_manifest_ids = []
+            for row in retry_exhausted_query:
+                metadata = row.metadata_json or {}
+                if metadata.get("retry_count", 0) >= max_retries:
+                    preempted_ids.add(row.manifest_id)
```
jbpratt, line 504: This sets `should_skip`, but `_index` only checks it after constructing `ManifestDataType` and calling `list_manifest_layers(..., include_placements=True)`. Exhausted rows therefore still incur manifest parsing and per-layer placement queries whenever selected. Please short-circuit before manifest/layer resolution; a dedicated exhaustion log/metric would also avoid reporting this permanent state as ordinary worker preemption.

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -118,6 +119,10 @@ def perform_indexing(self, start_token=None, batch_size=None):
         )
 
     def _find_and_claim_batch(self, batch_size, reindex_threshold, stale_threshold, indexer_hash):
+        max_retries = self.app.config.get(
+            "SECURITY_SCANNER_V2_MAX_SCAN_RETRIES", DEFAULT_MAX_SCAN_RETRIES
```
jbpratt: This V2 setting governs the same shared `metadata_json.retry_count` as V4, while `SecurityWorker` still schedules both legacy and V2 indexing. If V2 is set to 1 and V4 remains 3, V4 continues retrying through 3, so this knob does not enforce its configured ceiling. Please use one shared limit or make the scheduling/counters independent.

data/secscan_model/secscan_v4_model.py
```diff
@@ -412,6 +413,10 @@ def perform_indexing(self, start_token=None, batch_size=None):
         return ScanToken(max_id + 1)
 
     def _index(self, iterator, reindex_threshold):
+        max_retries = self.app.config.get(
+            "SECURITY_SCANNER_V4_MAX_SCAN_RETRIES", DEFAULT_MAX_SCAN_RETRIES
```
jbpratt: Neither new retry-limit key is declared in `DefaultConfig` or `CONFIG_SCHEMA`. Raw YAML values can therefore remain strings or null, and the later `int >= max_retries` comparison raises `TypeError`, aborting indexing cycles. Please add typed, bounded defaults/schema entries for both keys and update the supported config surfaces.

web/playwright/e2e/tags/security-scan.spec.ts
```diff
@@ -153,6 +153,29 @@ test.describe(
       await expect(vulnRows.first()).toBeVisible();
     });
 
+    test('security status API returns valid scan status', async ({
```
jbpratt: This test does not exercise retry limiting: it pushes a healthy image, calls an unchanged read endpoint, and accepts every ordinary status. The base implementation already satisfies these assertions, so retry-count, exhaustion, and recovery regressions still pass. Please remove this unrelated E2E case or replace it with a harness that induces repeated failures and proves an exhausted manifest no longer blocks eligible work.

data/secscan_model/secscan_v4_model.py
kleesc: APIRequestFailure handlers in both V1 and V2 no longer touch metadata_json — they mark FAILED
with indexer_hash="api_failure" but don't increment retry_count

data/secscan_model/secscan_v4_model_v2.py
kleesc: last_failed_hash is now stored in metadata_json. Previously-exhausted manifests get a new set of retries

data/secscan_model/secscan_v4_model_v2.py
kleesc: last_failed_hash should address this. After a indexer update, previously-exhausted rows become eligible again. For rows exhausted under the current hash, the last_indexed bump on exhausted rows pushes them to the back of the ORDER BY last_indexed ASC queue, so subsequent cycles pick newer rows

data/secscan_model/secscan_v4_model.py
kleesc: Done

data/secscan_model/secscan_v4_model_v2.py
kleesc: Both use SECURITY_SCANNER_MAX_SCAN_RETRIES now

data/secscan_model/secscan_v4_model.py
kleesc: Done. Defaults to 3

web/playwright/e2e/tags/security-scan.spec.ts
kleesc: Removed it

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -282,6 +322,19 @@ def _index_manifest_by_id(self, manifest_id, repository_id):
...
+                mss = ManifestSecurityStatus.get(ManifestSecurityStatus.manifest == candidate)
+                old_metadata = mss.metadata_json or {}
+                if old_metadata.get("last_failed_hash") != current_indexer_hash:
+                    metadata["retry_count"] = 1
+                else:
+                    metadata["retry_count"] = old_metadata.get("retry_count", 0) + 1
+            except ManifestSecurityStatus.DoesNotExist:
+                metadata["retry_count"] = 1
+            metadata["last_failed_hash"] = current_indexer_hash
+
```
lechuk47, line 351: Is it possible to avoid the `metadata_json` attribute? 

Given that related data is already stored as first-class columns (e.g., indexer_hash, indexer_version, last_indexed), introducing new data into metadata_json is inconsistent. It creates two tiers of data on the same model — one queryable and schema-enforced, the other opaque. If a field is worth storing, it should be a  proper column.

kleesc: In this case it would be ok to put this in metadata_json. Since this needs to go in a z-stream, it should be backward compatible, no migration if possible (@jbpratt). 

We can always move the same data in a new column if we find that we need to query based on these fields.

jbpratt: Yeah this needs to go back to 3.15 :disappointed:

data/secscan_model/secscan_v4_model_v2.py
```diff
@@ -250,7 +281,15 @@ def _index_manifest_by_id(self, manifest_id, repository_id):
             logger.exception("Failed to index: invalid content sent")
             return
         except APIRequestFailure as ex:
-            self._mark_failed(candidate.id, "api_failure", {"error": str(ex)})
+            ManifestSecurityStatus.update(
```
bcaton85, line 298: This handler was changed from calling `_mark_failed()` to a raw UPDATE that doesn't increment `retry_count`. The same pattern exists in `secscan_v4_model.py:641-651`.

The JIRA ticket (PROJQUAY-9998) specifically describes Clair returning HTTP 500 on each scan attempt as the root cause. HTTP 500 from Clair raises `APIRequestFailure` (via `Non200ResponseException` in `util/secscan/v4/api.py:262`). By excluding `APIRequestFailure` from retry counting, manifests that consistently fail with Clair 500s will still retry indefinitely — the exact scenario the customer escalated.

The distinction between transient API failures (network blip) and persistent ones (Clair can't process this manifest) isn't determinable at the point of failure. The retry limit should apply to all failure types to satisfy the ticket. Manifests that start succeeding after a transient failure will have their `retry_count` reset to `{}` on success, so counting API failures won't cause false exhaustion.

config.py
```diff
@@ -552,6 +552,9 @@ def create_transaction(db):
     # Minimum number of seconds before re-indexing a manifest with the security scanner.
     SECURITY_SCANNER_V4_REINDEX_THRESHOLD = 300
```
bcaton85, line 554: PROJQUAY-9998 says "Quay should try to scan an image at most 5 times." The default here is 3. While configurable, the default should match the ticket requirement unless there's a documented reason to differ.

util/config/schema.py
```diff
@@ -858,6 +858,11 @@
             "description": "The number of seconds between indexing cycles in the V2 security scanner. Defaults to 30.",
             "x-example": 30,
         },
+        "SECURITY_SCANNER_MAX_SCAN_RETRIES": {
+            "type": "number",
```
bcaton85: `"type": "number"` allows float values like 2.5 for a retry count. Should be `"type": "integer"`.

bcaton85:
The retry_count increment pattern (read MSS, check hash, increment/reset) is copy-pasted across 4 locations:
- `secscan_v4_model.py:748-757` (unknown_state handler)
- `secscan_v4_model.py:776-787` (Index_Error FAILED handler)
- `secscan_v4_model_v2.py:325-336` (general FAILED handler)
- `secscan_v4_model_v2.py:442-452` (`_mark_failed` method)

Consider extracting a `_build_retry_metadata(candidate_id, current_indexer_hash)` helper on each class. The v2 model already has `_mark_failed` but the Index_Error fall-through path bypasses it.

data/secscan_model/secscan_v4_model.py
```diff
@@ -735,6 +773,19 @@ def batched_iterator_with_preemption_check(iterator, batch_size=20):
                 )
                 continue
 
+            metadata = {}
+            if index_status == IndexStatus.FAILED:
+                try:
+                    mss = ManifestSecurityStatus.get(ManifestSecurityStatus.manifest == candidate)
```
bcaton85, line 811: The MSS row is read here to get the current `retry_count`, but the delete+create that consumes this value happens in a separate transaction at lines 789-801. Between the read and the transaction, another worker could modify the row, causing a lost update on `retry_count`. The same pattern exists in `secscan_v4_model_v2.py:327-350`.

Consider moving the read inside the transaction, or using an atomic UPDATE with a subquery instead of the delete+create pattern.

data/secscan_model/secscan_v4_model.py
```diff
@@ -479,7 +484,33 @@ def batch_preemption_check(candidate_ids):
...
+                metadata = row.metadata_json or {}
+                retry_count = metadata.get("retry_count", 0)
+                if metadata.get("last_failed_hash") != current_indexer_hash:
+                    retry_count = 0
+                if retry_count >= max_retries:
+                    preempted_ids.add(row.manifest_id)
+                    exhausted_manifest_ids.append(row.manifest_id)
+
+            if exhausted_manifest_ids:
+                ManifestSecurityStatus.update(
```
bcaton85, line 508: Exhausted manifests get their `last_indexed` bumped here to push them past the reindex threshold window, but they remain `IndexStatus.FAILED`. Every `reindex_threshold` cycle, they'll be selected again, checked, found exhausted, bumped, and dropped — a recurring no-op loop.

The JIRA ticket says "it must fail the image scanning and set the appropriate error message." Consider a terminal status (e.g., `IndexStatus.SCAN_RETRIES_EXHAUSTED`) or at minimum a distinct `indexer_hash` / `error_json` so operators and the UI can distinguish "failed, will retry" from "failed, gave up." The same applies to the v2 model's equivalent at `secscan_v4_model_v2.py:176-179`.

bcaton85 (changes requested):
Some bugs need to be addressed

data/secscan_model/secscan_v4_model_v2.py
kleesc: Retry counting only increments on report-level errors, where Clair actually evaluated the manifest and failed. APIRequestFailure (5xx, timeouts, connection errors) means Clair never rendered a verdict, so it marks FAILED without advancing the counter. The last_indexed bump still pushes these to the back of the queue each cycle so they don't block other work. Manifests Clair genuinely can't process come back as Index_Error and hit the 5-retry limit.

data/secscan_model/secscan_v4_model.py
kleesc: Two workers can't be processing the same manifest simultaneously. Even if that's the case, the worst case is that a manifest might get 6 retries instead on 5.

data/secscan_model/secscan_v4_model.py
kleesc: Exhausted manifests are now marked SCAN_RETRIES_EXHAUSTED, which removes them from the FAILED retry branch. When the Clair indexer hash changes, last_failed_hash resets the retry_count to 0.

data/secscan_model/secscan_v4_model_v2.py
bcaton85: will this actually resolve the issue described in the JIRA? it seems if they're getting 5XX's like the jira describes the retry logic won't even be reached

data/secscan_model/secscan_v4_model.py
bcaton85: this seems to be specific to v1, for v2 this isn't an issue. SECURITY_SCANNER_V4_LOCK fixes this but its set to false by default. Not sure why we don't have that enabled?

data/secscan_model/secscan_v4_model.py
bcaton85: we aren't rendering that the retries have exhausted in the api and ui anywhere. The jira asks for rendering a separate error message for exhausted retries. Is that worth implementing?

data/secscan_model/secscan_v4_model_v2.py
kleesc: I'm not sure that we should be counting 5XX as a retry. The external service returning 5XX means that something went wrong external to the actual manifest itself, and shouldn't count.

kleesc: Only 200s or 400s

data/secscan_model/secscan_v4_model.py
```diff
@@ -718,11 +777,22 @@ def batched_iterator_with_preemption_check(iterator, batch_size=20):
             elif report["state"] == IndexReportState.Index_Error:
                 index_status = IndexStatus.FAILED
             else:
-                # Unknown state - mark as FAILED instead of deleting to preserve scan history
+                try:
```
bcaton85, line 780: Nit: no test covers the `unknown_state` retry-tracking path. Index_Error and Non200ResponseException both have dedicated tests, but this is the 3rd inline copy of the retry logic in V1 and could silently diverge. A test like `test_unknown_state_increments_retry_count` that mocks a report with an unrecognized state value would close the gap.

data/secscan_model/secscan_v4_model.py
```diff
@@ -479,7 +485,34 @@ def batch_preemption_check(candidate_ids):
...
             )
 
-            return {row.manifest_id for row in preempted_query}
+            preempted_ids = {row.manifest_id for row in preempted_query}
+
+            retry_exhausted_query = ManifestSecurityStatus.select(
+                ManifestSecurityStatus.manifest_id,
+                ManifestSecurityStatus.metadata_json,
+            ).where(
+                ManifestSecurityStatus.manifest_id.in_(candidate_ids),
```
bcaton85, line 494: Should-fix: no test covers the `last_failed_hash` reset mechanism. Every retry-tracking path checks `if metadata.get("last_failed_hash") != current_indexer_hash` to reset `retry_count`, but no test exercises it. This is the safety valve that prevents permanent blacklisting after a Clair update — if it breaks, exhausted manifests would never get fresh retries even when the underlying Clair bug is fixed.

A test that fails a manifest 4 times with hash `"abc"`, then switches to hash `"def"` and confirms the manifest is retried (retry_count resets to 0 in the skip check), would cover this for both the batch_preemption_check path and the _find_and_claim_batch path.

data/secscan_model/secscan_v4_model.py
```diff
@@ -223,7 +225,7 @@ def load_security_information(
         except ManifestSecurityStatus.DoesNotExist:
             return SecurityInformationLookupResult.with_status(ScanLookupStatus.NOT_YET_INDEXED)
```
bcaton85, line 226: Nit: no test verifies that a manifest with `SCAN_RETRIES_EXHAUSTED` status returns `FAILED_TO_INDEX` from `load_security_information`. This is also the natural place to surface a distinct error message per the JIRA acceptance criteria ("set the appropriate error message") if that is planned as a follow-up.

bcaton85 (approved):

PR #6578, opened 2026-07-20
NO-ISSUE: fix(playwright): update BUSYBOX_IMAGE to valid manifest list digest

sridipta:
## Summary
- The pinned `quay.io/prometheus/busybox` digest (`467c84...`) was deleted upstream again, breaking all Playwright E2E tests that push images via skopeo
- Updated to the current manifest list digest (`065fa27...`) which resolves correctly for both single-platform (`--override-os/arch`) and multi-arch (`--all`) skopeo copies
- This is the same root cause as #6549 — the upstream image keeps getting garbage-collected

## JIRA
NO-ISSUE

## Test Plan
- [ ] CI Playwright E2E tests pass with the new digest
- [ ] Cherry-pick to `redhat-3.18` to unblock #6537

### DIFF
--- a/web/playwright/utils/container.ts
+++ b/web/playwright/utils/container.ts
@@ -21,7 +21,7 @@ const execFileAsync = promisify(execFile);
 const REGISTRY_HOST = new URL(API_URL).host;
 
 const BUSYBOX_IMAGE =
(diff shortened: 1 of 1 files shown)


sridipta:
Closing — superseded by #6566 which switched to `:latest` tag instead of pinned digest, avoiding the recurring digest-deletion problem entirely.

PR #6605, opened 2026-07-21
PROJQUAY-10587: fix(mirroring): prevent SSRF in repository sources

shaonrh:
- Prevent repository mirror sources from targeting private, reserved, or otherwise unsafe destinations unless allowed through `SSRF_ALLOWED_HOSTS`.
 - Validate sources when mirror configurations are created or updated and again immediately before Skopeo execution.
 - Return HTTP 400 for malformed `sync_start_date` values during mirror creation.

Existing mirrors using private or internal registries must add the source hostname or CIDR to `SSRF_ALLOWED_HOSTS` before upgrading. Otherwise, those mirrors will stop syncing. Public and already-allowlisted registries require no action.

### DIFF
--- a/endpoints/api/mirror.py
+++ b/endpoints/api/mirror.py
@@ -359,7 +359,10 @@ class RepoMirrorResource(RepositoryParamResource):
                 400,
             )
 
-        data["sync_start_date"] = self._string_to_dt(data["sync_start_date"])
+        try:
+            data["sync_start_date"] = self._string_to_dt(data["sync_start_date"])
+        except ValueError:
+            return {"detail": "Incorrect DateTime format for sync_start_date."}, 400
 
         rule = model.repo_mirror.create_rule(repo, data["root_rule"]["rule_value"])
         del data["root_rule"]
--- a/endpoints/api/mirror.py
+++ b/endpoints/api/mirror.py
@@ -25,6 +25,27 @@ from endpoints.api import (
 from endpoints.exception import InvalidRequest, NotFound
 from util.audit import track_and_log, wrap_repository
 from util.names import parse_robot_username
+from util.security.ssrf import SSRFBlockedError, validate_external_registry_reference
+
+SSRF_GENERIC_ERROR = "The provided registry location is not allowed"
+
+
+def _get_ssrf_allowed_hosts():
+    return app.config.get("SSRF_ALLOWED_HOSTS", [])
+
+
+def _validate_external_reference(reference):
+    """Validate a repository mirror source and normalize SSRF errors."""
+    try:
+        validate_external_registry_reference(
+            reference,
+            allowed_hosts=_get_ssrf_allowed_hosts(),
+        )
+    except SSRFBlockedError:
+        raise InvalidRequest(SSRF_GENERIC_ERROR)
+    except ValueError as e:
+        raise InvalidRequest(str(e))
+
 
 common_properties = {
     "is_enabled": {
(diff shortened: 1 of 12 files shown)


ibazulic:
How will not allowing the use of proxy help clients who need to use a proxy server to access external registries?

shaonrh:
> How will not allowing the use of proxy help clients who need to use a proxy server to access external registries?

@ibazulic The PR wasn't ready for review yet, the implementation you mentioned have been reverted. Can you please review to make sure your concern with proxy has been addressed. Thanks.

util/security/ssrf.py
```diff
@@ -128,6 +134,86 @@ def _is_allowed(hostname: str, ip_str: Optional[str], allowed_hosts: List[str])
     return False
 
 
+def validate_external_registry_reference(
```
Marcusk19, line 140: nit: I'll add a note that gpt caught that the `_is_ip_blocked()` helper function inside this does not account for ipv6 ranges. While I determined this isn't a blocker since it's not part of the diff, I'll leave it up to you if you want to address it in this PR or not

endpoints/api/mirror.py
```diff
@@ -327,44 +361,40 @@ def post(self, namespace_name, repository_name):
...
-                    return {
-                        "detail": "Architecture filtering requires FEATURE_SPARSE_INDEX to be enabled."
-                    }, 400
-                try:
-                    model.repo_mirror.validate_architecture_filter(arch_filter)
+            mirror = model.repo_mirror.enable_mirroring_for_repository(
+                repo,
+                root_rule=rule,
+                internal_robot=robot,
+                allowed_hosts=_get_ssrf_allowed_hosts(),
```
Marcusk19: **Apply the proxy restriction to mirror creation as well.**

`PUT` now rejects non-null `http_proxy`, `https_proxy`, and `no_proxy` values, but `POST` forwards `external_registry_config` here unchanged. The normal initial-setup path can therefore persist a private proxy such as `https_proxy: http://10.0.0.5:8080`; the worker later places it in the Skopeo environment, bypassing the restriction enforced for updates.

Please run the same proxy validation before this transaction for `POST` (preferably through a helper shared with `PUT`) and add a create-path regression test showing that non-null proxy values are rejected while `null` remains accepted.

endpoints/api/mirror.py
```diff
@@ -327,44 +361,40 @@ def post(self, namespace_name, repository_name):
...
-                "repo_mirror_config_changed",
-                wrap_repository(repo),
-                changed="external_reference",
-                to=data["external_reference"],
-            )
-            return "", 201
-        else:
-            # TODO: Determine appropriate Response
-            return {"detail": "RepoMirrorConfig already exists for this repository."}, 409
+                track_and_log(
```
Marcusk19: **Keep non-transactional audit side effects outside the database transaction.**

This call now runs inside `db_transaction()`, but `track_and_log()` can publish a Redis user event, queue Mixpanel analytics, and write through external log backends such as Splunk in addition to table-backed logging. Those side effects cannot roll back with Peewee: a later failure may roll back the mirror changes after an event was already emitted, and the transaction remains open while logging work runs.

Please collect the successful change events during the transaction and emit them after commit. If the table-backed action-log row must be atomic with the configuration, that database write should be separated from the Redis, analytics, and external-log publications. This applies to the corresponding calls in the PUT transaction as well.

Marcusk19 (changes requested):
Two potentially blocking issues caught by my agent during review. One other issue that is a gap for this fix, but imo would be fine if we address it in a separate PR. Inline comments below:

endpoints/api/mirror.py
shaon: Good catch. I've splitted proxy destination validation into PROJQUAY-12421 so we can handle repository and organization mirrors consistently across create, update, and worker execution. This PR is now limited to source-registry validation.

endpoints/api/mirror.py
shaon: Agreed. I removed the endpoint-wide POST/PUT transaction and restored the existing per-operation transaction behavior. Validation now runs before mutations, and each track_and_log() call runs after its corresponding model write completes, so external audit side effects are not executed inside an open endpoint transaction.

Marcusk19 (approved):
lgtm

PR #6645, opened 2026-07-24
PROJQUAY-12190: fix(mirroring): use separate authfiles for same-registry mirror

bcaton85:
## Summary

- Fix repository mirroring failure when source and destination are on the same Quay registry
- Replace single `--authfile` with `--src-authfile` and `--dest-authfile` skopeo flags in `copy()` and `copy_by_digest()`
- Add `_src_dest_authfiles()` context manager that creates independent temp files per credential set

## Root Cause

PR #5813 ([PROJQUAY-7340](https://redhat.atlassian.net/browse/PROJQUAY-7340)) replaced `--src-creds`/`--dest-creds` CLI flags with a single `--authfile` to prevent credentials from appearing in process listings. The authfile uses the registry hostname as the lookup key. When both source and destination share a hostname, `create_authfile_content()` produces a dict where the destination credentials overwrite the source credentials, causing `unauthorized` errors on the source repo.

## Fix

Skopeo supports `--src-authfile` and `--dest-authfile` flags (since v1.5.0, all Quay releases ship >= 1.22.2). Each side now gets its own temp file with exactly one credential entry, making key collisions structurally impossible. Single-credential methods (`tags`, `inspect`, `inspect_raw`) are unchanged.

## Test plan

- [x] Unit test documenting the bug (`test_create_authfile_content_same_registry_overwrites`)
- [x] Unit tests for `_src_dest_authfiles()` — distinct paths and same-registry credential isolation
- [x] Updated `_assert_skopeo_args()` to validate `--src-authfile`/`--dest-authfile` for copy commands and `--authfile` for non-copy commands
- [ ] CI unit tests
- [ ] CI worker integration tests

### DIFF
--- a/util/repomirror/skopeomirror.py
+++ b/util/repomirror/skopeomirror.py
@@ -94,6 +94,13 @@ def _authfile(entries: list) -> Iterator[str]:
         yield tmp.name
 
 
+@contextmanager
+def _src_dest_authfiles(src_entries: list, dest_entries: list) -> Iterator[tuple[str, str]]:
+    """Create separate source and destination authfiles, yielding both paths."""
+    with _authfile(src_entries) as src_path, _authfile(dest_entries) as dest_path:
+        yield src_path, dest_path
+
+
 class SkopeoMirror(object):
 
     # No DB calls here: This will be called from a separate worker that has no connection except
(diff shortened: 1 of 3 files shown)


sudipshil9862 (dismissed):

Marcusk19 (dismissed):
lgtm pending CI checks

Marcusk19 (approved):

PR #6648, opened 2026-07-27
PROJQUAY-12113: fix(ui): Quay UI does not display shield icon for images signed with cosign

nasonawa:
## Summary

Fixes missing cosign shield icons in both the legacy Angular UI and the V2 React UI for images signed with Cosign v3+.

Cosign v3 stores signatures as OCI referrers (`Manifest.subject` + Sigstore artifact types) without visible `sha256-<digest>.sig` tags. Both UIs previously detected signatures only by scanning sibling `.sig` tags client-side, so signed images never showed the shield.

This change:
- Adds server-side batch lookup of Cosign signatures (OCI referrers and classic `.sig` tags) and enriches the tags list API with `cosign_signature_manifest_digest` / `cosign_signature_tag`
- Updates Angular and React UIs to treat either field as “signed” when rendering the shield
- Preserves API-provided cosign fields in the React enricher, with client-side `.sig` scanning as a fallback
- `data/model/oci/manifest.py` — `lookup_cosign_signatures_for_digests()` for referrers + classic `.sig` tags
- `endpoints/api/tag.py` — enrich `ListRepositoryTags` responses with cosign fields
- `static/.../repo-panel-tags.{js,html}` — show shield when digest and/or tag is present
- `web/src/libs/cosign.ts` + `TagsTable.tsx` — same for V2 UI; link to `.sig` tag when available, otherwise show signature manifest digest

## Test plan
1. Open Quay UI and navigate to a repository you can push to
2. Push an image to the repository (for example with `podman push` / `docker push`)
3. Sign the image with Cosign v3 (for example `cosign sign $IMAGE`)
4. Confirm the signature with `cosign verify` (using the appropriate identity / issuer flags)
5. Reload the repository tags page in the **legacy UI** and confirm the shield icon appears next to the signed tag
6. Open the same repository in the **V2 UI** and confirm the shield icon appears next to the signed tag
7. Expand the signed tag row and confirm cosign signature details are shown
8. Confirm an unsigned tag in the same repository does **not** show a shield icon

### DIFF
--- a/web/playwright/e2e/tags/signatures.spec.ts
+++ b/web/playwright/e2e/tags/signatures.spec.ts
@@ -1,7 +1,11 @@
-import {test, expect} from '../../fixtures';
+import {test, expect, uniqueName} from '../../fixtures';
 import {TEST_USERS} from '../../global-setup';
 import {ApiClient} from '../../utils/api';
-import {pushImage} from '../../utils/container';
+import {
+  pushImage,
+  isCosignAvailable,
+  cosignSignWithKey,
+} from '../../utils/container';
 
 test.describe(
   'Tags - Show/Hide Signatures',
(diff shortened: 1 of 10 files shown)


static/directives/repo-view/repo-panel-tags.html
```diff
@@ -326,8 +326,12 @@ <h3 class="tab-header"><span class="hidden-xs">Repository </span>Tags</h3>
              manifest-digest="tag.manifest_digest" cache="labelCache"></div>
 
         <!-- Cosign Signatures -->
-        <div class="cosign-signature-row" ng-if="tag.cosign_signature_tag">
-          <cosign-link repository="repository" image-id="tag.image_id" manifest-digest="tag.cosign_signature_manifest_digest" tag-name="tag.cosign_signature_tag" tag-name="tag.cosign_signature_tag" signed-image-id="tag.image_id" signed-manifest-digest="tag.manifest_digest"></cosign-link>
+        <div class="cosign-signature-row" ng-if="tag.cosign_signature_tag || tag.cosign_signature_manifest_digest">
```
jbpratt: You can drop old UI changes, we don't modify it anymore unless major bug is found/introduced

web/playwright/e2e/tags/signatures.spec.ts
```diff
@@ -149,5 +150,135 @@ test.describe(
...
+        }
+        await new Promise((r) => setTimeout(r, 1000));
+      }
+      if (!digest) {
+        throw new Error('Pushed tag was not indexed after 10 attempts');
+      }
+
+      // Cosign v3-style referrer: sigstore artifact type, no visible .sig tag
+      const fixturesDir = path.resolve(__dirname, '../../fixtures/oras');
+      orasAttach(
```
jbpratt: It may be difficult, but it'd be nice if these were truly signed images (maybe we can push out a set of test images)

nasonawa: done, added a cosign install step in the CI,  now we use the cosign binary to sign the images.

jbpratt (changes requested):
Found three targeted issues:

1. `data/model/oci/manifest.py:229-230` treats every Sigstore bundle referrer as an image signature. Cosign v3 writes both `cosign sign` and `cosign attest` with `application/vnd.dev.sigstore.bundle.v0.3+json`; the intent is in `dev.sigstore.bundle.predicateType`, which this lookup does not inspect. Images with only attestations will receive the signed shield.

2. `web/playwright/e2e/tags/signatures.spec.ts:275-278` does not prove referrer-only storage. The tags UI hides `.sig` tags by default, while the API poll accepts `cosign_signature_manifest_digest` for both classic tags and referrers. Assert that `cosign_signature_tag` is absent from the API response, or show signatures before asserting no `.sig` tag.

3. `web/playwright/utils/container.ts:534-538` claims offline signing but leaves Cosign v3.0.2 `--tlog-upload` at its default `true`; the test depends on public Rekor. Pass `--tlog-upload=false`.

---

jbpratt (approved):

PR #6650, opened 2026-07-27
PROJQUAY-11682: fix(autoprune): Autopruner matching on .sig tags instead of filtering them

nasonawa:
## Summary

Cosign tag-schema signatures (`.sig` ) are ordinary tags (`hidden=false`, no OCI `subject`), so the autopruner treated them like image tags. That meant they consumed `NUMBER_OF_TAGS` slots, aged out under `CREATION_DATE`, and could leave orphans when the parent image (or the signature) was pruned.

This change:
- Excludes Cosign tag-schema names (`sha256-<digest>.{sig}`) from both autopruner fetch paths so they are not matched or counted
- Cascades Cosign tag cleanup when a non-Cosign tag is deleted
- Fixes `exclude_tags` filtering in `fetch_paginated_autoprune_repo_tags_by_number` (the `where()` result was previously discarded)

## Test plan

- Push image tags plus Cosign `.sig` tags, set a keep-N policy, run autopruner, and confirm Cosign tags are not counted or pruned
- Create old Cosign tags past a creation-date policy threshold, run autopruner, and confirm they are not pruned by age
- Delete an image tag and confirm its matching Cosign `.sig` tags are also removed
- Confirm normal non-Cosign tags still prune correctly under both policy types

### DIFF
--- a/data/model/oci/tag.py
+++ b/data/model/oci/tag.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import re
 import uuid
 from calendar import timegm
 
@@ -42,6 +43,51 @@ from util.timedeltastring import convert_to_timedelta
 logger = logging.getLogger(__name__)
 
 GC_CANDIDATE_COUNT = 500  # repositories
+
+COSIGN_TAG_SCHEMA_PATTERN = r"^sha256-[a-f0-9]{64}\.(sig|att|sbom)$"
+COSIGN_TAG_SCHEMA_RE = re.compile(COSIGN_TAG_SCHEMA_PATTERN)
+COSIGN_TAG_SCHEMA_PREFIX = "sha256-"
+COSIGN_TAG_SCHEMA_DIGEST_LEN = 64
+COSIGN_TAG_SCHEMA_SUFFIXES = ("sig", "att", "sbom")
+
+
+def is_cosign_tag_schema(tag_name):
+    return bool(COSIGN_TAG_SCHEMA_RE.fullmatch(tag_name))
+
+
+def _lowercase_hex_digest_clause(digest_expr):
+    """True when digest_expr is exactly lowercase hexadecimal (empty after stripping hex digits)."""
+    stripped = digest_expr
+    for ch in "0123456789abcdef":
+        stripped = fn.REPLACE(stripped, ch, "")
+    return stripped == ""
+
+
+def _not_cosign_tag_schema_clause(model=Tag):
+    """
+    SQL clause matching is_cosign_tag_schema without REGEXP (SQLite-safe).
+
+    Requires sha256- prefix, exactly 64 lowercase hex digest chars, and a supported suffix.
+    Uses SUBSTR equality so prefix/suffix checks are case-sensitive (unlike LIKE).
+    Accepts a Tag model or alias so the clause can be used in subqueries.
+    """
+    prefix = COSIGN_TAG_SCHEMA_PREFIX
+    digest_len = COSIGN_TAG_SCHEMA_DIGEST_LEN
+    digest_expr = fn.SUBSTR(model.name, len(prefix) + 1, digest_len)
+    hex_digest = _lowercase_hex_digest_clause(digest_expr)
+    prefix_match = fn.SUBSTR(model.name, 1, len(prefix)) == prefix
+    suffix_clauses = []
+    for suffix in COSIGN_TAG_SCHEMA_SUFFIXES:
+        expected_len = len(prefix) + digest_len + 1 + len(suffix)
+        suffix_start = len(prefix) + digest_len + 1
+        suffix_clauses.append(
+            (fn.LENGTH(model.name) == expected_len)
+            & (fn.SUBSTR(model.name, suffix_start, 1 + len(suffix)) == f".{suffix}")
+        )
+    suffix_match = suffix_clauses[0]
+    for clause in suffix_clauses[1:]:
+        suffix_match = suffix_match | clause
+    return ~(prefix_match & hex_digest & suffix_match)
 
 
 class RetargetTagException(Exception):
@@ -489,7 +535,24 @@ def retarget_tag(
                 return None
 
             delete_tag_notifications_for_tag(existing_tag)
+            displaced_manifest = existing_tag.manifest
             Tag.update(lifetime_end_ms=now_ms).where(Tag.id == existing_tag.id).execute()
+
+            # Cascade Cosign cleanup for the displaced subject digest. Autoprune
+            # excludes Cosign-schema names, so orphans left by retarget would
+            # otherwise persist forever.
+            if (
+                displaced_manifest is not None
+                and not is_cosign_tag_schema(existing_tag.name)
+                and displaced_manifest.id != manifest.id
+            ):
+                _expire_cosign_sibling_tags(
+                    existing_tag.repository,
+                    displaced_manifest.digest,
+                    now_ms,
+                    subject_manifest=displaced_manifest,
+                    exclude_tag_id=existing_tag.id,
+                )
 
         # Check if tag should be immutable based on manifest label or policies
         immutable = False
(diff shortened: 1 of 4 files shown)


data/model/oci/tag.py
```diff
@@ -597,6 +642,18 @@ def _delete_tag(tag, now_ms):
         if updated != 1:
             return None
 
+        # clean up cosign tags derived from the deleted image digest
+        if not is_cosign_tag_schema(tag.name) and tag.manifest is not None:
```
jbpratt: This cascade expires every `sha256-<digest>.{sig,att,sbom}` tag for `tag.manifest.digest` unconditionally, even when another alive tag still references the same manifest (e.g. deleting `v1.0` while `latest` points at the same image). Reproduced: two tags on one manifest, deleting one alias flips the `.sig` tag's `lifetime_end_ms` from `None` to an expiry timestamp even though the manifest is still tagged and live. Reachable via the tag-delete API and the autopruner. Check for another alive `Tag` referencing `tag.manifest` before cascading, or only cascade from `delete_tags_for_manifest` (which by definition removes every tag on the manifest).

data/model/oci/tag.py
```diff
@@ -597,6 +642,18 @@ def _delete_tag(tag, now_ms):
...
+            cosign_tag_prefix = tag.manifest.digest.replace(":", "-")
+            cosign_names = [
+                f"{cosign_tag_prefix}.{suffix}" for suffix in COSIGN_TAG_SCHEMA_SUFFIXES
+            ]
+            Tag.update(lifetime_end_ms=now_ms).where(
```
jbpratt: This update has no `Tag.immutable == False` predicate, unlike `delete_tag()`'s check for the primary tag (`features.IMMUTABLE_TAGS and tag.immutable` -> `ImmutableTagException`). If a Cosign sibling is matched by an immutability policy, it's silently expired here with no exception. Add `Tag.immutable == False` to this `where(...)` to match the primary-tag invariant.

data/model/oci/tag.py
```diff
@@ -597,6 +642,18 @@ def _delete_tag(tag, now_ms):
         if updated != 1:
             return None
 
+        # clean up cosign tags derived from the deleted image digest
```
jbpratt, line 703: `delete_tags_for_manifest` calls `_delete_tag` once per alive tag on a manifest, so this block re-runs an identical UPDATE for the same `cosign_names` once per alias -- redundant (idempotent, but wasted queries) on multi-tag-same-manifest deletes. Consider hoisting the cascade to run once per unique manifest in `delete_tags_for_manifest` instead of per tag.

data/model/oci/tag.py
```diff
@@ -597,6 +642,18 @@ def _delete_tag(tag, now_ms):
...
+        if not is_cosign_tag_schema(tag.name) and tag.manifest is not None:
+            cosign_tag_prefix = tag.manifest.digest.replace(":", "-")
+            cosign_names = [
+                f"{cosign_tag_prefix}.{suffix}" for suffix in COSIGN_TAG_SCHEMA_SUFFIXES
+            ]
+            Tag.update(lifetime_end_ms=now_ms).where(
+                Tag.repository == tag.repository,
+                Tag.name.in_(cosign_names),
+                (Tag.lifetime_end_ms >> None) | (Tag.lifetime_end_ms > now_ms),
+            ).execute()
```
jbpratt: No log line when this cascade actually expires a Cosign sibling. Suggest `logger.debug`/`logger.info` when the affected row count is >0, so operators have a trail when a signature disappears.

nasonawa: When Cosign signatures are expired for a given subject, we now log at info with the row count, subject digest, and repository id (`logger.info` in `_expire_cosign_sibling_tags` when `updated > 0`)

data/model/oci/tag.py
nasonawa: `delete_tags_for_manifest` now calls `_delete_tag(..., expire_cosign=False)` per alias and runs `_expire_cosign_sibling_tags` once after the loop (using a deleted tag’s manifest digest, since callers may pass a raw manifest id). This avoids repeating the same Cosign UPDATE for every alias on the subject.

data/model/oci/tag.py
nasonawa: Cosign signature expiry now runs only when no other visible, non-Cosign alive tag still references the subject manifest (`_manifest_has_other_alive_tags`). Deleting one alias (e.g. `v1.0` while `latest` remains) no longer expires the `.sig`/`.att`/`.sbom` tags.

data/model/oci/tag.py
nasonawa: `_expire_cosign_sibling_tags` now includes `Tag.immutable == False` in the UPDATE predicate, so Cosign siblings matched by an immutability policy are not silently expired.

data/model/oci/tag.py
```diff
@@ -597,6 +687,17 @@ def _delete_tag(tag, now_ms):
         if updated != 1:
             return None
 
+        # clean up cosign tags derived from the deleted image digest
+        if (
+            expire_cosign
+            and not is_cosign_tag_schema(tag.name)
+            and tag.manifest is not None
+            and not _manifest_has_other_alive_tags(
```
jbpratt: This adds a `SELECT EXISTS` round-trip to every non-cosign tag delete with a manifest, even in repos that never use Cosign, followed by a separate `UPDATE` when it returns false (the common case) -- 2 extra queries per delete vs. 0 before this PR (confirmed via `assert_query_count`: `test_delete_tag` 5->7). At autopruner scale this is a meaningful number of wasted round-trips. Consider collapsing into a single statement: `UPDATE tag SET lifetime_end_ms=... WHERE name IN (cosign_names) AND NOT EXISTS(<other-alive-tag subquery>)`, which keeps the same correctness guarantee with one query instead of two.

nasonawa: modified the '_expire_cosign_sibling_tags' function to combine the 'SELECT EXISTS ' and 'UPDATE' statements into a single query.

data/model/oci/tag.py
```diff
@@ -621,7 +717,20 @@ def delete_tags_for_manifest(manifest):
...
+        subject_tag = next(
+            (
+                tag
+                for tag in tags
+                if not is_cosign_tag_schema(tag.name) and tag.manifest is not None
+            ),
+            None,
+        )
+        if subject_tag is not None:
+            _expire_cosign_sibling_tags(subject_tag.repository, subject_tag.manifest.digest, now_ms)
```
jbpratt: This call omits `subject_manifest=`/`exclude_tag_id=`, so the `NOT EXISTS` guard added to fix the single-alias-still-tagged bug never applies here. It's safe only under the assumption that every alive tag on `manifest` is already in `tags` and expired above in the same transaction -- but `tags = list(query)` is read *before* `db_transaction()` opens, leaving a window where a tag retargeted/created onto this manifest between that read and the transaction start would be missed by the loop, stay alive, and still get its Cosign siblings wrongly expired here. Suggest passing `subject_manifest=subject_tag.manifest` (and moving the initial `Tag.select()` inside the transaction, or re-checking under the transaction) so this path gets the same guard as the single-tag path.

jbpratt:
**Outside this diff, `data/model/oci/tag.py:538` (`retarget_tag`)** -- P2, flagged during final review pass, not inline-commentable since the line is unmodified by this PR:

Retargeting an existing tag away from its manifest expires it directly here (`Tag.update(lifetime_end_ms=now_ms).where(Tag.id == existing_tag.id).execute()`), bypassing `_delete_tag`/`_expire_cosign_sibling_tags`, so no Cosign cleanup runs. Before this PR that was harmless -- an orphaned `.sig`/`.att`/`.sbom` tag would still eventually be caught by ordinary autoprune count/age matching. Now that both autoprune fetchers explicitly exclude Cosign-schema names, a Cosign tag orphaned via retarget (rather than delete) is never re-evaluated by anything and stays forever. Worth deciding whether retarget should also trigger the guarded Cosign cascade for the manifest being displaced.

data/model/oci/tag.py
```diff
@@ -489,8 +535,25 @@ def retarget_tag(
...
+                and not is_cosign_tag_schema(existing_tag.name)
+                and displaced_manifest.id != manifest.id
+            ):
+                _expire_cosign_sibling_tags(
+                    existing_tag.repository,
+                    displaced_manifest.digest,
+                    now_ms,
+                    subject_manifest=displaced_manifest,
+                    exclude_tag_id=existing_tag.id,
+                )
```
jbpratt, line 555: @nasonawa let's add additional e2e tests for deleting and retargeting tags to the playwright suite please! We currently are only covering autopruning but this modifies direct calls

nasonawa: Added Playwright coverage for the direct paths as well: deleteTag cascade (last alias + multi-alias keep), and retarget via createTag (displaced digest cascade + keep when another alias remains). Autoprune scenarios were already covered.

jbpratt (approved):

PR #6668, opened 2026-07-28
PROJQUAY-12278: feat(config): add configurable gunicorn worker timeouts

bcaton85:
## Summary
- Add GUNICORN_REGISTRY_TIMEOUT (default: 300s) and GUNICORN_WEB_TIMEOUT (default: 60s) config.yaml parameters
- supervisord.conf.jnj now passes --timeout to gunicorn-registry and gunicorn-web using these values
- supervisord_conf_create.py reads config.yaml at container init to extract timeout values
- Config schema validates both params as integers with minimum 30

## Motivation
Gunicorn hardcoded 30-second default timeout causes SIGKILL on legitimate long-running operations: proxy cache blob downloads (>500 MB), large image pushes, and API queries on large orgs with quota enabled. This affects at least 9 open PROJQUAY bugs (PROJQUAY-9103, PROJQUAY-8837, PROJQUAY-6564, PROJQUAY-8775, PROJQUAY-8765, PROJQUAY-8218, PROJQUAY-7462, PROJQUAY-6886, [PROJQUAY-8300](https://redhat.atlassian.net/browse/PROJQUAY-8300)).

The new defaults (300s registry, 60s web) provide immediate relief while a proper async fix for proxy cache is developed separately ([PROJQUAY-12311](https://redhat.atlassian.net/browse/PROJQUAY-12311)).

## Test plan
- [x] Schema validation tests: accepts valid integers >= 30, rejects values below minimum, rejects non-integers
- [x] Config default tests: verify 300s registry and 60s web defaults
- [x] Template rendering tests: default values, custom values, hotreload mode preserves 600s override
- [x] All 63 tests pass (57 schema + 6 supervisord)

### DIFF
--- a/conf/init/supervisord_conf_create.py
+++ b/conf/init/supervisord_conf_create.py
@@ -14,7 +14,7 @@ QUAYRUN_DIR = os.getenv("QUAYRUN", QUAYCONF_DIR)
 QUAY_LOGGING = os.getenv("QUAY_LOGGING", "stdout")  # or "syslog"
 QUAY_HOTRELOAD: bool = os.getenv("QUAY_HOTRELOAD", "false") == "true"
 
-MAX_GUNICORN_TIMEOUT = 1800
+MAX_GUNICORN_TIMEOUT = 300
 
 
 def _parse_csv_env(name):
@@ -121,9 +121,9 @@ if __name__ == "__main__":
         QUAY_LOGGING,
         QUAY_HOTRELOAD,
         gunicorn_registry_timeout=min(
-            app_config.get("GUNICORN_REGISTRY_TIMEOUT", 300), MAX_GUNICORN_TIMEOUT
+            app_config.get("GUNICORN_REGISTRY_TIMEOUT", 30), MAX_GUNICORN_TIMEOUT
         ),
         gunicorn_web_timeout=min(
-            app_config.get("GUNICORN_WEB_TIMEOUT", 60), MAX_GUNICORN_TIMEOUT
+            app_config.get("GUNICORN_WEB_TIMEOUT", 30), MAX_GUNICORN_TIMEOUT
         ),
     )
--- a/conf/init/supervisord_conf_create.py
+++ b/conf/init/supervisord_conf_create.py
@@ -4,6 +4,7 @@ import sys
 from typing import List
 
 import jinja2
+import yaml
 
 QUAYPATH = os.getenv("QUAYPATH", ".")
 QUAYDIR = os.getenv("QUAYDIR", "/")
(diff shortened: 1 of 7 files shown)


conf/init/supervisord_conf_create.py
```diff
@@ -13,6 +14,8 @@
 QUAY_LOGGING = os.getenv("QUAY_LOGGING", "stdout")  # or "syslog"
 QUAY_HOTRELOAD: bool = os.getenv("QUAY_HOTRELOAD", "false") == "true"
 
+MAX_GUNICORN_TIMEOUT = 1800
```
ibazulic: I think 30 minutes is a very long time. Once the worker stops emitting heartbeats, it will remain in a Schrodinger "live/not-alive" state for 30 minutes before gunicorn actually removes it and restarts. This long timeout can cause complete worker exhaustion if several blocking things happen at once, causing a performance impact and then complete unavailability of Quay. 

Gunicorn timeout IMHO should not be over 5 minutes, and even that is stretching. If an operation doesn't succeed in 5 minutes, it will never succeed.

conf/init/supervisord_conf_create.py
```diff
@@ -100,9 +113,17 @@ def override_services(config, override_services):
...
         os.path.join(QUAYCONF_DIR, "supervisord.conf"),
         config,
         QUAY_LOGGING,
         QUAY_HOTRELOAD,
+        gunicorn_registry_timeout=min(
+            app_config.get("GUNICORN_REGISTRY_TIMEOUT", 300), MAX_GUNICORN_TIMEOUT
+        ),
+        gunicorn_web_timeout=min(
+            app_config.get("GUNICORN_WEB_TIMEOUT", 60), MAX_GUNICORN_TIMEOUT
+        ),
```
ibazulic: IMHO default values, if not set, should be left at 30 seconds. This is to enable relative quick recycling of locked gunicorn processes.

util/config/schema.py
```diff
@@ -337,6 +337,20 @@
...
+            "description": "Timeout in seconds for gunicorn-registry workers. Workers that do not respond within this window are killed and restarted. Must be between 30 and 1800 (30 minutes). Defaults to 300",
+            "x-example": 300,
+        },
+        "GUNICORN_WEB_TIMEOUT": {
+            "type": "integer",
+            "minimum": 30,
+            "maximum": 1800,
+            "description": "Timeout in seconds for gunicorn-web workers. Workers that do not respond within this window are killed and restarted. Must be between 30 and 1800 (30 minutes). Defaults to 60",
+            "x-example": 60,
+        },
```
ibazulic, line 353: Maximum should be set to 300 seconds, with the default value set at 30.

bcaton85: updated

ibazulic (approved):

PR #6672, opened 2026-07-29
PROJQUAY-9692: fix(audit): log org_delete event when superuser deletes organization

LiZhang19817:
## Summary
- Adds missing `log_action("org_delete", ...)` call to the superuser organization delete endpoint (`SuperUserOrganizationManagement.delete` in `endpoints/api/superuser.py`)
- The normal org-delete path (`endpoints/api/organization.py`) already logs this event correctly; the superuser path was an oversight
- Follows the same pattern used by the superuser user-delete handler (`log_action("user_delete", ...)`) and org-rename handler (`log_action("org_change_name", ...)`) in the same file

## Test plan
- [ ] Delete an organization from the superuser admin panel and verify an `org_delete` audit log entry is created
- [ ] Delete an organization from the organization settings page and verify `org_delete` audit log still works as before
- [ ] Verify the audit log entry contains the correct `namespace` metadata field

### DIFF
--- a/endpoints/api/superuser.py
+++ b/endpoints/api/superuser.py
@@ -876,6 +876,11 @@ class SuperUserOrganizationManagement(ApiResource):
         Deletes the specified organization.
         """
         if SuperUserPermission().can():
+            try:
+                model.organization.get_organization(name)
+            except model.InvalidOrganizationException:
+                raise NotFound()
+
             log_action("org_delete", name, {"namespace": name})
             pre_oci_model.mark_organization_for_deletion(name)
             return "", 204
--- a/endpoints/api/superuser.py
+++ b/endpoints/api/superuser.py
@@ -876,6 +876,7 @@ class SuperUserOrganizationManagement(ApiResource):
         Deletes the specified organization.
         """
         if SuperUserPermission().can():
+            log_action("org_delete", name, {"namespace": name})
             pre_oci_model.mark_organization_for_deletion(name)
             return "", 204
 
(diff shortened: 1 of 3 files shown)


LiZhang19817:
@jbpratt @bcaton85 Pls help review and merge this fix.

endpoints/api/superuser.py
```diff
@@ -876,6 +876,12 @@ def delete(self, name):
         Deletes the specified organization.
         """
         if SuperUserPermission().can():
+            try:
+                model.organization.get_organization(name)
+            except model.InvalidOrganizationException:
+                raise NotFound()
```
jbpratt, line 882: @LiZhang19817 What is this addition for? Is this a different fix for something else?

LiZhang19817: This check is required so that `log_action("org_delete", ...)` can run **before** `mark_organization_for_deletion()`. The deletion internally calls `mark_namespace_for_deletion`, which renames the username to a random UUID — so if we log after deletion, `log_action` would fail with `User.DoesNotExist` when trying to look up the original namespace name.

This matches the pattern used by the superuser user-delete handler (`SuperUserManagement.delete`) and the org-admin delete handler in `endpoints/api/organization.py`, which both validate existence first, then log, then delete. The org-admin handler even has an explicit comment: `# we need to do this before the deletion, as the org will be gone after`.

Yes, `get_organization(name)` ends up being called twice (once here, once inside `mark_organization_for_deletion`), but the extra DB query is necessary to return a proper 404 for non-existent orgs and to ensure the audit log is recorded while the namespace still exists.

jbpratt (approved):

PR #6681, opened 2026-07-29
PROJQUAY-11868: fix(migrate): add fail-closed OMR source preflight

jbpratt:
Validate source database compatibility and authentication policy before
stopping OMR services. Open SQLite sources through an escaped read-only
URI, reject unsupported revisions and providers, and verify robot-token
and registry JWT key continuity.

Run the same preflight during dry runs so unsupported sources cannot
produce a viable migration plan. Remove the legacy bridge path because
external schemas remain disabled until artifact-derived compatibility
fixtures can prove they are safe to migrate.

Signed-off-by: Brady Pratt <bpratt@redhat.com>


internal/dal/dbcore/bridge.go
```diff
@@ -11,42 +11,9 @@ import (
...
-	"b1c2d3e4f5a6": true,
-	"15f06d00c4b3": true,
-	"414c5e2fc487": true,
-	// Previous target version (systems already bridged once)
-	"b30800b1d271": true,
-	// Quota notification tables
-	"6715e4719375": true,
-	"9fa37f66a9b6": true,
-}
+// ApprovedOMRSourceVersion is the only external OMR revision admitted by
```
harishsurf, line 14: **CRITICAL: Only OMR 2.0.11 revision accepted — all other OMR 2.0.x versions rejected**

Replacing the broad `knownOMRVersions` map (~20 revisions spanning OMR 2.0.0-2.0.11) with a single `ApprovedOMRSourceVersion = "3f8d7acdf7f9"` means customers on OMR 2.0.0 through 2.0.10 get "unsupported OMR source revision" and cannot migrate.

The PR description states this is intentional ("external schemas remain disabled until artifact-derived compatibility fixtures can prove they are safe"), which is a valid safety stance. But this is a significant behavior change — previously supported versions are now rejected. Customers must upgrade to OMR 2.0.11 before migrating to 3.0.

Should this be documented as a requirement? Or should the broader version support be retained until the fixtures are ready?

internal/dal/dbcore/bridge_test.go
```diff
@@ -1,272 +0,0 @@
-package dbcore
```
harishsurf, line 1: **MAJOR: 272 lines of bridge tests deleted with no replacement**

Tests for `ensureColumn`, `ensureNonUniqueIndex`, and `RunBridge` execution (already-current, unknown version, creates bridge tables) are removed. These functions still exist in `bridge.go` and are still called. The new `compatibility_test.go` only tests validation and rejection, not the actual bridge execution logic (column additions, index fixes, SQL application).

If `ensureColumn` or `ensureNonUniqueIndex` regress, no test catches it.

internal/dal/dbcore/bridge.go
```diff
@@ -69,25 +36,16 @@ var bridgeIndexFixes = []struct {
...
-// the standalone binary intentionally does not ship Python, Alembic, or Quay's
-// Python migration runtime. It validates the starting version, applies column
-// additions and index fixes that SQLite cannot express idempotently, then
-// executes the bridge SQL file.
+// RunBridge upgrades the approved OMR source schema to the Go binary's target
+// schema. Source preflight rejects unsupported revisions before the source is
+// stopped; this repeat guard ensures copied data cannot bypass that policy.
 func RunBridge(ctx context.Context, db *sql.DB, w io.Writer) error {
 	ver, err := SchemaVersion(ctx, db)
 	if err != nil {
```
harishsurf, line 44: **MAJOR: `RunBridge` no longer handles "already current" — errors on a database at TargetVersion**

The old code had `if ver == TargetVersion { return nil }` (no-op for already-migrated databases). The new code has `if ver != ApprovedOMRSourceVersion { return error }`. If `upgradeSchema()` calls `RunBridge` on a previously migrated database (already at TargetVersion), it returns an error instead of succeeding.

Verify the call chain: does `upgradeSchema` check for TargetVersion before calling `RunBridge`? If not, re-running migration on an already-migrated database will fail.

harishsurf (approved):

PR #6717, opened 2026-08-04
QUAYIO-1826: feat(gc): add configurable grace period for namespace deletion

kleesc:
Introduce NAMESPACE_GC_GRACE_PERIOD_S and NAMESPACE_GC_GRACE_PERIOD_NAMESPACES config options so that namespaces in the allowlist receive a queue-level delay before the GC worker permanently purges them. During the grace window, linked data (teams, robots, OAuth apps, federated logins) is preserved in the database so SREs can recover the namespace via direct DB operations.

Also fixes _delete_user_linked_data to resolve the original namespace name from the DeletedNamespace marker, so robot and quota lookups succeed even after the username has been replaced with a UUID during soft-delete.

### DIFF
--- a/config.py
+++ b/config.py
@@ -534,6 +534,16 @@ class DefaultConfig(ImmutableConfig):
     # How often the Garbage Collection worker runs.
     GARBAGE_COLLECTION_FREQUENCY = 30  # seconds
 
+    # Grace period (seconds) before deleted namespaces are permanently purged by GC.
+    # During this window the data remains in the database and can be recovered via
+    # direct DB operations. 0 means immediate purge (current behavior).
+    NAMESPACE_GC_GRACE_PERIOD_SECONDS = 0
+
+    # Allowlist of namespace names that receive the grace period. When non-empty,
+    # only namespaces in this list are protected. When empty, no grace period is
+    # applied (current behavior).
+    NAMESPACE_GC_GRACE_PERIOD_ALLOWLIST: list[str] = []
+
     # How long notifications will try to send before timing out.
     NOTIFICATION_SEND_TIMEOUT = 10
 
--- a/data/model/user.py
+++ b/data/model/user.py
@@ -157,13 +157,11 @@ def create_user_noverify(
     except User.DoesNotExist:
         pass
 
-    allowlist = config.app_config.get("NAMESPACE_GC_GRACE_PERIOD_ALLOWLIST", [])
-    if allowlist and username in allowlist:
-        try:
-            DeletedNamespace.get(DeletedNamespace.original_username == username)
-            raise InvalidUsernameException("Username is not available")
-        except DeletedNamespace.DoesNotExist:
-            pass
+    try:
+        DeletedNamespace.get(DeletedNamespace.original_username == username)
+        raise InvalidUsernameException("Username is not available")
+    except DeletedNamespace.DoesNotExist:
+        pass
 
     # Check email uniqueness only among non-organization users. Organizations are allowed
     # to share emails (enforced by the partial unique index on User.email).
@@ -1379,6 +1377,11 @@ def mark_namespace_for_deletion(
             allowlist = config.app_config.get("NAMESPACE_GC_GRACE_PERIOD_ALLOWLIST", [])
             if grace_period > 0 and allowlist and user.username in allowlist:
                 available_after = grace_period
+                logger.info(
+                    "Namespace %s marked for deletion with %ds grace period",
+                    user.username,
+                    grace_period,
+                )
             else:
                 available_after = 0
 
--- a/data/model/user.py
+++ b/data/model/user.py
@@ -157,6 +157,14 @@ def create_user_noverify(
     except User.DoesNotExist:
         pass
 
+    allowlist = config.app_config.get("NAMESPACE_GC_GRACE_PERIOD_ALLOWLIST", [])
+    if allowlist and username in allowlist:
+        try:
+            DeletedNamespace.get(DeletedNamespace.original_username == username)
+            raise InvalidUsernameException("Username is not available")
+        except DeletedNamespace.DoesNotExist:
+            pass
+
     # Check email uniqueness only among non-organization users. Organizations are allowed
     # to share emails (enforced by the partial unique index on User.email).
     if email_required:
@@ -1351,13 +1359,29 @@ def get_solely_admined_organizations(user_obj):
     return solely_admined
 
 
-def mark_namespace_for_deletion(user, queues, namespace_gc_queue, force=False):
+def mark_namespace_for_deletion(
+    user, queues, namespace_gc_queue, force=False, available_after=None
+):
     """
     Marks a namespace (as referenced by the given user) for deletion.
 
     A queue item will be added to delete the namespace's repositories and storage, while the
     namespace itself will be renamed, disabled, and delinked from other tables.
+
+    When available_after > 0 (grace period), linked data (teams, OAuth apps, robots, etc.) is
+    preserved so the namespace can be recovered via direct DB operations during the grace window.
     """
+    if available_after is None:
+        if force:
+            available_after = 0
+        else:
+            grace_period = config.app_config.get("NAMESPACE_GC_GRACE_PERIOD_SECONDS", 0)
+            allowlist = config.app_config.get("NAMESPACE_GC_GRACE_PERIOD_ALLOWLIST", [])
+            if grace_period > 0 and allowlist and user.username in allowlist:
+                available_after = grace_period
+            else:
+                available_after = 0
+
     if not user.enabled:
         return None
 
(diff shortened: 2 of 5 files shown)


data/model/user.py
```diff
@@ -1351,13 +1351,28 @@ def get_solely_admined_organizations(user_obj):
...
     """
     Marks a namespace (as referenced by the given user) for deletion.
 
     A queue item will be added to delete the namespace's repositories and storage, while the
     namespace itself will be renamed, disabled, and delinked from other tables.
+
+    When available_after > 0 (grace period), linked data (teams, OAuth apps, robots, etc.) is
+    preserved so the namespace can be recovered via direct DB operations during the grace window.
     """
+    if available_after is None:
```
deshpandevlab, line 1372: **Critical: `force=True` should bypass grace period**

The `available_after is None` check triggers the config lookup unconditionally, including when `force=True`. The superuser force-delete paths (`superuser_models_pre_oci.py:185,216`) pass `force=True` but no `available_after`, meaning a force-deleted namespace in the allowlist still gets the grace period.

At quay.io, superuser force-deletes are used for abuse/ToS violations and compliance takedowns. These **must not** be deferred.

Suggested fix:
```python
if available_after is None:
    if force:
        available_after = 0
    else:
        from app import app
        grace_period = app.config.get("NAMESPACE_GC_GRACE_PERIOD_S", 0)
        allowlist = app.config.get("NAMESPACE_GC_GRACE_PERIOD_NAMESPACES", [])
        if grace_period > 0 and allowlist and user.username in allowlist:
            available_after = grace_period
        else:
            available_after = 0
```

data/model/user.py
```diff
@@ -1381,8 +1396,10 @@ def mark_namespace_for_deletion(user, queues, namespace_gc_queue, force=False):
     for queue in queues:
         queue.delete_namespaced_items(user.username)
 
-    # Delete non-repository related items. This operation is very quick, so we can do so here.
-    _delete_user_linked_data(user)
+    # When a grace period is active, defer linked data deletion so the namespace can be
+    # recovered. The GC worker (delete_user) will clean up linked data at purge time.
+    if available_after <= 0:
```
deshpandevlab, line 1413: **Critical: Robot name collision / namespace squatting risk when grace period is active**

When a grace period is active, `_delete_user_linked_data` is skipped here, preserving robots with names like `originalname+robotname`. However, the namespace's `User.username` is immediately replaced with a UUID further below, freeing the original name for re-registration.

This creates a dangerous scenario at quay.io scale:

1. Org `bigcorp` is deleted with 14-day grace period
2. Robots `bigcorp+deployer`, `bigcorp+scanner` still exist in the `User` table
3. Someone registers a **new** namespace named `bigcorp`
4. The new `bigcorp` owner cannot create robots with the same short names (unique constraint on `User.username`)
5. Worse: `list_namespace_robots("bigcorp")` uses a prefix match (`User.username ** (entity_name + "+%")`), so the new namespace owner could **see the old orphaned robots** — this is a data leak

Suggested fixes (pick one):
- **(A)** Rename robots to `UUID+robotname` during grace-period marking (preserves recoverability since you can resolve via the marker)
- **(B)** Block re-registration of `DeletedNamespace.original_username` values during the grace window (add a check in `create_user`/`create_organization`)

Option B is probably cleaner since it also prevents confusion around team names, federated logins, etc. that also reference the original namespace identity.

deshpandevlab (commented):
Thanks for working on this — the grace period concept is sound and the queue/marker mechanics are solid. However, I found two critical issues that I believe should block merge given this affects GC at quay.io scale. See inline comments.

data/model/user.py
ibazulic: Another option is to just wipe robot accounts on namespace deletion immediately, regardless of the grace period. The robot accounts should not be valid when an org is deleted in any case. If the org or user account is resurrected, the clean approach would be to expect users to recreate robot account tokes.

Same thing goes for Oauth tokens and for app specific tokens (if we're talking about user accounts).

Edit: why deletions? Because in order to preserve robot accounts, you need to do one DB operation per robot account because you must preserve its user name. This is potentially hundreds of new db operations (`SELECT` to read robot accounts and then `UPDATE` each one). `DELETE` is potentially just one query, bulk deletetion for all entries in the `user` table that match `{ORGNAME}+{short_name}`.

config.py
```diff
@@ -534,6 +534,16 @@ def create_transaction(db):
     # How often the Garbage Collection worker runs.
     GARBAGE_COLLECTION_FREQUENCY = 30  # seconds
 
+    # Grace period (seconds) before deleted namespaces are permanently purged by GC.
+    # During this window the data remains in the database and can be recovered via
+    # direct DB operations. 0 means immediate purge (current behavior).
+    NAMESPACE_GC_GRACE_PERIOD_S = 0
```
deshpandevlab: Minor naming suggestion: the `_S` suffix for seconds doesn't match the codebase convention — the existing pattern is either `_SEC` (e.g., `PUSH_TEMP_TAG_EXPIRATION_SEC`, `SIGNED_GRANT_EXPIRATION_SEC`) or `_SECONDS` (e.g., `QUOTA_TOTAL_DELAY_SECONDS`, `REDIS_FLUSH_INTERVAL_SECONDS`). Would `NAMESPACE_GC_GRACE_PERIOD_SECONDS` (or `_SEC`) be worth considering for consistency?

Also, `NAMESPACE_GC_GRACE_PERIOD_NAMESPACES` reads a bit repetitively with "namespace" appearing twice. Something like `NAMESPACE_GC_GRACE_PERIOD_ALLOWLIST` might make the purpose clearer at a glance.

data/model/user.py
```diff
@@ -1351,13 +1351,28 @@ def get_solely_admined_organizations(user_obj):
...
 
     A queue item will be added to delete the namespace's repositories and storage, while the
     namespace itself will be renamed, disabled, and delinked from other tables.
+
+    When available_after > 0 (grace period), linked data (teams, OAuth apps, robots, etc.) is
+    preserved so the namespace can be recovered via direct DB operations during the grace window.
     """
+    if available_after is None:
+        from app import app
+
```
deshpandevlab: Worth noting: when `NAMESPACE_GC_GRACE_PERIOD_S` is set to a positive value but `NAMESPACE_GC_GRACE_PERIOD_NAMESPACES` is left empty (the default), the `and allowlist` check short-circuits here and no namespace receives protection — silently. This could be a footgun for operators who configure the grace period but forget the allowlist.

Would it make sense to either:
- Log a warning when `grace_period > 0` and `allowlist` is empty, so operators know the grace period isn't taking effect, or
- Note this interaction explicitly in the config schema description for `NAMESPACE_GC_GRACE_PERIOD_S`?

Either way, the current behavior should be clearly documented so it doesn't surprise anyone.

deshpandevlab (commented):
A few more observations on top of the inline threads:

**Request: test coverage for `force=True` + grace period interaction**

Since the `force` bypass is a correctness concern (see the thread at `user.py:1366`), it would be really helpful to have a dedicated test asserting that `force=True` always results in immediate deletion — even for namespaces on the allowlist. That way the fix stays protected against regressions.

**Non-blocker: double `_delete_user_linked_data` call in the no-grace-period path**

Small thing — in the existing flow (no grace period), `_delete_user_linked_data` is called eagerly in `mark_namespace_for_deletion`, and then again later when the GC worker runs `delete_user`. The second call is a harmless no-op since everything was already cleaned up, but it does run several queries against empty result sets. Not blocking at all — just noting for awareness. Could be guarded with a flag on the marker in the future if it matters for performance.

data/model/user.py
kleesc: Fixed by blocking recreation of allowlisted namespaces during the grace window. `create_user_noverify` now checks for a DeletedNamespace marker before allowing the name.
  
For now this check prevents the namespace squatting while preserving the option to restore everything.

data/model/user.py
```diff
@@ -157,6 +157,14 @@ def create_user_noverify(
     except User.DoesNotExist:
         pass
 
+    allowlist = config.app_config.get("NAMESPACE_GC_GRACE_PERIOD_ALLOWLIST", [])
+    if allowlist and username in allowlist:
```
deshpandevlab: can we check for the marker unconditionally? Like regardless of the current allowlist. If the allowlist changes between deletion and re-registration (admin removes a namespace from the list), the check is skipped but the DeletedNamespace marker and its linked data still exist. Is my understanding correct?

deshpandevlab:
@kleesc can we add logging when a grace period is applied? The GC worker logs "Got namespace GC queue item" but nothing distinguishes a deferred item from an immediate one. For prod confidence, it would be helpful to have something like:

```
"Namespace %s marked for deletion with %ds grace period", user.username, available_after)
```

deshpandevlab (approved):

PR #6740, opened 2026-08-05
PROJQUAY-12365: fix(api): hide robot account tokens from global readonly superusers

sudipshil9862:
## Summary
- Global readonly superusers could view robot account tokens via the `OrgRobotList` and `OrgRobot` GET endpoints. Robot tokens are credentials that enable authentication as the robot, breaking the read-only privilege boundary.
- Removed `allow_if_global_readonly_superuser()` from the `include_token` gate in `OrgRobotList.get` and conditionally set `include_token=False` in `OrgRobot.get` when the caller is only a global readonly superuser (not an org admin or full-access superuser).
- Global readonly superusers retain access to robot metadata (name, description, creation date) for auditing purposes.

## Test plan
- [x] Reproduced bug: global readonly superuser sees robot token via UI and API
- [x] Verified fix: global readonly superuser sees robot metadata but no token
- [x] Verified regression: org admin still sees robot tokens as before
- [x] Added unit tests for global readonly superuser token hiding (`test_robot.py`)
- [x] Updated existing test assertion in `test_api_usage.py`
- [ ] CI unit tests pass

### DIFF
--- a/endpoints/api/robot.py
+++ b/endpoints/api/robot.py
@@ -230,20 +230,21 @@ class OrgRobotList(ApiResource):
             or allow_if_global_readonly_superuser()
             or (features.SUPERUSERS_FULL_ACCESS and allow_if_superuser())
         ):
-            include_token = (
-                AdministerOrganizationPermission(orgname).can()
-                or allow_if_global_readonly_superuser()
-            ) and parsed_args.get("token", True)
+            is_org_admin = AdministerOrganizationPermission(orgname).can()
+            is_full_access_superuser = features.SUPERUSERS_FULL_ACCESS and allow_if_superuser()
+            include_token = (is_org_admin or is_full_access_superuser) and parsed_args.get(
+                "token", True
+            )
             include_permissions = (
-                AdministerOrganizationPermission(orgname).can()
-                or allow_if_global_readonly_superuser()
+                is_org_admin or is_full_access_superuser or allow_if_global_readonly_superuser()
             ) and parsed_args.get("permissions", False)
-            return robots_list(
+            result = robots_list(
                 orgname,
                 include_permissions=include_permissions,
                 include_token=include_token,
                 limit=parsed_args.get("limit"),
             )
+            return result
 
         raise Unauthorized()
 
@@ -270,14 +271,18 @@ class OrgRobot(ApiResource):
         Returns the organization's robot with the specified name.
         """
         permission = AdministerOrganizationPermission(orgname)
-        # Global readonly superusers can always view, regular superusers need FULL_ACCESS
+        is_org_admin = permission.can()
         if (
-            permission.can()
+            is_org_admin
             or allow_if_global_readonly_superuser()
             or (features.SUPERUSERS_FULL_ACCESS and allow_if_superuser())
         ):
             robot = model.get_org_robot(robot_shortname, orgname)
-            return robot.to_dict(include_metadata=True, include_token=True)
+            include_token = is_org_admin or (
+                features.SUPERUSERS_FULL_ACCESS and allow_if_superuser()
+            )
+            result = robot.to_dict(include_metadata=True, include_token=include_token)
+            return result
 
         raise Unauthorized()
 
--- a/endpoints/api/test/test_robot.py
+++ b/endpoints/api/test/test_robot.py
@@ -1,5 +1,5 @@
 import json
-from unittest.mock import Mock
+from unittest.mock import Mock, patch
 
 import pytest
 import requests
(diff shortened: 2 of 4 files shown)


endpoints/api/robot.py
```diff
@@ -230,13 +230,15 @@ def get(self, orgname, parsed_args):
...
+            is_full_access_superuser = (
+                features.SUPERUSERS_FULL_ACCESS and allow_if_superuser()
+            )
+            include_token = (is_org_admin or is_full_access_superuser) and parsed_args.get(
+                "token", True
+            )
             include_permissions = (
-                AdministerOrganizationPermission(orgname).can()
-                or allow_if_global_readonly_superuser()
+                is_org_admin or is_full_access_superuser
```
bcaton85: **blocker:** This also removes `permissions=true` support for global readonly superusers. The legacy robot manager requests `/robots?permissions=true&token=false` and then assumes each robot has `teams`/`repositories` (for example `robot.teams.map(...)` in `static/js/directives/ui/robots-manager.js`). With `include_permissions` now limited to org admins/full-access superusers, a global readonly user can still load the endpoint but gets a different response shape, which likely breaks the Robot Accounts UI and removes read-only audit data beyond the token. Can we keep tokens hidden while still allowing global readonly users to receive permissions metadata, e.g. include `is_global_readonly_superuser` in this `include_permissions` gate but not in `include_token`?

endpoints/api/test/test_robot.py
```diff
@@ -171,6 +171,72 @@ def test_duplicate_robot_creation(app):
         assert resp.json["error_message"] == "Existing robot with name: buynlarge+coolrobot"
 
 
+def test_globalreadonlysuperuser_org_robot_list_hides_token(app):
+    with client_with_identity("globalreadonlysuperuser", app) as cl:
+        result = conduct_api_call(
+            cl, OrgRobotList, "GET", {"orgname": "buynlarge", "token": "true"}, None
```
bcaton85, line 177: **should-fix:** This regression test only verifies token hiding for the default/list shape. Please also cover the `permissions=true` case for global readonly superusers and assert the read-only audit fields remain present, e.g. request `{"orgname": "buynlarge", "token": "true", "permissions": "true"}` and assert each robot has no `token` but still includes `teams` and `repositories`. That would catch the response-shape regression in `OrgRobotList.get`.

endpoints/api/test/test_robot.py
```diff
@@ -171,6 +171,72 @@ def test_duplicate_robot_creation(app):
...
+            "GET",
+            {"orgname": "buynlarge", "robot_shortname": "coolrobot"},
+            None,
+        )
+        assert result.json["token"] is not None
+
+
+def test_full_access_superuser_sees_robot_token(app):
+    original = app.config.get("SUPER_USERS", [])[:]
+    app.config["SUPER_USERS"] = ["freshuser"]
```
bcaton85: **should-fix:** Mutating `app.config["SUPER_USERS"]` here does not reliably make `freshuser` a superuser. Superuser membership is copied into `ConfigUserManager._superusers_array` when the app/user manager is initialized, and `SuperUserPermission().can()` checks `auth.permissions.usermanager.is_superuser(...)`, not this updated config list. Please patch `auth.permissions.usermanager.is_superuser` for this test (and keep `is_global_readonly_superuser` false), or use an already-configured superuser in a scenario where they are not also an org admin, so this actually exercises the full-access superuser path.

endpoints/api/test/test_robot.py
bcaton85: ^ GROSU should still be able to see permissions, you'll have to verify via the api

bcaton85 (changes requested):
Needs a couple of small changes

bcaton85 (dismissed):

endpoints/api/robot.py
```diff
@@ -230,20 +230,25 @@ def get(self, orgname, parsed_args):
...
+            is_global_readonly = allow_if_global_readonly_superuser()
+            result = robots_list(
                 orgname,
                 include_permissions=include_permissions,
                 include_token=include_token,
                 limit=parsed_args.get("limit"),
             )
+            if is_global_readonly and not include_token:
+                for robot in result.get("robots", []):
+                    robot["token"] = "Token hidden for global readonly superusers"
```
bcaton85: This isn't really the best way to do this. We would want this reported as an error message rather than returning it directly in the token field. We can forgo the placeholder text for now

bcaton85 (approved):

PR #6745, opened 2026-08-05
QUAYIO-2072: feat(secscan): add SECURITY_SCANNER_V4_INDEXING config to disable legacy indexer

kleesc:
Allow disabling the old V4 security scanner indexing operations while
keeping the V2 indexer and security scan query APIs functional.


Sunandadadi (approved):
lgtm

PR #6752, opened 2026-08-06
QUAYIO-2081: feat(deploy): add graceful shutdown and component-level logging for Envoy

Sunandadadi:
## Summary

- Add graceful connection draining for Envoy and RLS pods during rollouts
- Add configurable per-component debug logging for Envoy

## Problem

During pod rollouts, Envoy shuts down immediately on SIGTERM, dropping active connections. At 5% traffic (~1,410 req/s), we observed a 13% dip in request rate during rollouts as clients lost their connections and had to reconnect.

### Graceful shutdown
- `terminationGracePeriodSeconds: 45` on both Envoy and RLS pods
- `--drain-time-s 30` on the Envoy process
- `preStop` hook that calls `POST /drain_listeners?graceful` to explicitly trigger draining before SIGTERM

On pod termination, the sequence is:
1. Kubernetes triggers `preStop` hook
2. `POST /drain_listeners?graceful` → Envoy stops accepting new connections
3. `/ready` returns 503 → ALB detects unhealthy target, stops routing new traffic
4. Envoy sends `Connection: close` on idle keep-alive connections
5. In-flight requests complete normally
6. `sleep 30` → wait for drain period
7. SIGTERM sent → Envoy shuts down
8. 45s hard deadline → Kubernetes force-kills if needed (safety net)

Without the `preStop` hook, `--drain-time-s` only sets the drain duration but does not trigger draining on SIGTERM — Envoy closes listeners immediately.

### Component-level logging
- New `ENVOY_COMPONENT_LOG_LEVEL` parameter (default: empty)
- When set, enables debug logging for specific Envoy components without flooding all logs
- Example: `upstream:debug,connection:debug,pool:debug`
- Controlled via SaaS file parameter

### DIFF
--- a/deploy/openshift/rosa/quay-envoy-deploy-template.yaml
+++ b/deploy/openshift/rosa/quay-envoy-deploy-template.yaml
@@ -95,7 +95,7 @@ objects:
                 command:
                 - "/bin/sh"
                 - "-c"
-                - "curl -s -X POST http://localhost:9901/drain_listeners?graceful && sleep 30"
+                - "curl -sf --max-time 10 -X POST http://localhost:9901/drain_listeners?graceful ; sleep 30"
           env:
           - name: ENVOY_LOG_LEVEL
             value: "${ENVOY_LOG_LEVEL}"
--- a/deploy/openshift/rosa/quay-envoy-deploy-template.yaml
+++ b/deploy/openshift/rosa/quay-envoy-deploy-template.yaml
@@ -75,12 +75,25 @@ objects:
                   matchLabels:
                     app: quay-envoy
                 topologyKey: topology.kubernetes.io/zone
+        terminationGracePeriodSeconds: 45
         containers:
         - name: envoy
           image: ${ENVOY_IMAGE}:${ENVOY_IMAGE_TAG}
           imagePullPolicy: IfNotPresent
           command: ["/bin/sh", "-c"]
-          args: ["ulimit -n $(ulimit -Hn) && exec envoy -c /etc/envoy/envoy.yaml --log-level ${ENVOY_LOG_LEVEL}"]
+          args:
+          - |
+            ulimit -n $(ulimit -Hn)
+            ARGS="-c /etc/envoy/envoy.yaml --log-level $ENVOY_LOG_LEVEL --drain-time-s 30"
+            if [ -n "$ENVOY_COMPONENT_LOG_LEVEL" ]; then
+              ARGS="$ARGS --component-log-level $ENVOY_COMPONENT_LOG_LEVEL"
+            fi
+            exec envoy $ARGS
+          env:
+          - name: ENVOY_LOG_LEVEL
+            value: "${ENVOY_LOG_LEVEL}"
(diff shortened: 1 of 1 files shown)


deploy/openshift/rosa/quay-envoy-deploy-template.yaml
```diff
@@ -75,12 +75,25 @@ objects:
...
           command: ["/bin/sh", "-c"]
-          args: ["ulimit -n $(ulimit -Hn) && exec envoy -c /etc/envoy/envoy.yaml --log-level ${ENVOY_LOG_LEVEL}"]
+          args:
+          - |
+            ulimit -n $(ulimit -Hn)
+            ARGS="-c /etc/envoy/envoy.yaml --log-level $ENVOY_LOG_LEVEL --drain-time-s 30"
+            if [ -n "$ENVOY_COMPONENT_LOG_LEVEL" ]; then
+              ARGS="$ARGS --component-log-level $ENVOY_COMPONENT_LOG_LEVEL"
+            fi
+            exec envoy $ARGS
```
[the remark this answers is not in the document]
Sunandadadi: good catch! added

deploy/openshift/rosa/quay-envoy-deploy-template.yaml
```diff
@@ -181,6 +201,7 @@ objects:
         annotations:
           oneagent.dynatrace.com/inject: "${DYNATRACE_MONITORING}"
       spec:
+        terminationGracePeriodSeconds: 45
```
kleesc: How did you determine 45 sec?

Sunandadadi: Kubernetes default is 30s. So, to give some more room went with 45s. But, on second thought, I see that quay-app has 60s. I will update this to the same.

deploy/openshift/rosa/quay-envoy-deploy-template.yaml
```diff
@@ -75,12 +75,32 @@ objects:
...
+              ARGS="$ARGS --component-log-level $ENVOY_COMPONENT_LOG_LEVEL"
+            fi
+            exec envoy $ARGS
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                - "/bin/sh"
+                - "-c"
+                - "curl -s -X POST http://localhost:9901/drain_listeners?graceful && sleep 30"
```
cubismod: Does the envoy image have `curl` available?

Sunandadadi: yes - curl 7.76.1

cubismod:
```suggestion
                - "curl -sf --max-time 10 -X POST http://localhost:9901/drain_listeners?graceful ; sleep 30"
```

`-f`: fail flag so curl exits on 4xx/5xx, timeout so that curl doesn't wait endlessly, and the `;` so that a failure will still drain

kleesc (dismissed):

cubismod (approved):

PR #6754, opened 2026-08-06
PROJQUAY-12474: refactor(distribution): extract shared lifecycle runtime

jbpratt:
## Summary
Extract the Distribution lifecycle/runtime concerns from the mirror-registry-specific assembly. This keeps construction rollback, context cancellation, reverse-order cleanup, and idempotent shutdown reusable without forcing product-specific auth, storage, or middleware into a shared package.

## Related Issue
[PROJQUAY-12474](https://issues.redhat.com/browse/PROJQUAY-12474)

## Changes
- Add a product-neutral `internal/registry/distribution` lifecycle runtime with a synchronous LIFO `OnClose` cleanup contract.
- Move mirror-registry Distribution assembly, Quay database authorization, token handling, and constants to `internal/mirrorregistry/distribution`.
- Preserve the existing middleware, JWT, local storage, HTTP routes, metrics isolation, and upload-purger behavior.
- Add lifecycle, panic/error rollback, cleanup ordering, idempotency, and dependency-boundary tests.

## Testing
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go test -race ./internal/registry/distribution/... ./internal/mirrorregistry/... ./internal/oci/...`
- [x] `golangci-lint run --timeout=5m`
- [x] `git diff --check`

## Notes
- Draft PR for architectural review following the merged reusable mirror-registry composition work in #6730.
- Image-registry and Quay Lite implementations are intentionally out of scope; this PR provides only the reusable lifecycle seam.
- Product-specific Distribution configuration remains explicitly composed by each product adapter.


internal/mirrorregistry/distribution/token_handler.go
```diff
@@ -173,59 +180,105 @@ func parseRequestedScopes(r *http.Request) ([]jwtauth.ResourceActions, error) {
...
-			if err := h.policy.AuthorizeRepositoryAccess(r, principal, repositoryAccessItem(grant.Name, action), pullContext); err == nil {
-				allowed.Actions = append(allowed.Actions, action)
-			}
+		var err error
+		pullAllowed, err = h.authorizeAction(r, principal, repositoryAccessItem(grant.Name, repositoryPullAction), pullContext)
+		if err != nil {
+			return jwtauth.ResourceActions{}, err
 		}
-		if len(allowed.Actions) > 0 {
-			authorized = append(authorized, allowed)
```
harishsurf, line 225: **MODERATE — Unwrapped auth errors become 503 instead of silent deny**

The new `authorizeAction` classifies errors into three buckets: success, denied (sentinel errors), and infrastructure error (everything else → 503). This is a good improvement for real infra failures (DB down), but `authorizeDistributionAccess` has paths that return plain `fmt.Errorf` for unsupported resource types/actions — these become 503 instead of the old behavior of silent deny.

I verified `CanCreateRepository` (`authz.go:143`) already handles `sql.ErrNoRows` internally, so the main authorization paths are safe. The remaining unwrapped paths are edge cases (unsupported resource types, catalog access) that normal clients don't hit.

Affected paths in `authorizeDistributionAccess`:
- `"registry catalog is not supported by %s auth"` — unwrapped `fmt.Errorf`
- `"unsupported access resource type %q"` — unwrapped `fmt.Errorf`
- `"unsupported repository action %q"` in `authorizeMissingRepositoryAccess` — unwrapped `fmt.Errorf`

Suggested fix: wrap these returns with `errRepositoryAccessDenied`, or add a catch-all in `authorizeAction` that treats unknown policy-layer errors as denied + logs a warning.

jbpratt: Resolved in e61083977. Unsupported registry/catalog/resource and repository-action errors are now wrapped with errRepositoryAccessDenied, so token downscoping treats them as denied access while genuine infrastructure errors still propagate as HTTP 503. Added TestAuthorizeDistributionAccessClassifiesUnsupportedAccessAsDenied. Validation passed: go test ./..., go vet ./..., go test -race ./internal/registry/distribution/... ./internal/mirrorregistry/... ./internal/oci/..., and golangci-lint run --timeout=5m.

harishsurf (approved):

PR #6788, opened 2026-08-12
PROJQUAY-11682: fix(autoprune): re-land Cosign tag exclusion with race-safe cascade

nasonawa:
Re-lands [#6650](https://github.com/quay/quay/pull/6650) after it was reverted in [#6773](https://github.com/quay/quay/pull/6773).

- Excludes Cosign tag-schema names (`.sig`/`.att`/`.sbom`) from autopruner matching/counting, and fixes the `exclude_tags` `query.where(...)` assignment bug
- Cascades Cosign sibling cleanup on subject tag delete/retarget
- Closes the #6773 race by acquiring the same `compute_advisory_lock_id("retarget_tag", repository_id)` lock inside `_expire_cosign_sibling_tags` before the `NOT EXISTS` check/update (no feature flag)
- Adds lock-identity, re-entrancy, and PostgreSQL-only concurrency tests

## Test plan

- [] `TEST=true PYTHONPATH="." pytest data/model/oci/test/test_oci_tag.py -k "cosign or exclude_tags or cascade_serialized or advisory_lock or reacquires_same" -v`
- [] `TEST=true PYTHONPATH="." pytest data/model/oci/test/test_oci_tag.py::test_delete_tag data/model/oci/test/test_oci_tag.py::test_delete_tags_for_manifest -v`
- [ ] PostgreSQL: `TEST_DATABASE_URI=postgresql://... pytest ... -k cascade_serialized_with_concurrent_retarget -v`

### DIFF
--- a/data/model/oci/tag.py
+++ b/data/model/oci/tag.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+import re
 import uuid
 from calendar import timegm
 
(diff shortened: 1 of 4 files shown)


data/model/oci/tag.py
```diff
@@ -43,6 +44,51 @@
...
+        suffix_start = len(prefix) + digest_len + 1
+        suffix_clauses.append(
+            (fn.LENGTH(model.name) == expected_len)
+            & (fn.SUBSTR(model.name, suffix_start, 1 + len(suffix)) == f".{suffix}")
+        )
+    suffix_match = suffix_clauses[0]
+    for clause in suffix_clauses[1:]:
+        suffix_match = suffix_match | clause
+    return ~(prefix_match & hex_digest & suffix_match)
+
```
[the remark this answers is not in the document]
nasonawa: Mysql and MariaDB is not supported in the Quay since 3.10 version

data/model/oci/tag.py
```diff
@@ -557,6 +620,59 @@ def retarget_tag(
...
+            OtherTag.manifest == subject_manifest,
+            OtherTag.hidden == False,
+            _not_cosign_tag_schema_clause(OtherTag),
+            (OtherTag.lifetime_end_ms >> None) | (OtherTag.lifetime_end_ms > now_ms),
+        )
+        if exclude_tag_id is not None:
+            other_alive = other_alive.where(OtherTag.id != exclude_tag_id)
+        conditions.append(~fn.EXISTS(other_alive))
+
+    updated = Tag.update(lifetime_end_ms=now_ms).where(*conditions).execute()
```
cubismod: **Unique-constraint collision can crash the caller's transaction**

This bulk `UPDATE` sets `lifetime_end_ms = now_ms` on all matching Cosign sibling rows. `Tag` has a unique index on `(repository, name, lifetime_end_ms)` (`data/database.py:1915`). If a same-named historical row already has `lifetime_end_ms == now_ms` — e.g. two operations touching the same Cosign tag name land in the same millisecond, which is realistic under exactly the concurrent-retarget/delete races this PR is designed to handle — this raises an uncaught `peewee.IntegrityError`. Since this runs inside the caller's `db_transaction()` (`_delete_tag`/`retarget_tag`/`delete_tags_for_manifest`), that error rolls back the *entire* transaction and propagates up as a 500 on what should otherwise be a valid tag delete/push.

`remove_tag_from_timemachine()` elsewhere in this file hits the same index and avoids this by incrementing `lifetime_end_ms` per row instead of reusing one shared value across a bulk update. Worth applying the same pattern here (or catching `IntegrityError` and retrying with a bumped value) before this re-lands.

nasonawa: @cubismod I have added the requested changes to _expire_cosign_sibling_tags. It now follows the same pattern as remove_tag_from_timemachine.

cubismod (approved):

PR #6789, opened 2026-08-12
QUAYIO-2118: fix(ui): hide super:user scope when FEATURE_SUPER_USERS is disabled

Sunandadadi:
## Summary
- Hide the "Super User Access" (`super:user`) scope checkbox in the OAuth API token generation modal when `FEATURE_SUPER_USERS` is not enabled
- On quay.io (where `FEATURE_SUPER_USERS` is `false`), selecting this scope caused a 403 `insufficient_scope` error because `_can_mint_scope` calls `SuperUserPermission().can()` which always fails on non-superuser deployments
- Add vitest coverage for the three states: feature disabled, undefined, and enabled

### DIFF
--- a/web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
+++ b/web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
@@ -643,8 +643,13 @@ const APIAccessTokensTab: React.FC<APIAccessTokensTabProps> = (
           </FormGroup>
           <FormGroup label="Scopes" fieldId="api-token-scopes" isRequired>
             <Stack hasGutter>
-              {Object.entries(OAUTH_SCOPES).map(
-                ([scopeName, scopeInfo]: [string, OAuthScope]) => (
+              {Object.entries(OAUTH_SCOPES)
+                .filter(
+                  ([scopeName]) =>
+                    scopeName !== 'super:user' ||
+                    quayConfig?.features?.SUPER_USERS === true,
+                )
+                .map(([scopeName, scopeInfo]: [string, OAuthScope]) => (
                   <StackItem key={scopeName}>
                     <Checkbox
                       id={`api-token-scope-${scopeName}`}
--- a/web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
+++ b/web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
@@ -662,8 +662,7 @@ const APIAccessTokensTab: React.FC<APIAccessTokensTabProps> = (
                       data-testid={`api-token-scope-${scopeName}`}
                     />
                   </StackItem>
-                ),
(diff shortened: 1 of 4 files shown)


web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
```diff
@@ -643,8 +643,13 @@ const APIAccessTokensTab: React.FC<APIAccessTokensTabProps> = (
           </FormGroup>
           <FormGroup label="Scopes" fieldId="api-token-scopes" isRequired>
             <Stack hasGutter>
-              {Object.entries(OAUTH_SCOPES).map(
-                ([scopeName, scopeInfo]: [string, OAuthScope]) => (
+              {Object.entries(OAUTH_SCOPES)
+                .filter(
+                  ([scopeName]) =>
+                    scopeName !== 'super:user' ||
+                    quayConfig?.features?.SUPER_USERS === true,
```
shaon, line 650: Should we only rely on the feature flag?

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
```diff
@@ -643,8 +643,13 @@ const APIAccessTokensTab: React.FC<APIAccessTokensTabProps> = (
           </FormGroup>
           <FormGroup label="Scopes" fieldId="api-token-scopes" isRequired>
             <Stack hasGutter>
-              {Object.entries(OAUTH_SCOPES).map(
-                ([scopeName, scopeInfo]: [string, OAuthScope]) => (
+              {Object.entries(OAUTH_SCOPES)
```
deshpandevlab, line 646: on the side note: was looking for the source of OAUTH_SCOPES and looks like it is hardcoded. It duplicates in auth/scopes.py. Long-term, the frontend could consume the backend-provided scope list instead of maintaining its own copy to avoid diverging from backend provided value. This is out of scope for this PR though

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
Sunandadadi: I think so. what are you thinking?

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
Sunandadadi: good point

web/src/routes/OrganizationsList/Organization/Tabs/OAuthApplications/ManageOAuthApplicationTabs/APIAccessTokensTab.tsx
shaon: It's probably not straight-forward with static list of scopes. It's non-blocking.

shaon (approved):

PR #6799, opened 2026-08-13
PROJQUAY-12632: fix(migrate): discover and stop active OMR services when unit files not detected

harishsurf:
## Summary

- Add service discovery probe to `stopSourceServices()` when `UnitFiles` is empty
- Probe `systemctl is-active` in user then system scope to find running OMR services
- Stop discovered services before starting the new Quadlet

## Problem

When using manual path overrides (`--source-db`, `--source-storage`, `--source-certs`), auto-detection is skipped and `Source.UnitFiles` stays empty. `stopSourceServices()` sees no unit files and skips the stop entirely. The old OMR 2.0 services keep running on port 8443, causing the new Quadlet to fail with "Address already in use".

## Fix

`stopSourceServices()` in `schema.go` now calls `discoverOMRScope()` when `UnitFiles` is empty. This function probes `systemctl is-active --quiet quay-app.service` in user scope first, then system scope. If active services are found, it proceeds to stop all three OMR services (`quay-app`, `quay-redis`, `quay-pod`) in the discovered scope.

## Test plan

- [x] `TestStopSourceServices_DiscoversActiveServicesWhenUnitFilesEmpty` — probe triggers, all services stopped
- [x] `TestStopSourceServices_DiscoveryPrefersUserScope` — user scope checked first, stop uses `--user`
- [x] `TestStopSourceServices_DiscoveryFallsToSystemScope` — system scope fallback works
- [x] `TestStopSourceServices_SkipsWhenNoUnitFilesDetected` — still skips when no services active (updated)
- [x] All existing `schema_test.go` tests pass (no regression)
- [x] E2E: rootless RHEL 9 migration with manual overrides — services stopped, no port conflict
- [x] `go test ./internal/...` — all 20 packages pass

### DIFF
--- a/internal/migrate/schema.go
+++ b/internal/migrate/schema.go
@@ -60,14 +60,22 @@ func (m *Migrator) stopSourceServices(ctx context.Context) error { //nolint:unpa
 		slog.Info("no command runner, skipping service stop")
 		return nil
 	}
+
+	scope := m.Source.SystemdScope
 	if len(m.Source.UnitFiles) == 0 {
-		slog.Info("no OMR services detected, skipping stop")
-		return nil
+		discovered := m.discoverOMRScope(ctx)
+		if discovered == "" {
+			slog.Info("no OMR services detected, skipping stop")
+			return nil
+		}
+		scope = discovered
+		m.Source.SystemdScope = discovered
+		slog.Info("discovered running OMR services via probe", "scope", scope)
 	}
 
 	var scopeArgs []string
-	if m.Source.SystemdScope == scopeUser {
-		scopeArgs = []string{"--user"}
+	if scope == scopeUser {
+		scopeArgs = []string{systemdUserFlag}
 	}
 
 	for _, svc := range omrServiceNames {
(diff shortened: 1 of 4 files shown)


harishsurf:
Re: fullsend-ai SystemdScope writeback concern — already addressed in the current code. Line 72 of `internal/migrate/schema.go` explicitly writes back `m.Source.SystemdScope = discovered`, and `TestStopSourceServices_SkipsWhenNoUnitFilesDetected` (schema_test.go:151-153) asserts the discovered scope is persisted. Downstream consumers (`reloadSystemd`, `printPlan`) will see the correct value.

internal/migrate/schema.go
```diff
@@ -84,6 +92,24 @@ func (m *Migrator) stopSourceServices(ctx context.Context) error { //nolint:unpa
...
+func (m *Migrator) discoverOMRScope(ctx context.Context) string {
+	unit := omrServiceNames[0] + ".service"
+	for _, scope := range []struct {
+		name string
+		args []string
+	}{
+		{scopeSystem, []string{"is-active", "--quiet", unit}},
+		{scopeUser, []string{systemdUserFlag, "is-active", "--quiet", unit}},
+	} {
+		if err := m.Runner.Run(ctx, "systemctl", scope.args...); err == nil {
```
jbpratt, line 106: Seems we drop this error here

harishsurf: The error is intentionally dropped — systemctl is-active --quiet uses its exit code as a boolean:

exit 0 = service is active → return that scope
non-zero = not active → try the next scope

jbpratt (approved):

PR #6805, opened 2026-08-13
QUAYIO-2122: feat(secscan): add thread pool for concurrent V2 manifest indexing

kleesc:
Process claimed manifests concurrently using a ThreadPoolExecutor
(default 3 threads, configurable via SECURITY_SCANNER_V2_INDEX_THREAD_COUNT).

Only the Clair HTTP call runs in the thread pool; all Peewee DB operations
remain on the main greenlet where UseThenDisconnect manages connection
lifecycle, preventing connection leaks.


shaon (approved):
LGTM

PR #6819, opened 2026-08-14
PROJQUAY-12269: feat(web): implement search page in React UI

sudipshil9862:
## Summary

- The React UI had no `/search` route — navigating to `/search` was caught by the shorthand redirect which treated "search" as an org name and redirected to `/organization/search`
- Adds a fully functional search page at `/search` matching the Angular UI's search functionality
- Adds a global search bar to the masthead, matching the Angular UI's header search box (repositories + organizations, fuzzy matching)
- Supports both authenticated and anonymous users

## Changes

**New files:**
- `web/src/resources/SearchResource.ts` — TypeScript interfaces and fetch functions for `/api/v1/find/repositories` and `/api/v1/find/all`
- `web/src/hooks/UseSearch.ts` — `useSearch` (paginated results) and `useSearchSuggestions` (debounced typeahead) React Query hooks
- `web/src/components/header/HeaderSearchBar.tsx` / `.css` — Masthead search bar with typeahead
- `web/src/routes/Search/Search.css` — Search page styling
- `web/playwright/e2e/ui/search-page.spec.ts` — Playwright e2e tests for the search page
- `web/playwright/e2e/ui/header-search.spec.ts` — Playwright e2e tests for the header search bar

**Modified files:**
- `web/src/routes/NavigationPath.tsx` — Added `search = '/search'` to enum (also prevents shorthand redirect by registering "search" as a reserved prefix)
- `web/src/routes/StandaloneMain.tsx` — Added lazy import and route entry for Search component
- `web/src/routes/Search/Search.tsx` — Full rewrite from stub to functional search page
- `web/src/components/header/HeaderToolbar.tsx` — Renders the search bar, gated on `ANONYMOUS_ACCESS` or an authenticated user
- `conf/nginx/server-base.conf.jnj` — Route `/search` to the React app

## Features

- **Repository listing on load** — `/search` lists all visible repositories (public for anonymous, public + private for authenticated)
- **Typeahead suggestions** — Debounced (250ms) autocomplete dropdown after 3+ characters, calling `/api/v1/find/all` for repos, users, orgs, and teams
- **Header search bar** — Same typeahead in the masthead on every page; Enter goes to `/search?q=`, selecting a suggestion navigates straight to the repo or org. Hidden on `/search`, which has its own input
- **Search results** — Paginated results via `/api/v1/find/repositories` with star count, avatar, namespace/name link, description, last modified time, and signal-strength activity indicator
- **URL-driven state** — Search query and page stored in URL params (`?q=...&page=...`) for bookmarkable/shareable searches
- **PatternFly pagination** — Compact pagination in toolbar, full controls at bottom
- **Max page limit** — Respects `SEARCH_MAX_RESULT_PAGE_COUNT` config with help text at the limit; `page` is clamped client-side so an over-limit URL never reaches the API

## Accessibility

Both typeaheads implement the WAI-ARIA combobox pattern:

- ArrowUp / ArrowDown move through suggestions (wrapping at both ends), Enter selects the highlighted one, Escape closes the dropdown
- `role="combobox"`, `aria-expanded`, `aria-controls`, `aria-autocomplete`, and `aria-activedescendant` are set on the focusable `<input>` itself. PatternFly's `SearchInput` and `TextInputGroupMain` only forward a fixed set of props to the inner input and drop the rest onto a wrapper `<div>`, so these are passed through `inputProps`, with `isExpanded` used for `aria-expanded`
- The dropdown is plain `<ul role="listbox">` / `<li role="option">` markup rather than PatternFly's `Menu`. `MenuItem` applies `id` to an inner `role="menuitem"` button while `role` and `aria-selected` fall through to the `<li>`, which both broke the `aria-activedescendant` reference and nested a menuitem inside a listbox

## Test plan

All commands run from `web/`.

| Command | Result |
|---|---|
| `npx vitest run` | **Passed** — 173 files, 1287 tests |
| `npx playwright test playwright/e2e/ui/search-page.spec.ts playwright/e2e/ui/header-search.spec.ts` | **Passed** — 16/16, run with `PLAYWRIGHT_BASE_URL=http://localhost:9000` against the React dev server |
| `npx eslint` on changed source files | **Passed** |
| `npx prettier --check` on changed files | **Passed** |
| `npx tsc --noEmit` | **Passed** — no new errors (three pre-existing errors remain elsewhere: PatternFly `SVGIconProps` d.ts, `HeaderToolbar.test.tsx` `anonymous`, `AddNewTeamMemberDrawer` `isDisabled`) |

The full Playwright suite was **not run**; only the two search specs above were executed.

Playwright defaults to `http://localhost:8080`, which serves the container's prebuilt bundle. The search specs fail there until the bundle is rebuilt — this is a stale-artifact issue, not a test failure.

Manual browser testing: search page, typeahead, keyboard navigation, pagination, header search bar, anonymous access, and logged-in access.

**Playwright coverage — search page** (`web/playwright/e2e/ui/search-page.spec.ts`):
- Renders search page at `/search`
- Does not redirect to `/organization/search`
- Lists repos on load without query
- Searches repos by query with URL update
- Anonymous user can find public repos
- Typeahead suggestions appear while typing
- Combobox contract is exposed on the input, and `aria-activedescendant` resolves to the option element
- Empty state for no results

**Playwright coverage — header search bar** (`web/playwright/e2e/ui/header-search.spec.ts`):
- Present in the masthead on ordinary pages
- Hidden on the search page
- Submitting a term navigates to the search page
- Suggests repositories and navigates on click
- Suggests organizations as well as repositories
- Keyboard navigable, with `aria-activedescendant` resolving to the option element
- Escape closes the dropdown and it stays closed across a refetch
- Anonymous users get the header search bar

### DIFF
--- a/web/src/routes/Search/Search.tsx
+++ b/web/src/routes/Search/Search.tsx
@@ -38,10 +38,10 @@ export default function Search() {
   const navigate = useNavigate();
   const query = searchParams.get('q') || '';
   const parsedPage = parseInt(searchParams.get('page') || '1', 10);
-  const page = Number.isFinite(parsedPage) && parsedPage > 0 ? parsedPage : 1;
 
   const [inputValue, setInputValue] = useState(query);
   const [isDropdownOpen, setIsDropdownOpen] = useState(false);
+  const [activeSuggestionIndex, setActiveSuggestionIndex] = useState(-1);
 
   const inputRef = useRef<HTMLInputElement>(null);
   const searchBoxRef = useRef<HTMLDivElement>(null);
@@ -49,6 +49,11 @@ export default function Search() {
 
   const quayConfig = useQuayConfig();
   const maxPageCount = quayConfig?.config?.SEARCH_MAX_RESULT_PAGE_COUNT ?? 10;
+
+  const page = Math.min(
+    Number.isFinite(parsedPage) && parsedPage > 0 ? parsedPage : 1,
+    maxPageCount,
+  );
 
   const {results, hasAdditional, pageSize, startIndex, isLoading, error} =
     useSearch(query, page);
@@ -69,6 +74,10 @@ export default function Search() {
       isTyping && inputValue.trim().length >= 3 && suggestions.length > 0,
     );
   }, [suggestions, inputValue, query]);
+
+  useEffect(() => {
+    setActiveSuggestionIndex(-1);
+  }, [suggestions]);
 
   useEffect(() => {
     const handleClickOutside = (event: MouseEvent) => {
@@ -107,10 +116,30 @@ export default function Search() {
 
   const handleKeyDown = (event: React.KeyboardEvent) => {
     if (event.key === 'Enter') {
-      handleSearch();
+      if (
+        activeSuggestionIndex >= 0 &&
+        isDropdownOpen &&
+        suggestions[activeSuggestionIndex]
+      ) {
+        handleSuggestionSelect(suggestions[activeSuggestionIndex]);
+      } else {
+        handleSearch();
+      }
     }
     if (event.key === 'Escape') {
       setIsDropdownOpen(false);
+    }
+    if (event.key === 'ArrowDown' && isDropdownOpen && suggestions.length > 0) {
+      event.preventDefault();
+      setActiveSuggestionIndex((prev) =>
+        prev < suggestions.length - 1 ? prev + 1 : 0,
+      );
+    }
+    if (event.key === 'ArrowUp' && isDropdownOpen && suggestions.length > 0) {
+      event.preventDefault();
+      setActiveSuggestionIndex((prev) =>
+        prev > 0 ? prev - 1 : suggestions.length - 1,
+      );
     }
   };
 
@@ -153,6 +182,15 @@ export default function Search() {
           onKeyDown={handleKeyDown}
           ref={inputRef}
           data-testid="search-input"
+          role="combobox"
+          aria-expanded={isDropdownOpen}
+          aria-controls="search-suggestions-listbox"
+          aria-activedescendant={
+            activeSuggestionIndex >= 0
+              ? `search-suggestion-${activeSuggestionIndex}`
+              : undefined
+          }
+          aria-autocomplete="list"
         />
         <TextInputGroupUtilities>
           {inputValue && (
@@ -176,11 +214,17 @@ export default function Search() {
     <div ref={dropdownRef}>
       <Menu onSelect={() => setIsDropdownOpen(false)}>
         <MenuContent>
-          <MenuList>
+          <MenuList id="search-suggestions-listbox" role="listbox">
             {suggestions.map((s, i) => (
               <MenuItem
                 key={`${s.kind}-${s.name}-${i}`}
+                id={`search-suggestion-${i}`}
                 onClick={() => handleSuggestionSelect(s)}
+                className={
+                  i === activeSuggestionIndex ? 'search-suggestion-active' : ''
+                }
+                role="option"
+                aria-selected={i === activeSuggestionIndex}
               >
                 <Flex
                   alignItems={{default: 'alignItemsCenter'}}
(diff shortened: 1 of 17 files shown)


web/src/routes/Search/Search.tsx
```diff
@@ -1,13 +1,388 @@
...
+import RequestError from 'src/components/errors/RequestError';
+import {ISearchResult, ISearchSuggestion} from 'src/resources/SearchResource';
+import './Search.css';
 
 export default function Search() {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const navigate = useNavigate();
+  const query = searchParams.get('q') || '';
+  const parsedPage = parseInt(searchParams.get('page') || '1', 10);
+  const page = Number.isFinite(parsedPage) && parsedPage > 0 ? parsedPage : 1;
```
sridipta: `page` is validated for `NaN/negative/zero` but is not clamped to `maxPageCount`. A user navigating to `?page=500` sends that directly to the API through useSearch. The pagination UI won't show pages beyond the limit (since `effectiveTotal` is capped), but the API request still fires with the over-limit value.

Fix: clamp page before it reaches the hook:

```
const page = Math.min(
  Number.isFinite(parsedPage) && parsedPage > 0 ? parsedPage : 1,
  maxPageCount
);
```

web/src/routes/Search/Search.tsx
```diff
@@ -1,13 +1,388 @@
...
+          </Flex>
+        )}
+
+        {error && <RequestError message="Could not load search results" />}
+
+        {!isLoading && !error && results.length === 0 && (
+          <EmptyState
+            headingLevel="h2"
+            icon={SearchIcon}
+            titleText="No matching repositories found"
```
sridipta: Suggested improvement beyond building equivalence with the older Angular UI.

When `/search` loads without a query, before even the user submitting any query, saying "`No matching repositories found — Please try changing your query`" can be confusing. Branching on query to show something like "`Search for a repository to get started`" when empty would avoid the confusion. Better UX.

This is not mandatory. But we should improve on the shortcomings of the legacy UI whenever possible.

web/src/routes/Search/Search.tsx
```diff
@@ -1,13 +1,388 @@
...
+  };
+
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key === 'Enter') {
+      handleSearch();
+    }
+    if (event.key === 'Escape') {
+      setIsDropdownOpen(false);
+    }
+  };
```
sridipta, line 144: Again, not a mandatory change. But consider if we can improve on the UX when putting in new features in the new UI.

AI suggested this one. Give it a thought.

Typeahead not keyboard-accessible :accessibility:
`handleKeyDown` handles `Enter` and `Escape` only — no `ArrowDown`/`ArrowUp` for suggestion traversal. The input also lacks `role="combobox"`, `aria-expanded`, and `aria-activedescendant`. Keyboard-only and screen-reader users can't navigate suggestions at all. PatternFly's SearchInput component with expandable or a Select-based approach would give this for free, but at minimum wiring up arrow keys + ARIA attributes on the current custom implementation would close the gap.

web/src/routes/Search/Search.tsx
sudipshil9862: Fixed. Page is now clamped to maxPageCount before reaching the hook, so over-limit values never hit the API

web/src/routes/Search/Search.tsx
sudipshil9862: Good suggestion, implemented. The empty state now shows "Search for repositories" when no query is entered and "No matching repositories found" only when a search returns no results. Also handles the case where the user backspaces to clear the input.

web/src/routes/Search/Search.tsx
sudipshil9862: Agreed as in angular ui also we have this feature, implemented for react ui too. Added ArrowDown/ArrowUp for suggestion traversal, Enter to select the highlighted suggestion, and ARIA attributes (role="combobox", aria-expanded, aria-activedescendant, role="listbox", role="option") on the relevant elements.

sridipta (approved):
lgtm

PR #6831, opened 2026-08-17
PROJQUAY-12435: feat(migration): migrate OMR PostgreSQL sources

jbpratt:
## Summary

Add PostgreSQL-source support to the existing `quay migrate` command for the supported OMR v2.0.11 service topology. The migration converts the source once to SQLite, applies the existing compatibility bridge, and installs v3 without changing the current SQLite migration path.

## Related Issue

[PROJQUAY-12435](https://redhat.atlassian.net/browse/PROJQUAY-12435)

## Changes

- Freeze and validate the supported PostgreSQL schema/profile before copying all 96 source tables.
- Detect the KCS service, pod, and database topology; capture JWT material before stopping `quay-app`; and run the same binary inside `quay-pod`'s network namespace.
- Publish the converted database atomically, preserve PostgreSQL source data, and restore the old services if conversion or cutover fails.
- Complete the `3f8d7acdf7f9` to `9fa37f66a9b6` compatibility bridge and add a rootless PostgreSQL migration job to Go CI.

## Testing

- [x] `go test -count=1 ./cmd/quay ./internal/installer ./internal/migrate/... ./internal/dal/dbcore/...` passed.
- [x] `go test -race -count=1 ./internal/installer ./internal/migrate ./internal/migrate/postgres/... ./internal/dal/dbcore/...` passed.
- [x] `make go-test-postgres DOCKER=podman` passed against disposable PostgreSQL 16.
- [x] `go build ./...` passed.
- [x] `go vet ./...` passed.
- [x] `golangci-lint run ./... --timeout=5m` passed with 0 issues.
- [x] `go mod verify` passed.
- [x] `zizmor .github/workflows/ci-go.yaml` passed with no findings.
- [x] `git diff --check origin/master...HEAD` passed.
- [x] Live rootless OMR migration passed: PostgreSQL revision `3f8d7acdf7f9` migrated to SQLite revision `9fa37f66a9b6`; health returned 200, integrity was `ok`, and foreign-key violations were zero.
- [x] Live forced target-health failure passed: migration exited nonzero, removed the failed target Quadlet, and restored the old pod, PostgreSQL, Redis, and app to health.
- [ ] The new `e2e-migrate-postgres` job has not run yet; it requires GitHub Actions.

## Notes

- The CI job uses a pinned public S2I PostgreSQL 16 image because the exact KCS RHEL image requires credentials unavailable to pull-request workflows. The service, pod, mount, and URI topology remain KCS-compatible.
- PostgreSQL source data is intentionally retained after migration.


jbpratt:
/retest

@coderabbit review

internal/migrate/detect.go
```diff
@@ -251,6 +285,76 @@ func (m *Migrator) withImageArchive(src *OMRSource) OMRSource {
...
+		return "", fmt.Errorf("parse config: %w", err)
+	}
+	u, err := url.Parse(raw.DBURI)
+	if err != nil {
+		return "", fmt.Errorf("parse DB_URI")
+	}
+	switch u.Scheme {
+	case "sqlite":
+		return databaseSQLite, nil
+	case "postgresql":
```
harishsurf, line 302: PostgreSQL detection error is too generic. 

> `parsePostgresURI()` is doing several validations here, but every failure collapses into the same "DB_URI does not match the supported `quay-postgres` endpoint" error. Since this is the main operator-facing failure path for PG source detection, can we split this into specific errors (wrong scheme, wrong hostname, missing password, wrong DB path, unsupported query params, etc.)? That would make unsupported or partially supported source configs much easier to debug.

internal/migrate/postgres_source.go
```diff
@@ -0,0 +1,271 @@
...
+	if err := m.stopPostgresRemainingServices(ctx); err != nil {
+		return fmt.Errorf("stop old OMR at cutover: %w", err)
+	}
+	if err := m.install(ctx); err != nil {
+		return fmt.Errorf("install: %w", err)
+	}
+	return nil
+}
+
+func (m *Migrator) validatePostgresFlags() error {
```
harishsurf, line 80: (Note for self) We should call out in the docs that these flags are not supported for Postgres migration

internal/migrate/postgres_source.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+	if _, err := os.Stat(filepath.Join(m.DataDir, markerFile)); err == nil {
+		return fmt.Errorf("PostgreSQL migration does not support resuming a previous migration")
+	} else if !os.IsNotExist(err) {
+		return fmt.Errorf("stat migration marker: %w", err)
+	}
+	return nil
+}
+
+func (m *Migrator) hasTargetInstallation() (bool, error) {
```
harishsurf, line 143: Can we add an explicit preflight check that `-data-dir` is clean before stopping the source app? If -data-dir is already populated or corrupted (eg: from a previous failed run), we should fail early

harishsurf (commented):
Tested out the migration. Solid work! thanks for adding this! 
Left some minor comments, but otherwise LGTM

internal/dal/schema/sqlite/omr/3f8d7acdf7f9_schema.sql
```diff
@@ -0,0 +1,1260 @@
...
+	id INTEGER NOT NULL,
+	repository_id INTEGER NOT NULL,
+	uuid VARCHAR(255) NOT NULL,
+	byte_count BIGINT NOT NULL,
+	sha_state TEXT,
+	location_id INTEGER NOT NULL,
+	storage_metadata TEXT,
+	chunk_count INTEGER DEFAULT '0' NOT NULL,
+	uncompressed_byte_count BIGINT,
+	created DATETIME DEFAULT '2026-07-31 12:36:16' NOT NULL,
```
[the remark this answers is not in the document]
jbpratt: This file is a frozen schema snapshot matching OMR v2.0.11 Alembic revision `3f8d7acdf7f9`. We intentionally preserve the exact upstream DDL so that intermediate initialization and profile validation remain faithful to the approved schema contract. In practice, Peewee always supplies `created` explicitly on insert.

---

.github/workflows/ci-go.yaml
```diff
@@ -1345,3 +1345,349 @@ jobs:
...
+          sudo systemctl start quay-app.service
+          curl -kfsS --retry 60 --retry-delay 2 --retry-all-errors \
+            https://localhost:8443/health/instance
+
+          curl -kfsS --retry 30 --retry-delay 2 --retry-all-errors \
+            -X POST \
+            -H 'Content-Type: application/json' \
+            -d "{\"username\":\"init\",\"password\":\"$REGISTRY_PASSWORD\",\"email\":\"init@quay.io\"}" \
+            -o /dev/null \
+            https://localhost:8443/api/v1/user/initialize
```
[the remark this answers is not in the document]
jbpratt: Addressed in the latest push: removed retry options from the `/api/v1/user/initialize` curl invocation after instance health is confirmed.

---

internal/migrate/detect.go
jbpratt: Updated in the latest push: `parsePostgresURI` now returns specific, granular error diagnostics for invalid scheme, host, port, db path, missing user/password, and unsupported query parameters or fragments, while keeping credentials strictly redacted.

---

internal/migrate/postgres_source.go
jbpratt: Added `(SQLite only)` annotations to the flag descriptions in `newMigrateCmd()` for `-source-db`, `-dry-run`, `-cleanup`, and `-skip-install`, so `--help` and CLI usage clearly state the supported database scope for each flag.

---

internal/migrate/postgres_source.go
jbpratt: Added `validatePostgresTargetDir(m.DataDir)` inside `validatePostgresBeforeStop`. Any non-empty target directory (including leftover partial databases, previous migration markers, or corrupted files) is rejected immediately during preflight before `stopPostgresApp` is invoked, leaving the source services untouched.

---

harishsurf (approved):

PR #6971, opened 2026-08-25
fix(ui): fix keyword routing duplication (PROJQUAY-7011)

preetma-redhat:
fix(ui): fix keyword routing duplication ([PROJQUAY-7011](https://redhat.atlassian.net/browse/PROJQUAY-7011))


ibazulic:
@preetma-redhat Thanks for submitting the PR! 

Unfortunately, we won't change the current behavior of Quay. Using so-called "reserved words" in repository/organization names causes weird issues and weird API behaviour and it generally creates more problems than it actually solves.

PR #6972, opened 2026-08-25
PROJQUAY-12711: fix(mirrorregistry): avoid tag expiration timestamp collisions

jbpratt:
## Summary

- prevent concurrent same-tag manifest pushes from returning HTTP 500 when tag history timestamps collide
- choose a unique transaction-bound transition timestamp and use it for both the previous tag's end and the replacement tag's start
- resolve occupied timestamps without sleeping inside SQLite's sole write transaction
- keep pull and tag listing consistent when clock rollback produces future-relative history timestamps
- apply the same expiration path to tag replacement and deletion

## Testing

- `go test ./internal/dal/metastore -run '^TestExpireActiveTagWaitsForUnusedLifetimeEnd$' -count=10` — passed
- `go test ./internal/dal/metastore -count=1` — passed
- `go test ./internal/e2e/mirrorregistry -run '^TestRegistryConcurrentSameTagPushes$' -count=10` — passed
- `go test ./...` — passed (30 packages; 4 packages had no tests)
- `go test -race ./internal/dal/metastore -count=1` — passed
- `go test -race ./internal/e2e/mirrorregistry/...` — passed
- `go vet ./...` — passed
- `golangci-lint run --timeout=5m` — passed with 0 issues
- `sqlc generate` — passed

### DIFF
--- a/internal/dal/metastore/metastore_sqlite.go
+++ b/internal/dal/metastore/metastore_sqlite.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"strings"
 	"sync"
 	"time"
@@ -430,9 +431,9 @@ func (s *SQLiteStore) putTag(ctx context.Context, q *daldb.Queries, repoID, mani
 }
 
 // expireActiveTag chooses an unused transition millisecond no earlier than the
-// active tag's lifetime start. It begins with requestedEnd, then re-samples the
-// wall clock after a collision so forward progress and clock rollback are both
-// handled without one rollback-sized wait.
+// active tag's lifetime start. After a collision it follows the wall clock when
+// possible, or advances logically when rollback leaves the clock behind the
+// active generation.
 // q must be transaction-bound so the reads and update cannot interleave with
 // another tag mutation.
 func expireActiveTag(ctx context.Context, q *daldb.Queries, repoID int64, tag string, requestedEnd int64) (int64, error) {
@@ -469,9 +470,16 @@ func expireActiveTag(ctx context.Context, q *daldb.Queries, repoID int64, tag st
 		}
 
 		for {
-			next := max(time.Now().UnixMilli(), activeStart)
-			if next != candidate {
-				candidate = next
+			now := time.Now().UnixMilli()
+			if now < activeStart {
+				if candidate == math.MaxInt64 {
+					return 0, fmt.Errorf("no lifetime end available after %d", candidate)
+				}
+				candidate++
+				break
+			}
+			if now != candidate {
+				candidate = now
 				break
 			}
 
--- a/internal/dal/metastore/metastore_sqlite.go
+++ b/internal/dal/metastore/metastore_sqlite.go
@@ -431,9 +431,8 @@ func (s *SQLiteStore) putTag(ctx context.Context, q *daldb.Queries, repoID, mani
 }
 
 // expireActiveTag chooses an unused transition millisecond no earlier than the
-// active tag's lifetime start. After a collision it follows the wall clock when
-// possible, or advances logically when rollback leaves the clock behind the
-// active generation.
+// active tag's lifetime start. After the first collision it re-samples wall
+// time once, then advances logically while probing each candidate.
 // q must be transaction-bound so the reads and update cannot interleave with
 // another tag mutation.
 func expireActiveTag(ctx context.Context, q *daldb.Queries, repoID int64, tag string, requestedEnd int64) (int64, error) {
@@ -449,13 +448,7 @@ func expireActiveTag(ctx context.Context, q *daldb.Queries, repoID int64, tag st
 	}
 
 	candidate := max(requestedEnd, activeStart)
-	var poller *time.Ticker
-	defer func() {
-		if poller != nil {
-			poller.Stop()
-		}
-	}()
-
+	resampledWallClock := false
 	for {
 		exists, err := q.TagLifetimeEndExists(ctx, daldb.TagLifetimeEndExistsParams{
 			RepositoryID:  repoID,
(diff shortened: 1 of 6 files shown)


internal/dal/metastore/metastore_sqlite.go
```diff
@@ -434,19 +430,95 @@ func (s *SQLiteStore) putTag(ctx context.Context, q *daldb.Queries, repoID, mani
...
+
+	_, err = q.ExpireActiveTag(ctx, daldb.ExpireActiveTagParams{
+		LifetimeEndMs: sql.NullInt64{Int64: candidate, Valid: true},
 		RepositoryID:  repoID,
 		Name:          tag,
 	})
 	if err != nil {
+		return 0, err
+	}
+	return candidate, nil
```
harishsurf, line 487: On clock rollback, `candidate = max(requestedEnd, activeStart)` can be **in the future**. `DeleteTag` then writes that future `lifetime_end_ms`.

Those two queries then disagree:

- Pull / `GetActiveTagDigest`: `lifetime_end_ms IS NULL` → tag is gone
- `ListTags` / `GetTagsByRepository`: `lifetime_end_ms IS NULL OR lifetime_end_ms > now` → tag stays listed until the clock catches up

A later `PutTag` can leave **two list entries for the same name**: the zombie (`end > now`) and the new NULL-end row. The rollback tests in this PR create that data and never assert `ListTags` or `GetTagDigest`.

Python already keeps expires at `lifetime_end_ms <= now` so alive filters stay consistent (`data/model/oci/tag.py` around the Cosign expire loop).

Please either pick a unique end `<= now` (walk occupied slots, `candidate++` / `now-N`), or make list and pull use the same “alive” predicate. Don’t emit a future end and then treat it as alive in only one path.

internal/dal/metastore/metastore_sqlite_internal_test.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tx, err := db.BeginTx(ctx, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tx.Rollback() //nolint:errcheck // rollback after commit is a no-op
+	transitionMs, err := expireActiveTag(ctx, daldb.New(tx), repoID, tag, requestedEnd)
```
harishsurf, line 135: These unit tests call `expireActiveTag` directly, so a wiring bug (for example not passing `transitionMs` into `UpsertTag`) would not fail them.

`TestRegistryConcurrentSameTagPushes` also fires 50 concurrent PUTs of the **same digest**. With `MaxOpenConns=1` and HTTP overhead, those often land in different milliseconds, so that e2e can pass on unfixed master.

Please drive `store.PutTag` (and `PutManifest` with `Tag` set) against an already-occupied millisecond. After a rollback expire, assert pull 404 and a single/empty list entry. Keep the e2e, but don’t treat it as the regression.

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -434,19 +430,95 @@ func (s *SQLiteStore) putTag(ctx context.Context, q *daldb.Queries, repoID, mani
...
+			}
+
+			if poller == nil {
+				poller = time.NewTicker(5 * time.Millisecond)
+			}
+			select {
+			case <-ctx.Done():
+				return 0, ctx.Err()
+			case <-poller.C:
+			}
```
harishsurf: This 5ms ticker runs inside the caller’s transaction. `dbcore` sets `MaxOpenConns=1`, so a same-millisecond collision blocks every other metastore op (other repos, GC, blob metadata), not just this tag.

The check-then-update is only race-free because of that single connection. There is no UNIQUE retry, so raising `MaxOpenConns` later would bring the HTTP 500 back.

Please pick the next free millisecond in-process (`MAX(lifetime_end_ms)+1` or increment while occupied) with no sleep, and/or retry on UNIQUE. Don’t make correctness depend on `MaxOpenConns=1` plus a sleep.

harishsurf (changes requested):
The uniqueness fix addresses the original HTTP 500. Three items should be addressed before merge:

1. Clock-rollback expires can write a future `lifetime_end_ms`, so `ListTags` and pull disagree.
2. Tests call `expireActiveTag` directly and the e2e re-pushes the same digest, so they do not lock the product path.
3. Same-millisecond waits sleep inside the only SQLite write transaction.

Inline comments have details.

internal/dal/metastore/metastore_sqlite.go
jbpratt: Fixed in e3ec91e70. `ListTags` now selects only NULL-end rows, matching `GetTagDigest`, so future-relative history rows are no longer reported as active. The rollback coverage now verifies that deletion returns not-found from `GetTagDigest` and is absent from `ListTags`; the replacement cases verify exactly one visible entry.

---

internal/dal/metastore/metastore_sqlite_internal_test.go
jbpratt: Fixed in e3ec91e70. The SQLite regression now drives both `store.PutTag` and `store.PutManifest` with `Tag` set against a historical end equal to the future active start. It verifies the displaced end equals the replacement start, pull resolves the replacement digest, and listing contains one entry. The HTTP test now sends 50 distinct manifest bytes/digests by varying annotations instead of repeating one digest.

---

internal/dal/metastore/metastore_sqlite.go
jbpratt: Fixed in e3ec91e70. The ticker and all sleeping were removed from the transaction. After the first collision, expiration re-samples wall time once and then advances through logical milliseconds, probing the unique history index for every candidate. The current transaction-bound query path retains atomicity under OMR's configured single SQLite connection without blocking that connection on a timer.

---

harishsurf (dismissed):

harishsurf (approved):

PR #6986, opened 2026-08-25
PROJQUAY-12484: feat(auth): add kubernetes serviceaccount jwt trust validator

Marcusk19:
## Summary

- add local Kubernetes ServiceAccount JWT trust validation: issuer, signature, audience, validity, and bound identity claims (namespace/name/UID matching `sub`)
- fetch and cache OIDC discovery and JWKS using Quay's mounted ServiceAccount credential, pinned to the configured discovery origin with redirects disabled (SSRF / credential-exfiltration hardening)
- tolerate a bounded stale-key grace period (one extra `JWKS_CACHE_TTL_SECONDS` window) when a refresh fails, but fail closed immediately on an unknown key ID or a key an issuer has authoritatively rotated out
- never send the workload JWT to Kubernetes TokenReview/SAR — signature verification is entirely local

## Stack

This is the second PR in the Workload Identity Authentication for Quay Management API stack.

1. #6985 — administrator-facing workload identity configuration contract
2. **This PR** — Kubernetes ServiceAccount JWT trust validation

Parent epic: [PROJQUAY-12483](https://redhat.atlassian.net/browse/PROJQUAY-12483)
Enhancement: [quay/enhancements#45](https://github.com/quay/enhancements/pull/45)

## Scope

This PR performs local JWT trust verification only. Mapping a verified ServiceAccount identity to Quay authorization/scopes is a separate follow-up story.

## Test results

- `TEST=true PYTHONPATH="." pytest auth/test/test_kubernetes_sa.py -v` — passed, 36 tests
- `python -m mypy auth/kubernetes_sa.py` — no errors in this file (3 pre-existing errors in unrelated files)
- pre-commit hooks (black, isort, flake8, gitleaks) — passed

### DIFF
--- a/auth/kubernetes_sa.py
+++ b/auth/kubernetes_sa.py
@@ -1,0 +1,231 @@
+"""Validation of Kubernetes ServiceAccount JWTs for workload identity bootstrap."""
+
+import logging
+import re
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+from urllib.parse import urljoin, urlsplit
+
+import jwt
+from authlib.jose import JsonWebKey
+from jwt import InvalidTokenError
+
+from oauth.oidc import JWT_CLOCK_SKEW_SECONDS
+from util.security.jwtutil import decode
+
+logger = logging.getLogger(__name__)
+
+OIDC_WELL_KNOWN = ".well-known/openid-configuration"
+
+# Only RS256 is trusted for Kubernetes ServiceAccount JWTs. This is intentionally
+# separate from the generic OIDC login allow-list, which may support other algorithms.
+KUBERNETES_SA_ALLOWED_ALGORITHMS = ["RS256"]
+
+_SUBJECT_PATTERN = re.compile(r"^system:serviceaccount:([^:\s]+):([^:\s]+)$")
+
+
+class KubernetesSATokenValidationError(Exception):
+    """Raised when a Kubernetes ServiceAccount credential cannot be trusted."""
+
+
+@dataclass(frozen=True)
+class ValidatedKubernetesSA:
+    issuer: str
+    subject: str
+    claims: dict[str, Any]
+
+
+class KubernetesSATokenValidator:
+    """Validate Kubernetes ServiceAccount JWTs using OIDC discovery and JWKS.
+
+    This performs local trust verification only -- it never sends the workload
+    JWT to Kubernetes TokenReview/SAR. Mapping a verified identity to Quay
+    authorization is out of scope.
+    """
+
+    def __init__(self, config: dict[str, Any], http_client):
+        self._config = config
+        self._http_client = http_client
+
+    def validate(self, token: str) -> ValidatedKubernetesSA:
+        try:
+            unverified = jwt.decode(
+                token,
+                options={"verify_signature": False, "verify_exp": False, "verify_aud": False},
+                algorithms=KUBERNETES_SA_ALLOWED_ALGORITHMS,
+            )
+            headers = jwt.get_unverified_header(token)
+        except Exception as exc:
+            raise KubernetesSATokenValidationError(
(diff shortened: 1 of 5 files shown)


auth/kubernetes_sa.py
```diff
@@ -0,0 +1,313 @@
...
+    """Caches loader() results with a bounded grace period for refresh failures.
+
+    A value is reused without calling loader() until `ttl_seconds` elapses. Past
+    that, a refresh is attempted; if loader() raises, the previous value is reused
+    for one additional `ttl_seconds` window (bounded staleness) before the entry is
+    dropped entirely and callers must obtain a fresh value or fail. A successful
+    refresh always replaces the previous value, even if it no longer contains what
+    the caller is looking for -- there is no stale fallback once the issuer has
+    authoritatively responded.
+    """
```
ibazulic: Not sure if we should cache service accounts information. What if we don't cache it? That would be one extra API call right?

auth/kubernetes_sa.py
```diff
@@ -0,0 +1,313 @@
...
+                    return entry.value, True
+                self._entries[key] = _CacheEntry(value, now)
+                return value, False
+            # Past the bounded grace window: no more stale fallback is permitted.
+            self._entries.pop(key, None)
+
+        # Cold cache, forced refresh, or expired past the grace window: must succeed.
+        value = loader()
+        self._entries[key] = _CacheEntry(value, now)
+        return value, False
```
ibazulic: What happens if the service account gets deleted in Kubernetes while we're holding the cached response?

auth/kubernetes_sa.py
```diff
@@ -0,0 +1,313 @@
...
+        self._http_client = http_client
+        ttl = config.get("JWKS_CACHE_TTL_SECONDS", _DEFAULT_CACHE_TTL_SECONDS)
+        self._metadata_cache = _StaleTolerantCache(ttl)
+        self._jwks_cache = _StaleTolerantCache(ttl)
+
+    def validate(self, token: str) -> ValidatedKubernetesSA:
+        try:
+            unverified = jwt.decode(
+                token,
+                options={"verify_signature": False, "verify_exp": False, "verify_aud": False},
```
ibazulic, line 55: I think we should always verify that the JWT contains a proper audience, signature and expiration time. In this case, we would accept a JWT that is properly formatted but is missing any of the provided properties. For instance, we could accept a bearer token that has expired a very long time ago. Or has an improper signature. Or is missing an audience or has a completely wrong audience set.

auth/kubernetes_sa.py
```diff
@@ -0,0 +1,313 @@
...
+    def _get_key(self, issuer_config: dict[str, Any], kid: str):
+        issuer = issuer_config["ISSUER"]
+
+        keys_by_kid, stale = self._jwks_cache.get(issuer, lambda: self._fetch_jwks(issuer_config))
+        if stale:
+            logger.warning(
+                "Kubernetes ServiceAccount JWKS refresh failed for issuer '%s'; using cached "
+                "signing keys for the remainder of the bounded grace period",
+                issuer,
+            )
```
ibazulic: If we have a stale result, we should try to refresh immediately, not use the current key for the remainder of the grace period. We should only use keys that did not go stale.

auth/kubernetes_sa.py
Marcusk19: yeah I see a lot of your concerns as well are a result of caching. I'm going to opt to not use the cache in this case. saving the extra api call doesn't seem to be a good tradeoff for the security concerns

auth/kubernetes_sa.py
Marcusk19: this is the first part of a two-step verification where we get the `iss` and `kid` first then make sure it's listed in the config + fetch its current key. after that we do a more strict verification of audience, expiration, etc.

PR #7033, opened 2026-09-01
PROJQUAY-12913: fix(omr): reject index PUT when a child manifest is missing from SQLite

harishsurf:
## Summary

- On a tagged index PUT every child must already be in SQLite. `linkChild` is lookup-or-fail: a missing child returns `oci.ChildManifestUnknownError`, the whole `PutManifest` transaction rolls back, and the registry answers `400 MANIFEST_BLOB_UNKNOWN` naming the child digest. No `{}` placeholder rows, no parent-media-type fallback, no per-child temp tag on index PUT.
- `UpsertManifest` stays a plain upsert of real bytes; an index PUT never overwrites a child's `manifest_bytes`.
- Children keep the media type and the 1h `$temp-` protection from their own digest PUT (#7024), which is how `skopeo copy --all` pushes them.

## Context

`skopeo copy --all` PUTs each child by digest, then the tagged index. On master, if a GC cycle ran in between, the children were collected and the index PUT re-created them as `{}` stubs with no `manifestblob` rows (`linkChild` placeholder). Once `uploadedblob` expired, GC deleted the layers and pulls failed with `BLOB_UNKNOWN` while the tag still existed.

Reproduced live against a master binary with skopeo: children by digest → GC cycle → index PUT → `201` with six 2-byte stubs and `manifestblob=0` → `BLOB_UNKNOWN` after the upload window. With this branch the same index PUT is rejected with `400 MANIFEST_BLOB_UNKNOWN` and zero rows are written; a normal `skopeo copy --all` is unchanged and remains pullable after `uploadedblob` expiry and GC.

Earlier revisions of this PR materialized children from Distribution's blob store; that was dropped per review (two sources of truth). The DistDriver / DB-only serving change is #7054.

Stacked on #7024 ([PROJQUAY-12886](https://redhat.atlassian.net/browse/PROJQUAY-12886)); merge after it.

Related Jira: [PROJQUAY-12913](https://issues.redhat.com/browse/PROJQUAY-12913)

## Test plan

- [x] `go test ./internal/dal/metastore/ ./internal/registry/distribution/middleware/ ./internal/e2e/mirrorregistry/ -count=1`
- [x] `TestPutManifest_IndexWithMissingChildRejected` and `TestRegistryIndexWithMissingChildRejected` fail on #7024 alone and pass here; the e2e test asserts HTTP 400, `MANIFEST_BLOB_UNKNOWN`, the child digest, and `COUNT(*) FROM manifest = 0`
- [x] `TestManifestPut_MissingChild_ReturnsVerificationError`: the store error maps to `distribution.ErrManifestVerification{ErrManifestBlobUnknown{child}}` (a bare `ErrManifestBlobUnknown` is a 500 in distribution's handler)
- [x] `TestRegistryIndexPutAfterChildDigestDoesNotClobber` and `TestPutManifest_IndexDoesNotClobberNonEmptyChildBytes`
- [x] Live skopeo against built binaries: missing child → 400 (master: 201 + stubs); `skopeo copy --all` pullable after GC
- [x] `golangci-lint run` clean


ibazulic:
We should **not** store manifests as blobs in backend storage. As long as we store manifest information in two separate places, we will have to think about how any changes we do influence other things in the registry. It will be difficult to maintain and synchronize as we add more things to the rewrite. We must do things properly immediately.

If we determine that our source of truth is the database (which is why we have it), then manifest data (image metadata) must be stored there and only there and we should always load it from the database. Therefore, `PUT v2/{name}/manifests/{ref}` should be an internal db operation only.

internal/registry/distribution/middleware/manifest.go
```diff
@@ -52,15 +52,26 @@ func (ms *manifestService) Put(ctx context.Context, manifest distribution.Manife
...
 	if record.Subject != "" {
 		SetSubject(ctx, record.Subject)
 	}
 
 	// Classify references as blobs or child manifests based on parent media type.
 	if isIndexMediaType(mt) {
+		blobs := ms.repo.Repository.Blobs(ctx)
 		for _, ref := range manifest.References() {
 			record.ChildDigests = append(record.ChildDigests, ref.Digest)
+			if child, ok := loadChildFromCAS(ctx, blobs, ref); ok {
```
jbpratt, line 72: **Blocker.** Reading the child back from `Blobs()` only works because distribution wrote the manifest payload into blob storage, which is the two-sources-of-truth problem from the earlier comment. With the driver change this `Get` always misses.

The cleaner fix is one layer down, in [`distcompat.go`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/oci/storage/local/distcompat.go#L145-L180): resolve revision links from SQLite only and make the blob-path write a no-op.

```go
func (d *DistDriver) resolveManifestRevision(ctx context.Context, repoID int64, dgst digest.Digest) (digest.Digest, error) {
	return d.meta.GetManifestDigest(ctx, repoID, dgst)
}

// PutContent
case pathBlob:
	if _, err := digestFromBlobPath(path); err != nil {
		return err
	}
	return nil // manifest bytes live in manifest.manifest_bytes
```

Distribution only reaches blob-path `PutContent` from the three manifest handlers (layers go through the upload lifecycle and `Move`), and the existing blob-path read already falls through to `manifest_bytes`. With that in place, distribution's index validation (`Validation.Manifests.Indexes.Platforms` defaults to `all`) rejects an index whose child is not in the DB, so this branch can go back to collecting `ChildDigests` only and `ChildManifests` can be dropped from `oci.ManifestRecord`.

_AI-assisted review comment (Claude Code)._

internal/registry/distribution/middleware/manifest.go
```diff
@@ -129,3 +140,76 @@ func isIndexMediaType(mt string) bool {
...
+// its config/layers (or nested children). A miss is not fatal: Python only
+// fails the whole create on a hard load error, and distribution already
+// accepted the index. The metastore then either finds an existing row or
+// inserts a non-clobbering stub.
+func loadChildFromCAS(ctx context.Context, blobs distribution.BlobStore, ref v1.Descriptor) (oci.ManifestRecord, bool) { //nolint:gocritic // descriptor comes from distribution by value
+	if blobs == nil || ref.Digest == "" {
+		return oci.ManifestRecord{}, false
+	}
+	content, err := blobs.Get(ctx, ref.Digest)
+	if err != nil || len(content) == 0 {
```
jbpratt, line 154: **Major.** Every error path here, including real storage errors, is treated as "not found" and degrades to a `{}` stub with no `manifestblob` rows, which is the state this PR is fixing. CodeRabbit flagged the error handling; the deeper fix is that a child that cannot be found must fail the index write and roll back, see [`linkChild`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/dal/metastore/metastore_sqlite.go#L246-L271) on the alternative branch. Then this function is not needed.

_AI-assisted review comment (Claude Code)._

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -234,38 +240,166 @@ func (s *SQLiteStore) setSubjectAndProtect(ctx context.Context, q *daldb.Queries
...
+		seen[children[i].Digest] = struct{}{}
+		if err := s.linkChild(ctx, q, tx, repoID, parentID, parentMediaTypeID, &children[i]); err != nil {
+			return err
+		}
+	}
+	for _, dgst := range digests {
+		if _, ok := seen[dgst]; ok {
+			continue
+		}
+		stub := oci.ManifestRecord{Digest: dgst}
```
jbpratt, line 286: **Major.** Same point at the store layer: any digest without a CAS hit becomes a `{}` stub linked to the parent. Recommendation: `linkChild` looks the child up and fails when it is absent, which lets `linkChildren`, `upsertChildManifest` and `hasRealManifestBytes` go away:

```go
child, err := q.GetManifestByDigest(ctx, daldb.GetManifestByDigestParams{
	RepositoryID: repoID,
	Digest:       childDgst.String(),
})
if errors.Is(err, sql.ErrNoRows) {
	return fmt.Errorf("child manifest %s: %w", childDgst, oci.ErrNotExist)
}
```

The whole `PutManifest` transaction rolls back, so no index row, no stub, no tag. Full version: [`metastore_sqlite.go#L246-L271`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/dal/metastore/metastore_sqlite.go#L246-L271), with a store test at [`TestPutManifest_IndexWithMissingChildRejected`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/dal/metastore/metastore_temptag_test.go#L152-L182).

_AI-assisted review comment (Claude Code)._

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -234,38 +240,166 @@ func (s *SQLiteStore) setSubjectAndProtect(ctx context.Context, q *daldb.Queries
...
+	})
+	haveContent := hasRealManifestBytes(string(child.Content))
+	if errors.Is(err, sql.ErrNoRows) {
+		mtID := parentMediaTypeID
+		if child.MediaType != "" {
+			if id, merr := s.resolveMediaType(child.MediaType); merr == nil {
+				mtID = id
+			}
+		}
+		content := "{}"
```
jbpratt, line 355: **Major.** `content := "{}"` (and the parent-media-type fallback at 349 and 376) exist only to tolerate placeholders. Once children are never created during an index PUT, both go away and the child keeps the media type it was pushed with.

Related: a single place deciding how a manifest is protected is simpler than spreading it across `PutManifest`, `linkChild` and the middleware's `TempTagExpiration`:

```go
func (s *SQLiteStore) protectManifest(ctx context.Context, q *daldb.Queries, repoID, manifestID int64, m *oci.ManifestRecord) error {
	switch {
	case m.Tag != "":
		_, err := s.putTag(ctx, q, repoID, manifestID, m.Tag)
		return err
	case m.Subject != "":
		return s.setSubjectAndProtect(ctx, q, repoID, manifestID, m)
	default:
		return s.ensureTempTag(ctx, q, repoID, manifestID)
	}
}
```

([`metastore_sqlite.go#L189-L204`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/dal/metastore/metastore_sqlite.go#L189-L204).) That also keeps the store the owner of the policy, so `oci.ManifestRecord.TempTagExpiration` in `metadata.go` is not needed.

_AI-assisted review comment (Claude Code)._

internal/dal/queries/manifest.sql
```diff
@@ -7,7 +7,11 @@ WHERE repository_id = ? AND digest = ?;
 INSERT INTO manifest (repository_id, digest, media_type_id, manifest_bytes, subject, artifact_type, subject_backfilled, artifact_type_backfilled)
 VALUES (?, ?, ?, ?, ?, ?, 1, 1)
 ON CONFLICT (repository_id, digest) DO UPDATE
-  SET manifest_bytes = excluded.manifest_bytes,
+  SET manifest_bytes = CASE
+        WHEN excluded.manifest_bytes IN ('', '{}') AND manifest.manifest_bytes NOT IN ('', '{}')
```
jbpratt, line 11: **Major.** The `CASE` guard encodes stub semantics into the schema. If `PutManifest` never writes `{}`, `UpsertManifest` can stay a plain upsert of real bytes and this hunk (and the generated `manifest.sql.go` change) reverts.

_AI-assisted review comment (Claude Code)._

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -234,38 +240,166 @@ func (s *SQLiteStore) setSubjectAndProtect(ctx context.Context, q *daldb.Queries
...
-		ChildManifestID: child.ID,
+		ChildManifestID: childID,
 	}); err != nil {
-		return fmt.Errorf("link child manifest %s: %w", childDgst, err)
+		return fmt.Errorf("link child manifest %s: %w", child.Digest, err)
+	}
+
+	// Python get_or_create_manifest(..., for_tagging=False) attaches a 1h
+	// $temp- tag to children created during index PUT. Skip "{}" stubs.
+	if hasRealManifestBytes(string(child.Content)) {
```
jbpratt, line 334: **Minor.** Python attaches a temp tag during index processing because `_create_manifest` can create children there. If children are never created on index PUT, the only tag they need is the one from their own digest PUT (#7024), so this second `protectWithTempTag` call per child is redundant.

_AI-assisted review comment (Claude Code)._

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -234,38 +240,166 @@ func (s *SQLiteStore) setSubjectAndProtect(ctx context.Context, q *daldb.Queries
...
+// protectWithTempTag creates a hidden $temp- tag with lifetime_end_ms set to
+// now+expiration, unless a tag already protects the manifest for that window.
+// This is Python create_temporary_tag_if_necessary(expiration_sec) for
+// digest-only manifest PUTs (write_manifest_by_digest).
+func (s *SQLiteStore) protectWithTempTag(ctx context.Context, q *daldb.Queries, repoID, manifestID int64, expiration time.Duration) error {
+	now := time.Now()
+	endMs := now.Add(expiration).UnixMilli()
+	hasTag, err := q.HasProtectingTagForManifest(ctx, daldb.HasProtectingTagForManifestParams{
+		ManifestID:    sql.NullInt64{Int64: manifestID, Valid: true},
+		LifetimeEndMs: sql.NullInt64{Int64: now.UnixMilli(), Valid: true},
```
jbpratt, line 252: **Minor** (carried from #7024). Comparing against `now` means a tag expiring in one second counts as protection, so a re-push just before expiry gets nothing. CodeRabbit raised this too.

Two small changes fix it and avoid the row-per-repush that Python's version has: compare against the *new* end time, and extend an existing temp tag instead of skip-or-insert.

```sql
-- name: HasTagProtectingManifestUntil :one
SELECT EXISTS(
    SELECT 1 FROM tag
    WHERE manifest_id = ? AND (lifetime_end_ms IS NULL OR lifetime_end_ms >= ?)
) AS has_tag;

-- name: ExtendTempTag :execrows
UPDATE tag SET lifetime_end_ms = ?
WHERE manifest_id = ? AND hidden = 1 AND lifetime_end_ms IS NOT NULL AND lifetime_end_ms < ?;
```

Flow: protected until `end`? return. `ExtendTempTag` touched a row? return. Otherwise insert. See [`ensureTempTag`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/dal/metastore/metastore_sqlite.go#L273-L322) and [`tag.sql#L57-L77`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/dal/queries/tag.sql#L57-L77). Test: [`TestPutManifest_UntaggedGetsTempTag`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/dal/metastore/metastore_temptag_test.go#L57-L111) asserts one row after re-push and that the expiry only moves forward.

_AI-assisted review comment (Claude Code)._

internal/e2e/mirrorregistry/registry_test.go
```diff
@@ -235,25 +299,260 @@ func TestRegistryMultiArchGarbageCollectionCascade(t *testing.T) {
...
+
+	childManifest, err := json.Marshal(v1.Manifest{
+		Versioned: specs.Versioned{SchemaVersion: 2},
+		MediaType: v1.MediaTypeImageManifest,
+		Config:    v1.Descriptor{MediaType: v1.MediaTypeImageConfig, Digest: configDigest, Size: int64(len(configBytes))},
+		Layers:    []v1.Descriptor{{MediaType: v1.MediaTypeImageLayerGzip, Digest: layerDigest, Size: int64(len(layerBytes))}},
+	})
+	require.NoError(t, err)
+	childDigest := digest.FromBytes(childManifest)
+	_, err = h.Registry().PushBlob(ctx, repository, childManifest)
```
jbpratt, line 345: **Major** (same pattern at 411 and 509). `PushBlob(ctx, repository, childManifest)` uploads manifest JSON through the blob-upload endpoint. No OCI client does this; a manifest only reaches storage via distribution's own manifest write, which the driver change stops. So the scenario this test constructs cannot occur, and with DB-only manifests the index PUT here returns 400.

Recommendation:
- invert `TestRegistryIndexPutWhileChildRowMissingDoesNotClobber` to assert the 400 and that no rows were created;
- drop `TestRegistryIndexOnlyChildLayersSurviveUploadExpiry` and `TestRegistryIndexOnlyChildSurvivesIndexDeleteUntilTempTagExpires`;
- keep `TestRegistryDigestPushSurvivesCollectWithinUploadWindow`, `TestRegistryTaggedImageSurvivesUploadExpiry` and `TestRegistryIndexPutAfterChildDigestDoesNotClobber`, they are good regardless of which temp-tag commit lands.

For the inverted case there is a ready shape in [`TestRegistry_IndexWithMissingChildRejected`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/gc/registry_consistency_test.go#L276-L285): push the child by digest, delete its rows to simulate the GC, PUT the index, expect 400 and `COUNT(*) FROM manifest = 0`. `Harness.ExpireTempTags` is equivalent to [`ExpireTempTagProtection`](https://github.com/jbpratt/quay/blob/dc60cce7a525d591a84736775ccbaa50be41087c/internal/e2e/mirrorregistry/internal/e2etest/harness.go#L174-L187) on the other branch; whichever lands first, keep one.

_AI-assisted review comment (Claude Code)._

internal/gc/gc_test.go
```diff
@@ -1432,3 +1432,121 @@ func ctx(t *testing.T) context.Context {
...
+		Digest:            manifestDgst,
+		MediaType:         "application/vnd.oci.image.manifest.v1+json",
+		Content:           []byte(`{"schemaVersion":2}`),
+		BlobDigests:       []oci.BlobRef{{Digest: blobDgst, Size: int64(len(content))}},
+		TempTagExpiration: oci.PushTempTagExpiration,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	linkBlobToManifest(t, env, repoID, manifestID, blobID)
```
jbpratt, line 1533: **Minor.** `PutManifest` already linked this blob via `BlobDigests`, so linking it again means the test still passes if `PutManifest` stopped writing `manifestblob`. Drop the manual link so the test guards the path it targets.

_AI-assisted review comment (Claude Code)._

jbpratt (changes requested):
**Request changes.** This stack independently found both real bugs here (the `{}` placeholder children and the missing temp tag on digest-only pushes), and its e2e coverage is better than what `master` has.

The diagnosis is off by one step. The child JSON found in CAS was not left there by skopeo. Distribution writes every manifest payload into blob storage as an untracked blob, and the child's DB row was gone because the GC collects untagged manifests on its next 30 s cycle. This was reproduced on `master` with an HTTP-driven test (real quay driver, quaydb middleware, SQLite, collector): push a manifest by digest, run one collector cycle, the row is deleted while HEAD/GET still return 200 from storage, and a tagged index pushed afterwards gets a `{}` child with zero `manifestblob` rows. That is exactly the field state this PR patches around.

Given @ibazulic's point that manifests must not live in blob storage, materializing children from CAS builds on the thing that has to be removed. Once manifests are DB-only, `loadChildFromCAS` cannot find anything, and distribution rejects an index with an unknown child (400 `MANIFEST_BLOB_UNKNOWN`) before the middleware runs, which is what the spec requires and what Python does.

Alternative on [`jbpratt:fix/no-issue-omr-manifest-source-of-truth`](https://github.com/jbpratt/quay/tree/fix/no-issue-omr-manifest-source-of-truth) (3 commits on `master`, `go test ./internal/...` and `golangci-lint` green):
- [`e7a24a7e9`](https://github.com/jbpratt/quay/commit/e7a24a7e9b73b842fb24efc475e8f4b6ba9c9900) DB-only manifests: revision links resolve from SQLite only; blob-path `PutContent` is a no-op so manifest bytes never reach storage.
- [`93537b979`](https://github.com/jbpratt/quay/commit/93537b979ca0cef8a0a397042e305386236d823a) temp tag for every untagged non-referrer manifest at push time, renewed on re-push; missing index children are rejected, no placeholder.
- [`dc60cce7a`](https://github.com/jbpratt/quay/commit/dc60cce7a525d591a84736775ccbaa50be41087c) registry/GC consistency e2e tests covering the report's scenarios.

Proposed convergence: land `e7a24a7e9` first (independent, settles the design question), then one temp-tag implementation (#7024 or `93537b979`, maintainers' call), then rebase this PR down to its tests with the "child row missing" cases inverted to expect 400. Details inline.

Open decision for maintainers: hidden temp tags expire after 1 h but are only collected after the namespace's two-week `removed_tag_expiration_s`, matching Python. If OMR wants faster reclaim, `FindExpiredTags` could collect hidden tags at `lifetime_end_ms`. Neither branch changes that today.

---
_AI-assisted review (Claude Code). The reproduction, the alternative branch, and the test runs cited above were produced with the tool and checked before posting._

internal/dal/daldb/manifest.sql.go
```diff
@@ -279,7 +279,11 @@ const upsertManifest = `-- name: UpsertManifest :one
 INSERT INTO manifest (repository_id, digest, media_type_id, manifest_bytes, subject, artifact_type, subject_backfilled, artifact_type_backfilled)
 VALUES (?, ?, ?, ?, ?, ?, 1, 1)
 ON CONFLICT (repository_id, digest) DO UPDATE
-  SET manifest_bytes = excluded.manifest_bytes,
+  SET manifest_bytes = CASE
+        WHEN excluded.manifest_bytes IN ('', '{}') AND manifest.manifest_bytes NOT IN ('', '{}')
+        THEN manifest.manifest_bytes
+        ELSE excluded.manifest_bytes
+      END,
```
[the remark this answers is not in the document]
harishsurf: Done by removal, not by updating the CASE. Children are never created on index PUT now, so there is no stub backfill and no leftover parent index media type. `upsertChildManifest` is gone.

internal/dal/metastore/metastore_sqlite.go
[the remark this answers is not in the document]
harishsurf: Done (via #7024, this branch is rebased on it). `HasProtectingTagForManifest` is compared against the requested `endMs`, and an existing hidden tag is extended instead of skipped.

internal/gc/gc_test.go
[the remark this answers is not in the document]
harishsurf: Done on #7024 (this PR includes that commit). `TestCollect_DigestOnlyTempTagProtectsBlobsAfterUploadedBlobExpiry` now relies on `PutManifest` / `BlobDigests` only.

internal/registry/distribution/middleware/manifest.go
```diff
@@ -129,3 +140,76 @@ func isIndexMediaType(mt string) bool {
...
+		return oci.ManifestRecord{}, false
```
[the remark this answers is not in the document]
harishsurf: Done by removal. `loadChildFromCAS` is gone, so there is no Get-from-CAS error path that degrades to a `{}` stub. A missing child fails `linkChild` and the index PUT rolls back.

internal/registry/distribution/middleware/manifest.go
```diff
@@ -129,3 +140,76 @@ func isIndexMediaType(mt string) bool {
...
+	rec := parseChildRecord(ref, content)
+	if len(rec.ChildDigests) == 0 {
+		return rec, true
+	}
+	var nested struct {
+		Manifests []v1.Descriptor `json:"manifests"`
+	}
+	_ = json.Unmarshal(content, &nested)
+	for _, childRef := range nested.Manifests {
+		if child, ok := loadChildFromCAS(ctx, blobs, childRef); ok {
```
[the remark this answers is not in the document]
harishsurf: Done by removal. There is no recursive CAS walk anymore; index PUT only records `ChildDigests` and links existing SQLite rows.

internal/registry/distribution/middleware/manifest.go
harishsurf: Done for this PR: `loadChildFromCAS` is gone; index PUT no longer reads children from `Blobs()`. The DistDriver blob-path no-op / SQLite-only revision resolve is the catalog work (#7054), not this stack — we did not change `distcompat.go` here.

internal/registry/distribution/middleware/manifest.go
harishsurf: Done. `loadChildFromCAS` is gone. If the child is not in SQLite, `linkChild` returns `oci.ErrNotExist`, `PutManifest` rolls back, and middleware maps that to `distribution.ErrManifestBlobUnknown`.

internal/dal/metastore/metastore_sqlite.go
harishsurf: Done. `linkChild` is lookup-or-fail; `linkChildren`, `upsertChildManifest`, and `hasRealManifestBytes` are gone. `TestPutManifest_IndexWithMissingChildRejected` asserts the error and zero manifest rows.

internal/dal/metastore/metastore_sqlite.go
harishsurf: Done for the stub part: no `{}` insert and no parent-media-type fallback. The child keeps the media type from its own digest PUT. We did not fold everything into a single `protectManifest` switch in this PR; digest-only still uses `TempTagExpiration` from #7024.

internal/dal/queries/manifest.sql
harishsurf: Done. `UpsertManifest` is a plain upsert of real bytes again; the CASE guard and generated `manifest.sql.go` change are reverted.

internal/dal/metastore/metastore_sqlite.go
harishsurf: Done. Index PUT no longer calls `protectWithTempTag` per child. Children get the 1h tag from their own digest PUT (#7024).

internal/dal/metastore/metastore_sqlite.go
harishsurf: Done on #7024 (included here). Protection is checked against the requested end time, and `ExtendTempTag` renews the existing hidden row instead of inserting another.

internal/e2e/mirrorregistry/registry_test.go
harishsurf: Done. Dropped `TestRegistryIndexOnlyChildLayersSurviveUploadExpiry` and `TestRegistryIndexOnlyChildSurvivesIndexDeleteUntilTempTagExpires`. Replaced the race/blob-upload test with `TestRegistryIndexWithMissingChildRejected` (digest-PUT the child, wipe SQLite catalog, PUT index, expect error and `COUNT(*) FROM manifest = 0`). Kept `TestRegistryIndexPutAfterChildDigestDoesNotClobber`.

internal/gc/gc_test.go
harishsurf: Done on #7024 (this PR includes that commit). The extra `linkBlobToManifest` call is gone.

harishsurf:
@jbpratt addressed the write-path review on this stack (rebased on #7024).

**Done here**
- No `loadChildFromCAS` / no `{}` stubs. `linkChild` is lookup-or-fail; missing child rolls back the index PUT (`ErrManifestBlobUnknown`).
- `UpsertManifest` CASE reverted. No per-child `$temp-` on index PUT; digest-only protection + extend lives in #7024.
- Dropped the index-only blob-upload tests; added `TestRegistryIndexWithMissingChildRejected`; kept sequential child-then-index.

**Not in this PR**
- DistDriver blob-path no-op / SQLite-only catalog GET (`e7a24a7e9` / #7054).
- Consolidating tag/referrer/temp-tag into a single `protectManifest` switch.

Inline replies are on each comment.

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -234,26 +242,64 @@ func (s *SQLiteStore) setSubjectAndProtect(ctx context.Context, q *daldb.Queries
...
-			RepositoryID:  repoID,
-			Digest:        childDgst.String(),
-			MediaTypeID:   mtID,
-			ManifestBytes: "{}",
-		})
-		if err != nil {
-			return fmt.Errorf("ensure child manifest %s: %w", childDgst, err)
-		}
-	} else if err != nil {
+		return fmt.Errorf("child manifest %s: %w", childDgst, oci.ErrNotExist)
```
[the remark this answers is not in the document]
harishsurf: Not applying this. With `skopeo copy --all` (and every OCI client we know of) children are PUT by digest before the index, and #7024 keeps them alive with a 1h `$temp-` tag, so an index whose child is missing from SQLite means the catalog lost it (GC or a manual delete), not that the client skipped it. Re-materializing the child from Distribution's blob store would reintroduce the second source of truth that @ibazulic and @jbpratt asked us to remove; the agreed behavior is to reject the PUT and write nothing.

The latest commit makes that explicit: `linkChild` returns `oci.ChildManifestUnknownError{Digest}` and the middleware wraps it in `distribution.ErrManifestVerification`, so the client gets `400 MANIFEST_BLOB_UNKNOWN` naming the child (a bare `ErrManifestBlobUnknown` was a 500). Covered by `TestRegistryIndexWithMissingChildRejected` (HTTP 400 + child digest + zero rows) and `TestManifestPut_MissingChild_ReturnsVerificationError`.

harishsurf:
@jbpratt rebased on master + #7024 and pushed `8ccb66cbd`. What changed since your review, beyond the inline replies:

- **Missing child now answers `400 MANIFEST_BLOB_UNKNOWN` naming the child digest.** The previous revision returned a bare `distribution.ErrManifestBlobUnknown` with the parent digest; distribution's handler only maps that to 400 when it is wrapped in `ErrManifestVerification`, bare it fell through to `500 ErrorCodeUnknown` (the same thing you flagged on #7054). Verified live with skopeo against built binaries: master `201` + `{}` stubs, previous revision `500`, this one `400`. New `oci.ChildManifestUnknownError` carries the child digest and still satisfies `errors.Is(err, oci.ErrNotExist)`.
- **Tests:** `TestManifestPut_MissingChild_ReturnsVerificationError` (middleware mapping), and `TestRegistryIndexWithMissingChildRejected` now asserts HTTP 400, `MANIFEST_BLOB_UNKNOWN`, the child digest and zero manifest rows. Both missing-child tests fail on #7024 alone.
- **Dead scaffolding from the CAS approach removed:** `mockBlobStore.Get`, `Harness.ExpireTempTags`; added the `ManifestCount` doc comment that failed lint.
- Normal `skopeo copy --all` (children then index) is unchanged and stays pullable after `uploadedblob` expiry + GC, checked live on master, #7024 and this branch.

PR title and description updated to match the reject-on-missing-child design. The outdated threads above are all addressed in the current diff; happy for you to resolve or reopen any of them.

PR #7036, opened 2026-09-02
PROJQUAY-6259: feat(readonly): add operator-managed readonly lifecycle hooks

LiZhang19817:
## Summary

This PR adds the Quay application-side hooks required for the operator-managed read-only mode feature ([PROJQUAY-6259](https://issues.redhat.com/browse/PROJQUAY-6259)). The companion operator-side PR (against `quay/quay-operator`) will follow separately.

The feature enables a single `spec.readOnly` toggle on the `QuayRegistry` custom resource to transition Quay into and out of read-only mode for application-consistent backups.

These changes add the missing automation hooks that the operator needs — Quay's existing `REGISTRY_STATE: readonly` enforcement is reused unchanged.

### 1. Config: `INSTANCE_SERVICE_KEY_IMPORT_FROM_FILES` (config.py, util/config/schema.py)

- New boolean config key, defaults to `False`, added to `INTERNAL_ONLY_PROPERTIES`.
- Gates the file-based import path so existing deployments with default key files (`/conf/stack/quay.kid`, `/conf/stack/quay.pem`) do not accidentally trigger import.
- Operator renders this as `true` only when the readonly Secret is mounted.

### 2. Keyserver GET service filtering fix (endpoints/keyserver/\_\_init\_\_.py)

- **Bug fix:** The existing `GET /keys/services/<service>/keys/<kid>` endpoint accepted the `service` URL parameter but never passed it to the model lookup — keys were fetched by `kid` only.
- Now passes `service=service` to `model.get_service_key()`, so the operator can verify a key belongs to the expected service.

### 3. Keyserver `/status` endpoint (endpoints/keyserver/, data/model/service_keys.py)

- New `GET /keys/services/<service>/keys/<kid>/status` endpoint returning:
  ```json
  {"kid": "...", "service": "quay", "operator_managed": true, "expiration_date": "2026-09-16T03:54:00Z"}
  ```
- Returns `200` for expired and unapproved keys (unlike the raw GET which returns `403`/`409`), so the operator can read key metadata regardless of key state.
- Uses a new `get_service_key_for_status()` model function that **bypasses stale-expired filtering** in `_list_service_keys_query()`. This is required because keys expired longer than `EXPIRED_SERVICE_KEY_TTL_SEC` (7 days) are silently dropped by the existing query, but the operator needs to inspect them for cleanup.
- `operator_managed` is derived from `metadata.created_by == "quay-operator-readonly"` — no full metadata is exposed.
- `expiration_date` uses RFC 3339 UTC with explicit `Z` suffix for Go compatibility.

### 4. boot.py readonly DB-write guards (boot.py)

- Guards `sync_database_with_config()` and `set_region_release()` in `main()` when `REGISTRY_STATE == "readonly"`.
- Without this, both functions attempt DB writes (`ensure_image_locations()` does conditional INSERT, `set_region_release()` uses `get_or_create()`) that hit `ReadOnlyModeException` and crash the pod.
- `setup_bootstrap_token()` already had this guard.

### 5. boot.py file-based service key import (boot.py)

New `_import_service_key_from_files()` function for normal-mode boot when the operator has mounted a readonly Secret:

- Reads `kid` from `.kid` file, loads PEM, derives public JWK via authlib, verifies RFC 7638 thumbprint matches.
- **Direct lookup by `kid` before any path that triggers `_gc_expired()`** — prevents an expired operator-managed row from being physically deleted before import can refresh it.
- Validates existing rows: checks `service`, JWK `n`/`e`, and `metadata.created_by`.
- Backfills `created_by` on existing rows that lack it.
- Refreshes `expiration_date` on existing operator-managed rows.
- Auto-approves via `ServiceKeyApprovalType.AUTOMATIC`.
- Handles concurrent `IntegrityError` (pod race) and `ServiceKeyAlreadyApproved`.
- When import flag is `true` but files are missing: **fails boot** (no fallback to generating a replacement key).

### 6. boot.py readonly verification hardening (boot.py)

Rewritten `_verify_service_key()`:

- Now uses `approved_only=True`, `alive_only=True`, and passes `service` parameter (was `approved_only=False`, no service, bare `assert`).
- Loads PEM and verifies derived JWK `n`/`e` match DB JWK.
- Verifies computed RFC 7638 thumbprint matches `.kid` file.
- Returns `None` gracefully instead of raising `AssertionError` for missing files.

### 7. Cleanup CLI (tools/manage_servicekey.py — new file)

New `manage_servicekey.py` with `expire` subcommand for the operator cleanup Job:

```
manage_servicekey expire --kid <kid> --grace-seconds <seconds>
```

- **Direct DB lookup** bypassing stale-expired filtering (same pattern as `/status`).
- Missing key → exit 0 (idempotent).
- Wrong service → fail.
- `metadata.created_by != "quay-operator-readonly"` → fail (unconditional ownership check).
- Already expiring within grace window → exit 0 (idempotent).
- Sets `expiration_date = now + grace_seconds`.

### 8. Entrypoint mode (quay-entrypoint.sh)

- New `servicekey-expire` case matching the `migrate` mode pattern.
- Runs `certs_install.sh` and `client_certs.sh` before the CLI for DB TLS connectivity.
- Includes `shift` before `"$@"` — the existing entrypoint reads `$1` but never shifts, so without this the CLI would receive the mode name as its first argument.

### 9. Tests (test/test_readonly_lifecycle.py — new file)

Comprehensive test coverage for all changes:

- **Config:** flag in `INTERNAL_ONLY_PROPERTIES`, default is `False`.
- **Keyserver GET:** wrong service returns 404, correct service returns 200.
- **Keyserver /status:** 404 for unknown/wrong-service kid; correct fields returned; `operator_managed` true/false; 200 for expired keys (vs GET 403); 200 for unapproved keys (vs GET 409).
- **Model:** `get_service_key_for_status` bypasses stale filtering, filters by service.
- **Readonly guards:** readonly skips `sync_database_with_config` and `set_region_release`; normal mode calls both.
- **Import:** creates new key; refreshes existing; rejects wrong service, mismatched JWK, conflicting `created_by`; backfills metadata; flag-false does not trigger; flag-true with missing files fails.
- **Verify:** rejects unapproved, expired, wrong service, missing PEM; succeeds with valid key.
- **CLI:** missing key is noop; wrong service fails; non-operator key fails; operator key expires; already-expiring is idempotent.
- **Thumbprint:** PKCS#1 and PKCS#8 PEM produce identical thumbprints; deterministic for 2048-bit and 4096-bit keys; `as_dict()` returns only public fields.

## How it fits together

```
Operator creates Secret (.kid + .pem)
    ↓
Operator renders config with INSTANCE_SERVICE_KEY_IMPORT_FROM_FILES=true
    ↓
boot.py imports key from mounted files (PreparingKey phase)
    ↓
Operator verifies via GET /keys/.../status (operator_managed=true)
    ↓
Operator renders REGISTRY_STATE=readonly (EnteringReadOnly phase)
    ↓
boot.py verifies mounted key without DB writes (ReadOnly phase)
    ↓
Operator sets spec.readOnly=false (ExitingReadOnly phase)
    ↓
Cleanup Job runs: manage_servicekey expire --kid <kid> --grace-seconds 86400
```

## Test plan

- [ ] `TEST=true PYTHONPATH="." pytest test/test_readonly_lifecycle.py -v` — all new tests pass
- [ ] `TEST=true PYTHONPATH="." pytest workers/servicekeyworker/test/test_servicekeyworker.py -v` — existing service key worker test still passes (no regression)
- [ ] `TEST=true PYTHONPATH="." pytest test/test_api_usage.py -v` — existing API tests pass
- [ ] `bash -n quay-entrypoint.sh` — entrypoint syntax valid
- [ ] Verify `INSTANCE_SERVICE_KEY_IMPORT_FROM_FILES` appears in `INTERNAL_ONLY_PROPERTIES`
- [ ] Verify existing normal boot behavior is unchanged when import flag is `False`


boot.py
```diff
@@ -66,42 +226,55 @@ def _verify_service_key():
...
-    Creates a service key for quay.
+    Creates or imports a service key for quay.
     """
-    # Ensure we have an existing key if in read-only mode.
     if app.config.get("REGISTRY_STATE", "normal") == "readonly":
-        quay_key_id = _verify_service_key()
+        if app.config.get("INSTANCE_SERVICE_KEY_IMPORT_FROM_FILES", False):
+            quay_key_id = _verify_service_key()
+        else:
+            quay_key_id = _verify_legacy_service_key()
```
Marcusk19, line 242: nit: if this is the default I would refactor this to be clear that it's not a "legacy" service key per se

tools/manage_servicekey.py
```diff
@@ -0,0 +1,81 @@
...
+
+    metadata = key.metadata if isinstance(key.metadata, dict) else {}
+    if metadata.get("created_by") != OPERATOR_MANAGED_CREATED_BY:
+        print(
+            "Key '%s' is not operator-managed (created_by='%s'), refusing to expire."
+            % (kid, metadata.get("created_by"))
+        )
+        return 1
+
+    grace_deadline = datetime.utcnow() + timedelta(seconds=grace_seconds)
```
Marcusk19, line 47: nit: negative grace periods are excepted (eg. --grace-seconds -1) validate the argument is non-negative

tools/manage_servicekey.py
```diff
@@ -0,0 +1,81 @@
...
+    if key.expiration_date is not None and key.expiration_date <= grace_deadline:
+        print("Key '%s' already expiring within the requested grace window." % kid)
+        return 0
+
+    key.expiration_date = grace_deadline
+    key.save()
```
Marcusk19, line 53: nit: Cleanup CLI can overwrite concurrent fields

tools/manage_servicekey.py:52-53 uses key.save(), persisting the entire stale model instance. A concurrent update could
overwrite metadata, approval, or other columns. Prefer a partial update

Marcusk19 (approved):
left some nits inline, overall lgtm

PR #7040, opened 2026-09-02
NO-ISSUE: optimize: Minimize contention on database connection pool

ibazulic:
The logic introduced in the previous PR concerning pooled connections is sound but it has a critical problem: the retry/backoff loop ran inside `_connect()`, which Peewee invokes when holding the pool's lock. When a greenlet hit pool exhaustion and slept through the backoff, it held that lock for the whole duration, blocking other greenlets from acquiring or releasing connections during that process. Closing the connection via `close()` is the only way to return the connection to the pool, meaning that holding the lock during this time is genuinely harmful: the pool can never recover while the lock is held, which can cause cascading failures within a worker: one greenlet will block other greenlets until they start raising exceptions, degrading overall registry performance.

For this reason, the following changes are done:
1. the retry-and-backoff loop has been moved out of the lock-holding `_connect()` into the `connect()` wrapper; the lock is held only for a brief moment of acquiring a connection, but not while a greenlet is sleeping.
2. delay in the `_connect()` is completely removed, we immediately retry the connection instead of waiting 10 to 50 ms before the next attempt.
3. If the connection pool is exhausted, instead of raising a `500 Internal Server Error` (forcing the clients to back off immediately), a new handler is registered that returns a `503 Service Unavailable` with a `Retry-After` header. This signals clients that the issue is transient and they should retry the request after the alotted time.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/data/database.py
+++ b/data/database.py
@@ -503,26 +503,15 @@ class ObservablePooledDatabase(ObservableDatabase):
                 f"Unable to obtain healthy connection after {max_retries} attempts"
             )
 
-        try:
-            conn = super(ObservablePooledDatabase, self)._connect()
-        except MaxConnectionsExceeded:
-            # Pool exhausted - wait with exponential backoff before retrying
-            # Base delay: 10ms, max delay: 200ms to prevent long waits
-            delay = min(0.01 * (2**_retry_count), 0.2)
-            # Add jitter to prevent thundering herd (±25% randomization)
-            jitter = delay * (0.75 + uniform(0, 0.5))
-            logger.debug(
-                "Connection pool exhausted (attempt %d/%d), retrying after %.3fs",
-                _retry_count + 1,
-                max_retries,
-                jitter,
-            )
-            time.sleep(jitter)
-            return self._connect(_retry_count=_retry_count + 1)
+        # try and establish a connection
+        # if the call fails, the exception will propagate downwards, which is what we want
+        conn = super(ObservablePooledDatabase, self)._connect()
 
         try:
             with conn.cursor() as cursor:
                 cursor.execute("SELECT 1")
+                # roll back the request immediately so it doesn't end in IDLE_IN_TRANSACTION
+                conn.rollback()
             return conn  # Connection is healthy
         except Exception as e:
             # Catch ALL exceptions during liveness check - includes ProtocolViolation,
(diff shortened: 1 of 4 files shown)


endpoints/test/test_decorators.py
```diff
@@ -41,3 +43,20 @@ def test_require_xhr_from_browser(user_agent, include_header, expected_code, app
...
+    app.register_error_handler(MaxConnectionsExceeded, handle_max_connections_count)
+
+    with app.test_request_context("/v2/"):
+        try:
+            raise MaxConnectionsExceeded("pool full")
+        except Exception as e:
+            response = app.handle_user_exception(e)
+
+    assert response.status_code == 503
+    assert "Retry-After" in response.headers
```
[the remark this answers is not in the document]
ibazulic: This should be addressed in the new commit.

endpoints/decorated.py
```diff
@@ -75,3 +76,12 @@ def handle_not_implemented_error(ex):
 @app.errorhandler(RequestRedirect)
 def handle_bad_redirect(ex):
     return ex.get_response()
+
+
+@app.errorhandler(MaxConnectionsExceeded)
```
kleesc, line 81: If MaxConnectionsExceeded comes from a v2 endpoint, this will get raised, and change the format of the docker/oci spec error message.  Probably should add this in `endpoints/v2/__init__.py` too

data/database.py
```diff
@@ -503,26 +503,15 @@ def _connect(self, _retry_count=0):
...
-            return self._connect(_retry_count=_retry_count + 1)
+        # try and establish a connection
+        # if the call fails, the exception will propagate downwards, which is what we want
+        conn = super(ObservablePooledDatabase, self)._connect()
 
         try:
             with conn.cursor() as cursor:
                 cursor.execute("SELECT 1")
+                # roll back the request immediately so it doesn't end in IDLE_IN_TRANSACTION
+                conn.rollback()
```
kleesc, line 514: Can you move this back one indentation? This should be closed after the cursor closes

endpoints/decorated.py
```diff
@@ -75,3 +76,12 @@ def handle_not_implemented_error(ex):
...
+def handle_max_connections_count(ex):
+    logger.exception(ex)
+    response = jsonify({"message": "Service temporary unavailable due to high load. Please retry!"})
```
kleesc, line 84: I would just use Service temporary unavailable. The rest is implied from the 503

endpoints/decorated.py
```diff
@@ -75,3 +76,12 @@ def handle_not_implemented_error(ex):
...
+    response.status_code = 503
+    response.headers["Retry-After"] = 1
```
kleesc, line 86: Can you increase this a bit? Maybe 5s, since based on current attempts and jitter, it's already ~0.5s.

PR #7042, opened 2026-09-02
PROJQUAY-12486: feat(api): token exchange endpoint

Marcusk19:
## Summary

Adds the `/v1/bootstrap/exchange` OAuth 2.0 token-exchange endpoint for Kubernetes ServiceAccount workload identity. The endpoint validates the exchange request and ServiceAccount JWT, enforces configured identity and scope authorization, and issues a scoped bootstrap bearer token with structured errors. Playwright coverage is included for malformed, invalid, and unauthorized tokens.

### DIFF
--- a/endpoints/api/bootstrap.py
+++ b/endpoints/api/bootstrap.py
@@ -1,22 +1,32 @@
 import logging
 from datetime import UTC, datetime
+from urllib.parse import urlparse
 
 from flask import Request, request
 
 import features
-from app import app
+from app import app, model_cache
 from auth.auth_context import get_validated_oauth_token
+from auth.kubernetes_sa import (
+    KubernetesSATokenValidationError,
+    KubernetesSATokenValidator,
+)
 from data.database import OAuthAccessToken
 from data.model import db_transaction
 from data.model.oauth import (
-    create_bootstrap_oauth_api_token,
     delete_bootstrap_tokens,
     lock_bootstrap_token_operation,
     validate_bootstrap_token,
 )
 from endpoints.api import ApiResource, nickname, resource, show_if
 from endpoints.decorators import anon_allowed
-from endpoints.exception import InvalidToken, TokenRotationError, Unauthorized
+from endpoints.exception import (
+    ApiErrorType,
+    ApiException,
+    InvalidToken,
+    TokenRotationError,
+    Unauthorized,
+)
 from util.bootstrap_token import write_bootstrap_token
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


endpoints/api/bootstrap.py
```diff
@@ -49,6 +80,80 @@ def _is_local_bootstrap_renewal_request(req: Request) -> bool:
...
+        try:
+            validated = KubernetesSATokenValidator(
+                _exchange_config(), app.config["HTTPCLIENT"]
+            ).validate(raw)
+        except KubernetesSATokenValidationError:
+            _exchange_error(
+                "invalid_token", "Kubernetes ServiceAccount token failed validation", 401
+            )
+        issuer = validated.issuer
+        subject = validated.subject
```
ibazulic, line 132: Does this mean that anyone with a valid service account token (meaning valid issuer and subject) can get a bearer token from Quay? We have the `anon_allowed` decorator set. 

This seems a bit broad.

Marcusk19: It would need to be from an authorized subject in the quay config to obtain a token. The plumbing for that should be one more PR down the stack iirc

ibazulic: Sure, but authorized subject means anyone that creates a SA with that subject can mint  themselves a token, right? Maybe I'm overcautious, dunno.

Marcusk19: yeah that makes sense, but we're leaning on kubernetes RBAC here as well with putting some responsibility for security on the admin who specifies the allowed workload identities in the config. removing `anon_allowed` doesn't feel right to me here since the intended caller (k8s workload) wouldn't have a Quay bearer token to start with

endpoints/api/bootstrap.py
```diff
@@ -30,25 +43,143 @@ class BootstrapTokenCleanupError(Exception):
...
+            any(not values.get(key) for key in required)
+            or values.get("grant_type") != "urn:ietf:params:oauth:grant-type:token-exchange"
+            or values.get("subject_token_type") != "urn:ietf:params:oauth:token-type:jwt"
+        ):
+            _exchange_error("invalid_request", "invalid token exchange request", 400)
+        raw = values["subject_token"]
+        try:
+            validated = KubernetesSATokenValidator(
+                _exchange_config(), app.config["HTTPCLIENT"]
+            ).validate(raw)
```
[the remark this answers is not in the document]
Marcusk19: caching is done in #6986

endpoints/api/bootstrap.py
```diff
@@ -30,25 +43,143 @@ class BootstrapTokenCleanupError(Exception):
     pass
 
 
+class BootstrapExchangeError(ApiException):
+    def __init__(self, error, description, status_code):
+        super().__init__(
+            ApiErrorType.invalid_token if status_code == 401 else ApiErrorType.unauthorized,
```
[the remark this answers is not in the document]
Marcusk19: should be fixed, please verify

PR #7054, opened 2026-09-03
PROJQUAY-12887: fix: Decouple manifest storing in backend storage engine and make SQLite the single source of truth

ibazulic:
Previously, Distribution would write manifests to backend storage and would fall back to SQLite. This caused a split brain scenario with GC worker since GC uses data from the database as the single source of truth.

This PR does the following:
1. Adds new methods for manifest fetch that override Distribution functions and read manifests directly from the SQLite database
2. Completely removes storing of metadata in the backend storage; backend storage now only contains blob data
3. Changes the verification logic during push: no more orphaned data or placeholders (linkChild placeholder fabrication removed), manifest (be it image manifest or image index) cannot be committed unless all children and all blobs are stored in storage (follows standard Docker v2 API practice)
4. Preserves manifest children from being deleted by GC (creates temporary tags with 1 hour expiry on manifests that are pushed by digest only)
5. Removes the fallback functions that verify that manifests are stored in backend storage along with connected tests
6. Adds new tests covering new behaviour.

In addition to this, also fixes other boolean and type errors that were previously causing compile errors.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


internal/dal/metastore/metastore_sqlite.go
```diff
@@ -188,12 +212,81 @@ func (s *SQLiteStore) PutManifest(ctx context.Context, repoID int64, m oci.Manif
...
+				Name:            "$temp-" + uuid.NewString(),
+				RepositoryID:    repoID,
+				ManifestID:      sql.NullInt64{Int64: manifestID, Valid: true},
+				LifetimeStartMs: startMs,
+				LifetimeEndMs:   sql.NullInt64{Int64: expireMs, Valid: true},
+				TagKindID:       s.tagKindTag,
+				Hidden:          true,
+			}); err != nil {
+				return 0, fmt.Errorf("insert temporary protection tag: %w", err)
+			}
```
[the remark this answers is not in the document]
ibazulic: This should be addressed in the new commit.

internal/registry/distribution/middleware/manifest.go
```diff
@@ -129,3 +122,52 @@ func isIndexMediaType(mt string) bool {
...
+	// look up the manifest
+	content, mediaType, err := ms.repo.store.GetManifestForServing(ctx, repoID, digest)
+	if err != nil {
+		if errors.Is(err, oci.ErrNotExist) {
+			return nil, distribution.ErrManifestUnknownRevision{
+				Name:     ms.repo.Named().Name(),
+				Revision: digest,
+			}
+		}
+	}
```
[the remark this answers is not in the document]
ibazulic: This should be addressed in the new commit.

internal/gc/gc_test.go
```diff
@@ -118,14 +118,27 @@ func ensureRepo(t *testing.T, env *testEnv, ns, name string) int64 {
...
+			MediaType: "application/vnd.oci.image.manifest.v1+json",
+			Content:   []byte(`{"schemaVersion":2}`),
+			Tag:       tag,
+		})
+	} else {
+		id, err = env.store.PutManifest(t.Context(), repoID, oci.ManifestRecord{
+			Digest:    dgst,
+			MediaType: "application/vnd.oci.image.manifest.v1+json",
+			Content:   []byte(`{"schemaVersion":2}`),
+		})
```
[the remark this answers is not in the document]
ibazulic: This should be resolved in the new commit.

internal/gc/gc_test.go
```diff
@@ -170,14 +183,23 @@ func expireTag(t *testing.T, env *testEnv, repoID int64, name string, endMs int6
 	}
 }
 
+// expireTemporaryTag expires temporary tags by setting their expiration time to a specific value.
+func expireTemporaryTag(t *testing.T, env *testEnv, manifestID, pastMs int64) {
+	t.Helper()
+	query := `UPDATE tag SET lifetime_end_ms = ? WHERE manifest_id = ? AND hidden = 1 AND name LIKE '$temp-%'`
+	_, err := env.db.ExecContext(t.Context(), query, pastMs, manifestID)
```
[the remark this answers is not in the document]
ibazulic: This should be resolved in the new commit.

internal/gc/gc_test.go
```diff
@@ -419,13 +437,19 @@ func TestCollect_CascadeParentToChildren(t *testing.T) {
...
+	expireTemporaryTag(t, env, child1ID, pastMs)
+	expireTemporaryTag(t, env, child2ID, pastMs)
+
 	stats, err := env.collector.Collect(ctx)
 	if err != nil {
 		t.Fatal(err)
 	}
-	if stats.TagsExpired != 1 {
+
+	if stats.TagsExpired != 3 {
```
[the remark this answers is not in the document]
ibazulic: This should be resolved in the new commit.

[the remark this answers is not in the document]
ibazulic: Now it should be resolved in a new commit.

internal/gc/gc_test.go
```diff
@@ -335,6 +359,10 @@ func TestCollect_ManifestChildProtected(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// expire tag to explicity test that the tag is protected
```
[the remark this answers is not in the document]
ibazulic: Resolved in a new commit.

internal/gc/gc_test.go
```diff
@@ -335,6 +359,10 @@ func TestCollect_ManifestChildProtected(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// expire tag to explicitly test that the tag is protected
```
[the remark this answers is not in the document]
ibazulic: Should be resolved in new commit.

internal/gc/gc_test.go
```diff
@@ -863,14 +902,20 @@ func TestCollect_DeepCascade_ThreeLevels(t *testing.T) {
...
+	expireTemporaryTag(t, env, grandchildID, pastMs)
+
 	stats, err := env.collector.Collect(ctx)
 	if err != nil {
 		t.Fatal(err)
 	}
-	if stats.TagsExpired != 1 {
-		t.Fatalf("expected 1 expired tag, got %d", stats.TagsExpired)
+	if stats.TagsExpired != 3 {
+		t.Fatalf("expected 3 expired tag, got %d", stats.TagsExpired)
```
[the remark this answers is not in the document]
ibazulic: Should be resolved in new commit.

harishsurf:
This is the right design: SQLite is the catalog, no `{}` placeholders, index PUT fails if the child is not in the DB, digest-only PUTs get a 1h `$temp-`. That subsumes #7024's behavior and makes #7033's CAS materialization unnecessary.

Please treat [jbpratt's branch](https://github.com/jbpratt/quay/tree/fix/no-issue-omr-manifest-source-of-truth) (`e7a24a7e9` + `93537b979` + `dc60cce7a`) as the smaller shape of the same idea. A few gaps in this PR vs that:

1. `DistDriver` still `PutContent`s blob paths into CAS, and this PR deletes the SQLite fallback on blob GET/HEAD/Stat. Middleware skips inner `ManifestService.Put`, so the happy path may not write manifests to disk — but anything that still hits the driver will leak files, and `GET /v2/.../blobs/<manifest-digest>` 404s even when `manifest_bytes` exists. jbpratt's version no-ops blob-path `PutContent` and serves those reads from SQLite. That keeps Distribution's index validation (unknown child → 400) without replacing the whole `ManifestService`.

2. `insertTempTag` has the same now-vs-`endMs` skip as #7024. Re-push near expiry does not extend the window, and can insert a second `$temp-`. Extend the existing row.

3. Please drop unrelated churn from the landing commit: Go 1.26 pin, sqlc upgrade/revert, mediatype seed/schema reordering, authz bool drive-bys. Those make review and rebase harder than the actual catalog change.

4. `validateBlobs` / `GetBlobByChecksum`: `content_checksum` is not unique. If two `imagestorage` rows share a digest, this can attach the wrong id (`uploadedblob` vs a later insert). Reuse the existing `ensureBlob` / id-stability rule.

Happy to help fold #7024's tests and the inverted #7033 missing-child cases onto this branch once the driver contract matches "manifests are DB-only."

.github/workflows/ci-go.yaml
```diff
@@ -90,6 +90,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
+          go-version: "1.26"
```
jbpratt, line 93: Please remove this, this should be taken from the line below it using the `go.mod` file as the source of truth

internal/registry/distribution/middleware/manifest.go
```diff
@@ -16,28 +17,24 @@ import (
...
 	defer ms.repo.metrics.recordOp("manifest_put", time.Now(), &retErr)
 
-	dgst, err := ms.ManifestService.Put(ctx, manifest, options...)
+	mt, payload, err := manifest.Payload()
 	if err != nil {
-		return dgst, err
+		return "", logMetadataError("manifest_put", ms.repo.Named().Name(), "", err)
 	}
 
+	dgst := digest.FromBytes(payload)
```
jbpratt, line 31: At line 71 (the PutManifest call): validation failures from PutManifest ("blob unknown", "child manifest unknown") reach distribution's PutManifest handler as a MetadataWriteError, which maps to ErrorCodeUnknown -> HTTP 500. Before this change these cases returned 400 MANIFEST_BLOB_UNKNOWN via distribution's verifyManifest. Reproduced on e40b6d204: PUT of an index whose child row is absent -> 500; PUT of a manifest referencing never-uploaded blobs -> 500. Suggest a typed not-found error from the store, translated here into distribution.ErrManifestVerification{distribution.ErrManifestBlobUnknown{Digest: d}}, plus an HTTP-level test asserting 400.
AI-assisted review comment (Claude Code).

internal/registry/distribution/middleware/manifest.go
```diff
@@ -81,10 +78,6 @@ func (ms *manifestService) Put(ctx context.Context, manifest distribution.Manife
 func (ms *manifestService) Delete(ctx context.Context, dgst digest.Digest) (retErr error) {
 	defer ms.repo.metrics.recordOp("manifest_delete", time.Now(), &retErr)
 
-	if err := ms.ManifestService.Delete(ctx, dgst); err != nil {
-		return err
-	}
-
 	repoID, err := ms.repo.ensureRepo(ctx)
```
jbpratt, line 81: At line 86 (the DeleteManifest call): DeleteManifest returns nil when the row does not exist (metastore_sqlite.go:376), so DELETE of an unknown digest now answers 202 Accepted. The inner distribution Delete used to surface ErrBlobUnknown -> 404 MANIFEST_UNKNOWN. Reproduced on e40b6d204. Suggest returning distribution.ErrBlobUnknown when the store reports not-found, with a test.
AI-assisted review comment (Claude Code).

internal/registry/distribution/middleware/manifest.go
```diff
@@ -129,3 +122,53 @@ func isIndexMediaType(mt string) bool {
 	return mt == manifestlist.MediaTypeManifestList ||
 		mt == v1.MediaTypeImageIndex
 }
+
+// Exists checks if the manifest with the provided digest is stored in the database.
+func (ms *manifestService) Exists(ctx context.Context, dgst digest.Digest) (_ bool, retErr error) {
+	repoID, err := ms.repo.ensureRepo(ctx)
```
jbpratt, line 128: Exists (here) and Get (line 151) call ensureRepo, which runs EnsureUser + GetOrCreateRepository. A GET or HEAD by digest against a repository that does not exist now creates the namespace user and repository rows as a side effect of a read. The driver's read path uses GetRepositoryID (lookup only). Suggest the same here, mapping ErrNotExist to ErrManifestUnknownRevision.
AI-assisted review comment (Claude Code).

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -156,44 +176,150 @@ func (s *SQLiteStore) PutManifest(ctx context.Context, repoID int64, m oci.Manif
...
 	if err := tx.Commit(); err != nil {
 		return 0, fmt.Errorf("commit: %w", err)
 	}
 	return manifestID, nil
 }
 
+// validateBlobs validates that all blobs are committed to the database before the image manifest is stored.
+// This ensures consistency: manifests are stored only after all content for a particular image is stored,
+// with no orphaned blobs created.
+func (s *SQLiteStore) validateBlobs(ctx context.Context, q *daldb.Queries, m *oci.ManifestRecord) ([]int64, error) {
```
jbpratt, line 228: validateBlobs requires every References() entry to exist in imagestorage. Distribution's verifyManifest skipped the existence check for non-distributable / foreign layers that carry URLs (ocimanifesthandler.go, schema2manifesthandler.go). Windows images with foreign layers will now fail to push (and with the 500 above). Either skip descriptors that have URLs and a non-distributable media type, or document the intentional narrowing.
AI-assisted review comment (Claude Code).

jbpratt, line 228: The check is global (imagestorage by checksum), not scoped to the repository. Distribution's Stat went through the repository's linked blob store, so a manifest could only reference blobs already linked to (or mounted into) that repository. With this change a push to repo A can reference a blob only present in repo B; LinkManifestBlob then makes it pullable from A. Low impact for a single-admin mirror registry, but worth an explicit decision.
AI-assisted review comment (Claude Code).

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -156,44 +176,150 @@ func (s *SQLiteStore) PutManifest(ctx context.Context, repoID int64, m oci.Manif
...
+		}); err != nil {
+			return fmt.Errorf("link child manifest failed: %w", err)
+		}
+	}
+	return nil
+}
+
+// insertTempTag inserts a temporary tag for each manifest if that manifest is pushed by digest.
+func (s *SQLiteStore) insertTempTag(ctx context.Context, q *daldb.Queries, repoID, manifestID int64) error {
+	hasTemp, err := q.HasUnexpiredTemporaryTag(ctx, daldb.HasUnexpiredTemporaryTagParams{
```
jbpratt, line 294: HasUnexpiredTemporaryTag only matches "$temp-%" rows. Two consequences: a manifest that already has a live real tag gets an extra temp row on a by-digest re-push, and a re-push inside the window does not extend the existing temp tag's expiry (a manifest pushed at t=0 and re-pushed at t=59m still expires at t=60m). Python's create_temporary_tag_if_necessary checks any tag that protects at least as long and extends an existing temp tag. Suggest: SELECT EXISTS(... lifetime_end_ms IS NULL OR lifetime_end_ms >= ?), then UPDATE existing temp tag before INSERT.
AI-assisted review comment (Claude Code).

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -565,6 +659,24 @@ func (s *SQLiteStore) GetManifestDigest(ctx context.Context, repoID int64, dgst
...
+			return nil, "", fmt.Errorf("get manifest %s: %w", dgst, oci.ErrNotExist)
+		}
+		return nil, "", err
+	}
+
+	return []byte(row.ManifestBytes), row.MediaType, nil
+}
+
 // GetManifestContent returns manifest JSON stored in the database.
 func (s *SQLiteStore) GetManifestContent(ctx context.Context, dgst digest.Digest) ([]byte, error) {
```
jbpratt, line 681: GetManifestContent (also internal/oci/metadata.go:24) has no callers after the driver fallback was removed; consider dropping it from the interface and the stubs.
AI-assisted review comment (Claude Code).

internal/e2e/mirrorregistry/registry_test.go
```diff
@@ -233,6 +233,9 @@ func TestRegistryMultiArchGarbageCollectionCascade(t *testing.T) {
 	}
 
 	require.NoError(t, h.Registry().DeleteManifest(ctx, repository, indexResponse.Digest))
+
+	require.NoError(t, h.ExpireTemporaryTags(ctx))
+
 	stats, err := h.CollectGarbage(ctx)
 	require.NoError(t, err)
 	assert.Equal(t, 2, stats.ManifestsDeleted)
```
jbpratt, line 241: At line 251: the comment about revision links remaining readable is no longer true after this change, and the test does not assert the point of the PR: after GC the collected children should 404 from the registry, not only their blobs. Suggest assertRegistryMissing on each child digest after the second CollectGarbage, and asserting ManifestsDeleted == 0 on a cycle run while the temp tags are still live.
AI-assisted review comment (Claude Code).

internal/dal/schema/sqlite/seed_data.sql
```diff
@@ -212,10 +212,10 @@ INSERT INTO mediatype VALUES(9,'application/vnd.cnr.package.helm.v0.tar+gzip');
 INSERT INTO mediatype VALUES(10,'application/vnd.cnr.package.docker-compose.v0.tar+gzip');
 INSERT INTO mediatype VALUES(11,'application/vnd.cnr.manifests.v0.json');
 INSERT INTO mediatype VALUES(12,'application/vnd.cnr.manifest.list.v0.json');
-INSERT INTO mediatype VALUES(13,'application/vnd.docker.distribution.manifest.v1+json');
-INSERT INTO mediatype VALUES(14,'application/vnd.docker.distribution.manifest.v1+prettyjws');
-INSERT INTO mediatype VALUES(15,'application/vnd.docker.distribution.manifest.v2+json');
-INSERT INTO mediatype VALUES(16,'application/vnd.docker.distribution.manifest.list.v2+json');
+INSERT INTO mediatype VALUES(13,'application/vnd.docker.distribution.manifest.v1+prettyjws');
```
jbpratt, line 215: The regenerated seed swaps mediatype ids 13-16 (manifest.v1+json <-> v1+prettyjws, manifest.v2+json <-> manifest.list.v2+json) and quay_schema.sql reorders indexes/constraints and bumps blobupload.created DEFAULT (line 523). Fresh-DB only and fingerprint-checked, but it doubles the diff. If unintentional, regenerate from a clean state; if intentional, a note in the PR description would help.
AI-assisted review comment (Claude Code).

jbpratt (commented):
Review of e40b6d204. Two HIGH regressions against the pre-PR behavior: manifest validation failures (unknown blob or child) now return HTTP 500 instead of 400 MANIFEST_BLOB_UNKNOWN, and DELETE of an unknown manifest digest returns 202 instead of 404. Three MEDIUM items: GET/HEAD by digest on a nonexistent repository creates namespace and repository rows, foreign/non-distributable layers can no longer be pushed, and blob validation is no longer scoped to the repository. Remaining inline comments are LOW/nit (temp-tag renewal, dead GetManifestContent, e2e assertions, seed/schema churn, go-version pin). Both HIGH items were reproduced with ad-hoc pushes/deletes against the PR head.
AI-assisted review (Claude Code). The reproductions and test runs cited above were produced with the tool on the PR head e40b6d204 and checked before posting.

internal/dal/metastore/metastore_sqlite.go
ibazulic: Microsoft deprecated foreign layer support in 2022 and removed it in 2023. All Windows images now follow OCI standard completely so this issue is not relevant.

ibazulic:
@harishsurf I think I addressed all your issues. Here's the list:

1. DistDriver `PutContent` leaking manifest bytes into CAS: not sure if this is an issue. `PutContent` has a `default: return nil` branch. Manifest pushes do not fall under `/blobs/` or `/_uploads` paths, these are only for blob upload.
2. `insertTempTag` doesn't extend expiry on repush: Issue is fixed. A new query has been added that updates the table data with new expiry time the temporary tag already exists for a specific manifest. Also increased the expiry time from 1 hour to 6 hours by default to accomodate large multi-arch images (such as LLM models).
3. Unrelated churn (authz, go version pinning, schema reodering): all fixed. Now use go 1.26.0 and `sqlc` version 1.31.1 which does not create faulty bool types. Schema has been reset to what is in the `master` branch.

@jbpratt resolved issues:
1. HTTP 500 instead of 400: fixed. Two new errors added to the middleware, now we properly catch missing child manifests and missing blob manifests and wrap them in a proper HTTP 400 request.
2. `DELETE` unknown manifest retuns 202 instead of a 404: fixed. We return `oci.ErrNotExist` which middleware wraps into `ErrManifestUnknownRevision`.
3. `GET/HEAD`creates repos as a side-effect: fixed. `GET/HEAD` now invoke `lookupRepo` function which only checks for namespace/repo existence, instead of `ensureRepo` which creates them if they don't exist.
4. Foreign/non-distributable layers: this was deprecated by Microsoft in 2022. and removed in 2023. Now all Windows images follow OCI standard model. See [Microsoft's announcement](https://techcommunity.microsoft.com/blog/containers/announcing-removal-of-foreign-layers-from-windows-container-images/3846833) for details.
5. Blob validation not scoped to repository: fixed. Now `validateBlobs` checks both `ManifestBlob` and `UploadedBlob` tables to make sure that blobs which are part of the repo can be used or mounted.
6. `content_checksum` not unique: fixed. Added mutexes to repository, blob and manifest creations which should block creation of new rows until current database operation finishes. This eliminates duplicate entries. 
7. e2e regression test: fixed. Added an explicit GC run before expiry of tags is called to ensure that no deletions happen.

ibazulic:
Additional supporting evidence we don't leak any manifest bytes to storage:

~~~
# sqlite3 -readonly -json  quay.db "SELECT digest FROM manifest;" | jq '.[].digest' -r | cut -d':' -f2
0cc732607cf28ac1614c45afc8bb1258968f8ca4ce9aa0f4d25348eabeff3f23
10a700deef62eb9cd6ec24f536544a0366ea0df58c896caecbeab26d495059fe
11cad1844ac4d29f680eba2788d31faef83d7e8004cbca0e1f98d930a6264dd8
1263cca63f9f7e17b591c4a96a767139a8144f7101d180b9fc9e299ad0c117b2
30d2b893b4316d8fe1028e4cc2223833817fd3d4210cb0b6708da60577806055
8cc733917c2c3afcc06fef9e37e14cdf8152c27b9c7285ad920f1f6b86a118a5
91eed9d5c637c209d6a3eb5ddb56db40432aa9feb1bb83c46c587c89a52d9d8b
b0f379367120ae129811456d7b16bf76b041f4d1bd55510d409616d9f8ea526c
e261d1f43f20d23397063b348a2a91ab181693b019b5592dbb4ba8d443c49833
f30d896b3c88dfa59d8aa1600cc0a9dc49dad204281709229abffca5d08e3415
...
# sqlite3 -readonly -json  quay.db "SELECT digest FROM manifest;" | jq '.[].digest' -r | cut -d':' -f2 > manifest_digest.txt
# find storage/ -type f | grep -F -f manifest_digest.txt 
# echo $?
1
~~~

`grep` returns an exit status of `1` if nothing is matched. The only content that is stored in the filesystem are actual binary blobs.

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -156,44 +195,166 @@ func (s *SQLiteStore) PutManifest(ctx context.Context, repoID int64, m oci.Manif
...
+}
+
+// insertTempTag inserts a temporary tag for each manifest if that manifest is pushed by digest.
+func (s *SQLiteStore) insertTempTag(ctx context.Context, q *daldb.Queries, repoID, manifestID int64) error {
+	startMs := time.Now().UnixMilli()
+	// Python Quay sets this to 1 hour, but for LLM models and multiarch images this time might be too short
+	// so increase expiration to 6 hours to ensure big images are properly pushed without errors.
+	expireMs := time.Now().Add(6 * time.Hour).UnixMilli()
+
+	rows, err := q.ExtendTemporaryTag(ctx, daldb.ExtendTemporaryTagParams{
```
harishsurf, line 328: **Blocker.** `ExtendTemporaryTag` skips expired rows, so a digest re-push after expiry inserts a second `$temp-` for up to 2 weeks. It also doesn't check for an existing real or referrer tag. #7024's `protectWithTempTag` handles both (skip if any tag covers the requested end, else extend `lifetime_end_ms < ?`, else insert). Needs a re-push test; only creation is covered.
AI-assisted review comment (Claude Code).

internal/dal/metastore/metastore_sqlite.go
```diff
@@ -156,44 +195,166 @@ func (s *SQLiteStore) PutManifest(ctx context.Context, repoID int64, m oci.Manif
...
+	}
+	return nil
+}
+
+// insertTempTag inserts a temporary tag for each manifest if that manifest is pushed by digest.
+func (s *SQLiteStore) insertTempTag(ctx context.Context, q *daldb.Queries, repoID, manifestID int64) error {
+	startMs := time.Now().UnixMilli()
+	// Python Quay sets this to 1 hour, but for LLM models and multiarch images this time might be too short
+	// so increase expiration to 6 hours to ensure big images are properly pushed without errors.
+	expireMs := time.Now().Add(6 * time.Hour).UnixMilli()
```
harishsurf, line 326: 6h diverges from Python's 1h without a note, and line 230 still says "1 hour". It doesn't help slow pushes either: `uploadedblob` still expires at 1h and is purged globally, so layers go first. Keep 1h; handle the upload window separately.
AI-assisted review comment (Claude Code).

internal/dal/queries/storage.sql
```diff
@@ -44,3 +44,13 @@ WHERE s.content_checksum = ? AND (
   EXISTS (SELECT 1 FROM manifestblob mb WHERE mb.blob_id = s.id AND mb.repository_id = ?)
   OR EXISTS (SELECT 1 FROM uploadedblob ub WHERE ub.blob_id = s.id AND ub.repository_id = ? AND ub.expires_at > datetime('now'))
 );
+
+-- name: GetBlobByChecksumAndRepository :one
```
harishsurf, line 48: **Blocker.** `content_checksum` isn't unique, so `:one` returns an arbitrary duplicate. The mutex doesn't cover existing rows (migrated DBs). This also differs from `ensureBlob`, which `PutBlob` and the upload marker use, so `manifestblob` can link a different id than the marker protects and GC collects the wrong one. Use one shared lookup with a stable id.
AI-assisted review comment (Claude Code).

harishsurf (commented):
Two blockers inline: temp-tag duplicates without a re-push test, and the blob-id lookup on a non-unique column. One should-fix: the 6h temp-tag lifetime.
AI-assisted review (Claude Code).

internal/dal/queries/storage.sql
ibazulic: The amount of "duplicate" rows we have not important. Fetching of blob content checksums is gated behind the SHA digest (unique for same content) and `ManifestBlob` and that returns one row always. So regardless of which row is returned, the end result is correct. This stops future duplicate rows to appear. Removal of duplicate rows is not part of this PR, it should be resolved in a migration, not during pushes.

internal/dal/metastore/metastore_sqlite.go
ibazulic: I am unsure this makes sense. I need to investigate.

internal/dal/metastore/metastore_sqlite.go
ibazulic: Blob upload of 1 hour should be enough for most (although that could also be increased). Individual image upload rarely lasts for more than 1 hour. However, to get the OCI index or manifest list to be pushed, all children need to be preserved until the index is pushed. If each child requires 45 minutes to be pushed and you have a multiarch image with three different architectures, then 1st child image will expire much sooner than the index is pushed, which will fail the complete push. This is why it's increased to 6 hours. So we can guarantee that all children are present at the time when the index is committed. 

This should also be increased on the Python side.

harishsurf: on the flip side, if we want to keep the 6hrs, we should update: 
`internal/dal/queries/storage.sql:27
VALUES (?, ?, datetime('now'), datetime('now', '+1 hour'));`.  to 6hrs so that GC doesn't cleanup while layer 1 (after 1hr) while other layers are still being uploaded

internal/dal/metastore/metastore_sqlite.go
harishsurf: Looked again, downgrading this from blocker. The gap is real: `ExtendTemporaryTag` only touches rows with `lifetime_end_ms > now`, and GC keeps an expired tag for the namespace grace period (2 weeks by default), so a digest re-push in that window inserts a second `$temp-` row. But the newest row still protects the manifest and the stale one gets swept, so it's extra rows, not data loss. Same story for the missing "already has a real tag" check. Worth fixing while you're in there, plus a re-push test, but not something to hold the PR on.

internal/dal/queries/storage.sql
harishsurf: You're right, withdrawing this. I re-checked the GC path: a storage file is only removed once `CountBlobsByChecksum` hits zero, so as long as any row for that digest is referenced by a manifest or an unexpired marker the file stays. Linking row B while the upload marker sits on row A can't lose anything. The only leftover is that `validateBlobs` and `ensureBlob` use different lookups, which is untidy rather than unsafe. Fine to leave for a follow-up.

internal/dal/metastore/metastore_sqlite.go
ibazulic: @harishsurf I did do a check against the Python code and the solution proposed here follows the same code path. I'll see if I can make it better.

harishsurf: @ibazulic , this can be followed up as well, shouldn't be blocking this PR. Only the blob upload timeout (from the other comment) is what needs to be addressed

internal/dal/metastore/metastore_sqlite.go
ibazulic: @harishsurf this should now be addressed in the new commit.

internal/dal/metastore/metastore_sqlite.go
ibazulic: @harishsurf this should also be addressed in the new commit (decided to resolve it now so we don't have to do anything later).

harishsurf (approved):
LGTM, thanks for helping me out with this one :)

jbpratt (approved):

PR #7096, opened 2026-09-09
NO-ISSUE: fix(auth): downgrade basic auth parsing exception log level

lechuk47:
## Summary

- Downgrade `logger.exception` to `logger.debug` in `auth/basic.py` for basic auth header parsing failures
- Malformed `Authorization` headers from clients trigger `TypeError`, `UnicodeDecodeError`, or `ValueError` exceptions during base64 decoding — these are routine client errors, not server-side bugs
- `logger.exception` logs at ERROR level with a full traceback, which Sentry captures as reportable events, adding significant noise with no actionable value
- `logger.debug` preserves the diagnostic information for debug-level log consumers while keeping it out of Sentry

### DIFF
--- a/auth/basic.py
+++ b/auth/basic.py
@@ -55,7 +55,7 @@ def _parse_basic_auth_header(auth):
     try:
         credentials = [part.decode("utf-8") for part in b64decode(normalized[1]).split(b":", 1)]
     except (TypeError, UnicodeDecodeError, ValueError):
-        logger.exception("Exception when parsing basic auth header: %s", auth)
+        logger.debug("Exception when parsing basic auth header: %s", auth)
         return None, "Could not parse basic auth header"
 
     if len(credentials) != 2:


kleesc (approved):

ibazulic:
Exceptions can be used for investigating auth issues and identifying potential DDOS attacks where someone is brute forcing the registry. Downgrading that to mere debug log might cause downstream clients to lose that capability. 

The solution is not to downgrade exceptions but to set up Sentry to ignore these types of errors on Quay.io.

deshpandevlab (approved):

lechuk47:
> Exceptions can be used for investigating auth issues and identifying potential DDOS attacks where someone is brute forcing the registry. Downgrading that to mere debug log might cause downstream clients to lose that capability.
> 
> The solution is not to downgrade exceptions but to set up Sentry to ignore these types of errors on Quay.io.

Hey sorry, I missed your comment before merging. I think the way exceptions are logged with auth mechanisms / failures is not consistent. On the other hand, I don't think exception parsing in the logs is the right way of diagnosing DDOS attacks. If quay needs that feature, I would suggest adding metrics for that matter.

ibazulic:
Two things:

1. I'm talking about downstream customers who have alerts on exceptions. We're basically making all their alerts mute.
2. Turning auth failure into a `logger.debug` removes it from normal Quay logs except in a situation where `DEBUGLOG=true` is set. This complicates diagnosis for every downstream client. The auth failure exception is not logged for normal auth failures anyway, it is only logged in specific circumstances, like someone hammering a Docker v2 endpoint with improper authorization token. Quay.io has a higher degree of failures because it's a bigger registry.

PR #7101, opened 2026-09-10
NO-ISSUE: fix(auth): exclude basic auth parsing exception from Sentry

lechuk47:
## Summary

- Add `"exception when parsing basic auth header"` to Sentry `EXCLUDE_PATTERNS` so these events are dropped in `before_send` instead of being reported to Sentry
- The `logger.exception()` call in `auth/basic.py` remains unchanged — customers who parse logs for DDOS detection are unaffected
- This supersedes the approach in PR #7096, which downgraded the log level from `exception` to `error`.


Sunandadadi (approved):
lgtm

PR #7110, opened 2026-09-11
PROJQUAY-12562: feat(ci): add openshift ci-operator config generator

nasonawa:
A Python-based config generator that reads a compact `matrix.yaml` and produces ci-operator config files for each declared test cell. This is the core tooling that makes the CI matrix maintainable at scale.

- CLI Interface
```bash
python3 _generator/generate.py              # generate all configs
python3 _generator/generate.py --dry-run     # preview without writing
python3 _generator/generate.py --check       # verify configs are up-to-date
python3 _generator/generate.py --list        # show matrix expansion table
```


ci-generator/generate.py
```diff
@@ -0,0 +1,444 @@
...
+    if not isinstance(tests, list) or not tests or not isinstance(tests[0], dict):
+        raise ValueError("generated config is missing tests[0]")
+    test = tests[0]
+    steps = test.setdefault("steps", {})
+    if not isinstance(steps, dict):
+        raise ValueError("tests[0].steps must be a mapping")
+    if cell.tier == "presubmit":
+        test["always_run"] = True
+        steps.pop("cluster_profile", None)
+        steps["cluster_claim"] = {
```
jbpratt: `cluster_claim` belongs on the test entry, not inside `steps`. In ci-operator, `steps` is a
MultiStageTestConfiguration and has no `cluster_claim` field. Evidence from
openshift/release @ ced218c7905d: 542 config files match `^  cluster_claim:` and 0 match
`^    cluster_claim:`; `ci-operator/config/quay/quay/quay-quay-master__claim.yaml` puts it
at test level alongside `as` and `cron`. Suggestion: set
`test["cluster_claim"] = {...}` instead of `steps["cluster_claim"] = {...}`.
AI-assisted review comment (Claude Code).

ci-generator/generate.py
```diff
@@ -0,0 +1,444 @@
...
+def apply_tier_mapping(config: YamlMap, cell: Cell) -> YamlMap:
+    config = copy.deepcopy(config)
+    tests = config.get("tests")
+    if not isinstance(tests, list) or not tests or not isinstance(tests[0], dict):
+        raise ValueError("generated config is missing tests[0]")
+    test = tests[0]
+    steps = test.setdefault("steps", {})
+    if not isinstance(steps, dict):
+        raise ValueError("tests[0].steps must be a mapping")
+    if cell.tier == "presubmit":
```
jbpratt, line 265: The presubmit branch removes `cluster_profile` but leaves the cloud layer's
`workflow: ipi-aws` and `post: [quay-deprovision, chain: ipi-aws-post]` in place. A claimed
cluster must not run an IPI install workflow, and an `ipi-*` workflow without a
`cluster_profile` fails validation. `quay-quay-master__claim.yaml` uses
`workflow: generic-claim` with no post chain. Suggestion: have the presubmit mapping also
replace `workflow` and clear `post`, or move claim-vs-profile into a separate cloud/test
layer rather than patching it after the merge.
AI-assisted review comment (Claude Code).

ci-generator/generate.py
```diff
@@ -0,0 +1,444 @@
...
+    for _cell, filename, config in results:
+        path = output_dir / filename
+        if not path.exists():
+            print(f"missing {path}", file=sys.stderr)
+            failures += 1
+            continue
+        if path.read_text() != dump_config(config):
+            print(f"stale {path}", file=sys.stderr)
+            failures += 1
+    if output_dir.is_dir():
```
jbpratt, line 442: This loop reports any pre-existing `*.yaml` in the output directory as `unexpected`. Per
README, in-tree mode points the output directory at `ci-operator/config/quay/quay/`, which
already holds `quay-quay-master.yaml`, `quay-quay-master__claim.yaml`,
`quay-quay-master__omr-v2-upgrade-v3.yaml` and `quay-quay-master__omr-v3.yaml`. Reproduced
against a copy of that directory with all seven generated files byte-identical:

    unexpected .../quay-quay-master.yaml
    unexpected .../quay-quay-master__claim.yaml
    unexpected .../quay-quay-master__omr-v2-upgrade-v3.yaml
    unexpected .../quay-quay-master__omr-v3.yaml
    4 config(s) out of date

The documented CI gate cannot pass. Suggestion: restrict the orphan scan to filenames the
generator owns (a prefix derived from the matrix, or a generated-by marker in the file).
AI-assisted review comment (Claude Code).

ci-generator/generate.py
```diff
@@ -0,0 +1,444 @@
...
+def _check_configs(results: list[tuple[Cell, str, YamlMap]], output_dir: Path) -> int:
+    failures = 0
+    expected = {filename for _cell, filename, _config in results}
+    for _cell, filename, config in results:
+        path = output_dir / filename
+        if not path.exists():
+            print(f"missing {path}", file=sys.stderr)
+            failures += 1
+            continue
+        if path.read_text() != dump_config(config):
```
jbpratt: `--check` compares serialized text, and the serialization will not match what
openshift/release stores. `make ci-operator-config` there runs
`determinize-ci-operator --config-dir /ci-operator/config --confirm`, which rewrites every
config canonically: alphabetically ordered keys, `'*'` and `'@daily'` single-quoted, long
strings single-quote-folded. `dump_config` uses `sort_keys=False`, so it emits merge order
(`cron` after `steps`, `workflow` before `test`) and double-quote style. After the first
`make update` in the release repo every file reads as stale. Suggestion: either compare
parsed YAML here, or make the dumper match determinize's canonical form.
AI-assisted review comment (Claude Code).

ci-generator/templates/base.yaml
```diff
@@ -0,0 +1,36 @@
+build_root:
```
jbpratt, line 1: Both live redhat-3.18 configs in openshift/release carry a top-level

    prowgen:
      enable_secrets_store_csi_driver: true

which no template emits, so regenerating drops it. Suggestion: add it to `base.yaml`.
AI-assisted review comment (Claude Code).

ci-generator/overrides/phase0-aws-ocp422-e2e.yaml
```diff
@@ -0,0 +1,10 @@
+match:
+  quay: "3.18"
+  ocp: "4.22"
+  cloud: aws
```
jbpratt: `QUAY_BUILDER_IMAGE` is scoped to `cloud: aws`, but
`ci-operator/config/quay/quay/quay-quay-redhat-3.18__gcp-ocp422-e2e-install.yaml` carries
the same pinned digest today. The generated gcp and azure cells keep
`ENABLE_BUILD_SUPPORT: "true"` and the `quay-provisioning-builder` ref while losing the pin.
Suggestion: move `QUAY_BUILDER_IMAGE` into `templates/tests/e2e-install.yaml` next to
`ENABLE_BUILD_SUPPORT`, so the pin travels with build support.
AI-assisted review comment (Claude Code).

ci-generator/templates/tests/e2e-install.yaml
```diff
@@ -0,0 +1,35 @@
...
+        MAIL_USE_TLS: false
+        MAIL_USE_AUTH: false
+        MAIL_DEFAULT_SENDER: quay@quayqe.local
+        DEFAULT_UI: react
+        SUPER_USERS:
+          - admin
+        GLOBAL_READONLY_SUPER_USERS:
+          - readonly
+    test:
+    - ref: quay-enable-catalogsource
```
jbpratt, line 36: The live aws config also runs `- ref: quay-deploy-jaeger` between `quay-deploy-mailpit` and
the storage ref, `- ref: quay-gather-jaeger-traces` first in `post`, and carries
`FEATURE_OTEL_TRACING` plus an `OTEL_CONFIG` block in `QUAY_EXTRA_CONFIG`. None of it is in
the templates or the committed fixture, so regenerating removes tracing from a job that has
it today. Worth noting this is also the case the merge strategy handles worst: because
lists of mappings merge by index, inserting `quay-deploy-jaeger` at position 5 via an
override means restating the whole `test` list.
AI-assisted review comment (Claude Code).

ci-generator/overrides/phase0-gcp-redhat-3.17-e2e.yaml
```diff
@@ -0,0 +1,31 @@
+match:
+  quay: "3.17"
+  ocp: "4.22"
+  cloud: gcp
+  test: e2e-install
+tests:
+- as: gcp-gcs-3-16-daily-4-22
```
jbpratt: `as: gcp-gcs-3-16-daily-4-22` on a `quay: "3.17"` override. Same in
`phase0-azure-redhat-3.17-e2e.yaml:7`. `generate.py --list` shows the result: 3.17 gcp and
3.16 gcp both emit `gcp-gcs-3-16-daily-4-22`, 3.17 azure and 3.16 azure both emit
`azure-blob-3-16-daily-4-22`. The filenames differ so generation succeeds, but two Quay
releases produce identically named tests. Suggestion: drop `as` from these overrides
entirely - the derived `test_as` already produces the correct per-version name.
AI-assisted review comment (Claude Code).

ci-generator/matrix.yaml
```diff
@@ -0,0 +1,26 @@
...
+    jobs:
+      - tier: daily
+        clouds: [aws,gcp,azure]
+        ocp: ["4.22"]
+        test: e2e-install
+  - branch: redhat-3.17
+    jobs:
+      - tier: daily
+        clouds: [gcp,azure]
+        ocp: ["4.22"]
```
jbpratt: redhat-3.18 gcp is declared `tier: daily`, producing `cron: "@daily"` and
`as: gcp-gcs-3-18-daily-4-22`. The live config is `cron: '@weekly'` with
`as: gcp-gcs-3-18-nightly-4-22`. Adopting the generator therefore multiplies that job's
frequency by seven on a paid cluster profile and renames the Prow job, resetting its
history and any alerting keyed to the name. If that is intended it is worth stating in the
PR description; if not, `tier: weekly` reproduces today's behaviour.
AI-assisted review comment (Claude Code).

ci-generator/tests/test_generate.py
```diff
@@ -0,0 +1,211 @@
...
+    by_name = {filename: config for _cell, filename, config in results}
+    assert PHASE0_NAME in by_name
+    expected = yaml.safe_load(FIXTURE.read_text())
+    assert by_name[PHASE0_NAME] == expected
+
+
+def test_list_shows_phase0_row(capsys: object) -> None:
+    assert main(["--list"]) == 0
+    out = capsys.readouterr().out  # type: ignore[attr-defined]
+    assert "3.18" in out
```
jbpratt, line 164: This compares `yaml.safe_load(FIXTURE)` against the in-memory dict, so serialization is not
covered - which is what `--check` actually compares. The committed fixture is textually
different from `dump_config` output (`'*'` vs `"*"`, `cron` before `steps` rather than
after, env key order, `QUAY_BUILDER_IMAGE` position), so it cannot have been produced by
running the generator. Suggestion: regenerate the fixture from `dump_config` and assert
`dump_config(config) == FIXTURE.read_text()`; that also turns the quoting and key-order
drift into a visible test failure.
AI-assisted review comment (Claude Code).

ci-generator/README.md
```diff
@@ -0,0 +1,130 @@
...
+```bash
+python3 generate.py              # generate all configs
+python3 generate.py --dry-run    # preview without writing
+python3 generate.py --check      # verify configs are up to date
+python3 generate.py --list       # show matrix expansion table
+python3 generate.py --output DIR # write or check a specific directory
+```
+
+Standalone default output is `./out`. When this directory is named `_generator`, output is the parent directory (`ci-operator/config/quay/quay/`).
+
```
jbpratt, line 28: `--check` is documented as the CI verification step, but the PR adds no workflow under
`.github/`, so neither `--check` nor the 26 tests run anywhere. The generator can drift from
its fixture and from the matrix without CI noticing. Suggestion: add a small workflow that
runs `pytest` and `generate.py --check` on changes under `ci-generator/`.
AI-assisted review comment (Claude Code).

ci-generator/generate.py
```diff
@@ -0,0 +1,444 @@
...
+    for key, expected in match.items():
+        if key not in MATCH_FIELDS:
+            raise ValueError(f"unknown override match key {key!r}")
+        actual = context[MATCH_FIELDS[key]]
+        if str(actual) != str(expected):
+            return False
+    return True
+
+
+def apply_overrides(config: YamlMap, cell: Cell, overrides: list[Override]) -> YamlMap:
```
jbpratt: Three override outcomes are silent here. Two overrides with the same number of match keys
that both match a cell are both applied, with `path.name` deciding the winner - a rename
changes the output. An override whose `match` hits no cell is ignored with no message. And
an override can match more cells than the filename implies: `phase0-aws-ocp422-e2e.yaml`
has no `tier` key, so it would also rename a presubmit cell to `...-nightly-...`.
Suggestion: track which overrides applied and fail (or warn) on unused overrides and on
equal-specificity conflicts touching the same leaf.
AI-assisted review comment (Claude Code).

ci-generator/pyproject.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[project.optional-dependencies]
+test = [
+    "pytest>=7.0.0",
+]
+
+[tool.uv]
+package = false
+
+[tool.black]
```
jbpratt: `line-length = 120` here conflicts with the repo root's `line-length = 100`, and the
pre-commit isort hook is pinned to `--settings-path pyproject.toml` (repo root), so the
local `[tool.isort]` is dead config. black picks its configuration from the common ancestor
of the files it is given, so these files format at 120 when staged alone and at 100 when
staged alongside any other Python file. `black --check --line-length 100 generate.py
model.py` reformats `deep_merge`'s condition and two `raise` statements. Suggestion: drop
`[tool.black]`, `[tool.isort]` and `[tool.ruff]` here and inherit the repo root.
AI-assisted review comment (Claude Code).

ci-generator/pyproject.toml
```diff
@@ -0,0 +1,37 @@
...
+version = "0.1.0"
+description = "Generate ci-operator configs from a compact Quay CI matrix"
+readme = "README.md"
+requires-python = ">3.10.0"
+dependencies = [
+    "jinja2",
+    "pyyaml",
+]
+
+[project.optional-dependencies]
```
jbpratt, line 12: pytest is declared twice with disagreeing floors: `pytest>=7.0.0` in
`[project.optional-dependencies].test` and `pytest>=9.1.1` in `[dependency-groups].dev`.
The README documents `uv sync --extra test`. Suggestion: keep one declaration.
AI-assisted review comment (Claude Code).

ci-generator/pyproject.toml
```diff
@@ -0,0 +1,37 @@
+[project]
+name = "ci-generator"
+version = "0.1.0"
+description = "Generate ci-operator configs from a compact Quay CI matrix"
+readme = "README.md"
+requires-python = ">3.10.0"
```
jbpratt: `requires-python = ">3.10.0"` is an exclusive bound on a three-component version, so it
admits 3.10.1. `>=3.11` is the likely intent, and `uv.lock` inherits the same string.
AI-assisted review comment (Claude Code).

ci-generator/README.md
jbpratt: The generator lives in quay/quay while its output belongs in openshift/release, and nothing
in the PR establishes how it gets there - the README documents only a `_generator` directory
name convention. Worth stating the intended mechanism (vendored copy, sync job, or moving
the tool into the release repo), since items 3 and 4 above both turn on where `--check`
actually runs.
AI-assisted review comment (Claude Code).

ci-generator/generate.py
```diff
@@ -0,0 +1,444 @@
...
+    if not overrides_dir.is_dir():
+        return []
+    loaded: list[Override] = []
+    for path in sorted(overrides_dir.glob("*.yaml")):
+        data = load_yaml(path)
+        if data is None:
+            continue
+        if not isinstance(data, dict):
+            raise ValueError(f"{path} must be a YAML mapping")
+        match = match = data.get("match", {})
```
jbpratt: `match = match = data.get("match", {})` - duplicate assignment.
AI-assisted review comment (Claude Code).

ci-generator/generate.py
```diff
@@ -0,0 +1,444 @@
...
+    "ocp": "ocp_version",
+    "cloud": "cloud",
+    "test": "test",
+    "tier": "tier",
+    "branch": "branch",
+    "arch": "arch",
+}
+
+
+def default_output_dir(generator_dir: Path = GENERATOR_DIR) -> Path:
```
jbpratt, line 65: Behaviour switching on the directory being named `_generator` is invisible from the
checkout this ships in; a short comment naming the destination path would help.
AI-assisted review comment (Claude Code).

ci-generator/model.py
```diff
@@ -0,0 +1,119 @@
...
+    org: str
+    repo: str
+    branch: str
+    quay_version: str
+    ocp_version: str
+    cloud: str
+    test: str
+    tier: str
+    arch: str = "amd64"
+    image_source: str = "build"
```
jbpratt, line 41: `image_source` is carried through `Cell` and `context()` but no template references it.
AI-assisted review comment (Claude Code).

ci-generator/README.md
```diff
@@ -0,0 +1,130 @@
...
+Each override file has a `match:` block. Remaining keys are the patch. The generator sorts by specificity (fewer match keys first) so last-write-wins on deep merge. Known match keys: `quay`, `ocp`, `cloud`, `test`, `tier`, `branch`, `arch`.
+
+## Adding coverage
+
+Add a Quay release by appending to `quay:`:
+
+```yaml
+  - branch: redhat-3.17
+    jobs:
```
jbpratt, line 95: The table describes `nightly` as an alias for `daily`, but the only in-repo use is the aws
override renaming a `daily` job to `...-nightly-...`, which reads as the inverse. Either
drop the alias or say it exists to preserve legacy job names.
AI-assisted review comment (Claude Code).

ci-generator/overrides/phase0-gcp-redhat-3.16-e2e.yaml
```diff
@@ -0,0 +1,29 @@
+match:
+  quay: "3.16"
+  ocp: "4.22"
+  cloud: gcp
+  test: e2e-install
+tests:
```
jbpratt: `as:` here restates the value `test_as` already derives, and `QUAY_EXTRA_CONFIG` is
duplicated verbatim across all four 3.16/3.17 override files. A shared 3.16/3.17 override
matched on `quay` alone, plus per-cloud files for the genuine differences, would remove
both repetitions.
AI-assisted review comment (Claude Code).

ci-generator/generate.py
```diff
@@ -0,0 +1,444 @@
+#!/usr/bin/env python3
+"""
```
jbpratt: The emoji STOP banner is effective as a contributor sign but sits in the module docstring,
so it lands in `help()` and any generated docs. Consider a top-of-file comment block.
AI-assisted review comment (Claude Code).

jbpratt (changes requested):
The generator, the layering model and the test suite all read clearly, and the STOP banner
is the right instinct for a file people will be tempted to hardcode into. The blocking
issues below are about what the generated configs would do once they land in
openshift/release, not about the Python.

Reproduced locally against pr-7110 @ fbe58d93 and openshift/release @ ced218c7905d
(read-only; nothing was checked out in a shared tree). `pytest` passes 26/26 and
`generate.py --list/--check/--output` all behave as documented in isolation.

Nine blocking items:

1. `cluster_claim` is emitted inside `tests[0].steps`. In ci-operator it is a field of the
   test entry. In openshift/release today, 542 config files put it at test level and none
   nest it under `steps`, including this repo's own `quay-quay-master__claim.yaml`. Any
   presubmit cell produces a config that will not validate.
2. The presubmit path drops `cluster_profile` but keeps the cloud template's
   `workflow: ipi-aws` and `ipi-aws-post` chain. A claimed cluster must not run an IPI
   install workflow; `quay-quay-master__claim.yaml` uses `workflow: generic-claim`.
3. `--check` flags every pre-existing `*.yaml` in the output directory as `unexpected`.
   In the documented in-tree mode the output directory is `ci-operator/config/quay/quay/`,
   which already contains four `quay-quay-master*` configs. With all seven generated files
   byte-identical, `--check` still exits 1 with four `unexpected` lines.
4. openshift/release rewrites configs canonically via
   `determinize-ci-operator --confirm` (`make ci-operator-config`). The generator's dumper
   emits insertion order and different quoting, so the first `make update` in the release
   repo marks every file stale and `--check` can never pass again.
5. `prowgen: {enable_secrets_store_csi_driver: true}` is absent from the templates; both
   live redhat-3.18 configs carry it.
6. `QUAY_BUILDER_IMAGE` is scoped to `cloud: aws`, but the live gcp config carries the same
   pinned digest, and the generated gcp/azure cells still set `ENABLE_BUILD_SUPPORT: "true"`.
7. `FEATURE_OTEL_TRACING`/`OTEL_CONFIG`, `quay-deploy-jaeger` and `quay-gather-jaeger-traces`
   are live in the aws config today and absent from both the templates and the fixture.
8. The two redhat-3.17 overrides hardcode `3-16` in `as:`, so 3.17 and 3.16 emit identical
   test names on gcp and azure.
9. redhat-3.18 gcp moves from `@weekly` to `@daily` and is renamed `nightly` -> `daily`.
   That is a 7x frequency change on a paid cluster profile plus a Prow job rename, arriving
   as a side effect rather than as a stated decision.

The golden fixture compares parsed YAML, not bytes, which is why items 4 and 7 are not
caught by the suite: the committed fixture does not match what `dump_config` actually
writes. Adding a byte-level golden would catch both.

Full findings, including seven should-fix items and six nits, are in the inline comments.

AI-assisted review (Claude Code). The reproduction and test runs cited above were produced
with the tool and checked before posting.

jbpratt:
**Design note on `overrides/`: keep the grouped matrix, replace free-form override files with exact `cell_settings` plus named `env_profiles`.**

The five override files each match one cell today, but a `match:` block is a selector, not an identity: none constrains `tier` or `arch`, so a future presubmit or arm64 cell inherits the patch silently. Two overrides of equal specificity are ordered by filename (`generate.py:280`), a rule the README does not state, and an override that matches nothing is not reported. Patches target `tests[0]` by position, so `as:` in an override reads like a name selector but is not one, and inserting a step (the jaeger case) means restating the whole list. Four of the five files also carry the same `QUAY_EXTRA_CONFIG` body, differing only by two 3.17 flags.

Proposed shape, same seven cells and output as the PR, showing the Azure 3.16 exception:

```yaml
version: 2
global_defaults: {image_source: build, arch: amd64, repo: quay/quay}
quay:
  - branch: redhat-3.18
    jobs: [{tier: daily, clouds: [aws, gcp, azure], ocp: ["4.22"], test: e2e-install}]
  - branch: redhat-3.17
    jobs: [{tier: daily, clouds: [gcp, azure], ocp: ["4.22"], test: e2e-install}]
  - branch: redhat-3.16
    jobs: [{tier: daily, clouds: [gcp, azure], ocp: ["4.22"], test: e2e-install}]
env_profiles:
  quay-3.16-e2e:
    PLAYWRIGHT_GREP_INVERT: "@auth:OIDC|@auth:LDAP"
    QUAY_EXTRA_CONFIG: |
      # exact body from overrides/phase0-azure-redhat-3.16-e2e.yaml, unchanged
cell_settings:
  - cell: {branch: redhat-3.16, cloud: azure, ocp: "4.22", test: e2e-install, tier: daily, arch: amd64}
    env_profile: quay-3.16-e2e
    as: azure-blob-3-16-daily-4-22
```

Application order per cell: template and tier output, then the profile's env, then the entry's local `env`, with local keys replacing profile keys. `QUAY_EXTRA_CONFIG` stays one opaque string replaced whole, so an older profile cannot inherit a newer flag by accident. A profile may only supply `env`; it cannot rename a test or touch `steps`. Migrating the full PR is two profiles (`quay-3.16-e2e`, `quay-3.17-e2e`) referenced by four cells, plus one aws entry with a local `env: {QUAY_BUILDER_IMAGE: ...}`. `expand_cells`, the templates and the dumper are untouched; only override discovery and application change, and the golden output should be byte-identical before and after.

Validations this makes cheap, none of which the current merge can express:

- `env_profile` naming a profile that is not declared -> error.
- Two `cell_settings` entries with the same six-field identity -> error.
- A `cell` that does not match any expanded cell -> error, replacing today's silent no-op.
- A generated config with other than exactly one test when settings are applied -> error, instead of patching index 0.
- No filename precedence: each cell names at most one profile, so there is nothing to order.
- No positional list patching: settings apply to named env keys and `as` only.

Tradeoff: an editor makes one extra lookup from cell to profile, and a shared-profile edit fans out to every referencing cell, which `--list` should show. In exchange, coverage stays compact, exceptions are visible opt-in rather than inherited by selector, and the four duplicated environment bodies collapse to two. This is a representation change, separate from the blocking findings above; both can land as one PR or the override migration can follow.

AI-assisted review comment (Claude Code).

jbpratt:
This supersedes the [earlier override-shape proposal](https://github.com/quay/quay/pull/7110#issuecomment-5631996734).

Prefer branch-level `env` with job-local `env` and `as`. The shared 3.16/3.17 environments then live beside their coverage, with no separate cell identity or profile lookup:

```yaml
version: 2 # proposed syntax
global_defaults: {image_source: build, arch: amd64, repo: quay/quay}
quay:
  - branch: redhat-3.18
    jobs: [{tier: daily, clouds: [aws, gcp, azure], ocp: ["4.22"], test: e2e-install}]
  - branch: redhat-3.17
    env:
      PLAYWRIGHT_GREP_INVERT: "@auth:OIDC|@auth:LDAP"
      QUAY_EXTRA_CONFIG: |
        # exact 3.17 override body elided; shared by GCP and Azure
    jobs: [{tier: daily, clouds: [gcp, azure], ocp: ["4.22"], test: e2e-install}]
  - branch: redhat-3.16
    env:
      PLAYWRIGHT_GREP_INVERT: "@auth:OIDC|@auth:LDAP"
      QUAY_EXTRA_CONFIG: |
        # exact 3.16 override body elided; shared by GCP and Azure
    jobs:
      - {tier: daily, clouds: [gcp], ocp: ["4.22"], test: e2e-install}
      - tier: daily
        clouds: [azure]
        ocp: ["4.22"]
        test: e2e-install
        as: blob-daily
        # Optional env here replaces branch env keys for this job only.
```

This excerpt shows shared release env and the Azure 3.16 name. Retain other existing names and AWS's builder pin on their respective split job rows. Apply template/tier → branch → job, replacing env values whole; require one generated test. New test jobs inherit branch defaults deliberately. Migration still needs an output comparison.

AI-assisted review comment (Claude Code).

jbpratt:
The generated prefix already identifies Quay version, cloud, OCP version, and test. Keeping the current variant and filename while setting `test_as` to `f"{self.storage}-{self.tier}"` removes the repeated dimensions. All five override files also pin `as`; removing those keys and updating naming expectations makes the change effective. The existing 3.18 AWS job would be renamed, requiring coordinated history and alert updates. AI-assisted review comment (Claude Code).

jbpratt:
Also, please emit a large `# DO NOT EDIT ... <where to edit>` at the top of each generated file so they do not drift in openshift/release

nasonawa:
Thanks for the review — the blocking items and the later branch/job `env` design are in the next push.

**Landed**
- Presubmit: `cluster_claim` is on the test entry; `workflow: generic-claim`; `cluster_profile` and `post` are dropped.
- `--check` compares parsed YAML (so `determinize-ci-operator` quoting/key-order does not fail the gate) and only flags unexpected files under `{org}-{repo}-{branch}__` prefixes from the current matrix. Hand-written `quay-quay-master*` configs are ignored.
- Generated files get a `# DO NOT EDIT` header pointing at `matrix.yaml` / `templates/` / `generate.py`.
- `prowgen.enable_secrets_store_csi_driver`, `QUAY_BUILDER_IMAGE`, jaeger refs, and OTEL extra-config are in the shared templates.
- `overrides/` is gone. Shared 3.16/3.17 env lives on the branch in `matrix.yaml` v2; job-level `env` / `as` overlay that. Env values replace whole keys. Exactly one generated test is required.
- Live names/schedules kept: 3.18 AWS stays `aws-s3-3-18-nightly-4-22` `@daily`; 3.18 GCP stays `gcp-gcs-3-18-nightly-4-22` `@weekly`. 3.17 `as:` no longer copies `3-16`.
- Byte-level golden of `dump_config` output, generator pytest + `--check` workflow wired into sentinel, packaging nits (root line-length, one pytest pin, `requires-python = ">=3.11"`, STOP banner moved out of the module docstring).

Jaeger steps are template-owned, so 3.16/3.17/GCP/Azure e2e-install cells gain `quay-deploy-jaeger` / `quay-gather-jaeger-traces`. Older branches still replace `QUAY_EXTRA_CONFIG` wholesale and do not inherit OTEL flags.

**Deferred (replied on the comments)**
- Not shortening `test_as` to `{storage}-{tier}` — that would rename live 3.18 AWS and needs coordinated Prow history/alerts.
- Not implementing the superseded `cell_settings` / `env_profiles` shape; branch/job `env` is what landed.

AI-assisted comment ( cursor).

nasonawa:
Re https://github.com/quay/quay/pull/7110#issuecomment-5632279780: implemented. `matrix.yaml` is version 2 with branch-level `env` and optional job-level `env` / `as`. Merge order is template/tier → branch env → job env/as; env values replace whole keys; generated configs must have exactly one test. The `overrides/` files are removed. Live 3.18 AWS/GCP `as:` names and the GCP `@weekly` schedule are kept on split job rows.
AI-assisted comment ( cursor).

nasonawa:
Re https://github.com/quay/quay/pull/7110#issuecomment-5632280648: not doing the `{storage}-{tier}` `test_as` rename in this PR. The generated filename already carries version/cloud/OCP/test, but shortening `as:` would rename the live 3.18 AWS job (`aws-s3-3-18-nightly-4-22`) and needs coordinated Prow history and alert updates. Job-level `as:` is how we keep those live names.
AI-assisted comment ( cursor).

.github/workflows/ci-generator.yaml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          persist-credentials: false
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: pip install jinja2 pyyaml 'pytest>=7.0.0'
```
jbpratt: can we use uv here?

ci-generator/matrix.yaml
```diff
@@ -0,0 +1,74 @@
...
+        clouds: [aws, azure]
+        ocp: ["4.22"]
+        test: e2e-install
+      - tier: weekly
+        clouds: [gcp]
+        ocp: ["4.22"]
+        test: e2e-install
+  - branch: redhat-3.17
+    env:
+      PLAYWRIGHT_GREP_INVERT: "@auth:OIDC|@auth:LDAP"
```
jbpratt: Why does 3.17 need this but 3.18 doesn't?

nasonawa: the job template for e2e uses the  PLAYWRIGHT_GREP_INVERT: '@auth:OIDC|@auth:LDAP|@feature:QUOTA_NOTIFICATIONS|@webhook|saves and loads architecture filter with mirror configuration|loads existing architecture filter from saved mirror configuration' variable which is based on our first manual job config. 

modified the template for e2e-install.yaml to use  "PLAYWRIGHT_GREP_INVERT: "@auth:OIDC|@auth:LDAP"" as default value and added the override values for redhat-3.18 in the martrix.yaml

ci-generator/matrix.yaml
```diff
@@ -0,0 +1,74 @@
...
+        MAIL_SERVER: mailpit.quay-enterprise.svc.cluster.local
+        MAIL_PORT: 1025
+        MAIL_USE_TLS: false
+        MAIL_USE_AUTH: false
+        MAIL_DEFAULT_SENDER: quay@quayqe.local
+        DEFAULT_UI: react
+        SUPER_USERS:
+          - admin
+        GLOBAL_READONLY_SUPER_USERS:
+          - readonly
```
jbpratt: Let's move all of these we can to the default, the extra config can just be version specific or even no-ops

ci-generator/README.md
```diff
@@ -0,0 +1,134 @@
...
+quay:
+  - branch: redhat-3.18
+    jobs:
+      - {tier: daily, clouds: [aws, azure], ocp: ["4.22"], test: e2e-install}
+      - {tier: weekly, clouds: [gcp], ocp: ["4.22"], test: e2e-install}
+  - branch: redhat-3.17
+    env:
+      PLAYWRIGHT_GREP_INVERT: "@auth:OIDC|@auth:LDAP"
+      QUAY_EXTRA_CONFIG: |
+        # shared by every 3.17 job; replaces the template value whole
```
jbpratt: Oh gotcha, @nasonawa we should merge taking this as the override instead of replace

.github/workflows/sentinel.yaml
```diff
@@ -65,17 +70,40 @@ jobs:
...
               - '!.github/workflows/ci-lint.yaml'
+              - '!.github/workflows/sentinel.yaml'
 
   lint-ci:
     name: Lint
     uses: ./.github/workflows/ci-lint.yaml
     with:
       is-pr: ${{ github.event_name == 'pull_request' }}
 
+  actionlint:
```
jbpratt: @nasonawa let's remove this, there are lots of problems it seems - we'd need to do this in another PR

jbpratt (approved):

jbpratt:
Two further items from the reproduction pass that did not make the earlier review.

Feature-flag drift (should-fix). The e2e-install template sets `FEATURE_QUOTA_NOTIFICATIONS: true` for every generated cell. The live redhat-3.18 gcp config does not carry that flag, so regenerating turns it on for the gcp and new azure jobs as a side effect. Same class as the B5/B7/B9 drift already noted, but an addition rather than a drop. Either scope the flag to the cells that have it today or call the change out in the PR description.

Latent float stringification (nit). `_as_str_list` stringifies whatever PyYAML parsed. `4.22` round-trips exactly, but an unquoted `ocp: [4.20]` becomes `4.2` and silently yields `...-4-2` file and `as` names. Quoting the versions in matrix.yaml or rejecting non-string values would close it.

- ci-generator/templates/tests/e2e-install.yaml:11
- ci-generator/generate.py:141

jbpratt:
Two further items from the reproduction pass that did not make the earlier review.

Feature-flag drift (should-fix). The e2e-install template sets `FEATURE_QUOTA_NOTIFICATIONS: true` for every generated cell. The live redhat-3.18 gcp config does not carry that flag, so regenerating turns it on for the gcp and new azure jobs as a side effect. Same class as the B5/B7/B9 drift already noted, but an addition rather than a drop. Either scope the flag to the cells that have it today or call the change out in the PR description.

Latent float stringification (nit). `_as_str_list` stringifies whatever PyYAML parsed. `4.22` round-trips exactly, but an unquoted `ocp: [4.20]` becomes `4.2` and silently yields `...-4-2` file and `as` names. Quoting the versions in matrix.yaml or rejecting non-string values would close it.

- ci-generator/templates/tests/e2e-install.yaml:11
- ci-generator/generate.py:141

PR #7149, opened 2026-09-11
PROJQUAY-13202: fix(omr): purge API-deleted repositories in the GC cycle

harishsurf:
## Summary

- Adds a purge phase to the GC collector (after expired upload-marker cleanup, before tag expiry) that finds repositories in state 3 and deletes, in one transaction and in foreign-key order, every row referencing the repository, then the `deletedrepository` marker, its `queueitem`, and the repository row (only while still in state 3)
- Blobs are left to the existing orphaned-blob phase; with the links and the repository's upload markers gone they are reclaimed in the same cycle unless another repository references them
- `gc.Stats` gains `RepositoriesPurged`; the cycle log line reports it
- Adds sqlc queries for each delete, unit tests (purge, shared-blob survival, state guard), an e2e `DeleteRepository` client helper, and an e2e test that deletes through the API and checks rows, markers, reclamation, the surviving shared blob, and reuse of the name

## Context

`DELETE /api/v1/repository/<ns>/<name>` renames the repository to a UUID, sets `state = 3`, inserts a `deletedrepository` marker and a `queueitem`, and returns 204 — the first half of Quay's `mark_repository_for_deletion`. Quay drains that queue with `workers/repositorygcworker.py` → `model.gc.purge_repository`; the mirror registry had no consumer. Tags, manifests, `manifestblob` links, upload markers and blob files stayed forever, and a blob shared with a deleted repository could never be reclaimed because the hidden repository still referenced it. On the performance VM, deleting a repository holding a 1 GiB layer freed nothing after any number of cycles.

## Test plan

- [x] `go test ./internal/gc/ -count=1` (new `TestCollect_PurgesMarkedRepository`, `TestCollect_PurgeKeepsBlobsSharedWithLiveRepository`, `TestCollect_ActiveRepositoryIsNeverPurged`)
- [x] `go test ./internal/e2e/mirrorregistry/ -count=1` (new `TestRegistryDeleteRepositoryPurgesAndReclaimsStorage`)
- [x] `go test ./internal/... -count=1`, `golangci-lint run ./internal/gc/... ./internal/dal/... ./internal/e2e/...`
- [x] Manual, on the OMR VM: two repositories soft-deleted earlier with the master build (one holding a 1 GiB layer) were purged on the first cycle after swapping in the fixed binary — `repositories_purged=2 … blobs_deleted=2 bytes_reclaimed=1073834182`, storage dir shrank by 1024 MiB, marker and queue tables empty. A fresh push → API delete was purged on the next 30 s cycle; a sibling repository sharing the same layers still pulls

## Review follow-ups (2026-09-21)

- A `deletedrepository.queue_id` that does not parse as an integer is now an error from `FindMarkedRepositories`; the purge phase reports it instead of purging around the marker and leaving the queue row behind.
- `DeleteRepository` is `:execrows`. If it affects zero rows (another cycle purged the repository in the discovery gap, or it was reactivated) `PurgeRepository` returns `gc.ErrRepositoryNotMarked` without committing, so the dependent deletes roll back; the collector logs and skips it rather than failing the phase.
- Tests: `TestFindMarkedRepositories_MalformedQueueIDIsAnError`, `TestPurgeRepository_NoLongerMarkedRollsBackAndIsSkipped`. `go test -race ./internal/gc/` and `golangci-lint` clean on Linux; `internal/dal/daldb/gc.sql.go` regenerated with sqlc 1.30.0.

### DIFF
--- a/internal/dal/queries/gc.sql
+++ b/internal/dal/queries/gc.sql
@@ -79,3 +79,114 @@ SELECT COUNT(*) FROM uploadedblob WHERE blob_id = ? AND expires_at > datetime('n
 
 -- name: CountBlobsByChecksum :one
 SELECT COUNT(*) FROM imagestorage WHERE content_checksum = ?;
+
+-- Repository purge (PROJQUAY-13202). The repository API marks a repository as
+-- deleted (state = 3, renamed to a UUID, deletedrepository marker plus a
+-- queueitem) exactly like Quay's mark_repository_for_deletion. Quay drains
+-- that queue with repositorygcworker; the mirror registry purges marked
+-- repositories inside its GC cycle instead. Deletes run child tables first so
+-- SQLite foreign keys (PRAGMA foreign_keys=1) are satisfied.
+
+-- name: FindMarkedRepositories :many
+SELECT r.id, d.id AS marker_id, d.queue_id
+FROM repository r
+LEFT JOIN deletedrepository d ON d.repository_id = r.id
+WHERE r.state = 3
+ORDER BY r.id;
+
+-- name: CountManifestsByRepository :one
+SELECT COUNT(*) FROM manifest WHERE repository_id = ?;
+
+-- name: CountTagsByRepository :one
+SELECT COUNT(*) FROM tag WHERE repository_id = ?;
+
+-- name: DeleteTagNotificationsByRepository :exec
+DELETE FROM tagnotificationsuccess
+WHERE tag_id IN (SELECT id FROM tag WHERE repository_id = ?);
+
+-- name: DeleteTagsByRepository :exec
+DELETE FROM tag WHERE repository_id = ?;
+
+-- name: DeleteManifestLabelsByRepository :exec
+DELETE FROM manifestlabel WHERE repository_id = ?;
+
+-- name: DeleteManifestChildrenByRepository :exec
+DELETE FROM manifestchild WHERE repository_id = ?;
+
+-- name: DeleteManifestBlobsByRepository :exec
+DELETE FROM manifestblob WHERE repository_id = ?;
+
+-- name: DeleteManifestSecurityStatusByRepository :exec
+DELETE FROM manifestsecuritystatus WHERE repository_id = ?;
+
+-- name: DeleteManifestPullStatisticsByRepository :exec
+DELETE FROM manifestpullstatistics WHERE repository_id = ?;
+
+-- name: DeleteTagPullStatisticsByRepository :exec
+DELETE FROM tagpullstatistics WHERE repository_id = ?;
+
+-- name: DeleteManifestsByRepository :exec
+DELETE FROM manifest WHERE repository_id = ?;
+
+-- name: DeleteUploadedBlobsByRepository :exec
+DELETE FROM uploadedblob WHERE repository_id = ?;
+
+-- name: DeleteBlobUploadsByRepository :exec
+DELETE FROM blobupload WHERE repository_id = ?;
+
+-- name: DeleteRepositoryBuildsByRepository :exec
+DELETE FROM repositorybuild WHERE repository_id = ?;
+
+-- name: DeleteRepositoryBuildTriggersByRepository :exec
+DELETE FROM repositorybuildtrigger WHERE repository_id = ?;
+
+-- name: DeleteAccessTokensByRepository :exec
+DELETE FROM accesstoken WHERE repository_id = ?;
+
+-- name: DeleteRepoMirrorConfigByRepository :exec
+DELETE FROM repomirrorconfig WHERE repository_id = ?;
+
+-- name: DeleteRepoMirrorRulesByRepository :exec
+DELETE FROM repomirrorrule WHERE repository_id = ?;
+
+-- name: DeleteOrgMirrorRepositoriesByRepository :exec
+DELETE FROM orgmirrorrepository WHERE repository_id = ?;
+
+-- name: DeleteApprTagsByRepository :exec
+DELETE FROM apprtag WHERE repository_id = ?;
+
+-- name: DeleteRepositoryPermissionsByRepository :exec
+DELETE FROM repositorypermission WHERE repository_id = ?;
+
+-- name: DeleteRepositoryNotificationsByRepository :exec
+DELETE FROM repositorynotification WHERE repository_id = ?;
+
+-- name: DeleteRepositoryActionCountsByRepository :exec
+DELETE FROM repositoryactioncount WHERE repository_id = ?;
+
+-- name: DeleteRepositoryAuthorizedEmailsByRepository :exec
+DELETE FROM repositoryauthorizedemail WHERE repository_id = ?;
+
+-- name: DeleteRepositoryAutoPrunePoliciesByRepository :exec
+DELETE FROM repositoryautoprunepolicy WHERE repository_id = ?;
+
+-- name: DeleteRepositoryImmutabilityPoliciesByRepository :exec
+DELETE FROM repositoryimmutabilitypolicy WHERE repository_id = ?;
+
+-- name: DeleteRepositorySearchScoresByRepository :exec
+DELETE FROM repositorysearchscore WHERE repository_id = ?;
+
+-- name: DeleteQuotaRepositorySizesByRepository :exec
+DELETE FROM quotarepositorysize WHERE repository_id = ?;
+
+-- name: DeleteStarsByRepositoryForPurge :exec
+DELETE FROM star WHERE repository_id = ?;
+
+-- name: DeleteDeletedRepositoryMarkers :exec
+DELETE FROM deletedrepository WHERE repository_id = ?;
+
+-- name: DeleteQueueItem :exec
+DELETE FROM queueitem WHERE id = ?;
+
+-- name: DeleteRepository :exec
+DELETE FROM repository WHERE id = ? AND state = 3;
--- a/internal/gc/gc_test.go
+++ b/internal/gc/gc_test.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"fmt"
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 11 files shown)


internal/gc/sqlite_store.go
```diff
@@ -25,6 +26,109 @@ func (s *SQLiteStore) CleanExpiredUploadedBlobs(ctx context.Context) error {
...
+	for _, r := range rows {
+		repo := MarkedRepository{ID: r.ID}
+		if r.MarkerID.Valid {
+			repo.MarkerID = r.MarkerID.Int64
+		}
+		if r.QueueID.Valid {
+			if id, err := strconv.ParseInt(r.QueueID.String, 10, 64); err == nil {
+				repo.QueueID = id
+			}
+		}
```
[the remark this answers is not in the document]
harishsurf: Fixed. A `queue_id` that does not parse as an integer is now returned as an error from `FindMarkedRepositories` (`repository N: malformed deletedrepository.queue_id "..."`), so the purge phase reports it instead of purging around the marker and leaving the queue row behind. Covered by `TestFindMarkedRepositories_MalformedQueueIDIsAnError`, which also asserts `Collect` fails and the repository rows are untouched.

internal/gc/sqlite_store.go
```diff
@@ -25,6 +26,109 @@ func (s *SQLiteStore) CleanExpiredUploadedBlobs(ctx context.Context) error {
...
+	}
+	if repo.QueueID != 0 {
+		if err := q.DeleteQueueItem(ctx, repo.QueueID); err != nil {
+			return purge, fmt.Errorf("delete queue item %d for repository %d: %w", repo.QueueID, repo.ID, err)
+		}
+	}
+	if err := q.DeleteRepository(ctx, repo.ID); err != nil {
+		return purge, fmt.Errorf("delete repository %d: %w", repo.ID, err)
+	}
```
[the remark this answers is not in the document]
harishsurf: Fixed. `DeleteRepository` is now `:execrows` (regenerated with sqlc 1.30.0). When it affects zero rows `PurgeRepository` returns `gc.ErrRepositoryNotMarked` without committing, so the dependent deletes in the same transaction roll back too. The collector treats that error as a skip (logged, not counted) rather than failing the phase, since the usual cause is another cycle having purged the repository in the discovery gap. Covered by `TestPurgeRepository_NoLongerMarkedRollsBackAndIsSkipped`: tags, manifests, marker and repository all still present after the aborted purge, and a full `Collect` succeeds with 0 purged.

internal/dal/queries/gc.sql
```diff
@@ -79,3 +79,114 @@ SELECT COUNT(*) FROM uploadedblob WHERE blob_id = ? AND expires_at > datetime('n
...
+DELETE FROM quotarepositorysize WHERE repository_id = ?;
+
+-- name: DeleteStarsByRepositoryForPurge :exec
+DELETE FROM star WHERE repository_id = ?;
+
+-- name: DeleteDeletedRepositoryMarkers :exec
+DELETE FROM deletedrepository WHERE repository_id = ?;
+
+-- name: DeleteQueueItem :exec
+DELETE FROM queueitem WHERE id = ?;
```
jbpratt: `DELETE FROM queueitem WHERE id = ?` is the only delete in the purge not scoped to the repository. The id comes from `deletedrepository.queue_id`, a nullable `VARCHAR(255)` that nothing constrains to point at a live row of this queue.

No table in `internal/dal/schema/sqlite/quay_schema.sql` declares `AUTOINCREMENT`, so SQLite reuses the highest rowid. A marker whose queue item is already gone — the Python worker exhausting its retries leaves exactly that state, and the OMR v2 bridge copies `deletedrepository` and `queueitem` together with their ids — can therefore delete whichever `queueitem` later inherits that id, in the same transaction that drops the repository.

Suggested: `DELETE FROM queueitem WHERE id = ? AND queue_name LIKE 'repositorygc/%';`

internal/gc/sqlite_collector.go
```diff
@@ -75,6 +89,42 @@ func (c *collector) Collect(ctx context.Context) (Stats, error) {
...
+		return totals, err
+	}
+	for _, repo := range repos {
+		purge, err := c.store.PurgeRepository(ctx, repo)
+		if errors.Is(err, ErrRepositoryNotMarked) {
+			c.log.Info("gc: skipped repository purge, no longer marked for deletion", "repo_id", repo.ID)
+			continue
+		}
+		if err != nil {
+			return totals, fmt.Errorf("purge repository %d: %w", repo.ID, err)
```
jbpratt: Returning here aborts the entire cycle: `Collect` propagates the error, so tag expiry, manifest collection, blob reclamation and stale-upload cleanup never run for any repository. `FindMarkedRepositories` does the same one level up, failing the phase for a single malformed `queue_id` anywhere in the table.

The purge phase is new and runs first, so one repository that cannot be purged now stops all space reclamation indefinitely — the same symptom this PR fixes, reached from the other side. `TestFindMarkedRepositories_MalformedQueueIDIsAnError` pins that behaviour rather than flagging it.

This sharpens the resolved `queue_id` thread on `internal/gc/sqlite_store.go`: refusing to purge around a malformed marker is right, the blast radius is not.

Suggested: log the repository and continue the loop, returning the accumulated error after it, or move the purge phase after blob collection.

internal/gc/gc_test.go
```diff
@@ -1432,3 +1435,310 @@ func ctx(t *testing.T) context.Context {
...
+		t.Fatal(err)
+	}
+	id, err := env.store.PutBlob(t.Context(), oci.BlobRecord{Digest: dgst, Size: int64(len(content))})
+	if err != nil {
+		t.Fatal(err)
+	}
+	return id, dgst
+}
+
+// TestCollect_PurgesMarkedRepository covers PROJQUAY-13202: a repository the
```
jbpratt, line 1503: This test and the e2e test populate `tag`, `manifest`, `manifestblob` and `uploadedblob` only, so 23 of the 28 delete steps in `PurgeRepository` run against empty tables, where a wrong order still passes. The orderings that carry the risk are never exercised: builds before triggers before access tokens, mirror config before mirror rules, tag notifications before notifications.

Suggested: one fixture seeding a row in every table with a foreign key to `repository`, then purge and assert both zero remaining rows and a clean `PRAGMA foreign_key_check`.

internal/gc/gc_test.go
```diff
@@ -1432,3 +1435,310 @@ func ctx(t *testing.T) context.Context {
...
+	}
+	if n := countRows(t, env, `SELECT count(*) FROM tag WHERE repository_id = ? AND lifetime_end_ms IS NULL`, liveRepo); n != 1 {
+		t.Errorf("live repository lost its tag: %d live tags", n)
+	}
+	if n := countRows(t, env, `SELECT count(*) FROM manifestblob WHERE repository_id = ?`, liveRepo); n != 1 {
+		t.Errorf("live repository lost its blob link: %d rows", n)
+	}
+}
+
+// TestCollect_ActiveRepositoryIsNeverPurged guards the state = 3 condition:
```
jbpratt: This does not guard the `state = 3` condition on the delete. Removing `AND state = 3` from `DeleteRepository` leaves this test green; the repository is protected here by the `WHERE r.state = 3` filter in `FindMarkedRepositories`, and the delete guard is covered by `TestPurgeRepository_NoLongerMarkedRollsBackAndIsSkipped`.

Suggested: reword the comment to name the discovery filter, so a future change to the delete guard is not assumed to be covered here.

internal/dal/queries/gc.sql
```diff
@@ -79,3 +79,114 @@ SELECT COUNT(*) FROM uploadedblob WHERE blob_id = ? AND expires_at > datetime('n
...
+-- name: DeleteRepositoryImmutabilityPoliciesByRepository :exec
+DELETE FROM repositoryimmutabilitypolicy WHERE repository_id = ?;
+
+-- name: DeleteRepositorySearchScoresByRepository :exec
+DELETE FROM repositorysearchscore WHERE repository_id = ?;
+
+-- name: DeleteQuotaRepositorySizesByRepository :exec
+DELETE FROM quotarepositorysize WHERE repository_id = ?;
+
+-- name: DeleteStarsByRepositoryForPurge :exec
```
jbpratt: Byte-identical to `DeleteStarsByRepository` in `internal/dal/queries/repository.sql:70`, which produces a second generated method for the same statement. `Store.MarkDeleted` also deletes the repository's stars inside the mark transaction, so for anything the API marked this step has nothing left to remove.

Suggested: reuse the existing query.

jbpratt (commented):
The purge itself is sound on this schema. Every table with a foreign key to `repository` is covered, the delete order satisfies those constraints, and the `state = 3` condition on the final delete combined with the rollback on zero affected rows means a repository that leaves the deleted state mid-purge loses nothing.

Two things are worth changing before merge. The queue-item delete is the only statement in the purge not scoped to the repository. And a single repository that cannot be purged now blocks tag expiry, manifest collection and blob reclamation, because the purge phase runs first and returns on the first error — the reclamation failure this PR exists to fix, reintroduced by a different route.

The remaining comments are test coverage and duplication.

internal/dal/queries/gc.sql
harishsurf: Done. `DeleteQueueItem` now takes the id and a queue-name pattern (`repositorygc/%/<repository id>/`, the name the API writes), so a marker whose queue_id was reused by an item of another queue leaves that item alone. `TestPurgeRepository_DoesNotDeleteForeignQueueItem` deletes the real item, inserts a foreign-queue item under the same rowid, and asserts the purge succeeds and the foreign item survives.

internal/gc/sqlite_collector.go
harishsurf: Done. The purge loop now logs and continues on a failed repository and returns the joined errors after the loop, and `Collect` carries that error to the end of the cycle instead of returning early, so tag expiry, manifest collection, blob reclamation and stale-upload cleanup still run (a later phase error is joined with it, not lost). `queue_id` is also parsed in `PurgeRepository` rather than in discovery, so one malformed marker fails one repository. `TestCollect_MalformedQueueIDFailsOnlyThatRepository` has a bad marker, a healthy marker and an expired upload marker in one cycle: the healthy repository is purged, the orphaned blob is reclaimed, the bad repository is left intact and still marked, and the error names it. The worker logs the error and still reports the partial stats.

internal/gc/gc_test.go
harishsurf: Done. `TestPurgeRepository_RemovesEveryDependentRow` seeds one row in every table the purge deletes from (28 tables, including the FK chains build → trigger → access token, mirror config → mirror rule, tag notification success → notification/tag, and the label / org-mirror-config / manifest-child dependencies), purges, and asserts zero rows remain in each table plus an empty `PRAGMA foreign_key_check`. Seeding it surfaced no ordering bug in the existing step list.

internal/gc/gc_test.go
harishsurf: Done, comment reworded to say it guards the `WHERE r.state = 3` discovery filter and points at `TestPurgeRepository_NoLongerMarkedRollsBackAndIsSkipped` for the delete guard.

internal/dal/queries/gc.sql
harishsurf: Done. Dropped `DeleteStarsByRepositoryForPurge`; the purge calls the existing `DeleteStarsByRepository`. Kept the step so a marker created outside the API path (migration) is still cleaned.

internal/gc/gc_test.go
```diff
@@ -1432,3 +1435,489 @@ func ctx(t *testing.T) context.Context {
...
+	if n := countRows(t, env, `SELECT count(*) FROM manifestblob WHERE repository_id = ?`, liveRepo); n != 1 {
+		t.Errorf("live repository lost its blob link: %d rows", n)
+	}
+}
+
+// TestCollect_ActiveRepositoryIsNeverPurged guards the discovery filter
+// (WHERE r.state = 3 in FindMarkedRepositories): a repository that merely has
+// a stale deletedrepository marker but is still active is never selected. The
+// state = 3 guard on the final DELETE itself is covered by
+// TestPurgeRepository_NoLongerMarkedRollsBackAndIsSkipped.
```
jbpratt, line 1628: The rewording resolves the original inaccuracy but introduces the same kind in the other direction: this test does not pin the discovery filter either. Deleting `WHERE r.state = 3` from `FindMarkedRepositories` leaves `TestCollect_ActiveRepositoryIsNeverPurged` green — the repository is then selected, and the `AND state = 3` on the final `DELETE` rejects it and rolls the transaction back, so the assertions still hold. The test only goes red when both guards are removed together.

That wording came from the suggestion on comment 4066192858, so the imprecision originates there, not here.

Consequence: the same trap the original comment set, pointed at the discovery filter — a future change to that filter looks covered by this test and is not. No test in the package fails on its removal alone.

Suggested: describe what the test actually covers — the end-to-end invariant that an active repository with a stale marker is never purged, enforced by the discovery filter and the delete guard together — or add `FindMarkedRepositories` to the test and assert it returns nothing, which makes the current wording true.

jbpratt (commented):
All five comments from the previous pass are resolved in code, not just in replies. The two that carried risk — the unscoped `queueitem` delete and a single bad repository aborting the whole cycle — are fixed and each is now pinned by a test that fails when the fix is reverted.

Every delete in the purge is bounded to the target repository: 25 statements on `repository_id = ?`, `tagnotificationsuccess` through a subquery on that repository's tags, `queueitem` on `id` plus the `repositorygc/%/<repository id>/` name, and `repository` on `id AND state = 3` with the `:execrows` zero-row check rolling the transaction back. The purge cannot reach another repository's rows.

One leftover, minor and comment-only, below. It does not block merge.

Recommend approval once that comment is corrected or dropped.

internal/gc/gc_test.go
harishsurf: Agreed, the reworded comment overclaimed. The test now calls `FindMarkedRepositories` directly and asserts it returns nothing for the active repository, then runs the full cycle as before. Checked by mutation: changing the discovery filter to `WHERE 1 = 1` and regenerating makes it fail on the new assertion (`got [{ID:1 MarkerID:1 QueueID:}]`), with the delete guard still intact; restoring the filter makes the suite green again. Comment updated to say both guards are asserted on their own.

PR #7221, opened 2026-09-17
PROJQUAY-6560: feat(quota): add Prometheus metrics for quota consumption

rgordill:
Adds the following metrics related to the quota system:

- registry_total_used_bytes - Storage consumption in bytes for the entire registry
- namespace_stats_used_bytes - Storage consumption in bytes per namespace (labels: namespace, entity_type)
- namespace_quota_stats_capacity_bytes - Storage quota in bytes per namespace (labels: namespace, entity_type)
- namespace_quota_stats_available_bytes - Remaining storage capacity per quota in bytes per namespace (labels: namespace, entity_type)
- repository_stats_used_bytes - Storage consumption in bytes per repository (labels: repository, namespace)

Introduces a new config tunable QUOTA_METRICS (default: False). Only when enabled will the above metrics be emitted. GLOBAL_PROMETHEUS_STATS_FREQUENCY controls the refresh interval (default: 1 hour).

Database queries are paginated using keyset pagination to avoid full table scans on large installations.


workers/globalpromstats/globalpromstats.py
```diff
@@ -48,6 +75,105 @@ def get_robot_count():
...
+    still emit capacity and available-bytes metrics with usage treated as zero.
+
+    At most QUOTA_METRICS_NAMESPACES_LIMIT namespaces are processed per cycle
+    to prevent unbounded work, memory and metric cardinality on large registries.
+    """
+    namespace_stats_used_bytes._metrics.clear()
+    namespace_quota_stats_capacity_bytes._metrics.clear()
+    namespace_quota_stats_available_bytes._metrics.clear()
+
+    namespace_quotas = {row["id"]: row for row in model.namespacequota.get_namespaces_with_quotas()}
```
Marcusk19: namespace limit does not bound memory or database work 

Although emitted metric series are limited later, this consumes every `UserOrganizationQuota` row before applying
`QUOTA_METRICS_NAMESPACES_LIMIT`. On a large deployment, enabling metrics can still scan all quota namespaces and
materialize them in worker memory, defeating the stated bounded-work/cardinality design.

Fix: Apply the configured cap in the query/iterator itself, or drive collection from a single bounded query that
outer-joins quota-size data. Add a regression test proving iteration stops at the limit (not merely that emitted metrics
do).

Marcusk19 (changes requested):
adding an inline comment caught by my reviewer agent, seems like the initial coderabbit concern was not fully addressed. also I think at least the [major coderabbit finding](https://github.com/quay/quay/pull/7221#discussion_r4038532969) that is still open should be addressed as well

workers/globalpromstats/globalpromstats.py
rgordill: My bad, I checked other comments, looked that some were fixed but I didn't check this.

**Fix**
Pushed the limit into the generators themselves via a new max_rows parameter on all three data-layer functions:

get_namespaces_with_quotas(max_rows=limit) — stops fetching from DB after limit rows
get_all_namespace_sizes(max_rows=limit) — same
get_all_repository_sizes(max_rows=limit) — same
Each generator now computes fetch_limit = min(batch_size, remaining) so even the SQL LIMIT clause is tightened as the cap approaches.

**Tests updated**
Both test_repo_metrics_limit and test_namespace_metrics_limit now:

Use side_effect lambdas that respect max_rows (simulating the real generator behavior)
Assert that max_rows was correctly passed: mock_get_all_repository_sizes.assert_called_with(max_rows=3)

nasonawa:
@rgordill could you remove the rust-builddeps/target folder from commit.

rgordill:
I reviewed previous commits, and It was not there, don't know why it was in the latest one. Now fixed, and .gitignore added to avoid future issues.

PR #7298, opened 2026-09-21
QUAYIO-2198: fix(cve): bump vitest to 4.1.11 for CVE-2026-73653

harishsurf:
## Summary
- Bumps `vitest` and `@vitest/coverage-v8` from `^4.1.4` to `^4.1.11` to address **CVE-2026-73653**
- The vulnerability allows Browser Mode provider commands to bypass the file-access permission gate, enabling arbitrary file read/write/delete even when `allowWrite` is false
- Fixed upstream in vitest `3.2.7`, `4.1.10`, and `5.0.0-beta.6`

## Test plan
- [x] Verified `make local-dev-up` builds successfully (pnpm frozen-lockfile passes)
- [x] Verified UI loads correctly at localhost:8080
- [ ] CI `vitest run --coverage` passes (web-ci workflow)

## Jira
https://redhat.atlassian.net/browse/QUAYIO-2198


cubismod (approved):

web/pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+allowBuilds:
```
harishsurf, line 1: pnpm 10.7+ blocks post-install build scripts by default. `@parcel/watcher` (transitive dep via `sass` → `vite` → `vitest`) has a native build script that requires explicit approval, otherwise `pnpm install --frozen-lockfile` fails with `ERR_PNPM_IGNORED_BUILDS`. The Dockerfile uses `pnpm@10` (unpinned, resolves to latest), so this file is required for the container build to succeed.

Marcusk19 (approved):

PR #7299, opened 2026-09-21
PROJQUAY-11278: fix: Properly parse region_name on RadosGWStorage validation

ibazulic:
Previously, the region_name parameter was not read by the validator during storage validation, Minio tried to figure out what the region is automatically. Certain storage engines (like garage) reported a `400 Bad Request` because the initial request was malformed due to missing region. 

This PR adds explicit validation of the region name with RadosGWStorage driver. If region name is not present, validator will fall back to established behavior.


config-tool/pkg/lib/shared/storage_validators_test.go
```diff
@@ -1,9 +1,45 @@
...
 
+func makeGarageMockServer(t *testing.T) (url string, locationCalled *bool) {
+	called := false
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		t.Logf("Called method: %s", r.Method)
+		t.Logf("Endpoint: %s", r.RequestURI)
+		t.Logf("Params: %s", r.URL.RawQuery)
+		t.Logf("Headers: %v", r.Header)
+		if strings.Contains(r.URL.RawQuery, "location") {
+			called = true
```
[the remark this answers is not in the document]
ibazulic: I'm not sure this is necessary. Each test spins the test HTTP server and tears it down after test completes, all variables are instantiated only once and potentially only changed only once per test. There's no concurrency issue. The only thing we need to verify here is if a specific endpoint has or has not been called, and track it. And for that, a normal boolean flag is more than adequate.

config-tool/pkg/lib/shared/storage_validators_test.go
```diff
@@ -85,6 +121,79 @@ func TestBuildEndpoint(t *testing.T) {
...
+		Hostname:   host,
+		BucketName: "bucket",
+		RegionName: "failimmediately",
+	}
+
+	ok, err := ValidateStorage(Options{Mode: "Online"}, "test", "RadosGWStorage", args, "fg")
+
+	assert.False(t, ok)
+	t.Logf("received error: %s", err[0].Message)
+	assert.Contains(t, err[0].Message, "Bad Request")
```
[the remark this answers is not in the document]
ibazulic: Weird, the test passes and with a proper error message. Logs show that MinIO does report a 400 being raised. But I agree, the XML should be correct. Fixed in new commit.

config-tool/pkg/lib/shared/storage_validators.go
```diff
@@ -127,6 +127,13 @@ func ValidateStorage(opts Options, storageName string, storageType string, args
 			errors = append(errors, err)
 		}
 
+		// apply the region name if available, else skip
+		var region string
```
jbpratt: **Medium:** this case also covers `IBMCloudStorage`, but the registry's `IBMCloudStorage.__init__` (`storage/cloud.py:904`) has no `region_name` parameter and drivers are built with `driver_class(context, **parameters)`.
An `IBMCloudStorage` config with `region_name` now validates green against the endpoint with that region, while Quay raises `TypeError` at startup.
Suggest applying the region only for `RadosGWStorage`/`RHOCSStorage`, or reporting `region_name` as unsupported for `IBMCloudStorage`. The `var`/`if` pair can also be `region := args.RegionName`; the `!= ""` guard changes nothing.

config-tool/pkg/lib/fieldgroups/distributedstorage/distributedstorage_test.go
```diff
@@ -20,16 +28,56 @@ func TestValidateDistributedStorage(t *testing.T) {
...
-	// Define test data
-	var tests = []struct {
-		name   string
-		config map[string]interface{}
-		want   string
-	}{
-
-		{name: "MissingStorageConfig", config: map[string]interface{}{}, want: "invalid"},
+    storage_path: /datastorage/registry`), want: "valid"},
+		{name: "StorageWithRegionSet", config: []byte(`DISTRIBUTED_STORAGE_CONFIG:
```
jbpratt: **Medium:** removing the new `region_name` block from `NewDistributedStorageArgs` (`distributedstorage.go:145-150`) leaves every test in `./pkg/lib/shared/...` and `./pkg/lib/fieldgroups/distributedstorage/...` passing.
These cases run in `testing` mode, which now returns before any connection, and only check valid/invalid; the new `shared` tests build `DistributedStorageArgs` directly and bypass the parser.
The parser is the bug in the PR title, so it needs a direct assertion. Suggest checking `fg.DistributedStorageConfig["local_us"].Args.RegionName == "someregion"` for this case.

config-tool/pkg/lib/shared/storage_validators_test.go
```diff
@@ -85,6 +121,79 @@ func TestBuildEndpoint(t *testing.T) {
 	}
 }
 
+// Call validateMinioGateway with region set: bypasses GetBucketLocation on storage side and issues an immediate HEAD
+// on the bucket.
+func TestValidateMinioGatewayWithRegionSkipsLocationDetection(t *testing.T) {
```
jbpratt: **Medium:** CI does not run these tests. `.github/workflows/ci-config-tool.yaml` runs only `go test ./pkg/lib/fieldgroups/...`, so the four new `validateMinioGateway`/`ValidateStorage` tests never execute in CI.
They are the only tests that exercise the fix: dropping `Region: region` from `minio.New` fails three of the four locally.
Suggest adding `./pkg/lib/shared/...` to the CI test step, or moving a region test under `fieldgroups/`. Note that `./pkg/lib/shared/...` also contains `TestValidateCertPairWithHostname`, which fails on master locally with Go 1.26 (`x509: certificate relies on legacy Common Name field`) and would need fixing first.

config-tool/pkg/lib/fieldgroups/distributedstorage/distributedstorage.go
```diff
@@ -142,6 +142,13 @@ func NewDistributedStorageArgs(storageArgs map[string]interface{}) (*shared.Dist
 		}
 	}
 
+	if value, ok := storageArgs["region_name"]; ok {
+		newDistributedStorageArgs.RegionName, ok = value.(string)
```
jbpratt: **Low:** two `region_name` shapes now diverge from the registry.
`region_name:` with no value (YAML null) now fails with `region_name must be of type string`. Before this PR the key was ignored, and `RadosGWStorage(region_name=None)` falls back to `us-east-1`.
`region_name: ""` is accepted and treated as absent here, while boto3 signs with an empty region in the scope (`Credential=<key>/<date>//s3/aws4_request`), which a scope-checking backend such as Garage is likely to reject.
Suggest treating `nil` as absent and rejecting an empty string.

config-tool/pkg/lib/shared/storage_validators_test.go
```diff
@@ -1,9 +1,45 @@
...
+		t.Logf("Called method: %s", r.Method)
+		t.Logf("Endpoint: %s", r.RequestURI)
+		t.Logf("Params: %s", r.URL.RawQuery)
+		t.Logf("Headers: %v", r.Header)
+		if strings.Contains(r.URL.RawQuery, "location") {
+			called = true
+			// Garage returns 400 when scope is wrong
+			w.Header().Set("Content-type", "application/xml")
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte(`<?xml version="1.0"?><Error><Code>AuthorizationHeaderMalformed</code>` +
```
jbpratt: **Nit:** `<Code>AuthorizationHeaderMalformed</code>` closes with lowercase `</code>`, so the body is malformed XML and the mock never returns the error code this comment describes. Suggest `</Code>`.

jbpratt (changes requested):
Setting `region_name` on `RadosGWStorage` now pins the SigV4 scope and skips `GetBucketLocation`, which fixes the Garage case. Two gaps remain: no test that CI runs fails if the fix is reverted, and `IBMCloudStorage` now honors a key the registry rejects at startup. The `go.mod` testify line is needed by the new tests (`go mod tidy` is clean); the `commands/` `Fatalf` changes (they clear a `go vet` error that fails `go test ./commands` on master), the license-comment reindent and the `.gitignore` `vendor/` entry are unrelated to the fix and would fit a separate chore PR.

